PR #34, opened 2013-02-22
get_best_seen plugin

megatrond:
Simple plugin for -the imdb plugin- either of the lookup plugins (imdb, rottentomatoes).

Specify a timeframe, a number of movies and what score you want to check for and the plugin will accept the highest ranking entry with the highest quality within that timeframe. It will then reset and start over.

If an entry is accepted it will also reset and start over.

If used with a simple config, with either imdb_lookup or rottentomatoes_lookup, it will just download a number of entries within a timeframe, so it's good for getting i.e the 2 best movies every week.

### DIFF
--- a/flexget/plugins/filter/get_best_seen.py
+++ b/flexget/plugins/filter/get_best_seen.py
@@ -61,7 +61,6 @@ class GetBest(object):
 
 	Configuration:
 		get_best_seen:
-		  timeframe: <int≥
 		  timeframe: <int>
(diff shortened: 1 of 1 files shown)


gazpachoking:
@megatrond Do you want to take a stab at those last notes and update to work with newest develop? We are trying to clean up the PRs.

gazpachoking:
Tests would also be awesome.

PR #59, opened 2013-06-13
regexp queue

gazpachoking:
Adds a queue plugin similar to movie_queue for generic regex matches.

Still todo:
- [x] Cli interface for adding stuff, similar to `--movie-queue`
- [ ] Check if the CLI stuff might need to be rewritten to fit the existing code base
- [ ] Something like `queue_movies` to add to the regex queue from a task

Other stuff?

Maybe some more stuff can be put into queue_base for both plugins to use to implement above.

### DIFF
--- a/docs/api/flexget.rst
+++ b/docs/api/flexget.rst
@@ -65,16 +65,27 @@ flexget Package
     :undoc-members:
     :show-inheritance:
 
-:mod:`schema` Module
+:mod:`db_schema` Module
 --------------------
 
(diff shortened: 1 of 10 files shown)


stevezau:
@gazpachoking is this something the list_x plugins could solve??

gazpachoking:
I think this can be implemented more cleanly and easily if this happens.
http://flexget.com/wiki/Drafts/ListInterface
Closing for now, functionality will probably reappear at some point in a different form.

PR #60, opened 2013-06-19
Add mldonkey output plugin

Crupuk:
Work with mldonkey >= 3.0.6 for Ed2k link, local or distant torrent url.
Use telnet interface to add torrent to mldonkey

### DIFF
--- a/flexget/plugins/output/mldonkey.py
+++ b/flexget/plugins/output/mldonkey.py
@@ -10,10 +10,11 @@ class OutputMlDonkey(object):
     Example::
 
       mldonkey:
-        ip: "localhost"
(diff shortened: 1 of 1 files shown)


stevezau:
@Crupuk seems this PR fell through the cracks.

Is this PR still valid and working? If not i'll close the PR.

gazpachoking:
There are still several indentation issues, and I think it would be nicer to use https://github.com/ygrek/mldonkey/blob/api/docs/api.md than telnet. @Crupuk If you are no longer using this, going to close, and someone who is still using (and able to test/maintain) can open another PR.

PR #77, opened 2013-09-08
add search and input plugin for PublicHD.eu

mstrand:
http://www.publichd.eu input/search plugin


gazpachoking:
It looks like maybe we got this from somewhere else. Sorry to let this lapse @mstrand. Can you open a new one if there are still needed features/improvements?
https://github.com/Flexget/Flexget/blob/develop/flexget/plugins/search_publichd.py

cvium:
It would be better to delete the plugin entirely. The new PublicHD is incredibly bad and doesn't really seem to have anything to do with the old site.

PR #87, opened 2013-10-12
Overhaul of the ftp_download plugin

xetix:
Did a quite large overhaul of the ftp_download plugin, fixing couple of bugs and implemented some new stuff (See commit details). Code should be abit more robust now. 
Code is now also compliant to pep8 rules.


gazpachoking:
@xetix There are several changes to this plugin since this pull, can you update and open a new one against current develop if this is still relevant? (sorry)

PR #109, opened 2013-11-16
series_tasks config column and more emit_series search strings

Cyanogenoid:
Generate search strings for series with alternate names.
Use additional zero-padded episode numbers for series that are identified by sequence.


gazpachoking:
> Generate search strings for series with alternate names.
> Use additional zero-padded episode numbers for series that are identified by sequence.

I think we solved both of these with alternate pulls now.

PR #128, opened 2013-12-27
Inline email CSS before sending

KonstantinShkut:


stevezau:
@KonstantinShkut we are trying to clean up PR's. Can you please address @paranoidi last comment so we can merge or shall i just close this out?

PR #148, opened 2014-01-13
ffmpeg plugin

foux:
This plugin allows to encode entries with ffmpeg


stevezau:
@foux it seems this PR was missed.

Can you please confirm this plugin is still valid and working before we review the code?

I'm not sure how many use cases there are for this so if we don't hear back from you we will probabaly just close this PR

gazpachoking:
If this is still useful to anyone, please verify the code still works with latest flexget and open a new PR.

PR #191, opened 2014-01-30
Changed clean_symbols method to match on all non-word, non-digit symbols

Gargauth:
- publichd search plugin will use clean_title to remove non-word non-digit symbols from search queries 

Before this fix, release "Enders.Game.2013.etc.etc" wouldn't be found because search plugins will search for "Ender's Game" instead. (this should possibly do other plugins as well?)

### DIFF
--- a/flexget/plugins/search_publichd.py
+++ b/flexget/plugins/search_publichd.py
@@ -9,7 +9,7 @@ from flexget.event import event
 from flexget.config_schema import one_or_more
 from flexget.utils import requests
 from flexget.utils.soup import get_soup
-from flexget.utils.search import torrent_availability, normalize_unicode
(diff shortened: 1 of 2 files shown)


stevezau:
@gazpachoking this seems like it would be fairly easy to implement an optional flag to clean_title but it seems like it's a rare use case. Not sure if we should re-think this?

soloam:
this alteration plus the changes in #492 would clean all the special chars

stevezau:
@soloam thanks, i'll review that PR when i get a spare moment.

liiight:
@Gargauth is this still relevant?

liiight:
Please reopen if relevant

PR #213, opened 2014-02-11
move Plugin: 'to' to 'move_to'

tubedogg:
Per conversation in IRC: Change the ‘to’ option in the move plugin to ‘move_to’, and ensure that the plugin is looking in entry for that, not ‘path’.


stevezau:
@gazpachoking and @tubedogg 

I think this PR was missed. 

I agree with looking up move_to on the entry (Supporting jinja2) but this will need to be rebased to work with current master..

PR #224, opened 2014-02-25
Lookup for movies/series ratings on trakt

tarzasai:
Query trakt.tv for series and/or movies rated by the user. Set the
fields trakt_rating (love|hate) and trakt_rating_advanced (1..10) if the
user has actually rated the movie/show on trakt. Uses several fields
from metainfo plugins (series_name, tvdb_id or imdb_id for series,
title, tmdb_id or imdb_id for movies).

### DIFF
--- a/flexget/plugins/metainfo/trakt_ratings_lookup.py
+++ b/flexget/plugins/metainfo/trakt_ratings_lookup.py
@@ -109,8 +109,8 @@ class TraktWatched(object):
                 elif entry['title'] in movidx:
                     info = movies[movidx[entry['title']]]
             if info:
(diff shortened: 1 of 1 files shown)


gazpachoking:
I think we could look at this again now that the new api is out. Perhaps we could combine all the trakt_lookup plugins into the main trakt lookup plugin, and have the personal stuff be populated lazily if username/password is supplied.

tarzasai:
I'm sorry @gazpachoking, I've miss your last reply. Last time I checked (months ago) the api required for this code were still unavailable, so I'm not using trakt.tv anymore and I don't really want to work on this (sorry).

PR #232, opened 2014-03-01
Subtasks, take 2

gazpachoking:
The idea here is to let any plugin that allows entry generation via input subplugins, (discover and configure_series,) have a more general way to get their input, that works like a full task (with manipulation and filtering.)

Description of current implementation:
- Added 2 options for tasks `auto_accept` and `disable_builtins`
- Subtasks are just regular `Task`s initialized with different defaults:
  - They inherit db and requests session from their parent task
  - Both disable_builtins and auto_accept options are on by default
- To create them, call `parent_task.make_subtask`
- Name given to subtask will be appended to name of parent to create the full task name
- Plugins switched to use them:
  - crossmatch
  - discover
  - configure_series
  - if
  - inputs

The first try left several problems:
- [ ] Subtask names ended up really long and crappy in logs
- [ ] Warnings in subtasks about lack of filter and output plugins is not desired
- [x] emit_series totally went to shit

Possible TODOs/concerns:
- [x] ~~Make a new option for tasks, where every entry starts already accepted, rather than hacking in accept_all~~
Task now has auto_accept option, which subtasks turn on
- [x] ~~Right now, it makes and commits a new sqlalchemy session. Should it?~~
Added option to pass in sqlalchemy session to a task, task does not commit or close session when passed in
- [x] ~~requests session should maybe be transferred from main task?~~
Added option to init new Task with existing requests session
- [x] ~~discover plugin currently takes a list, rather than a plain dict. Change that, and just require use of `inputs` plugin for repeated input plugins? (we could even support backwards compatible configs if desired via inputs plugin)~~
Changed both discover and crossmatch to take a dict config that runs as a task.
- [x] ~~hrmmm... template/include plugin will not work in a subtask right now.~~ And.. I'm silly. They work fine.

### DIFF
--- a/flexget/entry.py
+++ b/flexget/entry.py
@@ -72,7 +72,7 @@ class Entry(dict):
         self.traces = []
         self.snapshots = {}
         self._state = 'undecided'
-        self._hooks = {'accept': [], 'reject': [], 'fail': [], 'complete': []}
+        self._hooks = {'accept': [], 'reject': [], 'fail': [], 'reset': [], 'complete': []}
         self.task = None
 
         if len(args) == 2:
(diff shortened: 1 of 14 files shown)


stevezau:
hey @gazpachoking.. task mater here.. what happened here?? :P

gazpachoking:
I still want this, but probably needs reworking again. I'll close PR for now until I get some motivation to revisit.

PR #244, opened 2014-03-15
Upcoming

thisirs:
Here is a new plugin: upcoming that outputs upcoming eps stored in tvrage db. To be used with html output plugin to produce a list of upcoming shows in your calendar... Tell what you think about it!


stevezau:
@thisirs I'm going through cleaning up PR's.

This plugin sounds cool and interesting but i'm wondering if it's still required?? Can't the same thing be achieved using trakt?

crawln45:
@stevezau yes but why not be able to do it locally if you so choose.

gazpachoking:
Outdated now, and tvrage is down. Open new PR if still relevant please.

PR #248, opened 2014-03-17
Update exists.py

Thermionix:
jinja expansion for paths in exists plugin


stevezau:
@Thermionix sorry for the late reply.

We are in the process of cleaning up PR's. 

Can you explain the use case for this??

PR #287, opened 2014-06-30
Log filter plugin

tarzasai:
Prevent entries with specific text from being logged.


stevezau:
@tarzasai is this still valid??

I really don't see the value in this plugin... log files are rotated and managed anyway.

tarzasai:
@stevezau you're right, I'm just lazy to look in big log files when I have to check why something has not been downloaded. I'm closing this.

stevezau:
@tarzasai i'm also writing a webui so hopefully that won't be a problem soon :)

It's a while off yet tho.

PR #328, opened 2014-09-05
Enable sqlite WAL to try to solve database locked problems.

gazpachoking:
There are several tickets involving crashes due to the database being locked. My theory is that it's when multiple threads are accessing the database at once. Hopefully enabling write ahead logging on the sqlite database will solve this. I'd like a little bit of testing for this change to see if it actually helps or causes any problems before I merge it in though. If anyone is getting errors like this:

`OperationalError: (OperationalError) database is locked`

If a few people could test this branch and give feedback it would be much appreciated.

### DIFF
--- a/flexget/manager.py
+++ b/flexget/manager.py
@@ -16,6 +16,7 @@ from datetime import datetime, timedelta
 import pkg_resources
 import sqlalchemy
 import yaml
+from sqlalchemy.engine import Engine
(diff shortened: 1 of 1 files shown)


antivirtel:
I had to enable this manually, since it was "delete". How is this possible, if I'm running the latest release?! (1.2.499)

PR #377, opened 2014-11-14
Fixed typo in tpb and alternate_name in emit_series

cvium:
This is my first pull request ever. I hope it works. I have implemented alternative_name handling in emit_series. 

Another solution that I toyed with is creating a new table "AlternativeNames", which has a many-to-one relationship with Series. Then the alternative names could be available for each show in SeriesTask obj. Or other variants of the same idea.

Right now I'm just fetching from series config.

I hope you like it and that it works (having problems running paver on windows)! At the very least I found a typo in urlrewrite_piratebay.

EDIT: ~~oops. Didn't mean to include my pushbullet changes. Guess I still have some stuff to learn about Git.~~ Fixed.

### DIFF
--- a/flexget/plugins/filter/series.py
+++ b/flexget/plugins/filter/series.py
@@ -230,7 +230,7 @@ class AlternateNames(Base):
     __tablename__ = 'series_alternate_names'
     id = Column(Integer, primary_key=True)
     _alt_name = Column('alt_name', Unicode)
-    _alt_name_normalized = Column('alt_name_lower', Unicode, index=True, unique=True)
+    _alt_name_normalized = Column('alt_name_normalized', Unicode, index=True, unique=True)
     series_id = Column(Integer, ForeignKey('series.id'), nullable=False)
 
     def name_setter(self, value):


flexget/plugins/filter/series.py
```diff
@@ -1435,12 +1476,26 @@ def on_task_start(self, task, config):
...
                     log.debug('adding series %s into db', series_name)
                     db_series = Series()
                     db_series.name = series_name
                     session.add(db_series)
+                    session.flush()  # flush to get id on series before creating alternate names
                     log.debug('-> added %s' % db_series)
+                    alts = series_config.get('alternate_name', [])
+                    if not isinstance(alts, list):
+                        alts = [alts]
+                    for alt in alts:
```
gazpachoking: I think this could just be untabbed once, so we don't have to repeat it in both the if and else sections.

flexget/plugins/filter/series.py
```diff
@@ -240,6 +274,8 @@ class Series(Base):
     episodes = relation('Episode', backref='series', cascade='all, delete, delete-orphan',
                         primaryjoin='Series.id == Episode.series_id')
     in_tasks = relation('SeriesTask', backref=backref('series', uselist=False), cascade='all, delete, delete-orphan')
+    alternate_names = relation('AlternateNames', backref='series', primaryjoin='Series.id == AlternateNames.series_id',
```
gazpachoking: Pretty sure we don't need the primaryjoin argument since there is one ForeignKey linking these tables, it should be detected automatically.

flexget/plugins/filter/series.py
```diff
@@ -1451,6 +1506,22 @@ def on_task_start(self, task, config):
                     except ValueError as e:
                         raise plugin.PluginError(e)
 
+def _add_alt_name(alt, db_series, series_name, session):
+    alt = unicode(alt)
+    db_series_alt = session.query(AlternateNames).join(Series).filter(AlternateNames.alt_name == alt).first()
```
gazpachoking: Don't think we need to join to Series table here, since we aren't filtering on anything from there.

gazpachoking:
Man, sorry I let this sit so long. I think it looks pretty good, just made a few more minor comments.

gazpachoking:
Oh, didn't realize we already merged. :tongue: Maybe these are notes for another cleanup pull.

cvium:
Ha. No problem. You can change it if you want or I can do it tomorrow. I can barely remember writing it though :p

gazpachoking:
Hehe. I may get around to it if you don't. Busy getting ready for a move tonight, and probably the rest of the week.

cvium:
I'll take a look at it tomorrow.

PR #389, opened 2014-11-23
Update subtitles_subliminal.py

camon:
added workaround for solving 'name_only = True' problem introduced by assignment to 'guessit.default_options' on flexget/plugins/parsers/parser_guessit.py

for more details on the bug check

http://discuss.flexget.com/t/flexget-subliminal-guessit/622/11


gazpachoking:
I think #412 fixes this a bit more cleanly. Thanks for looking at this though!

PR #392, opened 2014-12-02
imdb_plot_keywords

goto100:
add imdb_plot_keywords for imdb_lookup plugin

if:
      - "'xxx' in imdb_plot_keywords": reject


stevezau:
@goto100 any progress on this?? Feel join to join us on the IRC channel for help.

PR #394, opened 2014-12-08
subtitle queue

cvium:
I'm tired of having a separate program downloading subtitles (or scanning my files recursively), so I have created a very rough draft of a subtitle queue, which a subtitle plugin such as Subliminal can use as input.

This is the schema (`emit` is gonna be moved out of the dict, hence the `oneOf`):
```python
schema = {
        "oneOf": [
            {
                'type': 'object',
                'properties': {
                    'action': {'type': 'string', 'enum': ['add', 'remove', 'emit']},
                    'stop_after': {'type': 'string', 'format': 'interval'},
                    'languages': {"oneOf": [
                        {'type': 'array', 'items': {'type': 'string'}, 'minItems': 1},
                        {'type': 'string'},
                    ]},
                },
                'required': ['action'],
                'additionalProperties': False
            },
        ]
    }
```
Example tasks:
```yaml
populate_subtitle_queue:  # just run this once per session to keep queue uptodate
    seen: local
    find:
        path:
          - /media/TV Shows/
        regexp: '.*\.(avi|mkv|mp4)$'
    accept_all: yes
    regexp:
        reject:
          - '.*[sS]ample.*'
    subtitle_queue:
        action: add

torrents:
   discover:
        what:
            - emit_series:
                from_start: yes
        from:
            - piratebay: yes
        series:
           - "Arrow":
              begin: S02E01
   subtitle_queue:
        action: add
        languages:
           - eng
   set:
        path: /media/downloads
        movedone: "/media/TV Shows/{{series_name}}"
        content_filename: "{{ series_name }}"
   magnets: no

emit:
   subtitle_queue:
        action: emit
   subliminal:
        exact_match: yes
```
So what does this ugly code do?

The plugin has the ability to queue local files or torrents (does not support magnets). Add the contents of a torrent to the subtitle queue -- if torrent only contains 1 file, then that file is queued and if it contains more than 1 file, then the folder is queued instead.

What annoyed me most about something like this is that sorting tasks will mess with the queue. Therefore this plugin can be used in the same task as `move` and it will add the destination as an alternative filepath in the queue (same goes for torrents wrt. `path` and `movedone`).

`stop_after` is an interval that will drop queued items after the specified interval.

This is all still very rough around the edges and I'm not even sure if it works just yet, but it's at a stage where some third party input would be much appreciated.


gazpachoking:
@cvium Maybe call the field subtitles_missing, rather than missing_subtitles? We generally put the plugin name (or generic type) first, (series_X, movie_X, tvdb_X) with entry fields.

Other than that, how are you feeling about this?

gazpachoking:
@cvium Once you are feeling it's ready you can go ahead and merge, or put line comments about stuff you are wondering. It's a lot for me to parse at once. (But stick around for future bug reports :smile:)

cvium:
I think it's close to being ready for an open beta. People should be made aware that this is very untested :P there are many edge cases that are impossible to write unit tests for.

There are some small changes I want to make first though. I also have one question related to queueing in general. If an item is considered done, should it be removed from the queue or should its download flag simply be set to true?

cvium:
@gazpachoking I think it's close to being ready now. Just one last issue. If a task queues a bunch of files, then those entries will be marked as `seen` and hence the emitting task will reject those entries.

Is this just how it's supposed to work ie. users must set `seen: local` in the queueing task or should  I handle this somehow?

PR #397, opened 2014-12-13
Xbmc library scan/clean (experimental)

cvium:
I'm not sure this plugin should be a part of Flexget, but I have tried XBMC Watchdog and XBMC Library Auto Update addons and they both suck in some ways, so I think it would be nice to let Flexget handle it instead.

What I want is some automatic library scanning whenever new content is added to my library folders without having to re-scan the entire library. This plugin should be used with the `move` plugin or `find` etc. It can also be used in a task of its own, although Flexget will warn the user of missing filter plugins etc (and rightly so).

`dir` supports jinja. The JSONRPC API of XBMC is quite extensive. Plugin currently supports scan and clean of audio and video libraries. It may or may not support all the API functions in the future although I'm not sure I see any use for it ([Milhouse's TextureCache](https://github.com/MilhouseVH/texturecache.py) is much better for this anyway).

It's rather niche but I fully intend to use this plugin even if it does not get merged.

### DIFF
--- a/flexget/plugins/output/xbmc_library.py
+++ b/flexget/plugins/output/xbmc_library.py
@@ -22,7 +22,7 @@ class XBMCLibrary(object):
             {
                 'type': 'object',
                 'properties': {
-                    'action': {'type': 'string', 'enum': ['clean', 'scan', 'full_scan']},
+                    'action': {'type': 'string', 'enum': ['clean', 'scan', 'dir_scan']},
                     'category': {'type': 'string', 'enum': ['audio', 'video']},
                     'url': {'type': 'string'},
                     'port': {'type': 'integer'},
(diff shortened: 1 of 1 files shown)


cvium:
@gazpachoking I see you are trying to get "rid" of some of the pull requests. I'm just gonna close this one. It worked when I wrote it and I believe it still works in Kodi, but the limitations of the Kodi API is too annoying.

Kodi cannot scan a specific folder for new tv episodes unless the tv show already exists in the Kodi database.

gazpachoking:
@cvium Yeah, sorry this sat for so long without comment. And we aren't trying to get 'rid' of PRs, just get them down to a manageable level, so that it's easier to deal with new PRs as they come in, and not let them get lost in our backlog (like this did.)

cvium:
> And we aren't trying to get 'rid' of PRs, just get them down to a manageable level

I'm well aware. I was only joking :stuck_out_tongue:

PR #401, opened 2014-12-20
Update plugin_transmission.py to fix bug #2804

grrr2:
Try to fix bug #2804 by making RPC client object re-initiated for every task rather then doing it once for whole plugin.


gazpachoking:
Oops, sorry, didn't see you had made the change. Thanks!

PR #406, opened 2014-12-28
Change how emit_series tries on reruns. fix #2826

gazpachoking:
This prevents retries on a lot of series that shouldn't have been retried. Need to figure out a way to test some of the behavior though, as the old unit testing method doesn't work for a couple things though.


gazpachoking:
@paranoidi Might want to double check your emit_series tasks against this again, I think it's good stuff though.

PR #409, opened 2015-01-01
Update trakt plugins to the v2 api

gazpachoking:
- [x] trakt_list
- [x] trakt_add/trakt_remove
- [ ] trakt_emit:
    This is working, however it skips to next season too early when there are unaired episodes left in the current season.
- [ ] trakt_lookup


gazpachoking:
I think I'm going to push this now, even though it could use more work, as the plugins which have been updated are completely broken at the moment.

PR #411, opened 2015-01-03
fixed attribute error when downloading subtitles for movies

cvium:
Subliminal plugin would throw an exception when trying to download subtitles for movies as the movie object does not have the `series` attribute.


gazpachoking:
Thanks!

PR #412, opened 2015-01-03
ParserGuessit changes

cvium:
This should fix http://flexget.com/ticket/2775 and it was also discussed in https://github.com/Flexget/Flexget/pull/389

I am creating this pull request as the other one hasn't been changed in a while. I think it's time fix this, so the subliminal plugin can work correctly.


gazpachoking:
Cool, thanks!

PR #413, opened 2015-01-05
Switch simple persistence to load/flush from/to db outside task runs fix #2820, #2672, #2835

gazpachoking:
Simple persistence had an issue causing db locks when plugins were calling it while doing their own database read/writes. This branch changes the behavior so that the data is read from the database before a task is run, and written after.

http://flexget.com/ticket/2762
http://flexget.com/ticket/2820

### DIFF
--- a/flexget/manager.py
+++ b/flexget/manager.py
@@ -846,7 +846,7 @@ class Manager(object):
             log.info('Running database cleanup.')
             session = Session()
             try:
-                fire_event('manager.db_cleanup', session)
+                fire_event('manager.db_cleanup', self, session)
                 session.commit()
             finally:
                 session.close()
--- a/flexget/plugins/api_tvrage.py
+++ b/flexget/plugins/api_tvrage.py
@@ -28,7 +28,7 @@ tvrage.feeds.BASE_URL = 'http://services.tvrage.com/feeds/%s.php?%s=%s'
 
 
 @event('manager.db_cleanup')
-def db_cleanup(session):
+def db_cleanup(manager, session):
     value = datetime.datetime.now() - parse_timedelta('30 days')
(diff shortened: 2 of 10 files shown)


MikeyCanuck:
Using a config similar to the cookbook here: http://flexget.com/wiki/Cookbook/Series/DelugeThetvdbSeries

The issue appeared to be related to the tvdb_ep_name in the content_filename config section.

    set:
      path: /media/incomplete
      movedone: "/media/tv/{{ series_name }}/Season {{ series_season }}"
      content_filename: |
        {{ series_name }} - {{ series_id }}
        {% if tvdb_ep_name|default(False) %} - {{ tvdb_ep_name }} {% endif %} - {{ quality|upper }}
        {% if proper_count %} - proper{% if proper_count > 1 %}{{ proper_count }}{% endif %}{% endif %}

With the tvdb_ep_name if statement removed, all functioned normally, otherwise an error similar to the following was received.

2015-01-04 18:18 ERROR    set           FEED    Could not set content_filename for Example.Show.S05E10.HDTV.torrent: (OperationalError) (OperationalError) database is locked u'INSERT INTO simple_persistence (feed, plugin, "key", value, added) VALUES (?, ?, ?, ?, ?)' (None, u'api_tvdb', u'last_local', <read-only buffer for 0x7f34d83f1120, size -1, offset 0 at 0x7f34d83fc2f0>, '2015-01-04 18:17:22.714868')

The issue was repeatable with a --test run and clean database:

~/flexget-dev/bin/flexget -c ~/config.yml --test execute

Once updated with the simple_per_update branch the error was no longer received.

gazpachoking:
@paranoidi Is your simple_persistence table small if you exclude old tasks? This branch now cleans those up, and only loads into memory when a task is first run.

PR #414, opened 2015-01-05
Search for plugins in config_base

Ram-Z:
instead of the hardcoded ~/.flexget/plugins. This effectively allows to
put plugins in $XDG_CONFIG_HOME/flexget/plugins.

### DIFF
--- a/flexget/plugin.py
+++ b/flexget/plugin.py
@@ -412,12 +412,15 @@ def _load_plugins_from_dirs(dirs):
                       'point (before, after). Plugin is not working properly.' % (args[0], phase))
 
 
-def load_plugins(extra_dirs=[]):
+def load_plugins(extra_dirs=None):
     """
     Load plugins from the standard plugin paths.
     :param list extra_dirs: Extra directories from where plugins are loaded.
     """
     global plugins_loaded
+
+    if not extra_dirs:
+        extra_dirs = []
 
     # Add flexget.plugins directory (core plugins)
     extra_dirs.extend(_get_standard_plugins_path())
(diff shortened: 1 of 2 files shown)


flexget/plugin.py
```diff
@@ -356,14 +356,10 @@ def _get_standard_plugins_path():
     :returns: List of directories where plugins should be tried to load from.
     """
     # Get basic path from environment
+    paths = []
     env_path = os.environ.get('FLEXGET_PLUGIN_PATH')
     if env_path:
-        # Get rid of trailing slashes, since Python can't handle them when
-        # it tries to import modules.
```
gazpachoking: Does this comment not apply anymore? Does it break when the env path contains a trailing slash?

flexget/plugin.py
```diff
@@ -416,13 +412,19 @@ def _load_plugins_from_dirs(dirs):
                       'point (before, after). Plugin is not working properly.' % (args[0], phase))
 
 
-def load_plugins():
-    """Load plugins from the standard plugin paths."""
+def load_plugins(extra_dirs=[]):
```
gazpachoking: We can't use a mutable default argument (a list is mutable), as it won't reset if the function is called more than once. Usually, you'd set the default to `None`, then create your list in the body of the function.

flexget/plugin.py
Ram-Z: No, I removed the stripping in this function. The trailing slashes were being removed again further down [L377](https://github.com/Ram-Z/Flexget/blob/plugin-path/flexget/plugin.py#L377)

gazpachoking: Gotcha.

gazpachoking:
Cool, thanks! Perhaps we should put a little note about this in upgrade actions since it might affect a few people.

PR #416, opened 2015-01-06
Update scheduler.py

programatix:
Catch all type of exception instead of just pytz.UnknownTimeZoneError

### DIFF
--- a/flexget/plugins/daemon/scheduler.py
+++ b/flexget/plugins/daemon/scheduler.py
@@ -111,7 +111,7 @@ def setup_scheduler(manager):
             timezone = None
     except pytz.UnknownTimeZoneError:
         timezone = None
-    (except struct.error as e):
+    except struct.error as e:
         # Hiding exception that may occur in tzfile.py seen in entware
         log.warning('Hiding exception from tzlocal: %s', e)
         timezone = None
--- a/flexget/plugins/daemon/scheduler.py
+++ b/flexget/plugins/daemon/scheduler.py
@@ -109,7 +109,11 @@ def setup_scheduler(manager):
         timezone = tzlocal.get_localzone()
         if timezone.zone == 'local':
             timezone = None
-    except (pytz.UnknownTimeZoneError, struct.error):
+    except pytz.UnknownTimeZoneError:
+        timezone = None
+    except struct.error, e:
+        # Hiding exception that may occur in tzfile.py seen in entware
+        log.warning('Hiding exception from tzlocal: %s', e)
         timezone = None
     if not timezone:
         # The default sqlalchemy jobstore does not work when there isn't a name for the local timezone.
--- a/flexget/plugins/daemon/scheduler.py
+++ b/flexget/plugins/daemon/scheduler.py
@@ -4,6 +4,7 @@ import logging
 
 import pytz
 import tzlocal
+import struct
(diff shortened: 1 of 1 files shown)


gazpachoking:
What other exceptions are you getting? Bare excepts are not good practice.

flexget/plugins/daemon/scheduler.py
```diff
@@ -108,7 +108,7 @@ def setup_scheduler(manager):
         timezone = tzlocal.get_localzone()
         if timezone.zone == 'local':
             timezone = None
-    except pytz.UnknownTimeZoneError:
+    except:
```
paranoidi: Too wide exception catch, this is not good python. It will catch everything under the sun including aborting program with ctrl-c. What exception did you encounter? Add it into the list instead. eg

    except (pytz.UnknownTimeZoneError, SomethingElse):

programatix:
I'm getting the following error,

  File "/opt/bin/flexget", line 9, in <module>
    load_entry_point('FlexGet==1.2.253', 'console_scripts', 'flexget')()
  File "/opt/lib/python2.7/site-packages/flexget/__init__.py", line 36, in main
    manager.start()
  File "/opt/lib/python2.7/site-packages/flexget/manager.py", line 300, in start
    self.handle_cli()
  File "/opt/lib/python2.7/site-packages/flexget/manager.py", line 325, in handl                                                      e_cli
    self.daemon_command(options)
  File "/opt/lib/python2.7/site-packages/flexget/manager.py", line 387, in daemo                                                      n_command
    fire_event('manager.daemon.started', self)
  File "/opt/lib/python2.7/site-packages/flexget/event.py", line 103, in fire_ev                                                      ent
    result = event(*args, **kwargs)
  File "/opt/lib/python2.7/site-packages/flexget/event.py", line 22, in __call__
  File "/opt/lib/python2.7/site-packages/flexget/plugins/daemon/scheduler.py", l                                                      ine 108, in setup_scheduler
    timezone = tzlocal.get_localzone()
  File "/opt/lib/python2.7/site-packages/tzlocal/unix.py", line 108, in get_loca                                                      lzone
    _cache_tz = _get_localzone()
  File "/opt/lib/python2.7/site-packages/tzlocal/unix.py", line 100, in _get_loc                                                      alzone
    return pytz.tzfile.build_tzinfo('local', tzfile)
  File "/opt/lib/python2.7/site-packages/pytz/tzfile.py", line 30, in build_tzin                                                      fo
    typecnt, charcnt) =  unpack(head_fmt, fp.read(head_size))
struct.error: unpack requires a string argument of length 44

paranoidi:
Extremely weird, that looks like a problem with library providing tzlocal module. We should report it forward .. not hide and forget. Supressing it may be good idea while it gets fixed, but then it should be in a separate except catch and log some type of warning.

flexget/plugins/daemon/scheduler.py
```diff
@@ -110,6 +111,10 @@ def setup_scheduler(manager):
             timezone = None
     except pytz.UnknownTimeZoneError:
         timezone = None
+    except struct.error, e:
```
gazpachoking: Can we use the newer format for exceptions? (`except struct.error as e`)

paranoidi:
Looks good, thanks for contributing.

programatix:
Thank you all for the guidance and patient.


Sent from my Samsung device

-------- Original message --------

PR #418, opened 2015-01-07
Ratings support for trakt_list

brtkrbzhnv:
Support for getting all shows/episodes/movies with a particular rating.

### DIFF
--- a/flexget/plugins/input/trakt_list.py
+++ b/flexget/plugins/input/trakt_list.py
@@ -58,7 +58,7 @@ class TraktList(object):
       username: <value>
       password: <value>
       type: <shows|movies|episodes>
-      list: <collection|watchlist|watched|custom list name>
+      list: <collection|watchlist|watched|rating:<1|2|3|4|5|6|7|8|9|10>|custom list name>
       strip_dates: <yes|no>
 
     Options username, type and list are required. password is required for private lists.
@@ -70,7 +70,10 @@ class TraktList(object):
             'username': {'type': 'string'},
             'password': {'type': 'string'},
             'type': {'type': 'string', 'enum': ['shows', 'movies', 'episodes']},
-            'list': {'type': 'string'},
+            'list': {"oneOf": [
+              {'type': 'string'}, 
+              {"type": "object", "properties": {"rating": {"type": "integer", 'enum': [1, 2, 3, 4, 5, 6, 7, 8, 9, 10]}}
+            }]},
             'strip_dates': {'type': 'boolean', 'default': False}
         },
         'required': ['username', 'type', 'list'],
@@ -88,7 +91,9 @@ class TraktList(object):
     def on_task_input(self, task, config):
         session = get_session(config['username'], config.get('password'))
         endpoint = ['users', config['username']]
-        if config['list'] in ['collection', 'watchlist', 'watched']:
+        if type(config['list']) is dict:
+            endpoint += ('ratings', config['type'], config['list']['rating'])
+        elif config['list'] in ['collection', 'watchlist', 'watched']:
             endpoint += (config['list'], config['type'])
         else:
             endpoint += ('lists', make_list_slug(config['list']), 'items')
--- a/flexget/plugins/input/trakt_list.py
+++ b/flexget/plugins/input/trakt_list.py
@@ -59,7 +59,7 @@ class TraktList(object):
       username: <value>
       password: <value>
       type: <shows|movies|episodes>
-      list: <collection|watchlist|watched|custom list name>
+      list: <collection|watchlist|watched|ratings|rating:<1|2|3|4|5|6|7|8|9|10>|custom list name>
       strip_dates: <yes|no>
 
     Options username, type and list are required. password is required for private lists.
@@ -94,7 +94,7 @@ class TraktList(object):
         endpoint = ['users', config['username']]
         if type(config['list']) is dict:
             endpoint += ('ratings', config['type'], config['list']['rating'])
-        elif config['list'] in ['collection', 'watchlist', 'watched']:
+        elif config['list'] in ['collection', 'watchlist', 'watched', 'ratings']:
             endpoint += (config['list'], config['type'])
         else:
             endpoint += ('lists', make_list_slug(config['list']), 'items')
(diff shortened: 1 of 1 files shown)


gazpachoking:
Should we allow for the all ratings list too? Maybe just make that another text option 'ratings'

brtkrbzhnv:
Why not, I guess. I wouldn't use it, but maybe someone else would. Anyway, it would be great if we could get this merged.

stevezau:
@gazpachoking any reason you didn't merge this? seems ok to me?

snickers2k:
@brtkrbzhnv this looks great 
@gazpachoking please merge :)

snickers2k:
okay.. i just tested this and becoming the following error:

2015-09-01 02:50 CRITICAL manager                       mapping values are not allowed here in "<unicode string>", line 230, column 19: list: rating: 10 ^

OmgOhnoes:
I've been playing around with this and would it not be better to make it a separate setting instead of using the 'list' setting?

I'm not familiar with the trakt api (i tried but the documentation is lacklustre) but i would think it would be possible.

People use lists to manage their various shows, it would provide a lot of extra functionality to be able to filter by list AND rating instead of your global ratings.

For instance, i like a lot of shows and rate them 9, but maybe i only want the list of shows in X list that are rated 9 so i can set the quality to a higher level than shows rated 9 in other lists.

flexget/plugins/input/trakt_list.py
```diff
@@ -20,31 +20,32 @@
...
-        # TODO: Should this have series_name? We normally only have that for episodes
+        'title': lambda i: '%s (%s)' % (i['show']['title'], i['show']['year']),
+        'series_name': lambda i: '%s (%s)' % (i['show']['title'], i['show']['year']),
         'imdb_id': 'show.ids.imdb',
         'tvdb_id': 'show.ids.tvdb',
         'tvrage_id': 'show.ids.tvrage',
         'tmdb_id': 'show.ids.tmdb',
-        'trakt_id': 'show.ids.trakt',
+        'trakt_show_id': 'show.ids.trakt',
         'trakt_slug': 'show.ids.slug'
```
OmgOhnoes: Could we add ```'trakt_user_rating': 'rating'``` to this? That way we can filter lists by rating via crossmatch.

flexget/plugins/input/trakt_list.py
```diff
@@ -20,31 +20,32 @@
...
-        'tvdb_id': 'episode.ids.tvdb',
-        'tvrage_id': 'episode.ids.tvrage',
-        'trakt_id': 'show.ids.trakt',
-        'trakt_slug': 'show.ids.slug'
+        'imdb_id': 'show.ids.imdb',
+        'tvdb_id': 'show.ids.tvdb',
+        'tvrage_id': 'show.ids.tvrage',
+        'trakt_episode_id': 'episode.ids.trakt',
+        'trakt_show_id': 'show.ids.trakt',
+        'trakt_show_slug': 'show.ids.slug'
```
OmgOhnoes: Same thing here: ```'trakt_user_ep_rating': 'rating'```

snickers2k:
@OmgOhnoes great thoughts

Can somebody please take a quick look? I think it's almost done.. just needs some adjustments?

gazpachoking:
Merged in c7251688e6c8779dac5cbca2b34a13bc7338dbe4 @brtkrbzhnv Still around and want to update the wiki?

snickers2k:
like i said, it's not working.
when using

list: rating: 10

following error: mapping values are not allowed here in "<unicode string>", line 419, column 19: list: rating: 10 ^

when using 

list: ratings
rating: 10

"rating" is not an allowed value or something...

gazpachoking:
I think it should be:
```
list:
  rating: 10
```

snickers2k:
Hey, it's been quite here for a while. Is this still working? Because i can't find it in the wiki and it seems to be not in the trakt_list.py anymore?!

Could you please add some ability to interact with ratings, no matter for which database (imdb/trakt/tvdb)? I tried it with plex_userrating, but since plex isn't supported with Plex-Home it's not possible for me.
It could be really awesome to rate a movie with 10 Stars to load it in a higher quality or delete it with 1 Stars or something... 

thanks

gazpachoking:
It looks like we lost this feature when updating this to the new list style plugins. I think adding user rating support to trakt_lookup would be pretty nice, potentially nicer than being able to get lists of rated things with trakt_list.

snickers2k:
i would be fine with any solution :) but of course, trakt_lookup would be the better way.

cvium:
Should be available in the next release.

snickers2k:
this is awesome! thank you so much!

snickers2k:
is this already implemented? because it's not working for me.

>     if:
>      - trakt_movie_user_rating > '9': accept

cvium:
You're going to have to be more specific

snickers2k:
with this task:

`
  Update_Movie_Favorites:
    priority: 1
    template:
      - no_seen
      - no_global
    trakt_lookup: yes
    trakt_list:
      account: '{{ secrets.trakt.acc }}'
      list: collection
      type: movies
    crossmatch:
      from:
        - trakt_list:
            account: '{{ secrets.trakt.acc }}'
            list: '{{ secrets.trakt.Favorite_List }}'
            type: movies
      fields:
        - imdb_id
      action: reject
    if:
      - trakt_movie_user_rating > 9: accept
    list_add:
      - trakt_list:
          account: '{{ secrets.trakt.acc }}'
          list: '{{ secrets.trakt.Favorite_List }}'
          type: movies
`

it should accept a few dozens of movies rated above 9.. but instead

`2016-11-06 20:39 VERBOSE  details       Update_Movie_Favorites Summary - Accepted: 0 (Rejected: 0 Undecided: 1035 Failed: 0)
` 

the other way around; accept_all with "if below 9 -> reject" just accepts everything.

cvium:
Use triple backticks for code blocks
\`\`\`
test
  test2
\`\`\`

```
test
  test2
```

Anyway, you're using it wrong. `trakt_lookup: yes` can't get any user specific data. You need to specify account.

snickers2k:
oh, i overlooked that. thanks.

PR #419, opened 2015-01-08
search_cpasbien modified to use cloudflare-scrape

xfouloux:
As cpasbien switched to cloudflare, I had to modify the code.
So now, you have to have https://github.com/Anorov/cloudflare-scrape and all it's dependency to make it work.
Also, the file cfscrape.py must be put in the folder : /var/lib/flexget

But won't work until the download plugin use scrape too.

### DIFF
--- a/flexget/plugins/search_cpasbien.py
+++ b/flexget/plugins/search_cpasbien.py
@@ -1,4 +1,9 @@
 from __future__ import unicode_literals, division, absolute_import
+
+import sys
+sys.path.append("/var/lib/flexget")
+import cfscrape
+
 import logging
 import re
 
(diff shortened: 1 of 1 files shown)


gazpachoking:
> But won't work until the download plugin use scrape too.

I think that library works as a requests adapter. What we could do is make a cloudflare plugin, which when included in a task installed the cloudflare-scrape adapter on the task.requests session. Then the plugin doesn't have any tie to search_cpasbien, and could be used in the config for any site.

I guess we'd also want a way so that if a plugin that requires it was used, (like search_cpasbien,) it was automatically enabled.

stevezau:
@gazpachoking I agree this could be layered over requests..

@xfouloux have you made any progress on this?

xfouloux:
@stevezau I know that https://github.com/Anorov/cloudflare-scrape is now installable with PIP so it is better than before.

@gazpachoking I agree totally but my skills are limited in python, and I would not be able to do this.

cvium:
I have already made a cloudflare scrape plugin. Can be enabled in a task with "cfscraper: yes".

stevezau:
Closing as the same functionality is available via http://flexget.com/wiki/Plugins/cfscraper

cvium:
Should probably update the wiki so people know that cfscraper must be enabled.

PR #420, opened 2015-01-09
Scan for files and dirs, and allow different qualities.

Ram-Z:


flexget/plugins/filter/exists_movie.py
```diff
@@ -12,9 +11,7 @@
 
 log = logging.getLogger('exists_movie')
 
-
```
paranoidi: This belongs here, see PEP8

flexget/plugins/filter/exists_movie.py
```diff
@@ -23,17 +20,40 @@ class FilterExistsMovie(object):
...
+    schema = {
+        'anyOf': [
+            one_or_more({'type': 'string', 'format': 'path'}),
+            {
+                'type': 'object',
+                'properties': {
+                    'path': one_or_more({'type': 'string', 'format': 'path'}),
+                    'allow_different_qualities': {'enum': ['better', True, False], 'default': False},
+                    'type': {'enum': ['files', 'dirs'], 'default': 'dirs'},
+                    'imdb_lookup': {'type': 'boolean', 'default': False}
```
paranoidi: Perhaps change this to "lookup: imdb" so it will be forward compatible if we later add support for other lookup facilities?

flexget/plugins/filter/exists_movie.py
```diff
@@ -74,54 +94,76 @@ def on_task_filter(self, task, config):
...
-
-                try:
-                    imdb_id = imdb_lookup.imdb_id_lookup(movie_title=movie.name,
-                                                         raw_title=item.name,
-                                                         session=task.session)
-                    if imdb_id in path_ids:
-                        log.trace('duplicate %s' % item)
+            for root, dirs, files in os.walk(folder):
+                for item in eval(config['type']):
+                    if config['type'] == 'files' and not re.search(self.pattern, item, re.IGNORECASE):
```
paranoidi: Since this is in the loop the self.pattern should be precompiled (a lot faster when using same regexp many times).

flexget/plugins/filter/exists_movie.py
```diff
@@ -23,17 +20,40 @@ class FilterExistsMovie(object):
...
-    schema = one_or_more({'type': 'string', 'format': 'path'})
+    schema = {
+        'anyOf': [
+            one_or_more({'type': 'string', 'format': 'path'}),
+            {
+                'type': 'object',
+                'properties': {
+                    'path': one_or_more({'type': 'string', 'format': 'path'}),
+                    'allow_different_qualities': {'enum': ['better', True, False], 'default': False},
+                    'type': {'enum': ['files', 'dirs'], 'default': 'dirs'},
```
gazpachoking: We should put the 'type' in these, it helps give better error messages.

flexget/plugins/filter/exists_movie.py
```diff
@@ -74,54 +94,76 @@ def on_task_filter(self, task, config):
...
-                movie = get_plugin_by_name('parsing').instance.parse_movie(item)
-
-                try:
-                    imdb_id = imdb_lookup.imdb_id_lookup(movie_title=movie.name,
-                                                         raw_title=item.name,
-                                                         session=task.session)
-                    if imdb_id in path_ids:
-                        log.trace('duplicate %s' % item)
+            for root, dirs, files in os.walk(folder):
+                for item in eval(config['type']):
```
gazpachoking: What exactly is this doing? I don't think we need/should be using eval. Also, probably nicer to stick to path.py interface, so we have a constant way to deal with decoding.

Ram-Z: I switched to `os.walk()` because it iterates over both files and dirs in the same function instead of using `path.walkfiles()` and `path.walkdirs()`. Combined with the 'eval(config['type'])` to choose what variable `dirs/files` to iterate over, it seemed like a good way to avoid repetition.

However, when adding the patterns to skip (which are different for files and dirs) it wasn't so nice anymore. So I was planning on rewriting this anyway.

gazpachoking: You can use http://pythonhosted.org/path.py/api.html#path.Path.walk to walk files and dirs at the same time as needed, but since it's one or the other in the config, I'm not sure why we'd need to do that.

Ram-Z:
Just realized that one might want to search some paths for dirs and other paths for files, this is currently not supported. I'm thinking we could add the type to path instead:
```yaml
exists_movie:
  path:
    files:
      - /path/to/folder
    dirs:
      - /path/to/other/folder
```

gazpachoking:
Thanks!

PR #421, opened 2015-01-09
Encoding of path when content_filename is set

Bop4yN:
Attempt to fix ticket http://flexget.com/ticket/2824

### DIFF
--- a/flexget/plugins/plugin_transmission.py
+++ b/flexget/plugins/plugin_transmission.py
@@ -486,6 +486,7 @@ class PluginTransmission(TransmissionBase):
                                 download_dir = cli.get_session().download_dir
                             else:
                                 download_dir = options['add']['download_dir']
+                                download_dir = download_dir.encode('utf-8')
 
                             # Get new filename without ext
                             file_ext = os.path.splitext(fl[r.id][main_id]['name'])[1]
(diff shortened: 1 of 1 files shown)


gazpachoking:
Hmm, shouldn't that already be happening here? https://github.com/Flexget/Flexget/blob/develop/flexget/utils/titles/series.py#L20

Bop4yN:
Not sure how the loglevel can affect the problem. 
In any case, the latest update of series.py was October 27, and the problem is still there.

gazpachoking:
Man, that is not at all the line I meant to link. This is where I meant:
https://github.com/Flexget/Flexget/blob/develop/flexget/plugins/plugin_transmission.py#L326

Bop4yN:
That's more like it.:)
I think so too. Moreover, I have already written about it in http://flexget.com/ticket/2824
But somehow download_dir breaks in the line 492
https://github.com/Flexget/Flexget/blob/develop/flexget/plugins/plugin_transmission.py#L492
From error log:
File "/usr/local/lib/python2.7/dist-packages/flexget/plugins/plugin_transmission.py", line 487, in add_to_transmission
        file_path = os.path.dirname(os.path.join(download_dir, fl[r.id][main_id]['name']))

Bop4yN:
Hmmm... may be here is the problem
https://github.com/Flexget/Flexget/blob/develop/flexget/plugins/plugin_transmission.py#L438

And must be
fl = cli.get_files(r.id).encode('utf-8')
isn't it?

stevezau:
hey @Bop4yN  I assume this is still a problem?? can you describe the problem again with fresh logs and ill take a look? I am not clear on what the actual issue is and the cause.

gazpachoking:
This isn't quite the right fix. If somebody has more details please open up a new Github issue.

PR #424, opened 2015-01-10
trakt_emit: fixed parsing error for getting show data

grasdal:
See example of Trakt.tv json response data from
http://docs.trakt.apiary.io/#reference/users/watched/get-watched

Previous Flexget code was looking for a "type"-attribute that is no longer available in the response.

### DIFF
--- a/flexget/plugins/input/trakt_emit.py
+++ b/flexget/plugins/input/trakt_emit.py
@@ -61,7 +61,7 @@ class TraktEmit(object):
                 log.warning('The list "%s" is empty.' % config['list'])
                 return
             for item in data:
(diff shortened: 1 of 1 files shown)


grasdal:
Example of task I used that resulted in the parsing issue:

  trakt_series_begin:
    trakt_emit:
      username: <username>
      password: <password>
      context: watched
      list: watched
    accept_all: yes
    set_series_begin: yes

gazpachoking:
I wish trakt would fix this. We are supposed to get type for all items I believe, but it isn't working for watched or collected list. My only concern with this fix, is does it now count episodes in the list as shows?

PR #425, opened 2015-01-10
handle pluginerrors in configure_series

cvium:
Right now, configure_series will fail the task if one of the inputs fail. Is this the intended behaviour? The discover plugin will simply catch the plugin error exceptions and move on to the next input.

I wanted to hear from you, @gazpachoking, before just pushing random changes that conflict :) I feel it is an inconsistency though.


gazpachoking:
Sounds pretty good to me, thanks!

PR #430, opened 2015-01-13
Refactor set plugin to render templates lazily.

gazpachoking:
This changes the set plugin to render entries lazily. This has several advantages:
- We can do the set earlier, so that its fields can be used throughout the filter phase. (also less chance of people needing to use plugin_priority for advanced setups)
- Fields get set on all entries, instead of just accepted ones, so they can be used for filtering (or other plugins that act on non-accepted entries) if desired.
- Less chance of causing lookups that aren't needed, if the fields never get accessed.
- We can now reference other fields set by the set plugin, no matter the order.

@paranoidi Thoughts on this? I think the outcome is awesome, perhaps there are better ways to implement though, or unintended consequences.

### DIFF
--- a/flexget/entry.py
+++ b/flexget/entry.py
@@ -218,7 +218,7 @@ class Entry(dict):
 
         # url and original_url handling
         if key == 'url':
-            if not isinstance(value, basestring):
+            if not isinstance(value, (basestring, LazyField)):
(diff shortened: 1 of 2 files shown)


paranoidi:
Can't think any reason why not to do this, and it does look awesome :)

PR #431, opened 2015-01-14
separate authentication and regular http errors in trakt

cvium:
Right now the trakt error messages do not quite make sense. For instance, 500 error codes are server errors and have nothing to do with authentication. It seems to confuse some people.


gazpachoking:
Good call.

PR #432, opened 2015-01-15
Make auto_escape also work with spaces and single quotes

matrixagent:
I'm not sure if it might have already worked for Windows, so I didn't change the current behavior for that platform, but in Unix-based systems the option didn't work at all. Spaces were not escaped and single quotes were not escaped. There might be even more things that need escaping, but so far I've only ever come across spaces and single quotes. So far I've fixed it myself by jinja-replacing them ( ```{{ location|replace("'", "\\'")|replace(" ", "\\ ") }}``` ), and now that I haven't encountered any troubles with that in a while I figured it should be a worthwhile fix in the actual codebase.

### DIFF
--- a/flexget/plugins/output/exec.py
+++ b/flexget/plugins/output/exec.py
@@ -27,8 +27,12 @@ class EscapingDict(MutableMapping):
     def __getitem__(self, key):
         value = self._data[key]
         if isinstance(value, basestring):
-            # TODO: May need to be different depending on OS
-            value = value.replace('"', '\\"')
+            if sys.platform.startswith('win'):
+                value = value.replace('"', '\\"')
+            else:
+                value = value.replace('"', '\\"')
+                value = value.replace('\'', '\\\'')
+                value = value.replace(' ', '\\ ')
         return value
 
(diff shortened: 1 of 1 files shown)


gazpachoking:
Hmm, so the idea I had with auto escape was so that you could do something like `"{{title}} {{otherthing}}"` and even if the title or otherthing had quotes in them, it would work. Is your plan so that you don't even have to use quotes in the template string? That gets a bit more complicated, especially when you consider that different shells require different escaping.

matrixagent:
Well, apparently I'm an idiot. I almost can't believe I actually never tried adding double quotes around the jinja variables in my config during all those hours I spent trying to get it to work, but I just tested it to be certain and of course it did work with cases that caused problems earlier. I guess I was thrown off by the example config not having any double quotes but instead escaping spaces. I'll edit the wiki to clarify the usage and then we maybe can agree to never speak of this pull request again… ;)

gazpachoking:
@matrixagent Fixed up our documentation so that it is clear in the wiki how auto escape is supposed to work.

PR #435, opened 2015-01-18
Disabled.py crash when interval not met.

Lugoues:


gazpachoking:
Good catch. Perhaps it should be `None` though?

Lugoues:
Good point, updated it to None.

gazpachoking:
Thanks!

PR #436, opened 2015-01-24
Bootstrap update

Toilal:
I hit this bug when trying to run bootstrap on Ubuntu 14.04

https://bugs.launchpad.net/ubuntu/+source/python-pip/+bug/1306991

It's easily solved by running `sudo pip install setuptools pip --upgrade` to upgrade pip and setuptools in system python.

So maybe you could test on your system before merging.

### DIFF
--- a/bootstrap.py
+++ b/bootstrap.py
@@ -4,7 +4,7 @@
 """Create a "virtual" Python installation
 """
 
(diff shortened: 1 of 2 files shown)


gazpachoking:
Thanks!

PR #437, opened 2015-01-27
Added name/year id lookup

crawln45:
This should add the name/year search for the id's

### DIFF
--- a/flexget/plugins/api_trakt.py
+++ b/flexget/plugins/api_trakt.py
@@ -302,7 +302,7 @@ def get_id(type, name=None, year=None, trakt_slug=None, tmdb_id=None, imdb_id=No
         log.debug('ID Lookup failed on error: %s' % e)
     if results['type'] is type:
         for result in results:
-            if clean_title == result[type]['title']:
+            if clean_title == normalize_series_name(result[type]['title']):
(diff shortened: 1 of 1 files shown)


flexget/plugins/api_trakt.py
```diff
@@ -294,6 +294,37 @@ def get_id(type, name=None, year=None, trakt_slug=None, tmdb_id=None, imdb_id=No
...
 
+    clean_title = normalize_series_name(name)
+    found = None
+    try:
+        results = req_session.get(get_api_url('search'), params={'query': clean_title, 'type': type}).json()
+    except requests.RequestException as e:
+        log.debug('ID Lookup failed on error: %s' % e)
+    if results['type'] is type:
+        for result in results:
+            if clean_title == result[type]['title']:
```
gazpachoking: Hmm, isn't this going to give us trouble because the symbols are stripped out of clean_title?
On a separate note, I'm actually not so confident in the search by title at all until trakt fixes up results order.

crawln45: for all I care comment it out for now. I know I've done a few searches for some titles of stuff that I watch. and each one come up within the first grouping of searches. And I forgot to normalize the json title

PR #438, opened 2015-02-04
cleaning up alternate name in series

cvium:
@gazpachoking Is this what you had in mind? Any more changes?

And while we're at it. Have you thought about the sorting we talked about?


gazpachoking:
Looks good to me.

PR #439, opened 2015-02-05
added imdb_runtime information to imdb_lookup plugin and test parser

BrainDamage:


flexget/plugins/metainfo/imdb_lookup.py
```diff
@@ -448,6 +449,7 @@ def _parse_new_movie(self, imdb_url, session):
         movie.score = parser.score
         movie.votes = parser.votes
         movie.year = parser.year
+        movie.runtime = parser.runtime
```
crawln45: Is this going to be added to the db table, and did you upgraded the database schema

stevezau:
@BrainDamage what's the use case for this?

BrainDamage:
I should fix this pr
the use case would be rougly estimating an entry's bitrate by simply dividing the total size by the duration ( i want to reject / accept entries based off bitrate because it doesn't make much sense to me to deny based on total size since ofc longer movies / shows would naturally take more space for same quality )

gazpachoking:
@BrainDamage Please reopen if you update this.

PR #440, opened 2015-02-06
Add a logging plugin

Ram-Z:
This is basically a stripped down `exec` which writes arbitrary string to the log.

Currently it allows writing to different log levels. I'm thinking of adding the ability to set the name of the logger.

What do you think?

This has not been tested at all, apart from a single run to see if it actually works in one basic case.

### DIFF
--- a/flexget/plugins/output/log.py
+++ b/flexget/plugins/output/log.py
@@ -1,0 +1,150 @@
+from __future__ import unicode_literals, division, absolute_import
+import logging
+
+from flexget import plugin
+from flexget.event import event
+from flexget.utils.template import render_from_entry, render_from_task, RenderError
+
+log = logging.getLogger('log')
+
+
+class PluginLog(object):
+    """
+    Log entries
+
+    Simple example, log for entries that reach output::
+
+      log: 'found {{title}} at {{url}}'
+
+    Advanced Example::
+
+      log:
+        on_start:
+          phase: "Started"
+        on_input:
+          for_entries: 'got {{title}}'
+        on_output:
+          for_accepted: 'accepted {{title}} - {{url}}'
+
+    You can use all (available) entry fields in the command.
+    """
+
+    NAME = 'log'
+    HANDLED_PHASES = ['start', 'input', 'filter', 'output', 'exit']
+
+    schema = {
+        'oneOf': [
+            {'type': 'string'},
+            {
+                'type': 'object',
+                'properties': {
+                    'on_start': {'$ref': '#/definitions/phaseSettings'},
+                    'on_input': {'$ref': '#/definitions/phaseSettings'},
+                    'on_filter': {'$ref': '#/definitions/phaseSettings'},
+                    'on_output': {'$ref': '#/definitions/phaseSettings'},
+                    'on_exit': {'$ref': '#/definitions/phaseSettings'},
+                },
+                'additionalProperties': False
+            }
+        ],
+        'definitions': {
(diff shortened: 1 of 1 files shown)


gazpachoking:
Hmm. What would be the use case for this, mostly a debugging tool?

Ram-Z:
> Hmm. What would be the use case for this, mostly a debugging tool?

Yes, but also if you have external tool that parse the flexget logs you 
can facilitate their job by injecting easily parseable strings into the 
logs.

I currently use it to extract the destination name after a move task:

```yaml
  move-series: # {{{2
    # <snip>
    set:
      destination: '/home/ramsi/Series'
      folder: "{{tvdb_series_name}}/{% if tvdb_season == 0 %}Specials{% else %}Season {{tvdb_season}}{% endif %}"
      filename: "{{tvdb_season}}x{{tvdb_episode|pad(2)}} {{tvdb_ep_name|replace('/','-')}} ({{tvdb_ep_air_date|formatd❯
    move:
      to: "/home/ramsi/Series/{{folder}}"
      filename: "{{filename}}"
      clean_source: 100   # removes dir if < [N] MB

    log: "<<<{{folder}}/{{filename}}>>>"
```

```bash
if [[ "$postprocstatus" -eq 0 ]]; then
  # OK: try to move the file
  out=$(flexget execute --task move-$category)
  echo -e "$out\n"
  new_name=$(echo "$out" | sed -n 's/.*<<<\(.*\)>>>.*/\1/p')
  if [[ -n "$new_name" ]]; then
    echo "$new_name"
  else
    echo "Failed move!" >&2
    exit 1
  fi
fi
```

I was using `exec: echo "<<<{{folder}}/{{filename}}>>>"` before but that annoyingly outputs the string twice, and it's impossible to choose which output level the string is outputted at.

```
2015-03-24 16:34 VERBOSE  exec          move-series     Executing: echo "<<<Better Call Saul/Season 1/1x08 RICO (2015){1080p}{hdtv}{h264}[en]>>>"
2015-03-24 16:34 INFO     exec          move-series     Stdout: <<<Better Call Saul/Season 1/1x08 RICO (2015){1080p}{hdtv}{h264}[en]>>>
```

Ram-Z:
This plugin could also benefit from a `outputs` plugin as mentioned in [PR#475](https://github.com/Flexget/Flexget/pull/475#issue-64777649)

stevezau:
@Ram-Z what's the actual use case for this?? What are you doing with the outputted log file?? is it picked up by a script? what is that script doing?

FYI: This prob won't solve your use case but I'm creating a restful API PR https://github.com/Flexget/Flexget/pull/523 which will provide access to the server log.

Ram-Z:
@stevezau Sorry for the delay, this notification must have slipped through.

My use case is described above. I'm catching the output from `flexget execute --task` and parsing it. 

This is for a better SabNZB integration. When a download is finished, SabNZB can run a [script](https://github.com/Flexget/Flexget/pull/440#issuecomment-85632010) and show the output in its interface. The script above simply runs the particular task that tries to move the newly downloaded file into the correct position and `log`s the output if succesfull, the script then parses that output, searches for the log entry and prints out some message if the move failed/succeded.

I'm sure it can be used for other things, if for example the logging of a particular plugin is to verbose, it could be disabled and a personalized log message shown instead.

liiight:
There's an issue open to enhance logging capabilities, a more broad approach I believe #812.

PR #442, opened 2015-02-08
[WIP] search plugin for rarbg

cvium:
RarBG seems pretty decent and while their releases also float around on Torrentz etc. I think their site is a lot less noisy.

One thing though. They have some weird bot detection. If the cookie does not contain `'7fAY799j': 'VtdTzG69'`, then it will get caught by the bot filter after the ~5th consecutive request.

I'm not entirely sure they'd appreciate this plugin then, but their RSS feed sucks. Their anti-bot captcha is also extremely easy to circumvent with OCR.

### DIFF
--- a/flexget/plugins/search_rarbg.py
+++ b/flexget/plugins/search_rarbg.py
@@ -1,20 +1,18 @@
 from __future__ import unicode_literals, division, absolute_import
 import logging
 import urllib
-import re
 
 from flexget import plugin
 from flexget.entry import Entry
 from flexget.event import event
 from flexget.config_schema import one_or_more
-from flexget.utils import requests
 from flexget.utils.requests import Session
-from flexget.utils.soup import get_soup
-from flexget.utils.search import torrent_availability, normalize_unicode
+from flexget.utils.search import normalize_unicode
 
 log = logging.getLogger('rarbg')
 
-session = Session()
+requests = Session()
+requests.set_domain_delay('torrentapi.org', '10.3 seconds')  # they only allow 1 request per 10 seconds
 
 CATEGORIES = {
     'all': 0,
@@ -42,7 +40,7 @@ class SearchRarBG(object):
 
         To perform search against multiple categories:
 
-        publichd:
+        rarbg:
             category:
                 - x264 720p
                 - x264 1080p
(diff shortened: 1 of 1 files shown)


rarbg:
Hey :) Will a json api search work better ?

cvium:
Possibly. There are two kind of feeds I usually use. A continuously updated <category> rss feed and search specific rss feeds.

I believe KAT employs an XML search API eg. https://kickass.to/usearch/the%20walking%20dead/?rss=1.

But any publicly available search API that allows querying on categories, seeds etc. would be a welcome addition.

rarbg:
https://rarbg.com/pubapi/apidocs.txt
That looks fine to you ? I didnt have enough time to make it into something proper but should have everything you want ?

cvium:
It looks pretty good. I don't know what "The api has a 1req/min limit." means though.

rarbg:
1 request per minute max :)

cvium:
Ah. Quite a big limitation though. Not sure it will be of much use with Flexget. It will be fine for simple RSS downloading eg. watching the TV/Movie rss feed every hour.
It won't be of much use for direct searching though.

rarbg:
lowered the time to 1req/10s. That should be fine for searching ?

cvium:
Yeah, that'll work. I will rewrite my plugin when I get some time.

cvium:
Would it be possible to get magnet and .torrent options?

rarbg:
We cant really implement .torrent because of many reasons which i cant discuss over github :(

gazpachoking:
This still a WIP, or good to go?

cvium:
WIP. Been pretty swamped lately and their API is a little funky. I'll merge it when I get it done :P

gazpachoking:
@cvium Can you add this to the http://flexget.com/wiki/Searches page?

cvium:
@gazpachoking Done.

cvium:
@rarbg Would it be possible to search for specific episodes using tvdb or imdb ids? As it stands, older items can never be found if they are not within the latest 100 uploads.

rarbg:
@cvium 
From the api docs - The api searches only in the last 10000 torrents except for imdb,tvrage,tvdb.
Meaning if you use the parameteres search_imdb=xxxx / search_tvrage=xxxx / search_tvdb=xxxx you will be able to search in all the old uploads. 
example:
pubapi.php?mode=search&search_imdb=tt0944947&search_string=Game+Of+Thrones+S03E03&token=xxxx

cvium:
@rarbg  This one fails though: pubapi.php?mode=search&search_imdb=tt1839578&search_string=Person%20of%20Interest%20S02E01&token=gin6q1rtm0

but https://rarbgmirror.com/torrents.php?search=person+of+interest+s02e01 returns results.

rarbg:
@cvium 
you need to read the docs :)

by default the api will return only ranked torrents ( internal ) ex scene releases + -rarbg releases + -rartv releases.
If you want other groups included in the results use the ranked parameter with a value of 0 to get them included.

cvium:
@rarbg You are right, I wasn't thinking. Thanks for the clarification. All is well. I also see you have fixed min_seeders and min_leechers.

rarbg:
@cvium yes they are fixed a long time ago :)

PR #443, opened 2015-02-08
update TPB to new TLD and search API

X-dark:
back to .se

### DIFF
--- a/flexget/plugins/urlrewrite_piratebay.py
+++ b/flexget/plugins/urlrewrite_piratebay.py
@@ -13,7 +13,7 @@ from flexget.utils.search import torrent_availability, normalize_unicode
 
 log = logging.getLogger('piratebay')
 
-CUR_TLD = 'to'
+CUR_TLD = 'se'
 TLDS = 'com|org|sx|ac|pe|gy|se|%s' % CUR_TLD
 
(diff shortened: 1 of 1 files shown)


cvium:
Have you tested that it works? I get a RequestException because they seem to use cloudflare.

Plugin also needs to be updated to use new flexget search api.

X-dark:
Yes it does work for me.
I have seen about the new search API warning. Where are the changes documented?

cvium:
What is your config? It refuses to work for me because of SSL verification problems.

I am not sure about the specifics of the search API, but changing line 107 to `def search(self, task, entry, config=None):` and line 123 to `for search_string in entry.get('search_strings', [entry['title']]):` seems to do the trick.

X-dark:
Nothing particular in my config
```
    discover:
      what:
        - emit_series:
            from_start: yes
      from:
        - piratebay:
            category: 208
            sort_by: seeds
```

I don't see how you can get SSL verification problem with an HTTP URI.

cvium:
Maybe it's something Windows specific. But it definitely checks the SSL certificate. Anyway, if you fix the search api I think we can merge.

cvium:
Great, thanks.

PR #444, opened 2015-02-11
Adding language to qualitities

immenz:
I know there may be other/better ways, but could you take a look in this issue?

This would easily make it possible to define the language you want to download, without changing anything when no language is set. 
My problem is, that when i download an original episode and later on the synchronized one, it will be not accepted because they have the same quality, but different languages...


stevezau:
Hmm is there any chance this could mess up tites if eng was in the title of a tvshow for example?

I know guessit has lang support, perhaps we need to take a look how it works ?

liiight:
While I completely agree that language support is needed, I feel like qualities isn't the right place for it. It should be handled by the parser. 
You want to change this?

immenz:
I will take a look at this soon

antivirtel:
Great idea! What about the multi-language torrents? Will it support those also? Eg. HUN.ENG.

snickers2k:
anything new?! What was with guessit's language support? Maybe as a simple setting for guessit? would be awesome. thanks

PR #445, opened 2015-02-12
Extended Serienjunkies UrlRewriter

immenz:
This Version has better language recognition and allows multiple download urls (usable with pyload plugin)

### DIFF
--- a/flexget/plugins/urlrewrite_serienjunkies.py
+++ b/flexget/plugins/urlrewrite_serienjunkies.py
@@ -12,15 +12,15 @@ from bs4 import BeautifulSoup, NavigableString, Tag
 
 log = logging.getLogger('serienjunkies')
 
-reSingleEp  = re.compile(r'(S\d+E\d\d+)(?!-E)', re.I)
-reMultiEp   = re.compile(r'(?P<season>S\d\d)E(?P<startep>\d\d+)-E?(?P<stopep>\d\d+)', re.I)
-reSeason    = re.compile(r'(?<=\.|\-)S\d\d(?:[-\.]S\d\d)*(?!E\d\d+)', re.I)
+regex_single_ep = re.compile(r'(S\d+E\d\d+)(?!-E)', re.I)
+regex_multi_ep  = re.compile(r'(?P<season>S\d\d)E(?P<startep>\d\d+)-E?(?P<stopep>\d\d+)', re.I)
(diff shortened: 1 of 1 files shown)


paranoidi:
Looks good except variable names don't follow PEP8 standard (no camelCases). Want to fix that?-)

immenz:
Ok, will fix that later this day!

immenz:
Explanation:

First, language settings are different, so old configs have to be changed. instead of 'de' 'en' or 'both' it now goes with german, foreign, subtitle or dual. Option 'both' was useless anyway. This system is more aligned with the site serienjunkies, because it is a german site and therefore offers first of all german dubbed tv shows. All other languages are just bonus, so the combine to option foreign. Furthermore all Subtitles in form of bonus material (BD Rips etc) are ignored. The often are placed after a '+' in language field, so i ignore that. Only actual "needed" subtitles, when there is only one foreign audio channel, are evaluated. So:
German get german language and dual language (any foreign language)
foreign gives englisch, french, japanes etc (maybe regex has to be expanded some times) only
subtitle gets foreign language episode with subtitle (here no pattern is used, but I think there are only german subtitles here)
and dual (my setting) gets only episodes with 2 languages set, first have to be german.

Second change is that also multi episodes and seasons can be rewritten to multiple urls. Because I did'nt find another way, I put a list of these multiple links in description tag of the entry, but entry['url'] is also always filled, so there should be no error or problems with that. Because serienjunkies only provides och links, I think pyload is the only flexget plugin to take these kinds of links at the moment. 

So to use all the features, flexget has of course to be configured accordingly. For examples I have an special ep /ip regexp configured, so that multi episodes/ seasons are being accepted

      ep_regexp:
        - s(\d+)e(\d+)(-e?(\d+))?
        - s(\d+)ep(\d+)

    id_regexp:
      - (S\d\d)(?:.(S\d\d))*

immenz:
So what about this PR? I accidentally committed in this branch but want to use it for other things in future. As of now, I can't commit because there may be other Bugfixes or changes to make to this urlrewriter

gazpachoking:
@immenz Thanks, sorry for the delay. Can you update the wiki if needed?

immenz:
I doesn't know about a wiki where urlrewrite_* plugins are in...

But if you give me the url I'll could do that.

PR #451, opened 2015-02-23
Pushbullet - support multiple api keys and email destinations.

nikdoof:
Adds the ability to use multiple API keys in the Pushbullet output plugin, also the addition of email address destinations for messages.

Squashed version of my previous PR with fixes and cleanups.


gazpachoking:
Thanks! Sorry for the delay, want to update the wiki?

nikdoof:
Updated, added an example as well.

PR #452, opened 2015-02-23
Convert Magnet plugin

nikdoof:
This plugin uses libtorrent to convert a magnet URI torrent to a full torrent file by spinning up a small torrent client to pull from the DHT. This way magnets from non-torrent file sites can be processed using the content filters. 

Obviously the conversion is time consuming and in situations its possible for invalid torrents to take a lot of time to actually acquire the metadata, a parameter has been added to time out any long hanging queries.

This iteration has been running on my personal machine for a long period of time, suggestions are welcome.

### DIFF
--- a/flexget/plugins/modify/convert_magnet.py
+++ b/flexget/plugins/modify/convert_magnet.py
@@ -33,7 +33,7 @@ class ConvertMagnet(object):
         params = libtorrent.parse_magnet_uri(magnet_uri)
         session = libtorrent.session()
         handle = session.add_torrent(params)
-        log.debug('Aquiring torrent metadata for magnet {}'.format(magnet_uri))
+        log.debug('Acquiring torrent metadata for magnet {}'.format(magnet_uri))
         timeout_value = timeout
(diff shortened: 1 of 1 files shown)


cvium:
I think it looks interesting and seems like a decent way of not relying on torcache and what have you.

Have there been any particularly annoying problems?

nikdoof:
By far the biggest one is invalid torrents which don't exist in the DHT swarm which run the full timeout, hence why I added the configurable timeout. Generally it takes around 3-5sec to acquire a highly shared torrent, which most of the time will be connecting to the DHT in the first place. 

I'm sure some further configuration options could be added for the mini-client itself, but the defaults work well for direct connections and simple UPnP firewall setups.

cvium:
From what I gather this plugin will always rewrite magnet urls unless specifically disabled?

nikdoof:
Ahh yes, looks that way. I'm checking for ```config == False``` but the default config setup is enabling it. This isn't the intended result, i'll make changes to resolve it.

PR #454, opened 2015-02-26
Disabled SSL certificate validation to avoid SSLErrors in TPB

jesjimher:
Pirate bay discover plugin was failing with the following error:

    Error searching with piratebay: RequestException: hostname 'thepiratebay.se' doesn't match either of 'ssl2000.cloudflare.com', 'cloudflare.com', '*.cloudflare.com'

It seems that TPB's HTTPS certificate is not well configured. I just disabled host certificate validation, now it works.

### DIFF
--- a/flexget/plugins/urlrewrite_piratebay.py
+++ b/flexget/plugins/urlrewrite_piratebay.py
@@ -88,7 +88,7 @@ class UrlRewritePirateBay(object):
 
     @plugin.internet(log)
     def parse_download_page(self, url):
-        page = requests.get(url).content
+        page = requests.get(url,verify=False).content
         try:
             soup = get_soup(page)
             tag_div = soup.find('div', attrs={'class': 'download'})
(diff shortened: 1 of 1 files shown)


X-dark:
This is wrong. Certificate validation works well and the certificate is valid. You are using either an outdated version of Python or an outdated version of urllib that does not check the certificate subject alt name.
Additionally, this change makes a warning being thrown with each request:
``` InsecureRequestWarning: Unverified HTTPS request is being made. Adding certificate verification is strongly advised. See: https://urllib3.readthedocs.org/en/latest/security.html```

gazpachoking:
Hmm, I agree. I don't think we should turn verification off by default.

I believe this error happens not because of an invalid certificate, but when a server is using SNI, and the installed python/packages do not support it. Using python 2.7.9+, SNI should be supported natively, and on older pythons installing `requests[security]` will install the necessary packages.

PR #455, opened 2015-02-28
Plugin to list files from a transmission instance

ianstalk:
This a plugin that lists files in a Transmission instance in a similar fashion to the find plugin. I wasn't sure if it was appropriate to include in the main plugin_transmission.py module, so I made a new file. Feel free to merge them if my commit is accepted.


grrr2:
Hi,

If I may add two comments please:

1. As per transmission documentation https://trac.transmissionbt.com/browser/trunk/extras/rpc-spec.txt 	section "2.3.  Transport Mechanism" host, port and rpc path should be variable to allow "mapping and/or multiple daemons to run on a single server". So would be nice to add an rpc path variable to config.
2. If one runs multiple transmission daemons and create multiple tasks to use these different daemons once you create the client in your code it will be used for all tasks regardless if transmission config is different per task. There was a bug in the existing transmission plugin about it (#2804).

Thanks

ianstalk:
1. Would you suggest that I add a path parameter to TransmissionBase? Adding it to one Transmission plugin and not the rest seems like it'd be confusing to users,  not to mention a maintenance headache.

As it currently stands, you can specify the path in the host variable; I do that in my own personal config file: `host: 'https://{{secrets.transmission.host}}/transmission/rpc'`. This is consistent with how the transmissionrpc module itself does things.

2. Thanks for the tip; I'll make a similar change to my plugin.

grrr2:
Re 1. Nothing wrong in your code it is rather an advise to keep in mind that proper format is ```http(s)://host:port/path/rpc``` and not ```http(s)://host:port/transmission/rpc```, ie. /transmission is not hard coded can be different. So yes, setting ```host``` properly is usually enough as transmissionrpc library underneath will handle this special case too. Sorry if gave you wrong impression, it was just a note.

Thanks

PR #457, opened 2015-03-08
reverting changes from 761c2f206c... (pull request #453)

bozho:
Looks like the old code works again. I could have sworn that while
testing the patch, KAT was returning 404 if rss wasn't the first query
parameter. It was late, I may have been seeing things...

### DIFF
--- a/flexget/plugins/search_kat.py
+++ b/flexget/plugins/search_kat.py
@@ -48,7 +48,7 @@ class SearchKAT(object):
             params = {'rss': 1}
             if config.get('verified'):
                 search_string_url_fragment += ' verified:1'
-            url = 'http://kickass.to/usearch/%s/?rss=1' % urllib.quote(search_string_url_fragment.encode('utf-8'))
+            url = 'http://kickass.to/search/%s/' % urllib.quote(search_string_url_fragment.encode('utf-8'))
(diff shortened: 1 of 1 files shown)


patsissons:
Hey I just saw the change here, if you go to kickass.to and click on the magnifying glass in the search box it will take you to `usearch`. However, I believe both work all the same, so it probably doesn't matter. Just a heads up.

bozho:
On 25/03/2015 16:49, Pat Sissons wrote:
> Hey I just saw the change here, if you go to kickass.to and click on the
> magnifying glass in the search box it will take you to |usearch|. However, I
> believe both work all the same, so it probably doesn't matter. Just a heads up.

Yeah, I noticed that and noted it in the pull request in the original fix.

The strange thing is, the original fix was forcing 'rss=1' query parameter to
be the very first one in the URL and that seemed to have fixed 404's I was
seeing at the time - it's like KAT's request handler was accessing parsed
query parameters via an array rather than a dictionary/map and expecting the
RSS parameter to be the first one.

A few days later, I tried again and the original flexget code worked again, so
I reverted the changes.

Cheers,
-- 
Marko
ICQ: 5990814

I'm not under the alkafluence of inkahol
that some thinkle peep I am.
It's just the drunker I sit here the longer I get.

PR #459, opened 2015-03-12
Some PEP8 fixes

crawln45:
Just doing some PEP8 fixes

### DIFF
--- a/flexget/ui/plugins/movies/movies.py
+++ b/flexget/ui/plugins/movies/movies.py
@@ -44,8 +44,8 @@ def index():
             # this is probably not needed since non cached movies are retried also
             # in the cover function
             #
-            #import thread
-            #thread.start_new_thread(tmdb_lookup, (), {'imdb_id': item.imdb_id})
+#            #import thread
+#            #thread.start_new_thread(tmdb_lookup, (), {'imdb_id': item.imdb_id})
             continue
 
         # set thumb, but only if already in cache because retrieving is too slow here
--- a/flexget/utils/cached_input.py
+++ b/flexget/utils/cached_input.py
@@ -96,7 +96,7 @@ class cached(object):
 
             if api_ver == 1:
                 # get name for a cache from tasks configuration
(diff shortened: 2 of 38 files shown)


flexget/utils/cached_input.py
```diff
@@ -165,7 +165,8 @@ def wrapped_func(*args, **kwargs):
                     self.cache[cache_name] = copy.deepcopy(response)
                 except TypeError:
                     # might be caused because of backlog restoring some idiotic stuff, so not neccessarily a bug
-                    log.critical('Unable to save task content into cache, if problem persists longer than a day please report this as a bug')
+                    log.critical('Unable to save task content into cache, if problem persists longer than a day \
```
paranoidi: This adds \ into might with a lot empty spaces afterwards.

function('fkgivjfbf'
                'gkvkfbg')

Is best way to do this

flexget/ui/plugins/movies/movies.py
```diff
@@ -44,8 +44,8 @@ def index():
             # this is probably not needed since non cached movies are retried also
             # in the cover function
             #
-            #import thread
-            #thread.start_new_thread(tmdb_lookup, (), {'imdb_id': item.imdb_id})
+#            #import thread
```
paranoidi: I don't think that there was anything wrong with the original format

paranoidi:
Awesome fixes!

flexget/ui/plugins/movies/movies.py
crawln45: PEP8 yells about the "#" Not being at the beginning of the line only when it's actual Python code after the "#". Do not ask me I dunno so I fixed it to where it didn't show up.

flexget/utils/cached_input.py
crawln45: I didn't know if that would split the message into 2 parts. so that it was formatted on 2 lines or if it was just a continuation.

paranoidi:
These were mostly good fixes...

PR #460, opened 2015-03-15
Main file ratio

patsissons:
adding a new config key (`main_file_ratio`) to deluge and transmisson plugins so that when using the `main_file_only` config item, you are not forced to assume that the main file must be > 90% of the total torrent size, this ratio is now configurable. This helps a lot with shows that sometimes include samples which are potentially 10-15% of the total size.

I'm getting this PR started early as I want to start the review process right away (since I'm not super familiar with flexget plugins to begin with). I believe I got things right, or close to right at least, but I'm happy to fix style/consistency issues if there are any. I tried to add as little code as possible and keep existing code as unchanged as possible, as a result I may not be doing things the best way possible, but it will be as consistent as possible with the existing code base.

I'm testing this out on my server right now but won't really be able to see it's effectiveness for a few days.

### DIFF
--- a/flexget/plugins/plugin_deluge.py
+++ b/flexget/plugins/plugin_deluge.py
@@ -640,10 +640,10 @@ class OutputDeluge(DelugePlugin):
                                                      [(file['index'], new_name)]))
                         log.debug('File %s in %s renamed to %s' % (file['path'], entry['title'], new_name))
 
-                    # find a file that makes up more than 90% of the total size
+                    # find a file that makes up more than main_file_ratio (default: 90%) of the total size
                     main_file = None
                     for file in status['files']:
-                        if file['size'] > (status['total_size'] * 0.9):
+                        if file['size'] > (status['total_size'] * opts.get('main_file_ratio')):
                             main_file = file
                             break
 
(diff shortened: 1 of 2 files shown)


paranoidi:
Looking good!

gazpachoking:
Looks good to me, is this ready to go?

patsissons:
I believe so, I haven't run into any issues with it so far

gazpachoking:
Sweet, thanks! Want to update the wiki?

patsissons:
should I wait for the next master build first?

gazpachoking:
Don't think it matters if we jump the gun a bit, build will be out in 10 hours or so.

patsissons:
roger that. i'll post the updates later today.

olfway:
btw, it breaks deluge plugin:
  File "/usr/lib64/python2.7/site-packages/flexget/plugins/plugin_deluge.py", line 646, in on_get_torrent_status
    if file['size'] > (status['total_size'] * opts.get('main_file_ratio')):
exceptions.TypeError: unsupported operand type(s) for *: 'int' and 'NoneType'

patsissons:
thanks for the heads up. I don't have a deluge test environment. I'll try and fix this asap.

PR #461, opened 2015-03-16
Added the ability to the move plugin to clean the target location

kingamajick:
I often have the scenario where a better version of a file is made available to my library, for example when replacing DVDRips with BlurayRips.  The series plugin will allow upgrades in quality to occur, but currently this ends up with two copies of the file at the target of different qualities.  

This change allows a regex to be applied to the move plugin to remove files matching that pattern to be removed before the new file(s) are moved.

### DIFF
--- a/flexget/plugins/output/move.py
+++ b/flexget/plugins/output/move.py
@@ -3,6 +3,7 @@ import os
 import shutil
 import logging
 import time
+import re
 
 from flexget import plugin
 from flexget.event import event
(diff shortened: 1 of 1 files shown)


Ram-Z:
May I ask what happened to this PR? I would also be interested in this functionality.

kingamajick:
I ended up doing it as a independant plugin rather than just modifying the move plugin.  You can find it here https://gist.github.com/kingamajick/760f2a91a4197648533a I may resubmit a pull with this instead when I get a bit of time to get tests working correctly on my local machine.

Ram-Z:
Alright... even better.

Tests are failing on some machines, see https://github.com/kevin1024/vcrpy/issues/131.

PR #462, opened 2015-03-17
Transmission plugin magnetization waiting

patsissons:
This patch addresses [this issue (#2911)](http://flexget.com/ticket/2911) that was posted not too long ago, but was closed prematurely (due to a misunderstanding of the issue I assume).

I have only patched the transmission plugin, I'm unsure if the deluge plugin requires patching but I have no test environment to do any proper QA for a deluge patch. I have just deployed the patched file to my own environment, I should get confirmation if it is working over the next few days.

This patch simply checks for a (_new_) `magnetization_timeout` configuration value (in seconds, defaults to 0, i.e., disabled). If this timeout is set to any number > 0 then the add will poll the transmission RPC client every second (until the timeout elapses) to see if the torrent has been magnetized. Once magnetized, the file list will be populated and the `totalSize` will be re-fetched. This will allow proper pre-processing of the `main_file_only`, `content_filename`, `skip_files`, etc... when a torrent is added to transmission via a magnet URI.

Because the default value for `magnetization_timeout` is zero, the default behaviour for this feature is disabled entirely.

**NOTE**
I'm not a python programmer professionally (mostly C# these days), so my programming style may not be ideal. There may be some more optimal ways of accomplishing the tasks I have added in this patch. I'm open to constructive comments to improve the coding style and efficiency of the patch additions.

### DIFF
--- a/flexget/plugins/plugin_transmission.py
+++ b/flexget/plugins/plugin_transmission.py
@@ -413,6 +413,8 @@ class PluginTransmission(TransmissionBase):
                         filedump = base64.b64encode(f.read()).encode('utf-8')
                     r = cli.add_torrent(filedump, 30, **options['add'])
                 else:
+                    # we need to set paused to false so the magnetization begins immediately
+                    options['add']['paused'] = False
                     r = cli.add_torrent(entry['url'], timeout=30, **options['add'])
                 if r:
                     torrent = r
--- a/flexget/plugins/plugin_transmission.py
+++ b/flexget/plugins/plugin_transmission.py
@@ -456,6 +456,14 @@ class PluginTransmission(TransmissionBase):
                    'content_filename' in options['post'] or skip_files):
                         fl = cli.get_files(r.id)
                 
+                        if 'magnetization_timeout' in options['post'] and options['post']['magnetization_timeout'] > 0 and not downloaded and len(fl[r.id]) == 0:
+                            log.debug('Waiting %d seconds for "%s" to magnetize' % (options['post']['magnetization_timeout'], entry['title']))
+                            fl = _wait_for_files(cli, r, options['post']['magnetization_timeout'])
+                            if len(fl[r.id]) == 0:
+                                log.warning('"%s" did not magnetize before the timeout elapsed, file list unavailable for processing.' % entry['title'])
+                            else:
+                                total_size = cli.get_torrent(r.id, ['id', 'totalSize']).totalSize
+                
                         # Find files based on config
                         dl_list = []
                         skip_list = []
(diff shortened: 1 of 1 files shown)


patsissons:
received my first confirmation that this patch is working. Removing my debugging code in my test environment now to ensure the changes in this PR are good.

patsissons:
just FYI, it is still not clear if this patch is fully functional.

patsissons:
I now believe this is working, I have turned on a bit more logging to confirm all is well, but the last few items added were most definitely renamed properly.

There is a small caveat. Because magnetization requires the torrent is started immediately, there is a good chance that some additional files will be partially downloaded before the file list is processed. I have considered stopping the torrent immediately after magnetization (and then allowing it to start at the end again like regular torrents), but I'm not sure how well that would work (i.e. might still get a partial file anyways). Thoughts?

patsissons:
OK, confirmed working.

gazpachoking:
Cool! Want to bring the deluge plugin into parity? :wink:

patsissons:
i do, but i will do that as a separate PR. I also have the issue of no deluge test environment. So I would have to *code blind* so to speak. It's probably very similar to the transmission changes though.

cvium:
I use Deluge, so I could probably test it for you if you told me how.

patsissons:
Deluge looks like it will be a little more complicated because the file list is retrieved as the return value of adding the torrent (which I am assuming is empty for magnet links). I'm also having trouble tracking down the deluge python API so I can get some context on the API functions, anyone know where to find the docs?

crawln45:
http://deluge-torrent.org/docs/master/modules/ui/web/json_api.html
I think is what I was using to do some stuff not sure though.

gazpachoking:
Yeah, the readthedocs site and http://dev.deluge-torrent.org/wiki/Development/1.3/UIClient are helpful. That json api is something different provided by the webui, which we aren't using.

patsissons:
I found http://deluge-torrent.org/docs/master/modules/ui/web/json_api.html?highlight=get_torrent_info#deluge.ui.web.json_api.WebApi.get_torrent_files which returns the file list, I'm unsure if that's the right API to use though.

gazpachoking:
Here you can see how the deluge plugin is grabbing that list currently. https://github.com/Flexget/Flexget/blob/develop/flexget/plugins/plugin_deluge.py#L645

It's in the dictionary returned by `get_torrent_status`.

crawln45:
Well everything I've read leads me to believe that we are interacting with:
http://deluge.readthedocs.org/en/develop/core/rpc.html

I say this because well the namespaces mainly. am I wrong on this?

patsissons:
is there any way to call these API functions synchronously?

patsissons:
My next thought is to recursively call `get_torrent_status` in the callback if magnetization_timeout is set and the file list is empty until the timeout *expires* or the file list is populated. I think that might work, but I'm definitely stretching my grasp on python and the flexget arch.

patsissons:
OK, I've created #471 so we can move this discussion over there now :)

PR #463, opened 2015-03-18
Add search/urlrewrite plugin for extratorrent.cc

v17al:
- extratorrent doesn't mesh well with the search_rss plugin as
  the data doesn't alway contains a .torrent link (enclosures are often empty)
- Work around this issue by converting the .html link (always present)
  into a download link
- Add search plugin to make it easier for anyone else to get started with extratorrent.cc


v17al:
Thanks!

PR #464, opened 2015-03-20
fixing deluge plugin crash due to main_file_ratio config value not being present

patsissons:
this should fix the deluge plugin bug present in the latest 1.2.292 master branch. I don't have a test environment to confirm validity, but I'm fairly certain this was the missing code.

see https://github.com/Flexget/Flexget/pull/460#issuecomment-84051984

### DIFF
--- a/flexget/plugins/plugin_deluge.py
+++ b/flexget/plugins/plugin_deluge.py
@@ -802,6 +802,7 @@ class OutputDeluge(DelugePlugin):
                 modify_opts = {'label': format_label(entry.get('label', config['label'])),
                                'queuetotop': entry.get('queuetotop', config.get('queuetotop')),
                                'main_file_only': entry.get('main_file_only', config.get('main_file_only', False)),
+                               'main_file_ratio': entry.get('main_file_ratio', config.get('main_file_ratio')),
                                'hide_sparse_files': entry.get('hide_sparse_files', config.get('hide_sparse_files', True)),
                                'keep_subs': entry.get('keep_subs', config.get('keep_subs', True))
                 }


cvium:
I have tested it for you. It seems to work. Thanks.

patsissons:
great news!

I assumed that build issue was just an anomaly and not related to the patch.

cvium:
Yes, it crashed while setting up the test environment. Btw, I also fixed some transmission related errors.

patsissons:
should I make a similar change to prepare_config in #462 as well?

cvium:
I have no clue. I made some changes https://github.com/Flexget/Flexget/commit/5a4d0920746e617c99e8f0ca6e00c4b9c7fd0bd0 to plug a few holes, but I have nothing to do with that plugin otherwise.

patsissons:
ah ok, what was your reason for making the change at https://github.com/Flexget/Flexget/blob/5a4d0920746e617c99e8f0ca6e00c4b9c7fd0bd0/flexget/plugins/plugin_transmission.py#L63

instead of just keeping the change I made at 
https://github.com/Flexget/Flexget/blob/5a4d0920746e617c99e8f0ca6e00c4b9c7fd0bd0/flexget/plugins/plugin_transmission.py#L266

cvium:
https://github.com/Flexget/Flexget/blob/5a4d0920746e617c99e8f0ca6e00c4b9c7fd0bd0/flexget/plugins/plugin_transmission.py#L109

main_file_ratio is accessed in TransmissionBase, but it does not have a default value.

patsissons:
Ah good point. L266 would no longer be necessary then (I believe). Do you want me to do another PR to remove that line?

cvium:
There's no need imo.

patsissons:
yea it doesn't harm anything either way, but it may be confusing to anyone reading the code without any context why it is set twice.

crawln45:
Honestly I'm a little unsure why exactly there's so much configuration in the TransmissionPluginInput to begin with since 90% of it has nothing to do with the input side of the plugin. Any relevant configurations should be placed in my opinion in the corresponding Sections of the code or all together the latter being the better option so that everything is configured as one unit and then from there is applied to the different functions of the plugin.

PR #465, opened 2015-03-22
removing redundant config default setter for main_file_ratio

patsissons:
due to the change made in 5a4d092, the change originally made in 089f576 is no longer necessary as the `TransmissionBase.prepare_config` performs an identical `setdefault` call.


gazpachoking:
Sounds good, thanks! :smile:

PR #466, opened 2015-03-23
Add verify_ssl option to rss plugin.

Ram-Z:
This option allows to disable SSL Cert Verification.

### DIFF
--- a/flexget/plugins/input/rss.py
+++ b/flexget/plugins/input/rss.py
@@ -91,6 +91,15 @@ class InputRSS(object):
       rss:
         url: <url>
         group_links: yes
+
+    If the URLs SSL certificate cannot be verified but you trust the site, you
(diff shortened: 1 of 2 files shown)


Ram-Z:
Just saw there is a `verify_ssl_certificates` plugin which does the same on a per task basis.

This PR allows a more fine grained control over which SSL certs should be verified by adding an option to each instance of an `rss` setting. But feel free to close this unmerged it you feel it's not necessary.

cvium:
Tests using RSS are failing.

Ram-Z:
Fixed, I always forget that setting the default value in the schema is not sufficient to pass the tests.

gazpachoking:
I think verify_ssl_certificates plugin is good enough. If we need per-plugin control I'd like to think of a system better than adding this option to every single plugin.

PR #467, opened 2015-03-23
[WIP] Allow default settings for series.

Ram-Z:
Uses the special group_name `default` to set settings for every series,
including the ones generated from `configure_series`.

Would allow configs as followed to work:

```yaml
series:
  settings:
    default:
      identified_by: ep
    group1:
      identified_by: seq
    group2:
      quality: 720p

configure_series:
  from:
    [...]
  settings:
    quality: 1080p
```

All series will inherit the settings from group `default` and be identified by `ep`, except `group1` which overrides the default settings. `configure_series` inherits the default settings as well and would be able to override them if necessary.

### DIFF
--- a/flexget/plugins/filter/series.py
+++ b/flexget/plugins/filter/series.py
@@ -814,6 +814,8 @@ class FilterSeriesBase(object):
         for group_name in config:
             if group_name == 'settings':
                 continue
+            if group_name == "default":
+                continue
             group_series = []
             if isinstance(group_name, basestring):
                 # if group name is known quality, convenience create settings with that quality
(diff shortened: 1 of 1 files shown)


Ram-Z:
Currently doesn't work. 

My first attempt was to set the default values at [series.py#L882](https://github.com/Flexget/Flexget/blob/develop/flexget/plugins/filter/series.py#L882), but that function is unaware of any configs and the default configs are unavailable.

The current patch doesn't work for `configure_series`.

gazpachoking:
I think every plugin editing series config turns it into simple list form, at which point we lose any group info. Perhaps we need to use a different canonical form when merging in order to not lose this info.

liiight:
Is this still relevant?

liiight:
Please reopen if relevant

PR #469, opened 2015-03-24
Add the encoding parameter used to parse non ascii text files

Konubinix:


gazpachoking:
I'm thinking maybe we should default to UTF8 on this, as well as catch the exception if the file can't be encoded using the selected codec and give a proper error message to user.

gazpachoking:
I think requests should handle the decoding in the case of a url, so I think this would just apply to files. In that case, we could use `codecs.open` with the declared encoding rather than doing it line by line.

stevezau:
@Konubinix any progress on this?

PR #470, opened 2015-03-26
remove extra urllib3 INFO logging

patsissons:
It was clear in the code that the intent was to remove these annoying INFO log messages
> Starting new HTTP connection (1): ...

by setting the log level to WARN `logging.getLogger('requests.packages.urllib3').setLevel(logging.WARNING)`
But it isn't taking any effect in the current release.

I changed the logger name to just `urllib3`, which seems to have corrected the problem. I'm not sure if there are any implications for the change I have made, so any input is welcome in that department.

Sourced my patch from these two StackOverflow questions:
* http://stackoverflow.com/questions/21944445
* https://stackoverflow.com/questions/11029717

### DIFF
--- a/flexget/utils/requests.py
+++ b/flexget/utils/requests.py
@@ -15,7 +15,7 @@ from flexget.utils.tools import parse_timedelta, TimedDict
 log = logging.getLogger('requests')
 
 # Don't emit info level urllib3 log messages or below
-logging.getLogger('requests.packages.urllib3').setLevel(logging.WARNING)
+logging.getLogger('urllib3').setLevel(logging.WARNING)
 
 # Time to wait before trying an unresponsive site again
 WAIT_TIME = timedelta(seconds=60)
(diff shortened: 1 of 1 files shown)


patsissons:
Yea I didn't think this would have broken the CI build. if the build is fixed soon I can rebase.

crawln45:
All of emit_series is busted atm which is the reason for the build fails. I tested the logging internally using the unit tests where I have not included Paranoidi's new update to the backfill stuff. Nothing failed. so wouldn't worry about it.

gazpachoking:
Hmm, the old way still works for me (and with just 'urllib' it doesn't.) I'm guessing maybe you installed requests with your system package manager? Some package managers override the requests vendored urllib3 with the system package. I'm thinking we should just put both lines in.

patsissons:
I'm running this on ubuntu 14.04.2 LTS, if that means anything to you.

you're probably right though, I don't think there would be any harm including both lines.

gazpachoking:
Thanks!

PR #471, opened 2015-03-26
[WIP] Allow deluge plugin to wait for torrent magnetization

patsissons:
OK, the concept I'm going for here is to use `@defer.inlineCallbacks` to yield the file list out of the `get_torrent_status` API call. This should allow the plugin to semi-synchronously (not really) wait for the file list to be populated.

As you know I don't have a deluge test environment so I can only code this *blindly* and rely on others for testing. I'm also not terribly familiar with the `@defer.inlineCallbacks`, hopefully this concept works as I intend it to.

Let me know if you experience any crashes or issues. I've added (temporary) info level logging to simplify the testing of this patch.

### DIFF
--- a/flexget/plugins/plugin_deluge.py
+++ b/flexget/plugins/plugin_deluge.py
@@ -781,20 +781,23 @@ class OutputDeluge(DelugePlugin):
             for entry in task.accepted:
 
                 @defer.inlineCallbacks
-                def _wait_for_files(id, timeout):
+                def _wait_for_metadata(id, timeout):
                     from time import sleep
-                    log.info('wait_for_files(%s, %d)' % (id, timeout))
+                    log.info('Waiting %d seconds for "%s" to magnetize' % (timeout, entry['title']))
                     try:
                         while timeout > 0:
                             sleep(1)
                             status = yield client.core.get_torrent_status(id, ['files'])
                             if len(status['files']) > 0:
-                                defer.returnValue(True)
+                                log.info('"%s" magnetization successful' % (entry['title']))
+                                defer.returnValue(id)
                             else:
                                 timeout -= 1
                     except Exception as err:
-                        log.error('wait_for_files Error: %s' % err)
-                    defer.returnValue(False)
+                        log.error('wait_for_metadata Error: %s' % err)
+
+                    log.warning('"%s" did not magnetize before the timeout elapsed, file list unavailable for processing.' % entry['title'])
+                    defer.returnValue(id)
 
                 def add_entry(entry, opts):
                     """Adds an entry to the deluge session"""
@@ -811,24 +814,10 @@ class OutputDeluge(DelugePlugin):
 
                     log.verbose('Adding %s to deluge.' % entry['title'])
                     if magnet:
-                        # we need to set paused to false so the magnetization begins immediately
-                        # NOTE: is this the correct config property and value?
-                        opts['add_paused'] = False
-
-                        id = client.core.add_torrent_magnet(magnet, opts)
-
-                        magnetization_timeout = entry.get('magnetization_timeout', config.get('magnetization_timeout'))
-                        if magnetization_timeout > 0:
-                            log.info('Waiting %d seconds for "%s" to magnetize' % (magnetization_timeout, entry['title']))
-                            magnetized = _wait_for_files(id, magnetization_timeout).getResult()
-                            if not magnetized:
-                                log.warning('"%s" did not magnetize before the timeout elapsed, file list unavailable for processing.' % entry['title'])
-                            else:
-                                log.info('magnetization successful')
-                        else:
-                            log.info('not waiting for magnetization')
-
-                        return id
+                        d = client.core.add_torrent_magnet(magnet, opts)
+                        if config.get('magnetization_timeout'):
+                            d.addCallback(_wait_for_metadata, config['magnetization_timeout'])
+                        return d
                     else:
                         return client.core.add_torrent_file(entry['title'], filedump, opts)
 


patsissons:
**NOTE**: I'm creating this PR ***VERY*** early because I need others to test the patch. I'll post an update when I think it is code complete.

patsissons:
not sure how i can get the success out of `_wait_for_files`, maybe i don't need to (and just forgo the logging)...

flexget/plugins/plugin_deluge.py
```diff
@@ -780,7 +798,23 @@ def add_entry(entry, opts):
...
+                        # we need to set paused to false so the magnetization begins immediately
+                        # NOTE: is this the correct config property and value?
+                        opts['add_paused'] = False
+                        
+                        id = client.core.add_torrent_magnet(magnet, opts)
+                        
+                        magnetization_timeout = entry.get('magnetization_timeout', config.get('magnetization_timeout'))
+                        if magnetization_timeout > 0:
+                            log.info('Waiting %d seconds for "%s" to magnetize' % (magnetization_timeout, entry['title']))
+                            if _wait_for_files(id, magnetization_timeout) == False:
```
cvium: `if not _wait_for_files(id, magnetization_timeout)` is more pythonic

patsissons: haha yea, can you tell I don't use python very often :)

patsissons:
I'm rebasing this branch onto the latest dev version commit so I can hopefully see a successful build

cvium:
The build is failing because of the commit made by @paranoidi. No amount of rebasing is gonna help :p

patsissons:
it looked like 295 master is building, and im now based off the dev version commit right after it

patsissons:
ahhh I'm dumb. i realize now that travis is just building off a temporary merge of my branch back into upstream's develop, and that is why the build was failing. I'll rebase now that 693f23f has reverted 2bfe5b2.

jpts:
Bee testing and run across the following exception:
```  File "/home/smbuser/flexget-dev/flexget/plugins/plugin_deluge.py", line 773, in _wait_for_files
    log.info('wait_for_files(%d, %d)' % (id, timeout))
exceptions.TypeError: %d format: a number is required, not instance
```

patsissons:
i didn't realize the return value was the id as a string. all fixed now (i also rebased onto the latest develop version).

jpts:
I think it is a actually a Deferred object that is returned from core.add_torrent_magnet()
```
2015-04-26 14:48 INFO     deluge        add-movie-queue wait_for_files(<Deferred at 0x7f2771da7878>, 60)
2015-04-26 14:48 ERROR    deluge        add-movie-queue wait_for_files Error: <type 'instance'>
```

patsissons:
@jpts I believe that `get_torrent_status` returns a string (as per [the RPC docs](http://deluge-torrent.org/docs/1.2/core/rpc.html#deluge.__rpcapi.RpcApi.core.add_torrent_magnet)). The deferred object that you see should be due to the `@defer.inlineCallbacks` decorator on `_wait_for_files`. However this deferred object is *re-synced* using `getResult()`.

That said, I don't really do a lot of python programming, so I'm mostly drawing from my knowledge of how the **async/await** pattern operates in **C#**. Without a test environment for deluge I am truly coding in the dark for this feature.

jpts:
Ah yes, I see. I'm not familiar with deferred objects myself really.

Happy to do more testing / provide more logs/debug if that'd be helpful.

patsissons:
yea, i'll try and queue up another build as soon as the develop build is passing again.

gazpachoking:
@patsissons @jpts I didn't look too close yet, but how are you guys feeling about this?

patsissons:
hadn't looked at this for a while, so i just rebased onto a building develop. let me know how testing goes.

cvium:
Due to the new rTorrent plugin I have stopped using Deluge, so I hope someone else can test it.

patsissons:
I haven't put a whole lot of effort into this yet, so if the interest for this dies out it's no big deal for me. otherwise I'm happy to keep hacking away at it in the mean time.

MikeyCanuck:
I'd be interested in the feature and can probably help with the testing. I think I've retrieved and built it correctly since it doesn't complain about the parameter magnetization_timeout: 60

Here's the error message from the first attempt:

```
2015-06-09 18:27 VERBOSE  deluge        showRSS         Adding Silicon Valley 2x09 Binding Arbitration 720p to deluge.
2015-06-09 18:27 INFO     deluge        showRSS         Waiting 60 seconds for "Silicon Valley 2x09 Binding Arbitration 720p" to magnetize
2015-06-09 18:27 INFO     deluge        showRSS         wait_for_files(<Deferred at 0x7f9dcc245e18>, 60)
2015-06-09 18:27 ERROR    deluge        showRSS         wait_for_files Error: <type 'instance'>
Unhandled error in Deferred:
Unhandled Error
  File "/usr/lib64/python2.7/site-packages/twisted/protocols/policies.py", line 118, in dataReceived
    self.wrappedProtocol.dataReceived(data)
  File "/usr/lib/python2.7/site-packages/deluge/ui/client.py", line 179, in dataReceived
    d.callback(request[2])
  File "/usr/lib64/python2.7/site-packages/twisted/internet/defer.py", line 368, in callback
    self._startRunCallbacks(result)
  File "/usr/lib64/python2.7/site-packages/twisted/internet/defer.py", line 464, in _startRunCallbacks
    self._runCallbacks()
--- <exception caught here> ---
  File "/usr/lib64/python2.7/site-packages/twisted/internet/defer.py", line 551, in _runCallbacks
    current.result = callback(current.result, *args, **kw)
  File "/var/lib/deluge/test/Flexget/flexget/plugins/plugin_deluge.py", line 878, in on_get_session_state
    dlist.append(add_entry(entry, add_opts).addCallbacks(
  File "/var/lib/deluge/test/Flexget/flexget/plugins/plugin_deluge.py", line 823, in add_entry
    magnetized = _wait_for_files(id, magnetization_timeout).getResult()
exceptions.AttributeError: Deferred instance has no attribute 'getResult'

```

patsissons:
Thanks. I think i misread the deferred API. I'll try something a little different.

flexget/plugins/plugin_deluge.py
```diff
@@ -778,6 +780,22 @@ def on_get_session_state(torrent_ids):
...
+                @defer.inlineCallbacks
+                def _wait_for_files(id, timeout):
+                    from time import sleep
+                    log.info('wait_for_files(%s, %d)' % (id, timeout))
+                    try:
+                        while timeout > 0:
+                            sleep(1)
+                            status = yield client.core.get_torrent_status(id, ['files'])
+                            if len(status['files']) > 0:
+                                defer.returnValue(True)
```
gazpachoking: I think you just want `return True` here, and `return False` down below.

flexget/plugins/plugin_deluge.py
```diff
@@ -793,7 +811,24 @@ def add_entry(entry, opts):
...
+                        # we need to set paused to false so the magnetization begins immediately
+                        # NOTE: is this the correct config property and value?
+                        opts['add_paused'] = False
+
+                        id = client.core.add_torrent_magnet(magnet, opts)
+
+                        magnetization_timeout = entry.get('magnetization_timeout', config.get('magnetization_timeout'))
+                        if magnetization_timeout > 0:
+                            log.info('Waiting %d seconds for "%s" to magnetize' % (magnetization_timeout, entry['title']))
+                            magnetized = _wait_for_files(id, magnetization_timeout).getResult()
```
gazpachoking: You should be able to just call this function normally. i.e. drop the `.getResult()`

gazpachoking:
Okay, after actually reading the code and looking at the docs, I think what we want to do as just attach a callback to the `client.core.add_torrent_magnet` call which waits for the metadata before we return the deferred. The logging can be moved to that wait function, and no need to return anything from there anymore.

MikeyCanuck:
I've forked from @patsissons fork, made and tested some changes. It works, if not exactly as I expected. See comments in pull request here:  https://github.com/patsissons/Flexget/pull/1

gazpachoking:
This looks like the right way to do it to me, awesome nested PR action guys. :smile:

gazpachoking:
It still says WIP... but it looked good to me so I merged it. I also made some style tweaks here f7acd442f1507308ced11df39d841f864ad329c9
Hopefully I didn't break anything. :tongue: 
@patsissons @MikeyCanuck One of you want to update the wiki?

patsissons:
cleanup looks good to me. thanks for merging.

patsissons:
@gazpachoking wiki updated.

gazpachoking:
Ooh, maybe we should make the default for this option be something other than 0 when main_file_only or content_filename is used?

patsissons:
the reasoning for the 0 default was to make the option an opt-in for those upgrading. However, if you think there is more benefit to setting an active default value, I say go ahead. Being somewhat new to flexget dev, i didn't want to by default enable some code i just added.

gazpachoking:
Yeah, that reasoning is sound, I'm just thinking the desire to turn this option when setting content_filename or main_file_only is high, since they don't work without this. Probably high enough to put it as default in those cases.

PR #473, opened 2015-03-27
Abstract out torrent file renaming

tubedogg:
Create new plugin to handle renaming of files within torrents.
Currently only works with Deluge. Also render label before applying it.

New options (set under Deluge plugin config):
`keep_container` Default: True. Keeps existing container folder in torrent, if any. (If a multi-file torrent is added to Deluge, Deluge automatically creates a container folder using info.name from the torrent metadata. Setting this to false causes that part of the path to be removed. Similarly, if the torrent file structure itself had a top-level folder that contained all files in the torrent, setting this to False would cause it to be removed from the path.)
`container_directory` Default: blank. Create new container directory around all files within the torrent. Rendered before being applied.
`fix_year` Default: False. When this is enabled and a TV show is being processed, and the last four characters of the series name are a number (almost certainly a year), add parentheses around the year.
`rename_main_file_only` Default: False. Only rename the main file within the torrent. Note that `keep_container` and `container_directory` settings still apply even when this is set to True. In that case, the only filename changed will be the largest file (as determined using `main_file_ratio`), but the paths for the other files will still be changed.
`container_multiple` Default: True. Only applies the `container_directory` setting if there are multiple files within the torrent that will be downloaded. Useful if you have a task that processes both season packs and individual episodes.
`unlisted_filetype_default` Default: True. Whether to rename files whose extensions are not explicitly named in `filetypes` setting (see below). By default, renaming affects all files.
`filetypes` Default: Empty dict. Allows specifying yes/no (yes = rename, no = do not rename) or a different naming template on a per-file extension basis. Example: You want to rename all files except .nfo and .ogg files, and you want to change the template set in `content_filename`, but only for .avi files. Since `unlisted_filetype_default` defaults to True, all other files will be renamed using `content_filename` as a template.
```
deluge:
  filetypes:
    nfo: no
    ogg: no
    avi: '{{series_name}} (AVI)'
```

`content_filename` template from the existing Deluge config is used to change filenames (and paths, if `content_filename` contains directory separators), unless overridden on a per-file extension basis.
All options respect `hide_sparse_files` and `keep_subs` as provided by existing Deluge config.

### DIFF
--- a/flexget/plugins/modify/torrent_renamefiles.py
+++ b/flexget/plugins/modify/torrent_renamefiles.py
@@ -2,6 +2,7 @@ from __future__ import unicode_literals, division, absolute_import
 import logging
 import posixpath
 import os
+import re
 
 from flexget import plugin
 from flexget.event import event
@@ -22,6 +23,7 @@ class TorrentRenameFiles(object):
                 'properties': {
                     'keep_container': {'type': 'boolean'},
                     'container_directory': {'type': 'string'},
+                    'container_multiple': {'type': 'boolean'},
                     'rename_main_file_only': {'type': 'boolean'},
                     'fix_year': {'type': 'boolean'},
                 },
@@ -33,6 +35,7 @@ class TorrentRenameFiles(object):
     def prepare_config(self, config):
         config.setdefault('keep_container', True)
         config.setdefault('container_directory', '')
+        config.setdefault('container_multiple', '')
         config.setdefault('fix_year', False)
         config.setdefault('rename_main_file_only', False)
         return config
@@ -42,6 +45,11 @@ class TorrentRenameFiles(object):
         file_ratio = status['total_size'] * config.get('main_file_ratio')
         log.debug('File ratio: %s' % file_ratio)
         log.debug('Torrent size: %s' % status['total_size'])
+
+        if not entry['main_fileid']:
+            entry['main_fileid'] = None
+        if not entry['sub_fileid']:
+            entry['sub_fileid'] = None
 
         series_tokens = False
         if config.get('content_filename'):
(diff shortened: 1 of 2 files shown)


gazpachoking:
I like the idea, a few suggestions/ideas:
- I don't think the method should be called on_task_rename, since this isn't actually a task phase.
- Since this isn't a real task phase, perhaps the interface could be to pass it the entry and the list of paths and sizes directly, (rather than reading content_files from the entry,) and it  would return a list with the new name and any other options it's capable of setting. 
- We could move all these options under a new config key in both deluge and transmission plugin, then they could just use the schema directly from the util plugin for that config key.
- I don't think things like `main_fileid` should be stored in the entry dict

tubedogg:
I am leaving for the weekend in the morning so I committed my changes/fixes from the last few days. In regards to your specific comments:
- Renamed `on_task_rename` to `rename_files`
- The comment in torrent_renamefiles.py was misleading and hadn't been updated since I started coding this. Currently `entry[content_files]` is not used at all. Deluge's `status` dict and the current `entry` are passed to the plugin, which returns a dict in the form of `file[index] = new_filename` where `index` matches the file's index number from Deluge.
- I'm not sure exactly what you mean by "under a new config key" - meaning a new top-level plugin key, or a sub-key that is the same for both deluge and transmission? Either way my intention was to eventually move it to its own plugin key, since I knew from our discussions last year the intention was to use this with transmission too.
- I'm not sure where else to store main_fileid and subs_fileid, unless the `status` dict is modified before passing it to `rename_files`.

gazpachoking:
> I'm not sure where else to store main_fileid and subs_fileid, unless the status dict is modified before passing it to rename_files.

I was thinking we return a list of the new filenames along with do not download flags. Perhaps that's creeping in a bit of separate functionality, but the logic is often combined.

stevezau:
@tubedogg  any updates on this?

liiight:
Reopen if relevant

PR #474, opened 2015-03-27
Plugin to decompress Zip and RAR files

ianstalk:
This is a plugin for extracting Zip/RAR files within Flexget with an interface similar to the "move" plugin. To extract RAR files, it requires the rarfile Python package, and RARLabs Unrar, both of which are available for free.


flexget/plugins/plugin_rarfile.py
```diff
@@ -0,0 +1,196 @@
+from __future__ import unicode_literals, division, absolute_import
+import logging
+import os
+import rarfile
```
cvium: I think it'd be best to have some conditional importing so as not to break everything if `rarfile` is not installed.

ianstalk: Good point. I've fixed that in my latest commit.

flexget/plugins/plugin_rarfile.py
```diff
@@ -0,0 +1,211 @@
...
+
+from flexget import plugin
+from flexget.entry import Entry
+from flexget.event import event
+from flexget.utils.template import render_from_entry, RenderError
+
+try:
+    import rarfile
+except ImportError:
+    pass
```
gazpachoking: You can just set `rarfile = None` here, then test for that at runtime instead of the current hack. :wink:

gazpachoking:
Seems good to me. Would it be better though to make it a general extraction plugin? Perhaps this could be used https://pypi.python.org/pypi/pyunpack instead of rarfile. (not vouching for the quality of that particular library, just an idea)

gazpachoking:
It would also be cool to leverage #473 (once we figure out exactly how that should work) in here to rename the contents once they are extracted.

ianstalk:
I'll look into it. At the very least, it should be simple to add ZIP support since the rarfile API intentionally mirrors the built-in zipfile module. That'd get us 80% there in terms of what archives you're likely to encounter.

ianstalk:
Pyunpack and the underlying patool package it uses seem pretty limited it what they can do. You can extract a complete archive, but there's no way to extract files individually. regexp filtering and smart renaming would be out.

gazpachoking:
Cool, zip and rar sound good enough to me then.

gazpachoking:
Sweet! Want to make a wiki page?

gazpachoking:
Also, does this work in a task that has download plugin to extract the downloaded thing? If not it would be nice if we could make it.

gazpachoking:
Also, we should make sure the priorities are okay so that it can work combined with move plugin.

PR #475, opened 2015-03-27
Allow list of exec commands

Ram-Z:
Not sure if this is the best approach. Maybe add a new `outputs` plugin similar to `inputs`?

The diff is a bit misleading as I had to indent large blocks, here it is again with `-w`

```diff
diff --git a/flexget/plugins/output/exec.py b/flexget/plugins/output/exec.py
index 7efeda0..b9d04f9 100644
--- a/flexget/plugins/output/exec.py
+++ b/flexget/plugins/output/exec.py
@@ -6,6 +6,7 @@ import subprocess
 
 from flexget import plugin
 from flexget.event import event
+from flexget.config_schema import one_or_more
 from flexget.utils.template import render_from_entry, render_from_task, RenderError
 from flexget.utils.tools import io_encoding
 
@@ -67,7 +68,7 @@ class PluginExec(object):
 
     schema = {
         'oneOf': [
-            {'type': 'string'},
+            one_or_more({'type': 'string'}),
             {
                 'type': 'object',
                 'properties': {
@@ -88,11 +89,11 @@ class PluginExec(object):
             'phaseSettings': {
                 'type': 'object',
                 'properties': {
-                    'phase': {'type': 'string'},
-                    'for_entries': {'type': 'string'},
-                    'for_accepted': {'type': 'string'},
-                    'for_rejected': {'type': 'string'},
-                    'for_failed': {'type': 'string'}
+                    'phase': one_or_more({'type': 'string'}),
+                    'for_entries': one_or_more({'type': 'string'}),
+                    'for_accepted': one_or_more({'type': 'string'}),
+                    'for_rejected': one_or_more({'type': 'string'}),
+                    'for_failed': one_or_more({'type': 'string'})
                 },
                 'additionalProperties': False
             }
@@ -104,6 +105,12 @@ class PluginExec(object):
             config = {'on_output': {'for_accepted': config}}
         if not config.get('encoding'):
             config['encoding'] = io_encoding
+        for phase_name in config:
+            if phase_name.startswith('on_'):
+                for items_name in config[phase_name]:
+                    if isinstance(config[phase_name][items_name], basestring):
+                        config[phase_name][items_name] = [config[phase_name][items_name]]
+
         return config
 
     def execute_cmd(self, cmd, allow_background, encoding):
@@ -136,7 +143,7 @@ class PluginExec(object):
             log.debug('running phase_name: %s operation: %s entries: %s' % (phase_name, operation, len(entries)))
 
             for entry in entries:
-                cmd = config[phase_name][operation]
+                for cmd in config[phase_name][operation]:
                     entrydict = EscapingDict(entry) if config.get('auto_escape') else entry
                     # Do string replacement from entry, but make sure quotes get escaped
                     try:
@@ -167,7 +174,7 @@ class PluginExec(object):
 
         # phase keyword in this
         if 'phase' in config[phase_name]:
-            cmd = config[phase_name]['phase']
+            for cmd in config[phase_name]['phase']:
                 try:
                     cmd = render_from_task(cmd, task)
                 except RenderError as e:
```

### DIFF
--- a/flexget/plugins/output/exec.py
+++ b/flexget/plugins/output/exec.py
@@ -6,6 +6,7 @@ import subprocess
 
 from flexget import plugin
 from flexget.event import event
+from flexget.config_schema import one_or_more
 from flexget.utils.template import render_from_entry, render_from_task, RenderError
 from flexget.utils.tools import io_encoding
 
@@ -67,7 +68,7 @@ class PluginExec(object):
 
     schema = {
         'oneOf': [
-            {'type': 'string'},
+            one_or_more({'type': 'string'}),
             {
                 'type': 'object',
                 'properties': {
@@ -88,11 +89,11 @@ class PluginExec(object):
             'phaseSettings': {
                 'type': 'object',
                 'properties': {
-                    'phase': {'type': 'string'},
-                    'for_entries': {'type': 'string'},
-                    'for_accepted': {'type': 'string'},
-                    'for_rejected': {'type': 'string'},
-                    'for_failed': {'type': 'string'}
+                    'phase': one_or_more({'type': 'string'}),
+                    'for_entries': one_or_more({'type': 'string'}),
+                    'for_accepted': one_or_more({'type': 'string'}),
+                    'for_rejected': one_or_more({'type': 'string'}),
+                    'for_failed': one_or_more({'type': 'string'})
                 },
                 'additionalProperties': False
             }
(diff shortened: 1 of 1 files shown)


gazpachoking:
Yeah, I'd prefer we be able to just list the plugin twice, but I think this is fine for now. We can simplify a bunch of plugins later if we come up with a good way to use a plugin more than once in a task.

gazpachoking:
Wanna update the wiki? :smile:

flexget/plugins/output/exec.py
```diff
@@ -104,6 +105,12 @@ def prepare_config(self, config):
             config = {'on_output': {'for_accepted': config}}
         if not config.get('encoding'):
```
Ram-Z: Something is wrong here, needs to test if config is an array and deal with it accordingly before.

PR #478, opened 2015-04-04
trakt_list collection by episode

wakemaster39:
previously the trakt_list was unable to produce a list of episodes for the lists *collection* or *watched*. This enables that ability. 

Also the trakt_emit plugin always emitted a *series_name* that was equivalent to the trakt_list with the *strip_dates* field enabled. Added this field to trakt_emit to sync the result of the two plugins by default.

### DIFF
--- a/flexget/plugins/input/trakt_list.py
+++ b/flexget/plugins/input/trakt_list.py
@@ -46,6 +46,16 @@ field_maps = {
         'trakt_episode_id': 'episode.ids.trakt',
         'trakt_show_id': 'show.ids.trakt',
         'trakt_show_slug': 'show.ids.slug'
+    },
+    'collection_episodes': {
+        'title': lambda i: '%s (%s) S%02dE%02d' % (i['show']['title'], i['show']['year'], i['season'],
+                                                 i['episode']),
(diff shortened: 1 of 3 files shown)


flexget/plugins/input/trakt_list.py
```diff
@@ -46,6 +46,16 @@
...
+    },
+    'collection_episodes': {
+        'title': lambda i: '%s (%s) S%02dE%02d' % (i['show']['title'], i['show']['year'], i['season'],
+                                                 i['episode']),
+        'series_name': lambda i: '%s (%s)' % (i['show']['title'], i['show']['year']),
+        'series_season': 'season',
+        'series_episode': 'episode',
+        'series_id': lambda i: 'S%02dE%02d' % (i['season'], i['episode']),
+        'trakt_show_id': 'show.ids.trakt',
+        'trakt_show_slug': 'show.ids.slug'
```
gazpachoking: Why is this needed instead of just the normal episode type?

stevezau:
@wakemaster39 any updates on this?

liiight:
I don't think this is relevant anymore, is it?

PR #479, opened 2015-04-04
added yts

wakemaster39:
Added yts to the search group


wakemaster39:
Updated to include try around the torrents loop. Also added a verification to ensure entry is valid before appending it to the list.

gazpachoking:
Thanks! Want to update the wiki? http://flexget.com/wiki/Searches

crawln45:
Heh. Now that gazpachoking has pulled this request. might want to add in a test for this as well.

PR #481, opened 2015-04-08
Updated to use the download plugin

programatix:
I'm new to Python, so I'm not sure why the download will fail on private trackers. It gives the "maybe a login page" error.

I'm using the same config file I used on my router (with transmission plugin) and it failed. I'm currently running FlexGet on Windows. The task has form plugin.

### DIFF
--- a/flexget/plugins/output/utorrent.py
+++ b/flexget/plugins/output/utorrent.py
@@ -53,13 +53,18 @@ class PluginUtorrent(object):
         # our temp .torrent files
         if 'download' not in task.config:
             download = plugin.get_plugin_by_name('download')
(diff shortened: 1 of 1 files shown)


programatix:
I think I found out why the download failed. It seems that the urlrewrite is not working.

programatix:
If i remove the usage of download plugin, url rewrite plugin will kick in. Help from experience flexget coder needed.

gazpachoking:
Seems pretty good to me at first glance, is there something that is still not working as expected?

programatix:
Everything is working fine.


Sent from my Samsung device

-------- Original message --------

PR #482, opened 2015-04-11
Allow history to be filtered by task

matrixagent:
I'd love for history to get a few more options, maybe also a date-range instead of a count-based limit. Ideally, --search should check all (or most) fields and every field should be searchable on its own via dedicated parameter. What do you think?


gazpachoking:
Looks good to me, and I'm not opposed to adding more search options.

PR #485, opened 2015-04-14
Add what.cd user agent option

pR0Ps:
- Add back the option to use a custom user agent for searching on what.cd.
- Fix small oversight in response handling.


gazpachoking:
Don't we need a schema update as well?

pR0Ps:
Nope, it's [already there](https://github.com/Flexget/Flexget/blob/pR0Ps-add_whatcd_ua/flexget/plugins/input/whatcd.py#L179). It just wasn't removed when the use of the actual value was removed.

gazpachoking:
> Nope, it's already there. It just wasn't removed when the use of the actual value was removed.

Oh, gotcha. Thanks!

PR #486, opened 2015-04-14
Attach external site IDs to series table

Chemondelay:
Added new columns to the series table for tvdb_id, imdb_id, tmdb_id and trakt_show_id. These can be manually specified on a per-series basis through the config. This allows potentially more accurate searching through the discover plugin, since entries created by emit_series will have the equivalent fields populated if they are provided.

I don't know if it's worthwhile making any changes to the way the series filter parses returned entries, but hopefully this is useful as a start. I've tested it a bit (to make sure empty fields don't mess with searching for one), but let me know if there's anything I've missed.

### DIFF
--- a/flexget/plugins/filter/series.py
+++ b/flexget/plugins/filter/series.py
@@ -146,6 +146,13 @@ def upgrade(ver, session):
         from flexget.task import config_changed
         config_changed(session=session)
         ver = 12
+    if ver == 12:
+        log.info('Adding `tvdb_id`, `imdb_id`, `tmdb_id` and `trakt_show_id` columns to series table.')
+        table_add_column('series', 'tvdb_id', Integer, session)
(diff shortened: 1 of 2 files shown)


paranoidi:
This violates separation / coupling and makes it inconvenient adding new sites. More generic solution would be needed. One possibility would be adding columns like "site_id" and "site_name" .. but even that requires some more thinking ...

Chemondelay:
Totally agree, anything more complex is a bit beyond my abilities though. Open to any suggestions.

stevezau:
Hmm, i was discussing a similar idea with @gazpachoking .. We need a generic way to map a series to a site. Each site has it's own unique id and each site can name a series differently so what could we use for the unique id?

gazpachoking:
I'm still not sure best answer, but we can continue discussion if someone gets inspired.

PR #488, opened 2015-04-19
Plugins to download from SFTP

ianstalk:
SFTP plugins patterned after ftp_list and ftp_download. This plugin requires the pysftp Python modules and its dependencies.

### DIFF
--- a/flexget/plugins/plugin_sftp.py
+++ b/flexget/plugins/plugin_sftp.py
@@ -293,18 +293,11 @@ class SftpDownload(object):
         port = parsed.port or 22
 
         # get private key info if it exists
-        try:
-            private_key = entry['private_key']
-        except KeyError:
-            private_key = None
-
-        try:
-            private_key = entry['private_key_pass']
-        except KeyError:
-            private_key_pass = None
+        private_key = entry.get('private_key')
+        private_key_pass = entry.get('private_key_pass')
 
         if parsed.scheme == 'sftp':
-            config= ConnectionConfig(host, port, username, password, private_key, private_key_pass)
+            config = ConnectionConfig(host, port, username, password, private_key, private_key_pass)
         else:
             log.warn('Scheme does not match SFTP: %s' % entry['url'])
             config = None
--- a/flexget/plugins/plugin_sftp.py
+++ b/flexget/plugins/plugin_sftp.py
@@ -329,7 +329,7 @@ class SftpDownload(object):
         log.verbose('Downloading file %s to %s' % (path, destination))
 
         try:
-            sftp.get(path, destinationƒ)
+            sftp.get(path, destination)
         except Exception as e:
             log.error('Failed to download %s (%s)' % (path, e))
             if remotepath.exists(destination):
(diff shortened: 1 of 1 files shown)


flexget/plugins/plugin_sftp.py
```diff
@@ -0,0 +1,423 @@
...
+              get_size: True
+              files_only: False
+          dirs:
+              - '/path/to/list/'
+    """
+
+    schema = {
+        'type': 'object',
+        'properties': {
+            'config': {
```
gazpachoking: Do we need a separate key for this?

flexget/plugins/plugin_sftp.py
```diff
@@ -0,0 +1,423 @@
...
+        url_prefix = 'sftp://%s%s%s/' % (login_str, host, port_str)
+        
+        log.debug('Connecting to %s' % host)
+
+        conn_conf = ConenctionConfig(host, port, username, password, private_key, private_key_pass)
+        
+        try:
+            sftp = sftp_connect(conn_conf)
+        except Exception as e:
+            log.error('Failed to connect to %s (%s)' % (host, e))
```
gazpachoking: Maybe this should raise a PluginError

flexget/plugins/plugin_sftp.py
```diff
@@ -0,0 +1,423 @@
...
+            log.warn('Skipping unknown file %s' % path)
+
+    def on_task_download(self, task, config):
+        """
+        Task handler for sftp_download plugin
+        """
+        dependency_check()
+        config = self.prepare_config(config, task)
+
+        # Download entries by host so we can reuse the connection
```
gazpachoking: Should we be filtering to only try to process entries using the sftp uri scheme?

flexget/plugins/plugin_sftp.py
ianstalk: I just did it that way to keep it similar to the existing FTP/FTPS plugin. Not sure why the author of that plugin did it that way, so I'm not particularly attached to the idea.

flexget/plugins/plugin_sftp.py
```diff
@@ -0,0 +1,461 @@
...
+        host = parsed.hostname
+        username = parsed.username or None
+        password = parsed.password or None
+        port = parsed.port or 22
+
+        # get private key info if it exists
+        try:
+            private_key = entry['private_key']
+        except KeyError:
+            private_key = None
```
gazpachoking: Dicts already have a method for this, `private_key = entry.get('private_key')`

flexget/plugins/plugin_sftp.py
```diff
@@ -0,0 +1,461 @@
...
+
+        try:
+            private_key = entry['private_key_pass']
```
gazpachoking: This meant to be `private_key_pass = ...`? Also, same note as above.

flexget/plugins/plugin_sftp.py
```diff
@@ -0,0 +1,461 @@
...
+        except KeyError:
+            private_key_pass = None
+
+        if parsed.scheme == 'sftp':
+            config= ConnectionConfig(host, port, username, password, private_key, private_key_pass)
```
gazpachoking: Spaces around operators.

gazpachoking:
Thanks!

PR #489, opened 2015-04-21
Fix deluge sparse files/movedone functionality

jpts:
Fixed http://flexget.com/ticket/2881

Makes sure there is only one top level directory, enabling correct movedone functionality in deluge.

This is accomplished by ensuring the content_filename option doesn't pop the main file / sub file out of folder and also by changing the location of the sparse files (top_torrent_folder/.sparse_files)

### DIFF
--- a/flexget/plugins/plugin_deluge.py
+++ b/flexget/plugins/plugin_deluge.py
@@ -663,13 +663,15 @@ class OutputDeluge(DelugePlugin):
 
                         if opts.get('content_filename'):
                             # rename the main file
-                            big_file_name = opts['content_filename'] + os.path.splitext(main_file['path'])[1]
+                            big_file_name = os.path.dirname(main_file['path']) + "/"\
+                                            + opts['content_filename'] \
+                                            + os.path.splitext(main_file['path'])[1]
                             big_file_name = unused_name(big_file_name)
                             rename(main_file, big_file_name)
 
                             # rename subs along with the main file
                             if sub_file is not None and keep_subs:
-                                sub_file_name = os.path.splitext(big_file_name   )[0] \
+                                sub_file_name = os.path.splitext(big_file_name)[0] \
                                               + os.path.splitext(sub_file['path'])[1]
                                 rename(sub_file, sub_file_name)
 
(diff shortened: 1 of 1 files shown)


gazpachoking:
Cool, I think this is the right fix.

Would it be good to make it so that if content_filename contains a directory, it renames the top level torrent folder?

jpts:
Good idea! I was actually in the middle of writing something to that effect anyway, using another config option. Keeping it all in content_filename is more elegant I feel.

A a side effect this supports multiple directories theoretically - don't know if there is a depth limit on the deluge side of things.

gazpachoking:
Thanks!

ratoaq2:
This fix broke my configuration.

This is my configuration:
```
    deluge:
      host: localhost
      port: 58846
      username: **
      password: **
      main_file_only: true
      hide_sparse_files: true
      content_filename: "{{trakt_series_name|default(series_name)|replace(':',' -')}} - {{series_id}} - {{tvdb_ep_name|default('Unknown')}}"
      movedone: "/home/osmc/Public/TV Shows/{{trakt_series_name|default(series_name)|replace(':',' -')}}/Season {{series_season|pad(2)}}/"
```

Before it was correctly moving only the main file to the correct place:
`home/osmc/Public/TV Shows/CSI - Cyber/Season 01/CSI - Cyber - S01E12 - Bit by Bit.mkv`

But now, it's moving the main file but with its parent folder and this is definitely not the behaviour that I was expecting:
`home/osmc/Public/TV Shows/CSI - Cyber/Season 01/CSI.Cyber.S01E12.720p.HDTV.X264-**/CSI - Cyber - S01E12 - Bit by Bit.mkv `

jpts:
Although the previous behaviour (of using multiple top level files) is probably preferential in the long run, it isn't currently supported in deluge ([ticket](http://flexget.com/ticket/2881)) and seems to provide inconsistent functionality when used in combination with *movedone*.

This fix changes FlexGet functionality to work in line with the [documentation](http://flexget.com/wiki/Plugins/deluge) This covers two points:
* making sure *main_file_only* doesn't touch the file structure of the torrent, only setting the *do_not_download* status of individual files
* sparse files are moved to a subdirectory instead of a top level one

Having said this, it does seem surprising the you had the previous code working consistently - perhaps it wasn't as broken as previously thought? What happened to the sparse files from your torrents before?

The fix also extends functonality by accepting paths in *content_filename*. This means you can rename the top folder in the torrent (see [docs](http://flexget.com/wiki/Plugins/deluge)).

I currently use a combination of this and a script triggered by the deluge execute plugin to move the main file up one level once complete.

ratoaq2:
I understand your point but you need to agree it must have a way to simply configure flexget to move the main file to the correct place with the correct name, no matter the torrent folder structure.

I have been using a deluge plugin to delete the sparse files (they're no problem for me)

nirajp13:
@gazpachoking Sorry for dragging this one up, but I would appreciate it if you could share that deluge execute script.

nirajp13:
@jpts Sorry for dragging this one up, but I would appreciate it if you could share that execute script.

jpts:
I have this set up to execute on torrent removed https://gist.github.com/jpts/e99347902976f81c1570723af6ffac2e

PR #490, opened 2015-04-23
Add a Basic alternate_name to Series

soloam:
Add's a Simple alternate_name to Series without special chars, this will improve the matching in discover and in series matching

I'm a long time programmer, but never worked in python before. This change will solve the problems in discover, when the name contains special chars. Making obsolete the use os Series to define "alternate_name".

### DIFF
--- a/flexget/plugins/filter/series.py
+++ b/flexget/plugins/filter/series.py
@@ -3,6 +3,8 @@ import argparse
 import logging
 import re
 import time
+import unicodedata
+import string
 from copy import copy
 from datetime import datetime, timedelta
 
@@ -1010,6 +1012,9 @@ class FilterSeries(FilterSeriesBase):
                         alts = [alts]
                     for alt in alts:
                         _add_alt_name(alt, db_series, series_name, session)
+
+                    _add_alt_name_dummy(db_series, series_name, session)
+
                 if not series_name in found_series:
                     continue
                 series_entries = {}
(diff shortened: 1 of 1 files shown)


soloam:
I don't know if this will be accepted or not, but I see this as a temporary solution, to face the problem in discover and other plugins that can't handle the special characters. A good solution would be to do this replacement before discovering a title or testing the title match.

I'm not a python programmer, I must admit my limitations on this mater, this is a fast as workable workaround.

Cheers

gazpachoking:
I'd say that each search plugin should be responsible for sanitizing the search string, since different sites will have different limitations on what characters are allowed. Am I understanding the issue you are trying to solve here?

cvium:
One issue that comes to mind is when a title is Some Show's Apostrophe when search engines support apostrophes but the uploaders have "decided" to exclude the apostrophe eg. America's Next Top Model.

This issue is not really related to any one site and needing a custom alternate name just for this is cumbersome.

gazpachoking:
If our issue is having a standard way to normalize the name, I don't think we need to store it as a separate alternate name, we could just have a utility function which search plugins could use if they wanted. We already have some utility functions for this here https://github.com/Flexget/Flexget/blob/develop/flexget/utils/search.py#L9-L31

soloam:
that seems to be the solution from what I see not all the search plugins use it

soloam:
This can be closed, a new change is in #492 and #493 

Thank you

soloam:
This can be closed, a new change is in #492 and #493 

Thank you

PR #492, opened 2015-04-30
Add's some special cases to clean_title

soloam:
Add's some special cases to clean_title and removes all special characters

### DIFF
--- a/flexget/utils/search.py
+++ b/flexget/utils/search.py
@@ -12,12 +12,8 @@ def clean_symbols(text):
     if isinstance(result, unicode):
         result = normalize('NFKD', result)
     result = re.sub('[ \(\)\-_\[\]\.]+', ' ', result).lower()
-    
-    """Special cases"""
-    result = re.sub("(\'s)+", 's', result)
-    result = re.sub("(\'t)+", 't', result)
 
-    """Leftovers"""
+	# Leftovers
     result = re.sub(r"[^a-zA-Z0-9 ]", "", result)
 
     return result
@@ -41,7 +37,6 @@ def normalize_unicode(text):
 
 def torrent_availability(seeds, leeches):
     """Returns a rating based on seeds and leeches for a given torrent.
-
     :param seeds: Number of seeds on the torrent
     :param leeches: Number of leeches on the torrent
     :return: A numeric rating
--- a/flexget/utils/search.py
+++ b/flexget/utils/search.py
@@ -11,7 +11,16 @@ def clean_symbols(text):
     result = text
     if isinstance(result, unicode):
         result = normalize('NFKD', result)
-    return re.sub('[ \(\)\-_\[\]\.]+', ' ', result).lower()
+    result = re.sub('[ \(\)\-_\[\]\.]+', ' ', result).lower()
+    
+    """Special cases"""
+    result = re.sub("(\'s)+", 's', result)
+    result = re.sub("(\'t)+", 't', result)
+
+    """Leftovers"""
+    result = re.sub(r"[^a-zA-Z0-9 ]", "", result)
+
+    return result
 
 
 def clean_title(title):


soloam:
Sorry #492 and #493 where supposed to be in the same patch, but they ended up separated, one is depended of the other.

flexget/utils/search.py
```diff
@@ -11,7 +11,16 @@ def clean_symbols(text):
...
         result = normalize('NFKD', result)
-    return re.sub('[ \(\)\-_\[\]\.]+', ' ', result).lower()
+    result = re.sub('[ \(\)\-_\[\]\.]+', ' ', result).lower()
+    
+    """Special cases"""
+    result = re.sub("(\'s)+", 's', result)
+    result = re.sub("(\'t)+", 't', result)
+
+    """Leftovers"""
+    result = re.sub(r"[^a-zA-Z0-9 ]", "", result)
```
gazpachoking: Will this strip out non-ascii characters?

flexget/utils/search.py
```diff
@@ -11,7 +11,16 @@ def clean_symbols(text):
     result = text
     if isinstance(result, unicode):
         result = normalize('NFKD', result)
-    return re.sub('[ \(\)\-_\[\]\.]+', ' ', result).lower()
+    result = re.sub('[ \(\)\-_\[\]\.]+', ' ', result).lower()
+    
+    """Special cases"""
```
gazpachoking: Regular comments should just begin with a `#`. And I'm still not quite convinced this is the best approach in what is meant to be a general utility. Might not some (less smart) search engines have a problem giving results like `marvel's` when the input was `marvels`?

flexget/utils/search.py
soloam: This will remove all special chars like "à" "ã" and replace it by basic form "a" for example, replace  basic chars like "_" "-" by space (to accommodate words like Spider-Man and make it Spider Man). and in the end remove all special chars, making words like Marvel's to Marvel.

flexget/utils/search.py
soloam: A lot of the searches on flexget like "Kat" and "PiratBay" are giving errors because they can't find shows like Marvel's etc and Grey's Anatomy

flexget/utils/search.py
gazpachoking: Sure about that?
```
>>> re.sub('[^a-zA-Z0-9 ]', '', 'bàãb')
'bb'
```

soloam: shouldn't result = normalize('NFKD', result) handle that?

gazpachoking: Ahh you are right. In that case, this whole thing ends up being just the same as `re.sub(r"[^a-zA-Z0-9 ]", "", normalize('NFKD', text))`, no? I don't think any of the other stuff does anything.

gazpachoking: I guess I'm fine with the extra cleaning then. If other sites turn out to need different kinds of cleaning we can revisit and make stuff less or more specific here as needed.

soloam: normalize('NFKD', text)) would remove á ã etc etc

result = re.sub('[ \(\)\-_\[\]\.]+', ' ', result).lower() will remove chars like - and _ and replace them with spaces

result = re.sub(r"[^a-zA-Z0-9 ]", "", result) will trim out the remaining special chars.

request #493 will add this functionality to kat, I'll make the same change to PiratBay

soloam:
I uniformed the changes to the code, clean it up.

I made some tests with my changes in kat searches and it works #493

Interesting fact. If you go to kat search and look for:
https://kat.cr/usearch/marvels%20daredevil%20s01e01%20verified%3A1/
https://kat.cr/usearch/marvel%20s%20daredevil%20s01e01%20verified%3A1/

The results are not the same. Makes me wonder if the approach in #490 isn't the best one. It would be common to all searches, it would work in all scenarios and it would return more reliable results, eliminating bad search engines coding.

But for now the #492 and #493 will do it

soloam:

Thank You

soloam:
hello all. any news regarding this?

PR #493, opened 2015-04-30
Strips Special Charters from KAT

soloam:
Removes special charters from KAT search


soloam:
Sorry #492 and #493 where supposed to be in the same patch, but they ended up separated, one is depended of the other.

soloam:

Thank You

liiight:
@soloam is this still relevant?

soloam:
@liiight hello, yes, this would add the clean_title to kat search! The one #492.

PR #495, opened 2015-05-03
Rtorrent plugin

stevezau:
Initial release of RTorrent Plugin.

 - Currently supports output to rtorrent over scgi or http(s).
 - No third party libs required.
 - Plugin will also utilize options from entry fields

Example:

>      rtorrent:
>        url: scgi://localhost:5000
>        username: myusername (http(s) Only)
>        password: mypassword (http(s) Only)
>        path: /some/download/folder
>        priority: off/low/medium/high


stevezau:
Handing over to @RScully to complete

PR #496, opened 2015-05-03
fix for titles with non-ascii characters

sergiogiogio:
This search plugin does not handle non-ascii titles (fails with error UnicodeDecodeError: 'ascii' codec can't decode byte 0xc3 in position XXX: ordinal not in range(128)). This fix makes sure we encode the title properly to be used in a url search string.


gazpachoking:
Thanks!

PR #497, opened 2015-05-04
 path_select plugin allows setting a field to a folder based on it's spa...

stevezau:
Allows setting a field to a folder based on it's space. Plugin will randomly select a drive If one or more drive meets the criteria.

Example:
Find path with highest percent free space and set path to <path>/TVHD

```
  path_select:
    select: most_free_percent # or most_free, most_used, most_used_percent
    threshold: 2 # Include any drives that are within 2% difference of the most free drive
    to_field: path # default
    paths:
      - /drive1/
      - /drive2/
      - /drive3/
    set:
      path: '{{ path }}/TVHD'
```

### DIFF
--- a/flexget/plugins/modify/path_select.py
+++ b/flexget/plugins/modify/path_select.py
@@ -6,53 +6,10 @@ from collections import namedtuple
 
 from flexget import plugin
 from flexget.event import event
+from flexget.config_schema import parse_size, parse_percent
 from flexget.config_schema import one_or_more
 
 log = logging.getLogger('path_select')
-
-SYMBOLS = {
-    'customary': ('B', 'K', 'M', 'G', 'T', 'P', 'E', 'Z', 'Y'),
-    'customary_ext': ('byte', 'kilo', 'mega', 'giga', 'tera', 'peta', 'exa', 'zetta', 'iotta'),
-    'iec': ('Bi', 'Ki', 'Mi', 'Gi', 'Ti', 'Pi', 'Ei', 'Zi', 'Yi'),
-    'iec_ext': ('byte', 'kibi', 'mebi', 'gibi', 'tebi', 'pebi', 'exbi', 'zebi', 'yobi'),
-}
-
-
-def human2bytes(s):
-    """
-    Attempts to guess the string format based on default symbols
-    set and return the corresponding bytes as an integer.
-    When unable to recognize the format ValueError is raised.
-    """
-
-    init = s
-    num = ""
-    while s and s[0:1].isdigit() or s[0:1] == '.':
-        num += s[0]
-        s = s[1:]
-    num = float(num)
-    letter = s.strip()
-    for name, sset in SYMBOLS.items():
-        if letter in sset:
-            break
-    else:
-        if letter == 'k':
-            # treat 'k' as an alias for 'K' as per: http://goo.gl/kTQMs
-            sset = SYMBOLS['customary']
-            letter = letter.upper()
-        else:
-            raise ValueError("can't interpret %r" % init)
-    prefix = {sset[0]: 1}
-    for i, s in enumerate(sset[1:]):
-        prefix[s] = 1 << (i+1)*10
-    return int(num * prefix[letter])
-
-
-def percentage(part, whole):
-    try:
-        return 100 * part/whole
-    except ZeroDivisionError:
-        return 0.0
 
 
 disk_stats_tuple = namedtuple(
@@ -62,8 +19,8 @@ disk_stats_tuple = namedtuple(
 )
 
 
-def get_disk_stats(folder):
-    """ Return folder/drive stats in megabytes (free, used, total) """
+def os_disk_stats(folder):
+    """ Return drive free, used and total bytes """
     if os.name == 'nt':
         import ctypes
         free_bytes = ctypes.c_ulonglong(0)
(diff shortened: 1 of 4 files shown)


flexget/plugins/modify/path_select.py
```diff
@@ -0,0 +1,136 @@
...
+from flexget import plugin
+from flexget.event import event
+from flexget.config_schema import one_or_more
+
+log = logging.getLogger('path_select')
+
+
+def percentage(part, whole):
+    try:
+        return 100 * float(part)/float(whole)
```
gazpachoking: Doesn't importing division from future mean we don't have to explicitly turn these in to floats?

flexget/plugins/modify/path_select.py
```diff
@@ -0,0 +1,136 @@
...
+    except ZeroDivisionError:
+        return float(0.0)
```
gazpachoking: `0.0` is already a float literal, no need to use `float` here.

flexget/plugins/modify/path_select.py
```diff
@@ -0,0 +1,136 @@
...
+
+    return free_bytes_mb, used_bytes_mb, total_bytes_mb
+
+
+def get_free_space(folder):
+    return get_disk_stats(folder)[0]
+
+
+def get_used_space(folder):
+    return get_disk_stats(folder)[1]
```
gazpachoking: Maybe make `get_disk_stats` return a named tuple rather than have these access methods?

gazpachoking:
I can't quite decide if this is the nicest method to do this, but it doesn

stevezau:
@gazpachoking I'd like to write some tests for this but not sure how since it needs a disk drive to test against?

stevezau:
Also thinking i should change threshold to be human readable.. 2%, 6GB, 700MB etc..

Is there someway you can validate size/percent in the schema?

gazpachoking:
> Is there someway you can validate size/percent in the schema?

You can use the `pattern` keyword to validate a string value against a regex pattern.

http://json-schema.org/latest/json-schema-validation.html#anchor33

gazpachoking:
> @gazpachoking I'd like to write some tests for this but not sure how since it needs a disk drive to test against?

You could use the [mock](https://mock.readthedocs.org) library (already required for our tests) to patch `os.statvfs` and `ctypes.windll.kernel32.GetDiskFreeSpaceExW` functions, or something along those lines.

gazpachoking:
> You could use the mock library (already required for our tests) to patch os.statvfs and ctypes.windll.kernel32.GetDiskFreeSpaceExW functions, or something along those lines.

Oh, that's silly now that I think about it. You can just mock your `get_disk_stats` function.

stevezau:
@gazpachoking added the test cases and cleaned up the code.

flexget/plugins/modify/path_select.py
```diff
@@ -0,0 +1,216 @@
...
+        stats = os.statvfs(folder)
+        free_bytes = stats.f_bavail * stats.f_frsize
+        total_bytes = stats.f_blocks * stats.f_frsize
+
+    free_bytes = int(free_bytes)
+    total_bytes = int(total_bytes)
+    used_bytes = total_bytes - free_bytes
+
+    free_percent = 0.0 if total_bytes == 0 else percentage(free_bytes, total_bytes)
+    used_percent = 0.0 if total_bytes == 0 else percentage(used_bytes, total_bytes)
```
gazpachoking: I think any added logic should be separate from the function that grabs the actual disk stats from the OS. As is, this logic isn't being tested when running the tests, since the whole function is mocked away. (Granted, it's simple logic in this case, so it probably doesn't matter.) Also, now that you are checking for divide by 0 manually, the `percentage` function seems a bit superfluous.

flexget/plugins/modify/path_select.py
```diff
@@ -0,0 +1,216 @@
...
+
+SYMBOLS = {
+    'customary': ('B', 'K', 'M', 'G', 'T', 'P', 'E', 'Z', 'Y'),
+    'customary_ext': ('byte', 'kilo', 'mega', 'giga', 'tera', 'peta', 'exa', 'zetta', 'iotta'),
+    'iec': ('Bi', 'Ki', 'Mi', 'Gi', 'Ti', 'Pi', 'Ei', 'Zi', 'Yi'),
+    'iec_ext': ('byte', 'kibi', 'mebi', 'gibi', 'tebi', 'pebi', 'exbi', 'zebi', 'yobi'),
+}
+
+
+def human2bytes(s):
```
gazpachoking: Perhaps this should be a utility function for any plugin wishing to accept a file size in the config? We have a similar thing for intervals here
https://github.com/Flexget/Flexget/blob/develop/flexget/utils/tools.py#L320
I guess if we do that we should have a standard way to define one of these in a config schema, once again probably similar to how we do intervals:
https://github.com/Flexget/Flexget/blob/develop/flexget/config_schema.py#L167-L171

flexget/plugins/modify/path_select.py
```diff
@@ -0,0 +1,216 @@
...
+from flexget.event import event
+from flexget.config_schema import one_or_more
+
+log = logging.getLogger('path_select')
+
+SYMBOLS = {
+    'customary': ('B', 'K', 'M', 'G', 'T', 'P', 'E', 'Z', 'Y'),
+    'customary_ext': ('byte', 'kilo', 'mega', 'giga', 'tera', 'peta', 'exa', 'zetta', 'iotta'),
+    'iec': ('Bi', 'Ki', 'Mi', 'Gi', 'Ti', 'Pi', 'Ei', 'Zi', 'Yi'),
+    'iec_ext': ('byte', 'kibi', 'mebi', 'gibi', 'tebi', 'pebi', 'exbi', 'zebi', 'yobi'),
```
gazpachoking: I don't think we need all these possibilities, I think we should just define a standard and document it. The one thing I was also wondering about is what to do about decimal vs binary units... mostly, whether MB should actually mean MiB or not...

stevezau:
@gazpachoking i cant seem to get it to fail again.. not sure why it failed on the previous run..

gazpachoking:
Awesome! Double thanks for all the added tests. :cake:

PR #498, opened 2015-05-04
Add's a advanced Matching System to crossmatch

soloam:
This add's a advanced mode to crossmatch that will make a Regex Search to match strings.

This can be very useful in for example downloading Ebooks or content that can have mixed names, example:
```
  Ebooks:
    rss:
      url: 'http://my_rff_feed
    crossmatch:
      from:
        - listdir:
          - /mnt/Downloads/Ebooks
      fields:
        - title
      action: accept
      advanced: yes
    transmission:
      host: 'localhost'
      path: '/mnt/Downloads/Ebooks/{{ match }}'
```
On my folder I can use:
-- Downloads
------ Ebooks
---------- Python
---------- PHP
---------- Spider Man

This will match on the feed for example:
- Advanced Python programming
- Python_Article_May15
- PHP form Dummies
- The Spider-Man Adventures
- The Amazing spiderMan

A match property is added to the entry to use for example in download path

### DIFF
--- a/flexget/plugins/filter/crossmatch.py
+++ b/flexget/plugins/filter/crossmatch.py
@@ -101,12 +101,12 @@ class CrossMatch(object):
                 v1Re = v1.lower()
                 v1Re = unicodedata.normalize('NFKD', v1Re).encode('ASCII', 'ignore')
                 v1Re = re.sub("[^a-zA-Z0-9 ]", " ", v1Re)
-                v1Re = v1Re.replace(' ','.*')
+                v1Re = v1Re.replace(' ','.?')
                 v1Re = '.*' + v1Re + '.*'
                 v2Re = v2.lower()
                 v2Re = unicodedata.normalize('NFKD', v2Re).encode('ASCII', 'ignore')
                 v2Re = re.sub("[^a-zA-Z0-9 ]", " ", v2Re)
-                v2Re = v2Re.replace(' ','.*')
+                v2Re = v2Re.replace(' ','.?')
                 v2Re = '.*' + v2Re + '.*'
                 
                 log.debug('Performing flexible Regex search %s and %s' % (v1Re, v2Re))
--- a/flexget/plugins/filter/crossmatch.py
+++ b/flexget/plugins/filter/crossmatch.py
@@ -11,9 +11,7 @@ log = logging.getLogger('crossmatch')
 class CrossMatch(object):
     """
     Perform action based on item on current task and other inputs.
-
     Example::
-
       crossmatch:
         from:
           - rss: http://example.com/
@@ -28,7 +26,7 @@ class CrossMatch(object):
             'fields': {'type': 'array', 'items': {'type': 'string'}},
             'action': {'enum': ['accept', 'reject']},
             'from': {'type': 'array', 'items': {'$ref': '/schema/plugins?phase=input'}},
-            'advanced': {'type': 'boolean', 'default': False}
+            'mode': {'enum': ['standard', 'flexible'], 'default': 'standard'},
         },
         'required': ['fields', 'action', 'from'],
         'additionalProperties': False
@@ -38,7 +36,7 @@ class CrossMatch(object):
 
         fields = config['fields']
         action = config['action']
-        advanced = config['advanced']
+        mode = config['mode']
 
         match_entries = []
 
@@ -67,7 +65,7 @@ class CrossMatch(object):
         for entry in task.entries:
             for generated_entry in match_entries:
                 log.trace('checking if %s matches %s' % (entry['title'], generated_entry['title']))
-                common = self.entry_intersects(entry, generated_entry, fields, advanced)
+                common = self.entry_intersects(entry, generated_entry, fields, mode)
                 if common:
                     msg = 'intersects with %s on field(s) %s' % \
                           (generated_entry['title'], ', '.join(common))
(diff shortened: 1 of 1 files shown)


soloam:
I'm open to recommendations, I added a new "advanced" to the properties to keep actual configurations working as intended. It can be removed to make regex default behavior if needed.

soloam:
Added code to normalize fields removing special characters and diacritics

soloam:
I don't know if this change is going to be merged as it is, but I'm open to suggestions to make it better, I think that the crossmatch needs something like this to make it more powerful. I have it running in my system to download magazines and ebooks, but it can be used for other applications.

Thank You

flexget/plugins/filter/crossmatch.py
```diff
@@ -94,10 +99,31 @@ def entry_intersects(self, e1, e2, fields=None):
...
+                v1Re = v1.lower()
+                v1Re = unicodedata.normalize('NFKD', v1Re).encode('ASCII', 'ignore')
+                v1Re = re.sub("[^a-zA-Z0-9 ]", " ", v1Re)
+                v1Re = v1Re.replace(' ','.*')
+                v1Re = '.*' + v1Re + '.*'
+                v2Re = v2.lower()
+                v2Re = unicodedata.normalize('NFKD', v2Re).encode('ASCII', 'ignore')
+                v2Re = re.sub("[^a-zA-Z0-9 ]", " ", v2Re)
+                v2Re = v2Re.replace(' ','.*')
+                v2Re = '.*' + v2Re + '.*'
```
gazpachoking: My worry is that this seems awfully specific to just hide behind an option called 'advanced'

soloam: Idea:

mode: standard (default)
mode: flexible (new mode)

Leaves possibility to future modes.

soloam:
Added mode option, more flexible with futures changes, we can add mode modes upon need

flexget/plugins/filter/crossmatch.py
```diff
@@ -94,10 +97,31 @@ def entry_intersects(self, e1, e2, fields=None):
             log.trace('v1: %r' % v1)
             log.trace('v2: %r' % v2)
 
-            if v1 == v2:
-                common_fields.append(field)
+            if mode == "flexible":
+                v1Re = v1.lower()
+                v1Re = unicodedata.normalize('NFKD', v1Re).encode('ASCII', 'ignore')
+                v1Re = re.sub("[^a-zA-Z0-9 ]", " ", v1Re)
+                v1Re = v1Re.replace(' ','.*')
```
ianstalk: Isn't this going to be way too way too greedy? `Spider.*Man` would match both "Spider-Man" and "Spiders are the bane of man". It also strikes me as a bad idea to treat all non-alphanumeric characters as noise. This is going to cause a lot of false positives for non-English entries.

soloam: I can add a extra mode

mode: standard (default)
mode: flexible (new mode)
mode: no_greedy

In no_greedy I'll replace spaces with .?, this will match Spider Man, Spider-Man and SpiderMan, but not "Spiders Are Man Best Friend".

ianstalk: I don't think adding new modes is the correct solution, it's just going to make this plugin really confusing to configure. Why not use a more specific character class like `\W` so you only match on whitespace?

I'm also still concerned about your choice to strip Unicode characters; this mode will be detrimental for, say, Russian content.

soloam: \W is not the solution, using the same example If I have Spider Man it would reject Spider-Man and SpliderMan. I think that the solution is to replace with ".?", this will match Spider-Man, Spider Man, but not Spiders are Man Best Friend. 

Best Regards

soloam:
Ok, I made the change so that the matching is not that greedy. Replacing ".*" on the black spaces with ".?", this way:

Spider Man would match:
 - Spider Man
 - Spider-Man
 - Spider_Man
 - SpiderMan
 - The Amazing Spider-Man Rocking on Flexget
 - The Amazing Spider Man Vs The Amazing Fly
 - etc
But it would not match:
 - Spiders Are Man Best Friend
 - My Spiders are good for Man

Best Regards

soloam:
This is my first contribution to a project in GitHub, I don't know all the steps, if there is something that I need to do to follow with the merge process, please tell me :)

stevezau:
@gazpachoking @ianstalk thoughts on this?

soloam:

Thank You

soloam:
Hi all, any use for this?

gazpachoking:
@soloam Sorry for the long delay. What if we called this mode 'fuzzy_match' instead of advanced, or something like that. I was also wondering if we could simplify it a bit, maybe do some small normalization on both strings, then use `difflib.SequenceMatcher` to compute a similarity.

soloam:
I had removed the "advanced" option already, now it's called "mode", allowing "standard" and "flexible", this way we can add more mode's in the future. Regex make the matching more flexible to some variations on the middle of words! I used this to auto download e-books and worked very well.

Open to suggestions!

flexget/plugins/filter/crossmatch.py
```diff
@@ -94,10 +97,31 @@ def entry_intersects(self, e1, e2, fields=None):
             log.trace('v1: %r' % v1)
             log.trace('v2: %r' % v2)
 
-            if v1 == v2:
-                common_fields.append(field)
+            if mode == "flexible":
+                v1Re = v1.lower()
```
paranoidi: does not follor pep8

flexget/plugins/filter/crossmatch.py
```diff
@@ -94,10 +97,31 @@ def entry_intersects(self, e1, e2, fields=None):
...
+                v1Re = unicodedata.normalize('NFKD', v1Re).encode('ASCII', 'ignore')
```
paranoidi: will break unicode without any good readon

paranoidi:
Breaks unicode, adds questionable normalization to crossmatch. Coul

PR #500, opened 2015-05-05
Added search SpanisTracker

malkavi:
Added search SpanishTracker using cookies to download torrents. (Based on torrent411)


gazpachoking:
Could we switch these to use requests rather than urllib?

gazpachoking:
Could probably do with a rebase as well to get rid of all the merge commits.

malkavi:
I will cancel this PR, spanishtracker is closed now... Sorry.

PR #502, opened 2015-05-05
Better shields

rbellamy:
Crate.io shield is broken. They seem to be having issues. Since we're using PyPi, let's see those stats.


gazpachoking:
Hmm, those badges don't seem to be working for me either. Also, it's not case sensitive? We are 'FlexGet' on pypi.

rbellamy:
That's fascinating - they were absolutely working for me when I created them. Looks like pypi isn't responding to shields.io - I went to https://img.shields.io/pypi/v/FlexGet.svg and got "vendor|unresponsive" as a badge.

gazpachoking:
Looks like it's working again now, thanks!

PR #503, opened 2015-05-05
Decompress: Close file handles after file is extracted

ianstalk:
Dangling file handles were causing tests to fail under Windows. This is a fix for: http://flexget.com/ticket/3003

### DIFF
--- a/flexget/plugins/plugin_decompress.py
+++ b/flexget/plugins/plugin_decompress.py
@@ -216,6 +216,7 @@ class Decompress(object):
             if not os.path.exists(destination):
                 success = False
                 error_message = ''
+                source = None
 
                 log.debug('Attempting to extract: %s to %s' % (path, destination))
                 try:
@@ -223,7 +224,6 @@ class Decompress(object):
                     source = archive.open(path)
                     with open(destination, 'wb') as target:
                         shutil.copyfileobj(source, target)
-                    source.close()
 
                     log.verbose('Extracted: %s' % path )
                     success = True
@@ -236,6 +236,9 @@ class Decompress(object):
                 except Exception as e:
                     error_message = 'Unexpected error while extracting %s from %s (%s)' % \
                                     (path, archive_path, e)
+                finally:
+                    if source and not source.closed:
+                        source.close()
 
                 if not success:
                     log.error(error_message)
--- a/flexget/plugins/plugin_decompress.py
+++ b/flexget/plugins/plugin_decompress.py
@@ -4,6 +4,7 @@ import os
 import re
 import shutil
 import zipfile
+import contextlib
 
 from flexget import plugin
 from flexget.entry import Entry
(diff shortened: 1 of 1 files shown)


gazpachoking:
Why not just use `shutil.copyfile`?

gazpachoking:
> Why not just use shutil.copyfile?

Oh I see, the archive is the source.

gazpachoking:
It looks like there is a lot of stuff in that try block though, and the exception it's catching isn't specific enough. Could we improve that?

ianstalk:
I'll work on it. Do you suggest multiple try/except blocks, or more specific exception trapping?

crawln45:
Could probably do a single try block with multiple exceptions.

gazpachoking:
I suggest multiple try blocks, otherwise, how do you know what call the error came from? How do you clean up properly if you don't know which handles successfully opened?

gazpachoking:
That would be this, if you haven't used a context manager to open a file before:
```
with open(destination, 'wb') as target:
    shutil.copyfileobj(source, target)
```

ianstalk:
The "with" syntax wasn't working with Python 2.6

gazpachoking:
> The "with" syntax wasn't working with Python 2.6

Should do, we are using it quite a few places already. https://github.com/Flexget/Flexget/blob/develop/flexget/plugins/output/html.py#L41-L42

gazpachoking:
That is absolutely how this should be handled. Unfortunately you can't do multiple clauses in a with statement in 2.6 :disappointed: I think just nesting the 2 with statements is what we'll have to do. And we should still stick some more specific errors to catch on that except block.

ianstalk:
You can't use "with" on ZipFile.open() in Python 2.6. Guess we go with the ugly solution.

flexget/plugins/plugin_decompress.py
```diff
@@ -212,22 +214,39 @@ def handle_entry(self, entry, config):
...
+                    error_message = 'OS error while creating file: %s (%s)' % \
+                                    (destination, e)
+                except (zipfile.BadZipfile, rarfile.Error) as e:
                     error_message = 'Failed to extract file: %s in %s (%s)' % \
                                     (path, archive_path, e)
+                except Exception as e:
+                    error_message = 'Unexpected error while extracting %s from %s (%s)' % \
+                                    (path, archive_path, e)
+                finally:
+                    if source and not source.closed:
```
gazpachoking: If the `archive.open` call fails, source will never be defined, and this will cause another error. This might be the way if context managers aren't supported by `ZipFile.open` on 2.6 https://docs.python.org/2/library/contextlib.html#contextlib.closing

flexget/plugins/plugin_decompress.py
```diff
@@ -212,22 +214,39 @@ def handle_entry(self, entry, config):
...
+                    success = True
+                except (IOError, os.error) as e:
+                    error_message = 'OS error while creating file: %s (%s)' % \
+                                    (destination, e)
+                except (zipfile.BadZipfile, rarfile.Error) as e:
                     error_message = 'Failed to extract file: %s in %s (%s)' % \
                                     (path, archive_path, e)
+                except Exception as e:
+                    error_message = 'Unexpected error while extracting %s from %s (%s)' % \
+                                    (path, archive_path, e)
```
gazpachoking: We have a utility now to log a more detailed report for debugging if you want. Makes it easy for users to post debug logs and traceback should this happen. https://github.com/Flexget/Flexget/blob/develop/flexget/manager.py#L899

flexget/plugins/plugin_decompress.py
ianstalk: Line 219 should take care of that.

gazpachoking:
Awesome, thanks!

PR #504, opened 2015-05-06
UI Enhancements

stevezau:
Load UI plugins from <user dir>/ui_plugins

Load UI plugins from env variable FLEXGET_UI_PLUGIN_PATH


stevezau:
@gazpachoking can you please create a ui_enhancements branch or something similar on flexget?

gazpachoking:
I'm thinking the webui should be moved to be a plugin is `flexget.plugins.daemon` package, and webui plugins could move to `flexget.plugins.webui`. Then we could just eliminate the duplicated plugin import machinery.
@paranoidi You agree?

stevezau:
scrap this and move to rest_api branch

PR #505, opened 2015-05-08
remove useless extra TLDs

TauPan:
DNS requests for torrentz.eu lead to the same two IP-Adresses as those for .ch, .in, .me and those extra requests make it hard to figure out working domain_delay, limit and interval settings for torrentz.eu (receiving a timeout from torrentz.eu and then hammering on the other domains might add penalty to an existing ban).

(I intend to file a separate issue/PR for "uncivilised behaviour" of the torrentz plugin, possibly resulting in bans. I need to investigate with this branch first, though.)

### DIFF
--- a/flexget/plugins/urlrewrite_torrentz.py
+++ b/flexget/plugins/urlrewrite_torrentz.py
@@ -61,7 +61,7 @@ class UrlRewriteTorrentz(object):
         entries = set()
         for search_string in entry.get('search_strings', [entry['title']]):
             query = normalize_unicode(search_string+config.get('extra_terms', ''))
-            for domain in ['eu', 'me', 'ch', 'in']:
+            for domain in ['eu']:  # , 'me', 'ch', 'in' point to the same two addresses!
                 # urllib.quote will crash if the unicode string has non ascii characters, so encode in utf-8 beforehand
                 url = 'http://torrentz.%s/%s?q=%s' % (domain, feed, urllib.quote(query.encode('utf-8')))
                 log.debug('requesting: %s' % url)


TauPan:
After some testing, it appears that I can successfully configure the torrentz plugin to behave "civilised" enough in order to find some working settings:

{{{!yaml
domain_delay:
  torrentz.eu: 5 minutes
  torrentsmirror.com: 5 minutes
...
discover:
  interval: 2 hours
}}}

Those are *way* too conservative, I guess, but it appears my IPs are not banned any more, now. Will file proper ticket with better explanation when I find the time (a lot of real life going on right now for me).

cvium:
How did you come to the realisation that you were banned? I've never had any issues.

TauPan:
Timeouts were gone after switching IP-Addresses, but reappeared after a while.

Did you use any timeout/limit/interval settings or just the defaults? How many entries did your discover settings create? I regularly see the warning/error that I have over 500.

Ban is still a hypothesis, but I have further evidence (email exchange with torrentz admin) that supports it.

cvium:
My discover rarely creates more than a handful of entries per run up to 5 reruns. I use 5 second intervals.

gazpachoking:
Might these have been there because some of the domains are blocked from some countries? Any chance we'll be messing stuff up for some people by doing this? If we aren't worried about it, I'd ditch the loop completely, as it's a bit silly to loop over one thing.

TauPan:
You're talking about a purely DNS based block?

That makes sense. I'm beginning to doubt that my patch does anything useful at all.

I still see timeouts on multiple networks while others work. Changing IP-adresses alone does not help, so there might also be routing problems, congestions etc. as alternative explanations.


Maybe it would make more sense to catch the exception and only retry on resolver errors, to avoid hammering on 502 and timeouts, but still avoiding DNS blocks? 

Also I propose that the torrentz plugin should have a useful default for domain_delay, say 5 seconds, so that the user doesn't have to configure it.

TauPan:
I meant 403 ;)... Also note that in a previous post I said that switching IP-Adresses would help.

Doing so in the same network decidedly does NOT fix timeouts etc. for me. What does help is switching to a different network (via vpn or proxy).

That makes a ban seem highly unlikely.

gazpachoking:
> You're talking about a purely DNS based block?

Yeah. I'm not certain that's the reason though, just guessing.



> Maybe it would make more sense to catch the exception and only retry on resolver errors, to avoid hammering on 502 and timeouts, but still avoiding DNS blocks?

If they all resolve to the same ip, this seems like a good idea to me.

> Also I propose that the torrentz plugin should have a useful default for domain_delay, say 5 seconds, so that the user doesn't have to configure it.

If they have a policy like that, it sounds good to me.

TauPan:
> > You're talking about a purely DNS based block?
> Yeah. I'm not certain that's the reason though, just guessing.

I am guessing too. Judging from http://en.wikipedia.org/wiki/Torrentz#Censorship that might be the case.

 
> > Maybe it would make more sense to catch the exception and only retry on resolver errors, to avoid hammering on 502 and timeouts, but still avoiding DNS blocks?
> If they all resolve to the same ip, this seems like a good idea to me.

They certainly do, at least here.
 
> > Also I propose that the torrentz plugin should have a useful default for domain_delay, say 5 seconds, so that the user doesn't have to configure it.
> If they have a policy like that, it sounds good to me.

I can't know for sure. I asked them about their policy, but didn't receive a definitive reply. Here's the mail exchange:

My initial request to admin@torrentz.eu:
```
Hi!

Since the pirate bay rss feeds went down, I'm trying to get my torrents via flexget's discover / search plugins.

However since a couple of months or so, my requests always time out.

After a couple of tries I figured out that I could access the feeds from an ip address in a different country, so I switched my configuration to that IP.

Very quickly, requests started to time out for that IP as well, while the old one worked again.

This leads me to the suspicion that the plugin behaves in a way that causes an automatic block from the server providing the rss feeds.

It would be helpful if you could provide me with terms of use for automatic access, so I can either configure the plugin appropriately or fix it, file a ticket, etc.

Kind regards and thanks in advance! 
```

Reply from admin:
```
Your RSS client is broken. Civilized software works just fine with our
feeds. Please try another client.
```

Note that this is neither a confirmation or denial that there is any anti-bot policy in place at all.

I decided not to give up and (hopfully politely) ask again:

```
I'm asking you to specify how civilized software should behave and in what way I can fix my broken rss client.

The polling interval in flexget is configurable but currently I find my requests timing out after the 5th query, so it seems the code needs to be fixed.

You don't need to specify the exact configuration, just what would be acceptable to you.

Thanks and best regards
```

Not sure what kind of "civilized" behaviour is meant, if it's simply respecting robots.txt or some timing requirements from the rss client specification. It might be that our usecase does not coincide with the intended usecase of the rss feeds, so admin doesn't care about us.

Or maybe the effects I see are not under the control of torrentz.eu at all. No idea how to proceed here.

Maybe this issue should be closed as null and void, if cvium and others don't have any issues.

cvium:
It would help if you could provide a config that definitely fails. My config doesn't seem to be triggering anything, so it makes it hard to verify what the problem is.

TauPan:
> It would help if you could provide a config that definitely fails. My config doesn't seem to be triggering anything, so it makes it hard to verify what the problem is.

I'm not sure if I can come up with a config that definitely fails in a reasonable amount of time.

At some point I was watching a log with --discover-now running and after the 5th search request to torrentz.eu, the 6th was timing out.

I also remember two subsequent curl requests to a rss search url, and the second (identical one) timed out.

I can start experimenting based on that.

gazpachoking:
@TauPan I implemented the idea about not trying more domains when it wasn't a dns problem. Feel free to open another PR if you have some more ideas, or a bug ticket on the tracker if you just want to track before a PR. Going to close this for now.

PR #507, opened 2015-05-16
Refactor decompress

ianstalk:
Clean up decompress and move it to the output package


flexget/plugins/output/plugin_decompress.py
```diff
@@ -86,36 +86,30 @@ def open_archive(self, entry):
...
+        elif rarfile and rarfile.is_rarfile(archive_path):
+            try:
+                archive = rarfile.RarFile(rarfile=archive_path)
+                log.debug('Successfully opened RAR: %s' % archive_path)
+            except rarfile.BadRarfile:
+                error_message = 'Bad RAR file: %s' % archive_path
+                log.error(error_message)
+                entry.fail(error_message)
+            except rarfile.RarWarning as error:
+                log.warn('Nonfatal error: %s (%s)' % (archive_path, error))
```
gazpachoking: I still don't understand how this is a nonfatal error. If it happens, `archive` will be None. Certainly that's fatal?

flexget/plugins/output/plugin_decompress.py
```diff
@@ -243,16 +236,16 @@ def handle_entry(self, entry, config):
 
                     if os.path.exists(destination):
                         error_message = log.debug('Cleaning up partially extracted file: %s' % destination) % \
```
gazpachoking: Oops, `error_message = ` looks extra here.

flexget/plugins/output/plugin_decompress.py
```diff
@@ -227,12 +220,12 @@ def handle_entry(self, entry, config):
 
                     log.verbose('Extracted: %s' % path )
                     success = True
-                except (IOError, os.error) as e:
+                except (IOError, os.error) as error:
                     error_message = 'OS error while creating file: %s (%s)' % \
-                                    (destination, e)
-                except (zipfile.BadZipfile, rarfile.Error) as e:
+                                    (destination, error)
+                except (zipfile.BadZipfile, rarfile.Error) as error:
```
gazpachoking: Can this even happen? Aren't they already caught by the try blocks in the `open_archive` function?

flexget/plugins/output/decompress.py
```diff
@@ -146,13 +138,12 @@ def is_dir(self, info):
...
-            base = os.path.basename(info.filename)
-            return not bool(base)
-        elif isinstance(info, rarfile.RarInfo):
+
+        if hasattr(info, 'isdir'):
             return info.isdir()
         else:
-            raise ValueError('Not a RarInfo or ZipInfo object.')
+            base = os.path.basename(info.filename)
+            return not bool(base)
```
gazpachoking: No need to manually `bool` this, the `not` operator will do that for us. Another style note, a `try` block with `AttributeError` is probably more pythonic than `hasattr`.
https://docs.python.org/2/glossary.html#term-eafp

flexget/plugins/output/decompress.py
```diff
@@ -227,12 +218,12 @@ def handle_entry(self, entry, config):
...
-                except (IOError, os.error) as e:
+                except (IOError, os.error) as error:
                     error_message = 'OS error while creating file: %s (%s)' % \
-                                    (destination, e)
-                except (zipfile.BadZipfile, rarfile.Error) as e:
+                                    (destination, error)
+                except (zipfile.BadZipfile, rarfile.Error) as error:
                     error_message = 'Failed to extract file: %s in %s (%s)' % \
-                                    (path, archive_path, e)
+                                    (path, archive_path, error)
```
gazpachoking: We allow up to 120 line length, I'd like to avoid the `\` line extension where possible.

flexget/plugins/output/decompress.py
```diff
@@ -242,17 +233,17 @@ def handle_entry(self, entry, config):
                     entry.fail(error_message)
 
                     if os.path.exists(destination):
-                        error_message = log.debug('Cleaning up partially extracted file: %s' % destination) % \
-                                        (path, archive_path, e)
+                        log.debug('Cleaning up partially extracted file: %s' % destination) % \
+                                  (path, archive_path, error)
```
gazpachoking: Hmm, the `% (path, archive_path, error)` also seems extra.

flexget/plugins/output/decompress.py
```diff
@@ -242,17 +233,16 @@ def handle_entry(self, entry, config):
                     entry.fail(error_message)
 
                     if os.path.exists(destination):
-                        error_message = log.debug('Cleaning up partially extracted file: %s' % destination) % \
-                                        (path, archive_path, e)
+                        log.debug('Cleaning up partially extracted file: %s' % destination) % path
```
gazpachoking: Ahh, still extra % outside the `log.debug` call. :tongue:

gazpachoking:
Thanks!

PR #509, opened 2015-05-21
Added two action to the filter

PhyberApex:
Added the actions: "reject_like" and "accept_like" to compare the fields and allow filtering if one contains the other (Both ways)

### DIFF
--- a/flexget/plugins/filter/crossmatch.py
+++ b/flexget/plugins/filter/crossmatch.py
@@ -25,7 +25,7 @@ class CrossMatch(object):
         'type': 'object',
         'properties': {
             'fields': {'type': 'array', 'items': {'type': 'string'}},
-            'action': {'enum': ['accept', 'reject']},
+            'action': {'enum': ['accept', 'reject', 'reject_like', 'accept_like']},
             'from': {'type': 'array', 'items': {'$ref': '/schema/plugins?phase=input'}}
         },
         'required': ['fields', 'action', 'from'],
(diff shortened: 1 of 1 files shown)


soloam:
I have a change #498 waiting to be accepted also, looking at your code I think that is possible to integrate my change (if it's accepted) in yours, or vice versa.

Best Regards

PhyberApex:
Hey I just checked out your change. I don't think I really get what you are doing there I'm sorry. I'm not that good with regex. Also I can't seem to find the code where you add the "match" field to the entry.

soloam:
Basically I replace all the spaces with " .* " and add " .* " to the beginning and end of the string. That would make for example:

- Spider Man would become .*Spider.*Man.*, this on regex would match any char making it possible to have a positive match with:
   - Spider-Man
   - The Amazing Spider Man
   - The Spider is the Man Best Friend
   - etc...

The main change is:
  common = self.entry_intersects(entry, generated_entry, fields, mode)

And the changes to entry_intersects()

Best Regards

PhyberApex:
Ah okay got it. I will think about it and maybe add it to my change if I find a suitable way to configure it.

soloam:
If yours is accepted before mine I'll edit my change request to make it compatible with yours.

liiight:
@PhyberApex is this still relevant?

PhyberApex:
@liiight Yes I think this is something that should be implemented. Maybe in a better way then I did but I still think it is needed if you don't want to change your config file all too often.

E.g. having an RSS feed with some sort of regex for video game titles and then match that against a RSS feed of a private tracker or something like this.

liiight:
Ok, I'll review it too asap. I'm going over all Prs and pinging them

liiight:
Can you provide an example for this? (if it'll be in the form of a nosetest it wont be the worst thing in the world either :wink)

PhyberApex:
Sorry I am not able to write a nose test, as I don't have any experience with Python what so ever xD

An exaple would be:

- First RSS Feed is a RSS feed for torrent downloads which mostly have the release name in their title - tag.
- Second RSS Feed is a personal feed where I manage e.g. game titles I want to download as soon as they are available. Of course I don't know at this point of time what the release name will be.

- I want entries from the first feed to be accepted if their title contains the words of the second rss feed. Even better would be to if you would be able to use the title of the second feed as a regexp to match the title of the first feed.

Does this help?

liiight:
It does but I'm not sure this implementation will do that. Have you got this working with any scenario? 

The field in field check is too literal imo. Like you said, trying to match regex of a normalized string should be more useful.

PhyberApex:
Sadly I do not have a working solution for this as of now.

You are free to reject this specific patch and maybe at a later time I will
try to implement it in a way that is more generally useful and link to this
discussion in that pull request.

2016-02-01 9:50 GMT+01:00 Or Carmi <notifications@github.com>:

lepinkainen:
Reopen if you manage get it working again.

PR #511, opened 2015-05-24
plugin path_select tests fix

stevezau:
-Types in path_select tests
-Will now abort a path cannot be selected

### DIFF
--- a/flexget/plugins/modify/path_select.py
+++ b/flexget/plugins/modify/path_select.py
@@ -56,15 +56,7 @@ def _path_selector(paths, threshold, stat_attr, reverse=True):
 
     valid_paths = [paths_stats[0].path]
 
-    if isinstance(threshold, float):
-        # Percentage
-        valid_paths.extend([
-            path_stat.path for path_stat in paths_stats[1:]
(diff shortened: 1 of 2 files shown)


flexget/plugins/modify/path_select.py
```diff
@@ -56,15 +56,7 @@ def _path_selector(paths, threshold, stat_attr, reverse=True):
...
-    if isinstance(threshold, float):
-        # Percentage
-        valid_paths.extend([
-            path_stat.path for path_stat in paths_stats[1:]
-            if abs(getattr(path_stat, stat_attr) - getattr(paths_stats[0], stat_attr)) <= threshold
-        ])
-
-    elif isinstance(threshold, int) and threshold > 0:
-        # Size in bytes
+    if threshold > 0:
```
gazpachoking: How are percent thresholds being differentiated from size ones?

gazpachoking:
@stevezau Wanted to get those tests working, but I think we should revisit the threshold ambiguity. Is there anything stopping an user from making a config that doesn't get interpreted properly right now?

PR #512, opened 2015-05-27
Sonarr & Sickbead plugins

liiight:
This plugin allows retreival of all shows configured under Sonarr.
Syntax:

sonarr:
  base_url: http://localhost
  port: 8989
  api_key: API123

Can be used with the discover or configure_series plugins for example.


flexget/plugins/sonarr.py
```diff
@@ -0,0 +1,43 @@
...
+    }
+
+    def on_task_input(self, task, config):
+        url = '%s:%s/api/series' % (config['base_url'], config['port'])
+        headers = {'X-Api-Key': config['api_key']}
+        json = task.requests.get(url, headers=headers).json()
+        entries = []
+        for show in json:
+            showName = show["title"]
+            entry = Entry(title = '%s' % (showName),
```
cvium: Two things: We usually use single quotes for strings and everything should be PEP8-compliant. It could just as well be `title=show['title']`.

gazpachoking:
Can we add tvdb_id and tvrage_id to entries if available from sonarr?

liiight:
Sure, I'll resubmit soon including pep8 compatibility

liiight:
Oops, I accidentally included another plugin to this pull request...
Sorry for that. Is that OK though, or should I split those two?

flexget/plugins/input/sonarr.py
```diff
@@ -0,0 +1,46 @@
...
+        'required': ['api_key', 'base_url', 'port'],
+        'additionalProperties': False
+    }
+
+    def on_task_input(self, task, config):
+        url = '%s:%s/api/series' % (config['base_url'], config['port'])
+        headers = {'X-Api-Key': config['api_key']}
+        json = task.requests.get(url, headers=headers).json()
+        entries = []
+        for show in json:
```
midraal: maybe a check could be added to see if the show is monitored in sonarr and hasn't ended by adding something like

    if show["monitored"] and show["status"] != "ended"

this would cut down on searching and would respect the sonarr settings

liiight: Good idea, i'll add it

gazpachoking:
> Oops, I accidentally included another plugin to this pull request...
> Sorry for that. Is that OK though, or should I split those two?

I'd prefer it in two, but I think these are simple enough to just pull them in as is.

gazpachoking:
Want to add some wiki pages?

liiight:
Sure thing, but I don't see myself doing it until Sunday, so if it can wait
I'll do it then.

PR #513, opened 2015-05-27
add couchpotato plugin

midraal:
This plugin allows retrieval of the couchpotato wanted movie list

example usage:
```
   queue_movies_couchpotato:
     couchpotato:
       base_url: http://localhost
       port: 5050
       api_key: <your key here>
     accept_all: yes
     movie_queue: add
```

### DIFF
--- a/flexget/plugins/input/couchpotato.py
+++ b/flexget/plugins/input/couchpotato.py
@@ -1,0 +1,57 @@
+from __future__ import unicode_literals, division, absolute_import
+import logging
+import requests
+
+from flexget import plugin
+from flexget.event import event
+from flexget.entry import Entry
+
+log = logging.getLogger('couchpotato')
+
+
+class CouchPotato(object):
+
+    schema = {
+        'type': 'object',
+        'properties': {
+            'base_url': {'tpye': 'string'},
+            'port': {'type': 'number'},
+            'api_key': {'type': 'string'}
+        },
+        'required': ['api_key', 'base_url', 'port'],
+        'additionalProperties': False
+    }
+
+    def on_task_input(self, task, config):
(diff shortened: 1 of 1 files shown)


flexget/plugins/input/couchpotato.py
```diff
@@ -0,0 +1,54 @@
...
+
+    Options base_url, port and api_key are required.
+    """
+        url = '%s:%s/api/%s/movie.list?status=active' % (config['base_url'], config['port'], config['api_key'])
+        json = task.requests.get(url).json()
+        entries = []
+        for movie in json['movies']:
+            if movie['status'] == 'active':
+              title = movie["title"]
+              entry = Entry(title = '%s' % (title),
```
cvium: We aim to have all new code PEP8 compliant, which states that keyword arguments shouldn't have any spaces ie. it should be `title='%s'`. Is it even necessary to do it like that? Wouldn't it work with `title=title`?

midraal:
it now runs through the pep8 script without errors

flexget/plugins/input/couchpotato.py
```diff
@@ -0,0 +1,57 @@
...
+import requests
+
+from flexget import plugin
+from flexget.event import event
+from flexget.entry import Entry
+
+log = logging.getLogger('couchpotato')
+
+
+class CouchPotato(object):
```
ianstalk: Usage info should go here as a docstring. This is what's printed when the --doc option is used from the command line.

gazpachoking:
Does the couchpotato api provide any movie ids? If so, can we add them to the entries with appropriate names? (imdb_id, tmdb_id)

gazpachoking:
Thanks! Want to make a wiki page for it?

PR #514, opened 2015-05-28
couchpotato plugin: fixed whitespace for pep8

midraal:


gazpachoking:
Good catch, thanks!

midraal:
forgot to run pep8 script before pushing the original plugin

PR #517, opened 2015-05-28
NewPCT fixed

malkavi:
Fixed urlrewrite. You can use newpct or newpct1 feed, for users that have newpct blocked by their ISP.

### DIFF
--- a/flexget/plugins/urlrewrite_newpct.py
+++ b/flexget/plugins/urlrewrite_newpct.py
@@ -8,11 +8,20 @@ from flexget.plugins.plugin_urlrewriting import UrlRewritingError
 from flexget.utils.requests import Session
 from flexget.utils.soup import get_soup
 
+from flexget.utils.search import normalize_unicode
+
+from requests.auth import AuthBase
+
 log = logging.getLogger('newpct')
 
-requests = Session()
-requests.headers.update({'User-Agent': 'Mozilla/4.0 (compatible; MSIE 5.5; Windows NT)'})
-requests.set_domain_delay('imdb.com', '2 seconds')
+
+class NewPCTAuth(AuthBase):
+    """Attaches HTTP Authentication to the given Request object."""
+
+    def __call__(self, r):
+        headers = {'user-agent': 'Mozilla/5.0 (Macintosh; Intel Mac OS X 10.9; rv:32.0) Gecko/20100101 Firefox/32.0'}
+        r.prepare_headers(headers)
+        return r
 
 
 class UrlRewriteNewPCT(object):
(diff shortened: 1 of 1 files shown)


flexget/plugins/urlrewrite_newpct.py
```diff
@@ -20,27 +29,56 @@ class UrlRewriteNewPCT(object):
...
-        torrent_id_prog = re.compile("'(?:torrentID|id)'\s*:\s*'(\d+)'")
-        torrent_ids = soup.findAll(text=torrent_id_prog)
-        if len(torrent_ids) == 0:
-            raise UrlRewritingError('Unable to locate torrent ID from url %s' % url)
-        torrent_id = torrent_id_prog.search(torrent_ids[0]).group(1)
-        return 'http://www.newpct.com/descargar/torrent/%s/dummy.html' % torrent_id
+        torrent_span = soup.find('span', attrs={'id': 'content-torrent'})
+        try:
+            torrent_tag = torrent_span.findAll('a', href=True)
+        except Exception as e:
```
gazpachoking: We should use specific exceptions we want to catch. I suspect here you are actually just catching the case that torrent_span is None.

flexget/plugins/urlrewrite_newpct.py
```diff
@@ -20,27 +29,56 @@ class UrlRewriteNewPCT(object):
...
+            log.error('Error findAll')
+        if len(torrent_tag) == 0:
```
gazpachoking: If the above exception happens, `torrent_tag` will not be defined, and this will cause an exception.

flexget/plugins/urlrewrite_newpct.py
```diff
@@ -8,11 +8,20 @@
...
+from requests.auth import AuthBase
+
 log = logging.getLogger('newpct')
 
-requests = Session()
-requests.headers.update({'User-Agent': 'Mozilla/4.0 (compatible; MSIE 5.5; Windows NT)'})
-requests.set_domain_delay('imdb.com', '2 seconds')
+
+class NewPCTAuth(AuthBase):
+    """Attaches HTTP Authentication to the given Request object."""
```
gazpachoking: Can we fix this docstring to say what this method actually does? This isn't http authentication, this is user-agent spoofing.

flexget/plugins/urlrewrite_newpct.py
```diff
@@ -20,27 +29,56 @@ class UrlRewriteNewPCT(object):
 
     # urlrewriter API
     def url_rewritable(self, task, entry):
-        url = entry['url']
-        rewritable_regex='^http:\/\/(www.)?newpct1?.com\/.*'
-        return re.match(rewritable_regex,url) and not url.startswith('http://www.newpct.com/descargar/')
+        if entry.get('urls'):
```
gazpachoking: urls[0] should always be the same as the regular url, so I don't think this is needed. Especially since the actual rewrite function doesn't handle it.

flexget/plugins/urlrewrite_newpct.py
malkavi: removing the try exception would fix this problem? Whitout the try In the error case, torrent_tag's lenght would be 0.

flexget/plugins/urlrewrite_newpct.py
malkavi: I know that It has no sense, but without this I could not make it work.

stevezau:
@malkavi any updates on this?

malkavi:
I can't make it work without using entry['urls'] instead of entry['url']
I don't know where is the problem.

gazpachoking:
We need an update to latest develop and a bit more investigation in the comments if we are going to merge this.

PR #523, opened 2015-05-31
[WIP] RESTful API and WebUI

stevezau:
The aim to the create a full RESTful API for flexget and build a FlexGet WebUI using the API's

http://flexget.com/wiki/Drafts/RESTApi
http://flexget.com/wiki/Drafts/json_schema

TODO:
- [ ] Schedules API should return list of schedules with attributes (last_run, next_run etc)


stevezau:
@gazpachoking and @paranoidi I think the wsgi server should be embedded and started with the daemon..

flexget.api will provide the core API's.

Plugins can expose api's within their implementaion ie:

{{{
@api.route('/plugins/archive', methods=['GET'])
def archive():
    //logic
    return jsonify(history_items)
}}}

flexget/api.py
```diff
@@ -22,41 +25,111 @@ def attach_schema(response):
...
+@api.route('/version/')
 def version():
     return jsonify(flexget_version=flexget.__version__, api_version=API_VERSION)
 
 
+# Execution API
+
 exec_parser = get_parser('execute')
 
+def _get_task(task_id):
```
gazpachoking: I think if we are going to be looking anything up about task executions after the fact, we need to store the info here. Such that when someone calls the /execution/ endpoint we generate and store the unique_id -> task info mapping here. I'm not sure how long we need to hold on to that though...

stevezau: @gazpachoking see my latest commit. As per our discussions task_queue now stores task data for at least an hr.

Still need to add the task_id to the log file for streaming to work.

flexget/plugins/daemon/http_server.py
```diff
@@ -0,0 +1,92 @@
...
+
+log = logging.getLogger('webui')
+
+main_schema = {
+    'type': 'object',
+    'properties': {
+        'bind': {'type': 'string', 'format': 'ipv4', 'default': '0.0.0.0'},
+        'port': {'type': 'integer', 'default': 5050},
+        'autoreload': {'type': 'boolean', 'default': False},
+        'authentication': {
```
gazpachoking: Hmm. This is currently implemented by a plugin. We may have to consider how webui plugins can go in config.

stevezau: yes I agree. I left it as a plugin for now until we make a decision on if the api is baked in.

If it's baked in then there should be an option to point to a register an alternative webui...?

stevezau:
@gazpachoking  currently the webui is not loaded. We need to decide on a way to register ui's.

I was thinking about:

flexget.webserver.register_ui(blueprint)
flexget.webserver.register_home(route)

The default UI will route to /ui.. Other ui's must specify the url ie: /custom

@app.route('/') will route to what ever was set at register_home.

I'm not sure how to handle loading of the flexget.webui or custom ui's modules tho?

flexget/api.py
```diff
@@ -0,0 +1,635 @@
...
+task_api_schema = api.schema('task', task_api_schema)
+
+
+def _load_config():
+    with codecs.open(manager.config_path, 'rb', 'utf-8') as f:
+        try:
+            raw_config = f.read()
+        except UnicodeDecodeError:
+            raise ValueError('Config file is not UTF-8 encoded')
+    return yaml.safe_load(raw_config) or {}
```
gazpachoking: Hmm, I don't think we should be loading this from disk every time, I think it should be from manager.config.

stevezau: yeah i agree.. I was just worried about memory but the config is quite small..

I'll store manager.user_config and manager.config (with defaults).

It also makes a backup on save. I might change that to only keep the last 10 changes.

gazpachoking: Yeah, and the config is already in memory at `manager.config` that's where everything gets its config from ultimately.

stevezau:
@gazpachoking can u test the latest on android. works ok for me on ios

gazpachoking:
@stevezau I've added in a hack that lets us use oneOf and anyOf with schema-form. Hopefully you can figure out where to extract it to, I've just got it plugged straight into schema-form at the moment. Still needs a bit of improving, but should let us get things running before we get full oneOf/anyOf support.

PR #524, opened 2015-05-31
Added more options to sickbeard plugin

liiight:
Added include_ended and only_monitored option to plugin

### DIFF
--- a/flexget/plugins/input/sickbeard.py
+++ b/flexget/plugins/input/sickbeard.py
@@ -16,7 +16,9 @@ class Sickbeard(object):
         'properties': {
             'base_url': {'type': 'string'},
             'port': {'type': 'number'},
-            'api_key': {'type': 'string'}
+            'api_key': {'type': 'string'},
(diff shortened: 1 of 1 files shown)


flexget/plugins/input/sickbeard.py
```diff
@@ -63,11 +65,13 @@ def on_task_input(self, task, config):
         json = task.requests.get(url).json()
         entries = []
         for id, show in json['data'].items():
-            entry = Entry(title=show['show_name'],
-                          url='',
-                          series_name=show['show_name'],
-                          tvdb_id=show['tvdbid'],
-                          tvrage_id=show['tvrage_id'])
+            if show['paused'] == 0 or not config.get('only_monitored'):
```
cvium: I assume 'paused' is either 1 or 0? In that case it'd be more pythonic to write `if not show['paused']`.

liiight: sure thing, i'll change it,

PR #525, opened 2015-05-31
enhance url composition in sonarr couchpotato and sickbeard plugins

midraal:
should fix the use case where these plugins are not located at the server root path or behind a http proxy

### DIFF
--- a/flexget/plugins/input/couchpotato.py
+++ b/flexget/plugins/input/couchpotato.py
@@ -1,4 +1,5 @@
 from __future__ import unicode_literals, division, absolute_import
+from urlparse import urlparse
 import logging
 import requests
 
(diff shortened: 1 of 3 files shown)


cvium:
is this ready to be merged then?

midraal:
yes, in manual testing it seems to work, but don't have an install wich is like @grrr2

liiight:
I'm not sure this'll do the trick. In @grrr2 example, the user can set a URL that doesn't even include the port setting as it's behind a proxy http:///ur/sickbeard/api.
Hence, adding the port setting may not be required and can even break the URL (unless the user inputs the port as 80 which is redundant.

I think the best choice is to let to user to enter the entirety of the URL including the port number as a single parameter.

midraal:
changing the default port to 80 should make the use case of being behind a proxy slightly more elegant

flexget/plugins/input/couchpotato.py
```diff
@@ -15,7 +16,7 @@ class CouchPotato(object):
         'type': 'object',
         'properties': {
             'base_url': {'tpye': 'string'},
-            'port': {'type': 'number'},
+            'port': {'type': 'number', 'default': 80},
             'api_key': {'type': 'string'}
         },
         'required': ['api_key', 'base_url', 'port'],
```
liiight: you should remove the port from the required properties

midraal: can't believe i missed that

liiight:
Any chance merging this? I want to add some stuff to those plugins as well

cvium:
Well, did you agree on how to handle the various urls?

liiight:
yeah, the proposed fix is the agreed upon implementation

cvium:
There are some conflicts though because of a recent merge. If @midraal fixes it I'll merge it. I'd rather not fiddle around with the cli.

midraal:
should be okay to merge now

PR #526, opened 2015-06-03
update code to follow pep8

midraal:
11 errors remaining on paver pep8, but not sure how to fix them


flexget/plugins/api_tmdb.py
```diff
@@ -77,7 +77,8 @@ def update_from_dict(self, update_dict):
     def update_from_object(self, update_object):
         """Populates any simple (string or number) attributes from object attributes"""
         for col in self.__table__.columns:
-            if hasattr(update_object, col.name) and isinstance(getattr(update_object, col.name), (basestring, int, float)):
+            if (hasattr(update_object, col.name) and
+            isinstance(getattr(update_object, col.name), (basestring, int, float))):
```
gazpachoking: I think better would be to put 2 indents here, so it doesn't line up with either the previous or next lines.

midraal:
i have been using this patch for about a week now, should be okay to merge

gazpachoking:
Thanks!

PR #527, opened 2015-06-03
Pushalot support

liiight:
Pushalot service support


cvium:
Is it ready now?

liiight:
The only thing that's not yet tested is multiple tokens. I think you can merge it, and if push comes to shove, I'll add a fix.

PR #528, opened 2015-06-04
Fix incorrectly named global in extratorrent url rewrite plugin

nalipaz:
I am getting errors whilst trying to use the plugin, this should fix it, but I was not able to test further at the moment.


chewy74:
Getting message: 
2015-06-27 12:03 WARNING  discover      movies          Search plugin extratorrent does not support latest search api

PR #529, opened 2015-06-05
Update html.py to support jinja entry

tyjtyj:


stevezau:
@tyjtyj it looks like the email was sent at the task level but your changes are at the entry level??

tyjtyj:
not sure the different entry and task level. i was assuming output is the correct make_html output 

File flexget/plugins/output/html.py 

Drop it/Correct it if it is not in correct place. 

Thanks for reviewing it.

gazpachoking:
@tyjtyj What exactly is this trying to accomplish? I think you can already do what you need, but since the email template renders at the task level, you'll have to loop through task.accepted in the jinja template rather than in python.

gazpachoking:
@tyjtyj Closing this PR. Feel free to discuss with us though if you feel there is more functionality needed.

PR #535, opened 2015-06-12
Add input plugin for Letterboxd lists

Chemondelay:
Creates entries for input from any **public** user list on Letterboxd. The plugin uses BeautifulSoup to parse the list for site-specific metainfo (e.g. user scores, where applicable), extracts the `tmdb_id` and average score from the page of each film included on the list, then queries the TMDb API to fill the rest of the entry.

The plugin is configured with the following settings:
* **`username`**: Your or another user's Letterboxd username, which can be found in the profile URL.
* **`list`**: The name of a custom list, or one of the built-in lists, i.e. `watchlist`, `diary`, `likes`, `rated` or `watched`. It should be entered as it appears in the list URL; so for example, the a list at `http://letterboxd.com/user/list/some-list/` should be entered as `some-list`.
* **`sort_by`** (optional): Sort films within the list according to one of the parameters allowed by Letterboxd, or leave unspecified to use the default sorting for the list type. The available options are:
    * `name`: Sorted by film name, in ascending alphabetical order.
    * `added`: Sorted by the date the film was added to the list, in reverse chronological order.
    * `popularity`: Sorted in descending order of popularity among Letterboxd users.
    * `length-ascending` or `length-descending`: Sorted by runtime.
    * `rating-ascending` or `rating-descending`: Sorted by average rating or user rating, depending on the list context (e.g. watchlists will be sorted by the average, rated lists by the individual).
    * `release-ascending` or `release-descending`: Sorted by release date.
* **`max_results`** (optional): An integer specifying the maximum number of results that the plugin will loop over.

The `username` and `list` strings are used to fill in the URL for the initial parse, and so should be entered exactly as they appear in the URL for the list. So for example, to pull entries from a list at
````
http://letterboxd.com/some-user/list/some-users-list/
````
you would enter the `username` as `some-user`, and the `list` as `some-users-list`, not 'Some User' and 'Some user's list' (or whatever).

The plugin populates the following entry fields:
* **`title`**: Film Name (Year)
* **`url`**: The film's page on Letterboxd.
* **`imdb_id`** and **`tmdb_id`**
* **`letterboxd_list`**: In the format list (username), included in case the plugin is used as one of several input plugins, and you want to apply filtering against these.
* **`letterboxd_score`** (where available): A floating-point number representing the average score (out of 10) given to the film by Letterboxd users.
* **`letterboxd_uscore`** (where available): An integer representing the score (out of 10) given to the film by the list owner.
* **`movie_name`** and **`movie_year`**.

Results are cached for two hours to avoid swamping the site, and there is a short domain delay between requests. And that's about it. If anyone sees any problems, please let me know.

### DIFF
--- a/flexget/plugins/input/letterboxd.py
+++ b/flexget/plugins/input/letterboxd.py
@@ -1,0 +1,152 @@
+from __future__ import unicode_literals, division, absolute_import
+import logging, re
+
+from flexget import plugin
+from flexget.entry import Entry
+from flexget.event import event
+from flexget.plugin import get_plugin_by_name
+from flexget.utils.cached_input import cached
+from flexget.utils.requests import RequestException, Session
+from flexget.utils.soup import get_soup
+
+log = logging.getLogger('letterboxd')
+logging.getLogger('api_tmdb').setLevel(logging.CRITICAL)
+
+requests = Session(max_retries=5)
+requests.set_domain_delay('letterboxd.com', '1 seconds')
+base_url = 'http://letterboxd.com'
+
+SLUGS = {
+    'default': {
+        'p_slug': '/%(user)s/list/%(list)s/',
+        'f_slug': 'data-film-slug'},
+    'diary': {
+        'p_slug': '/%(user)s/films/diary/',
+        'f_slug': 'data-film-slug'},
+    'likes': {
+        'p_slug': '/%(user)s/likes/films/',
+        'f_slug': 'data-film-link'},
+    'rated': {
+        'p_slug': '/%(user)s/films/ratings/',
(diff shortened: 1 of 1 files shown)


gazpachoking:
Thanks! Want to make a wiki page?

Chemondelay:
Yeah, sure thing. I haven't contributed before; are there any specific guidelines I should know about for the wiki?

gazpachoking:
Not really, just make a page at http://flexget.com/wiki/Plugins/letterboxd to document plugin usage and options, then make a link to it from the input section of http://flexget.com/wiki/Plugins

Chemondelay:
No problem, I'll get on it. Thanks!

PR #536, opened 2015-06-13
Fix two issues with rottentomatoes plug-in

jeroenl:


flexget/plugins/api_rottentomatoes.py
```diff
@@ -508,7 +509,7 @@ def lists(list_type, list_name, limit=20, page_limit=20, page=None, api_key=None
 
 def movies_search(q, page_limit=None, page=None, api_key=None):
     if isinstance(q, basestring):
-        q = q.replace(' ', '+').encode('utf-8')
+        q = q.replace(' ', '+').encode('ascii', errors='ignore')
```
gazpachoking: I think the correct fix for this without stripping non-ascii would be `q = urllib.quote_plus(q.encode('utf-8'))` can you give that a test?

jeroenl: Hi Chase,

The problem isn't in the method of quoting, it's with search strings that
contain two languages of the movie title. Since RT only contains the
english names, this allows these items to be found. If you don't agree with
this solution, shall I give a pull request for just the 'movie_id cannot be
NULL' fix?

Regards,
Jeroen

Op za 13 jun. 2015 om 17:43 schreef Chase Sterling <notifications@github.com

gazpachoking: Does this work if user searches for something like `Léon`? Wouldn't the search end up being `Lon`?

gazpachoking: If I search for `Léon` on the web, the query string comes out like `L%E9on`. Based on that, I think the correct line might be `urllib.quote_plus(q.encode('latin1', errors='ignore'))`

jeroenl: I agree. Was just testing that one.
How do I change this? Cancel pull, fix it and send a new one?

gazpachoking: You can just make another commit in the same branch and it will update the pull.

gazpachoking:
Cool, thanks!

PR #537, opened 2015-06-13
Allow for caps in uTorrent destination path

jeroenl:
Paths are now only normalized (lowercased, on case-insensitive
platforms) when checking if directory is allowed, not when passing on
the directory name to uTorrent.


gazpachoking:
Looks good to me, thanks!

PR #539, opened 2015-06-14
Add output plugins (add and remove) for Letterboxd

Chemondelay:
This PR includes two new output plugins for Letterboxd: `letterboxd_add` and `letterboxd_remove`. I've followed the lead of the Trakt output plugins by including them both as subclasses within the same file, `plugins/services/letterboxd_submit.py`.

Both are configured the same way:
````
tasks:
  letterboxd_[add|remove]:
    username: *****
    password: *****
    list: [watchlist|watched]
````

The plugins uses re.search() to grab the URL slug for the film from the search results page. This means that the `imdb_id` field is required, since it's the only reliable way to isolate a single film in the results. The slug is then used to fill out various POST requests for modifying the list entries.

Only the `watchlist` and `watched` lists are supported for now. Adding or removing a film from a custom list will involve messing around with Javascript commands on the list's edit page, and since I'm basically figuring out what I'm doing as I go along, I'm stuck on this for the moment. (Incidentally, if anyone can point me in the right direction on this, I'd really appreciate it.)

I've also renamed the existing Letterboxd plugin to `letterboxd_list` to distinguish it from the new ones. The name is long enough to mess up the log formatting, not sure if this is a problem.

### CODE REVIEW

**Comment by Chemondelay** [comment]
Turns out it needs some more work.

File: flexget/plugins/services/letterboxd_submit.py [hunk 1]
```diff
@@ -0,0 +1,113 @@
...
+        elif self.remove:
+            command = LISTS[config['list']]['remove_command']
+            log_str = LISTS[config['list']]['remove_log']
+
+        for entry in task.accepted:
+            if any(field in entry for field in ['imdb_id', 'tmdb_id', 'movie_name']):
+                if 'imdb_id' in entry:
+                    film = self.parse_film(entry['imdb_id'])
+                    try:
+                        r = requests.post('%s/film/%s/%s/' % (base_url, film, command), data=params)
```
gazpachoking: Our requests session raises an exception by default on non success status codes, which you'll have to turn off. https://github.com/Flexget/Flexget/blob/develop/flexget/utils/requests.py#L132

gazpachoking:
Maybe you can get early access to the api and start a plugin against that? http://letterboxd.com/api-coming-soon/

flexget/plugins/services/letterboxd_submit.py
Chemondelay: Thanks, that explains something that was happening earlier (or rather wasn't happening).

Chemondelay:
Yeah good idea. I shot them an email, we'll see what they come back with. I resolved a problem I had with incorrect matches with the last commit. Stupidly I didn't think to narrow the search to just films, so occasionally it would pick up a review or something that mentioned the imdb_id. It works now.

flexget/plugins/services/letterboxd_submit.py
gazpachoking: Looks like you put the raise_status on the wrong post call.

Chemondelay: Whoops. That's what happens when you don't know what you're doing.

PR #541, opened 2015-06-14
Add "highres tv" to piratebay search plugin

oxc:
When using piratebay with 'tv' search, it won't find e.g. 1080p results. Instead, the 'highres tv' section must be used, which is added by this commit.


gazpachoking:
Thanks! Can you update the wiki page? http://flexget.com/wiki/Searches/piratebay

oxc:
Done.

PR #542, opened 2015-06-14
Add input plugin for myepisodes.com show list

oxc:
This pull request contains the file I've already attached to [trac-ticket #243](http://flexget.com/ticket/243:comment9):

myepisodes_list.py​ contains an input plugin that lists all ever-watched shows from myepisodes.com (all that show up on the "Time I've Wasted" page). The plugin has been based on code from the myepisodes service plugin and other input plugins, so I hope I got the coding style right.

However, I did not abstract any myepisodes api that can be shared between the two plugins, something that should probably be done (but rather by someone with more flexget experience than me).

### DIFF
--- a/flexget/plugins/input/myepisodes_list.py
+++ b/flexget/plugins/input/myepisodes_list.py
@@ -1,0 +1,101 @@
+from __future__ import unicode_literals, division, absolute_import
+import logging
+import re
+from itertools import chain
+
+from flexget import plugin
+from flexget.entry import Entry
+from flexget.event import event
+from flexget.utils.cached_input import cached
+from flexget.utils.soup import get_soup
+
+log = logging.getLogger('myepisodes')
+
+URL = 'http://www.myepisodes.com/'
+
+class MyEpisodesList(object):
+    """Creates an entry for each item in your myepisodes.com show list.
+
+    Syntax:
+
+    myepisodes_list:
+      username: <value>
+      password: <value>
+      strip_dates: <yes|no>
+      include_ignored: <yes|no>
+
+    Options username and password are required.
+    """
+
+    schema = {
+        'type': 'object',
+        'properties': {
+            'username': {'type': 'string'},
+            'password': {'type': 'string'},
+            'strip_dates': {'type': 'boolean', 'default': False},
+            'include_ignored': {'type': 'boolean', 'default': False},
+        },
+        'required': ['username', 'password'],
+        'additionalProperties': False,
+    }
+
+    @cached('myepisodes_list')
+    @plugin.internet(log)
+    def on_task_input(self, task, config):
+        if not task.requests.cookies:
+            username = config['username']
+            password = config['password']
+
+            log.debug("Logging in to %s ..." % URL)
+            params={
+                'username': username,
+                'password': password,
+                'action': 'Login'
+            }
+            loginsrc = task.requests.post(URL + 'login.php', data=params).content
+            if str(username) not in loginsrc:
+                raise plugin.PluginWarning(('Login to myepisodes.com failed, please check '
+                                 'your account data or see if the site is down.'), log)
+
+
+        page = task.requests.get(URL + "shows.php?type=manage").content
+        try:
+            soup = get_soup(page)
+        except Exception as e:
+            raise plugin.PluginError("Unable to parse myepisodes.com page: %s" % (e,))
+
+        entries = []
+
+        def show_list(select_id):
+            return soup.find('select', {'id': select_id}).findAll('option')
+
+        options = show_list('shows')
+        if config['include_ignored']:
+            options = chain(options, show_list('ignored_shows'))
+        for option in options:
+            name = option.text
+            if config.get('strip_dates'):
+                # Remove year from end of name if present
+                name = re.sub(r'\s+\(\d{4}\)$', '', name)
+            showid = option.get('value')
+            url = '%sviews.php?type=epsbyshow&showid=%s' % (URL, showid)
+
+            entry = Entry()
+            entry['title'] = name
+            entry['url'] = url
+            entry['series_name'] = name
+            entry['myepisodes_id'] = showid
+
+            if entry.isvalid():
+                entries.append(entry)
+            else:
+                log.debug('Invalid entry created? %s' % entry)
+
+        if not entries:
+            log.warn("No shows found on myepisodes.com list. Maybe you need to add some first?")
+
+        return entries
+
+@event('plugin.register')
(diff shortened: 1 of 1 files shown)


flexget/plugins/input/myepisodes_list.py
```diff
@@ -0,0 +1,98 @@
...
+            loginsrc = session.post(URL + 'login.php', data=params).content
+            if str(username) not in loginsrc:
+                raise plugin.PluginWarning(('Login to myepisodes.com failed, please check '
+                                 'your account data or see if the site is down.'), log)
+
+        # Populate shows with the list of show_ids in our account
+        wasted_url = "%s/%s" % (URL, "life_wasted.php")
+        log.debug('Fetching "Time I\'ve wasted" page: %s' % wasted_url)
+
+        page = session.get(wasted_url).content
```
stevezau: any reason why task.requests is not used here?

flexget/plugins/input/myepisodes_list.py
```diff
@@ -0,0 +1,98 @@
...
+        soup = get_soup(page)
```
stevezau: Do we need some exception handling here?

stevezau:
@oxc i've made a few comments, other then that the code looks ok.

oxc:
@stevezau thanks for the review. I've updated the change to include your comments. The reason I did not use `task.requests` was that I didn't know about it ;) I've added a rather defensive error handling, let me know if this is too much.

Also, I've changed the source of the list to a different page on myepisodes.com, which is much more suitable for this task. It includes new shows that have not been watched yet at all. Additionally, it allows me to omit the ignored shows, which is the new default; they can be added back by setting the new `include_ignored` to True.

Please review, thanks.

oxc:
Forgot to include the new option in the docstring, this is now fixed.

oxc:
Cleaned up some imports and simplified the shows URL creation

flexget/plugins/input/myepisodes_list.py
```diff
@@ -0,0 +1,99 @@
...
+            username = config['username']
+            password = config['password']
+
+            log.debug("Logging in to %s ..." % URL)
+            params={
+                'username': username,
+                'password': password,
+                'action': 'Login'
+            }
+            loginsrc = task.requests.post(URL + 'login.php', data=params).content
```
stevezau: Need some exception handling here... What if we get a http 5xx, 3xx or timeout error?

flexget/plugins/input/myepisodes_list.py
```diff
@@ -0,0 +1,99 @@
...
+            if str(username) not in loginsrc:
+                raise plugin.PluginWarning(('Login to myepisodes.com failed, please check '
+                                 'your account data or see if the site is down.'), log)
+
+
+        page = task.requests.get(URL + "shows.php?type=manage").content
```
stevezau: exception handling needed for http errors and timeouts

flexget/plugins/input/myepisodes_list.py
```diff
@@ -0,0 +1,99 @@
...
+                entry['url'] = url
+                entry['series_name'] = name
+                entry['myepisodes_id'] = showid
+
+                if entry.isvalid():
+                    entries.append(entry)
+                else:
+                    log.debug('Invalid entry created? %s' % entry)
+
+        except Exception as e:
```
stevezau: We shouldn't use blanket exception handling. check here for more info on soup excpetions

http://www.crummy.com/software/BeautifulSoup/bs4/doc/#errors-when-parsing-a-document

flexget/plugins/input/myepisodes_list.py
stevezau: FYI http://docs.python-requests.org/en/latest/user/quickstart/#errors-and-exceptions

oxc:
Isn't the @plugin.internet decorator supposed to handle the request exceptions you mentioned?

There's really nothing the plugin can do if any of the requests or parsing the site fails. Returning an empty list in that case does not sound right either.

stevezau:
Yes you are right, sorry didn't see @plugin.internet decorator.

Probably should change that blanket exception though?

oxc:
Yes, I will do that.

stevezau:
Also, what are the advantages of using www.myepisodes.com over trakt or something similar? The www.myepisodes.com website looks outdated, the last accouncement was almost 10 years ago.

oxc:
The advantage is that it has my show list in it :-)

I even tried migrating it to Trakt just for Flexget, but it has a different workflow and the interface, albeit beautiful, is much more unprecise and cluttered when it comes to simply having a number of shows of which you want to track acquiring and watching each episode.

I will try again if the site goes down, but until then I'm perfectly happy with myepisdoes.com

stevezau:
@oxc i'm really not sure/convinced this should added to the flexget base as i think you might be the only user using it which suggested it might be better as a user plugin?

FYI @gazpachoking is working on a way to create and track movie/series lists internally to flexget.. I'm working on a webui and RESTful api which will eventually allow you to manage your own lists via a mobily friendly ui.. Might be a better solution for you?..

bfd:
Well, there is a ticket for it ( http://flexget.com/ticket/243 ), and there already is a myepisodes scrobbler plugin. You should throw out that as well then.
Me for my part, I'm happy to maintain this for my own. I just thought I'd share...

oxc:
Well, there is a ticket for it ( http://flexget.com/ticket/243 ), and there already is a myepisodes scrobbler plugin. You should throw out that as well then.
Me for my part, I'm happy to maintain this for my own. I just thought I'd share...

stevezau:
@oxc that's all the convincing I need.. Fix that blanket exception and ill merge.

oxc:
I've narrowed the scope of the try-block, but I have not changed the exception type. It seems very unreliable what kind of exceptions beautifulsoup actually throws, depending on the parser used. I could not found a single mention of HTMLParser or HTMLParseError in the whole flexget project.

Otherwise, I've only added a warning when the collected entries are empty, since this could mean that the user has not added any shows on myepisodes.com and might be confused by the default flexget message for empty output.

flexget/plugins/input/myepisodes_list.py
gazpachoking: The internet decorator will actually handle it as is, though the error message might not be as nice as explicit error handling.

gazpachoking:
Thanks! Want to add it to the wiki?

oxc:
Done: http://flexget.com/wiki/Plugins/myepisodes_list

PR #543, opened 2015-06-15
do not skip SSL verification on plugin piratebay

X-dark:
There is no reason to disable SSL verification.


gazpachoking:
Thanks, yeah, I didn't like that being off by default.

If this breaks anything for anyone, you probably need python 2.7.9, or to install the dependencies provided by `requests[security]` package on older pythons.

PR #545, opened 2015-06-17
Add search plugin for PTP

Chemondelay:
Configured like so:
```
discover:
  ...
  from:
    - ptp:
        username: *****
        password: *****
        passkey: *****   # Found after `torrent_pass=` in any torrent download link on the site
```

The plugin populates the usual entry fields, as well as three unique boolean fields that can be filtered against:
 * `ptp_checked`, indicating whether the torrent has been checked by staff;
 * `ptp_gp`, indicating a 'Golden Popcorn' (i.e. verified high-quality) release; and
 * `ptp_scene`, indicating a Scene release.

### DIFF
--- a/flexget/plugins/search_ptp.py
+++ b/flexget/plugins/search_ptp.py
@@ -16,7 +16,7 @@ session.set_domain_delay('tls.passthepopcorn.me', '9 seconds')
 base_url = 'https://tls.passthepopcorn.me'
 
 
-class SearchPTP2(object):
+class SearchPTP(object):
 
     schema = {
         'type': 'object',
--- a/flexget/plugins/search_ptp.py
+++ b/flexget/plugins/search_ptp.py
@@ -1,0 +1,79 @@
+from __future__ import unicode_literals, division, absolute_import
+import logging
+
+from flexget import plugin
+from flexget.entry import Entry
+from flexget.event import event
+from flexget.utils import requests, json
+from flexget.utils.search import torrent_availability
+
+log = logging.getLogger('ptp')
+
+session = requests.Session()
+# The site will block your IP for a couple of hours if too many requests are made within a certain period, which seems
+# to be quite short, so the domain delay has been set low enough to allow continuous requesting.
+session.set_domain_delay('tls.passthepopcorn.me', '9 seconds')
+base_url = 'https://tls.passthepopcorn.me'
+
+
+class SearchPTP2(object):
+
+    schema = {
+        'type': 'object',
+        'properties': {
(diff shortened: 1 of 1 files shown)


flexget/plugins/search_ptp.py
```diff
@@ -0,0 +1,79 @@
...
+
+from flexget import plugin
+from flexget.entry import Entry
+from flexget.event import event
+from flexget.utils import requests, json
+from flexget.utils.search import torrent_availability
+
+log = logging.getLogger('ptp')
+
+session = requests.Session()
```
stevezau: You should be able to use 

task.session and task.requests

flexget/plugins/search_ptp.py
```diff
@@ -0,0 +1,79 @@
...
+            'username': {'type': 'string'},
+            'password': {'type': 'string'},
+            'passkey': {'type': 'string'}
+        },
+        'required': ['username', 'password', 'passkey'],
+        'additionalProperties': False
+    }
+
+    def search(self, task, entry, config):
+        session.post(
```
stevezau: Perhaps it should check for failed credentials?

flexget/plugins/search_ptp.py
```diff
@@ -0,0 +1,79 @@
...
+
+        try:
+            r = session.post('%s/torrents.php?json=1&searchstr=%s' % (base_url, search))
+        except RequestException as e:
+            log.error('Error searching PTP: %s' % e)
+            return
+        try:
+            content = json.loads(r.content)
+        except ValueError as e:
+            log.debug('No results from PTP.')
```
stevezau: json.loads(r.content) should work even if it's empty json.. I guess this should say invalid response from server or something similar?

flexget/plugins/search_ptp.py
```diff
@@ -0,0 +1,79 @@
...
+            return
+
+        results = set()
+        if 'Torrents' in content:
+            for item in content['Torrents']:
```
stevezau: could just do 

for item in content.get('Torrents', []):

flexget/plugins/search_ptp.py
nikdoof: This has probably been done to catch error 500s and cloud flare issues, but this isn't the right way to do it. We should be checking ```r.status_code``` and making use of ```r.json```, let requests do the work for us.

liiight:
@Chemondelay is this still relevant?

liiight:
Reopen if relevant

PR #547, opened 2015-06-24
Added include_data parameter & changed only_monitored default to true

liiight:
Added ability to set show quality via Sonarr's 'cutoff' quality parameter.
Also changed the default of only_monitored to true

### DIFF
--- a/flexget/plugins/input/sonarr_emit.py
+++ b/flexget/plugins/input/sonarr_emit.py
@@ -1,6 +1,7 @@
 from __future__ import unicode_literals, division, absolute_import
 from urlparse import urlparse
 import logging
+import math
 from requests import RequestException
 
 from flexget import plugin
@@ -91,7 +92,7 @@ class SonarrEmit(object):
     def on_task_input(self, task, config):
         json = self.get_page(task, config, 1)
         entries = []
-        pages = int(json['totalRecords'] / config.get('page_size')) + (json['totalRecords'] % config.get('page_size') > 0)  # Sets number of requested pages
+        pages = int(math.ceil(json['totalRecords'] / config.get('page_size')))  # Sets number of requested pages
         current_series_id = 0  # Initializes current series parameter
         for page in range(2, pages):
             for record in json['records']:
--- a/flexget/plugins/input/sonarr_emit.py
+++ b/flexget/plugins/input/sonarr_emit.py
@@ -1,0 +1,125 @@
+from __future__ import unicode_literals, division, absolute_import
+from urlparse import urlparse
+import logging
+import requests
+
+from flexget import plugin
+from flexget.event import event
+from flexget.entry import Entry
+
+log = logging.getLogger('sonarr_emit')
+
+
+class Sonarr_emit(object):
+
+    '''
+    This plugin return the 1st missing episode of every show configures in Sonarr.
+    This can be used with the discover plugin or set_series_begin plugin to
+    get the relevant data from Sonarr.
+
+    Syntax:
+
+    sonarr_emit:
+      base_url=<value> (Required)
+      port=<value> (Default is 80)
+      api_key=<value> (Required)
+      include_ended=<yes|no> (Default is yes)
+      only_monitored=<yes|no> (Default is yes)
+      page_size=<value> (Default is 50)
+
+    Page size determines the amount of results per each API call. 
+    Higher value means a bigger response. Lower value means more calls.
+    Should be changed if there are performance issues.
+
+
+    Usage: (Example with discover)
+
+    discover_from_sonarr_task:
+      discover:
+        what:
+          - sonarr_emit:
+              base_url: '{{ secrets.credentials.sonarr.url }}'
+              port: 8989
+              api_key: '{{ secrets.credentials.sonarr.api_key }}'
+              include_ended: false
+        from:
+          - kat:
+              verified: yes
+      all_series: yes
+      download: c:\bla\
+
+    Usage: (Example with set_series_begin)
+
+    set-series-begin-from-sonarr:
+      sonarr_emit:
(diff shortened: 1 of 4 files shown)


liiight:
Could you point me in the direction of why this build failed? I couldn't make sense of it

flexget/plugins/input/sonarr_emit.py
```diff
@@ -0,0 +1,125 @@
...
+        'required': ['api_key', 'base_url'],
+        'additionalProperties': False
+    }
+
+    def get_page(self, task, config, page_number):  # Function that gets a page number and page size and returns the responding result json
+        parsedurl = urlparse(config.get('base_url'))
+        url = '%s://%s:%s%s/api/wanted/missing?page=%d&pageSize=%d&sortKey=series.title&sortdir=asc' % (parsedurl.scheme, parsedurl.netloc, config.get('port'),
+                                                                                                        parsedurl.path, page_number, config.get('page_size'))
+        headers = {'X-Api-Key': config['api_key']}
+        json = task.requests.get(url, headers=headers).json()
```
stevezau: Perhaps we need some exception handling here?

liiight: You're right obviously, i'll add it

flexget/plugins/input/sonarr_emit.py
```diff
@@ -0,0 +1,125 @@
...
+                                  series_season=season,
+                                  series_episode=episode,
+                                  series_id='S%02dE%02d' % (season, episode),
+                                  tvdb_id=record['series']['tvdbId'],
+                                  tvrage_id=record['series']['tvRageId'],
+                                  title=record['series']['title'] + ' ' + 'S%02dE%02d' % (season, episode))
+                    if entry.isvalid():
+                        entries.append(entry)
+                    else:
+                        log.debug('Invalid entry created? %s' % entry)
```
stevezau: Maybe make this log.error?

flexget/plugins/input/sonarr_emit.py
```diff
@@ -0,0 +1,125 @@
...
+                        log.info("    Title: %s" % entry["title"])
+                        log.info("    URL: %s" % entry["url"])
+                        log.info("    Show name: %s" % entry["series_name"])
+                        log.info("    TVDB ID: %s" % entry["tvdb_id"])
+                        log.info("    TVRAGE ID: %s" % entry["tvrage_id"])
+                        log.info("    Season: %s" % entry["series_season"])
+                        log.info("    Episode: %s" % entry["series_episode"])
+                        log.info("    Series ID: %s" % entry["series_id"])
+                    continue
+            json = self.get_page(task, config, page)
```
stevezau: Just a suggestion but perhaps it's better to build json then loop through it.. such as

json[pagenum][entries]

that way you can handle all exceptions for requests before you add the entries

liiight: You mean run get all data into local json and then operate on it in its entirety? I though of that but data can be potentially huge. Image dozens of shows, each one with several seasons of unwatched episodes.
Also note that there is a preliminary API call before the loop to get the total pages value. If there is a outstanding connection error that is derived from bad config, the plugin should stop there.

stevezau: Wouldn't it have to do it either way??

stevezau:
@gazpachoking i'm not 100% sure on this one, code looks ok to me but would prefer you to review before i merge.

liiight:
I appreciate the caution, it's always best to have more eyes look at everything, but as I said before, this PR contains a critical bug fix for the Sickbeard plugin, as it is currently not working (causes all download paths to be null).

stevezau:
@liiight This PR seems to contain alot more then just a bug fix.. Perhaps you can open a new PR for only the bug fix? I'll merge that one in straight away.

liiight:
Sound like a good idea, but i'm not sure how I do that. Can you help?

stevezau:
@liiight might be faster if you point out the code fixes and ill do it manually

liiight:
@stevezau you can just merge sickbeard.py then. It contains error handling and the bug fix (removal of any reference to 'path')

flexget/plugins/input/sonarr_emit.py
```diff
@@ -0,0 +1,132 @@
...
+        except RequestException as e:
+            raise plugin.PluginError('Unable to connect to Sonarr at %s://%s:%s%s. Error: %s'
+                                     % (parsedurl.scheme, parsedurl.netloc, config.get('port'),
+                                        parsedurl.path, e))
+        return json
+
+    def on_task_input(self, task, config):
+        json = self.get_page(task, config, 1)
+        entries = []
+        pages = int(json['totalRecords'] / config.get('page_size')) + (json['totalRecords'] % config.get('page_size') > 0)  # Sets number of requested pages
```
gazpachoking: I think this would be nicer if we do `math.ceil(json['totalRecords'] / config.get('page_size'))`

liiight: I think there was a reason I did it like that, but I can't figure out why. I'll change this.

flexget/plugins/input/sonarr_emit.py
liiight: But I wouldn't have to loop over everything twice (even though, 2nd loop is a local representation)

liiight:
Any update with merging this PR?

gazpachoking:
I would prefer the new plugin be in a separate PR to fixes next time, but it all looks good to me. Thanks!

gazpachoking:
@liiight If you can update the wiki as needed that would be great. :smile:

liiight:
Wiki already updated. I'll make sure new plugins will be separated next time, thanks!

PR #548, opened 2015-06-29
Fixes in trakt_emit: unused imports, UnboundLocalError of var `eps` #3033

antivirtel:
Responsible for testing is qvazzler.

### DIFF
--- a/flexget/plugins/input/trakt_emit.py
+++ b/flexget/plugins/input/trakt_emit.py
@@ -101,12 +101,12 @@ class TraktEmit(object):
                             else:
                                 epn += 1
                         break
-                    else:
-                      if config['position'] == 'next':
-                        eps = epn = 1
-                      else:
-                        # There were no watched/collected episodes, nothing to emit in 'last' mode
-                        continue
+                else:
+                  if config['position'] == 'next':
+                    eps = epn = 1
+                  else:
+                    # There were no watched/collected episodes, nothing to emit in 'last' mode
+                    continue
             if eps and epn:
                 entry = self.make_entry(fields, eps, epn)
                 entries.append(entry)
--- a/flexget/plugins/input/trakt_emit.py
+++ b/flexget/plugins/input/trakt_emit.py
@@ -1,5 +1,4 @@
 from __future__ import unicode_literals, division, absolute_import
-import hashlib
 import logging
 from urlparse import urljoin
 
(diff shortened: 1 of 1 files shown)


antivirtel:
Maybe an `eps = null` would be better, since rarely there used to be 0th episodes, pilots, etc...

qvazzler:
Hi Gaz, please merge so I can use trakt_emit again. :+1:

flexget/plugins/input/trakt_emit.py
```diff
@@ -103,6 +101,8 @@ def on_task_input(self, task, config):
                             else:
                                 epn += 1
                         break
+                    else:
+                        eps = 0;
```
stevezau: not sure why this line ends in a semicolon?

Also you could cleanup the above logic 

IE:
```
if config['position'] == 'next':
                             if seas['completed'] >= seas['aired']:
```

to:
```
if config['position'] == 'next' and seas['completed'] >= seas['aired']:
```

antivirtel:
The semicolon is just a routine for me @stevezau, mainly, I'm Java developer. I'll remove it in the next commit, but I'm still waiting for @gazpachoking's help, to figure out what to do with that `eps = 0`.

gazpachoking:
I think probably something like this would be okay as an `else` to the for loop.
```
                else:
                    if config['position'] == 'next':
                        eps = epn = 1
                    else:
                        # There were no watched/collected episodes, nothing to emit in 'last' mode
                        continue
```

gazpachoking:
Hmm, I meant the else block on the for loop, rather than that if statement. @cvium Did you want to give this a shot so @antivirtel isn't doing it blind?

cvium:
Sure... once it's been updated.

antivirtel:
@gazpachoking did you want to place that code block into this `for` loop's 1st or 2nd `if`?( https://github.com/Flexget/Flexget/blob/develop/flexget/plugins/input/trakt_emit.py#L92 or https://github.com/Flexget/Flexget/blob/develop/flexget/plugins/input/trakt_emit.py#L97 )

cvium:
@gazpachoking is suggesting a `for-else` loop. It should not be inside the `for`-loop.

antivirtel:
Ah I hope I get it now (https://docs.python.org/2/tutorial/controlflow.html#break-and-continue-statements-and-else-clauses-on-loops) - so an `else` into this `for`: https://github.com/Flexget/Flexget/blob/develop/flexget/plugins/input/trakt_emit.py#L90 !

cvium:
Yes.

cvium:
I am unable to test this for the next 2 weeks, so if someone else could handle this it'd be cool.

qvazzler:
Really crammed up at work around this time guys.. Will try when I get the chance.

qvazzler:
Just tested your PR again Antivirtel, it's working! :+1: :+1: :+1:

gazpachoking:
Thanks everyone!

PR #549, opened 2015-06-29
add option to allow unverified SSL requests to pyload

manacker:
I like to run pyload in HTTPS mode. Since the python version on my NAS doesn't allow proper certificate validation (InsecurePlatformWarning) and it doesn't make much sense for https://localhost anyways the API calls to pyload failed. 

So I added an option to disable SSL certificate verification in the requests.request call. It still spams a lot of warnings, but at least it works now.

The default behavior is as current, you have to explicitly set "verify_ssl: no" to deactivate SSL verification.

### DIFF
--- a/flexget/plugins/output/pyload.py
+++ b/flexget/plugins/output/pyload.py
@@ -32,6 +32,7 @@ class PluginPyLoad(object):
         parse_url: no
         multiple_hoster: yes
         enabled: yes
+        verify_ssl: yes
 
     Default values for the config elements::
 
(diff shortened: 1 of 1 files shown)


stevezau:
You should be able to do this already via

 http://www.flexget.com/wiki/Plugins/verify_ssl_certificates

??

manacker:
I tried that, but it didn't work. Might be because the pyload plugin doesn't use the task.requests session, but its own instance. 
So I guess its better if I create a patch to respect the existing verify_ssl_certificates plugin instead of building my own implementation.

stevezau:
@manacker yep, i suggest you change the pyload plugin to use tasks.requests.

I will close this PR so you can open a new PR.

PR #551, opened 2015-07-09
Add SFTP socket timeout

ianstalk:
Add SFTP socket timeout to fix hung SFTP sessions; a couple of other minor tweaks.


gazpachoking:
Thanks!

PR #554, opened 2015-07-17
T411.io: Adapted SUB_CATEGORIES to the new site inputs

jmcabrera:
Just passing by :)
The format of the inputs on t411.io has changed since jul 8th approx.
This PR is to reflect these changes.
Keep up the good work!


stevezau:
I'm assuming this is right as i don't use t411.io..

PR #555, opened 2015-07-21
Freshontv search plugin

OmgOhnoes:
Adds a Freshon.tv search plugin.

Supports pagination (limited to 10 pages as users can set their pages to up to 100 items each).
Has a 100% and 50% torrent filter as well as HD and WebDL filters.

Downloads via RSS download link using the users passkey.

Note: throws a SSL error when getting the pages, not sure how to fix that as every recommendation doesn't seem to work including "verify_ssl_certificates: no"

### DIFF
--- a/flexget/plugins/search_freshon.py
+++ b/flexget/plugins/search_freshon.py
@@ -1,0 +1,167 @@
+from __future__ import unicode_literals, division, absolute_import
+import logging
+import re
+import urllib 
+
+from flexget import plugin
+from flexget.entry import Entry
+from flexget.event import event
+from flexget.utils import requests
+from flexget.utils.soup import get_soup
+from flexget.utils.search import torrent_availability, normalize_unicode
+
+log = logging.getLogger('search_freshon')
+
+session = requests.Session()
+
+
+class SearchFreshon(object):
+
+    schema = {
+        'type': 'object',
+        'properties':
+        {
+            'username': {'type': 'string'},
+            'password': {'type': 'string'},
+            'passkey': {'type': 'string'},
+            'category': {
+                'type': 'string',
+                'enum': ['hd', 'webdl', 'all']
+            },
+            'freeleech': {
+                'type': 'string',
+                'enum': ['free', 'half', 'all']
+            },
+        },
+        'required': ['username', 'password', 'passkey'],
+        'additionalProperties': False
+    }
+
+    def prepare_config(self, config):
+        config.setdefault('freeleech', 'all')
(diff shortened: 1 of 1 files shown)


flexget/plugins/search_freshon.py
```diff
@@ -0,0 +1,197 @@
...
+                'type': 'string',
+                'enum': ['free', 'half', 'all']
+            },
+        },
+        'required': ['username', 'password', 'passkey'],
+        'additionalProperties': False
+    }
+
+    def prepare_config(self, config):
+        config.setdefault('freeleech', 'all')
```
stevezau: You could use schema defaults here.

```
'category': {
    'type': 'string',
    'default': 'all',
    'enum': ['hd', 'webdl', 'all']
  },
```

flexget/plugins/search_freshon.py
```diff
@@ -0,0 +1,197 @@
...
+        if config['freeleech'] == 'all':
+            freeleech = 0
+        elif config['freeleech'] == 'free': 
+            freeleech = 3
+        elif config['freeleech'] == 'half': 
+            freeleech = 4
+        entries = set()
+
+        for search_string in entry.get('search_strings', [entry['title']]):
+            search_string = search_string.replace(' ', '-').lower()
```
stevezau: regex replace might be more efficient here.

flexget/plugins/search_freshon.py
```diff
@@ -0,0 +1,197 @@
...
+            search_string = search_string.replace('(', '')
+            search_string = search_string.replace(')', '')
+            query = normalize_unicode(search_string)
+            query_url_fragment = urllib.quote_plus(query.encode('utf-8'))
+            str_url = (base_url, search_fragment)
+            url = '/'.join(str_url) % (query_url_fragment, config['category'], str(freeleech))
+            log.debug('Search url: %s' % url)
+            f = session.get(url).content
```
stevezau: Could use task.session instead as it will solve your ssl issues

OmgOhnoes:
I appreciate the feedback, still learning python and Flexget is a beast so i'm doing my best with previous plugin examples.

I've done some of the changes you've suggested and optimized the code a bit.

Using task.session does not seem to have fixed the SSL errors even while using "verify_ssl_certificates: no." On the plus side it seems to respect domain delays now.

gazpachoking:
Thanks! Want to add a wiki page to searches section? http://flexget.com/wiki/Searches

OmgOhnoes:
Will do, thanks!

PR #556, opened 2015-07-22
Input plugin for AniDB wishlist

radek-sprta:
An input plugin that generates a list of movie entries from user's AniDB (http://anidb.net) wishlist. The wishlist must be public.


gazpachoking:
@radek-sprta Sorry it took so long. Want to update the wiki with this plugin?

PR #557, opened 2015-07-29
Optional regex module

gazpachoking:
This branch uses the `regex` module from here if installed:
https://pypi.python.org/pypi/regex

This allows users to override regex flags in the config when this module is installed. e.g.
`(?i)something` to ignore case when the plugin doesn't by default
`(?V1-i)someTHING` to turn on case sensitivity even if plugin doesn't by default. (The V1 is necessary for toggling off flags)

TODO:
- [x] Add a better validator message if inline flags are attempted in config without regex module installed

### DIFF
--- a/flexget/config_schema.py
+++ b/flexget/config_schema.py
@@ -1,7 +1,6 @@
 from __future__ import absolute_import, division, unicode_literals
 
 import os
-import re
 import urlparse
 from collections import defaultdict
 from datetime import datetime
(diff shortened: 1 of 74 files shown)


gazpachoking:
@paranoidi This seem like a good idea to you? Is having everything import from `flexget.utils.regex` the nicest way to accomplish this?

gazpachoking:
Not much feedback here. I don't want to introduce complication if not needed, so closing.

PR #559, opened 2015-07-29
Added ability to pass quality via CP plugin, pep8 fixes

liiight:
Added ability to pass quality via CP plugin, pep8 fixes, handled connection exceptions


gazpachoking:
@liiight So are these actually quality requirements that it's filling? Maybe we should put them in `quality_requirements` field, make it actual `QualtyRequirements` object rather than `Quality` and change `movie_queue` `add` mode to look at that field first.

flexget/plugins/input/couchpotato.py
```diff
@@ -30,43 +30,85 @@ def on_task_input(self, task, config):
...
         for movie in json['movies']:
+            quality = ''
             if movie['status'] == 'active':
+                if config.get('include_data'):
+                    for profile in profile_json['list']:
+                        if profile['_id'] == movie['profile_id']:  # Matches movie profile with profile JSON
+                            # Creates a string of flexget qualities from CP's qualities list
+                            converted_list = ', '.join([cp_to_flexget[quality] for quality in profile['qualities']])
+                            try:
+                                quality = qualities.Quality (converted_list)  # Return the best components from the list
```
gazpachoking: Got an extra space in here before the parens.

liiight:
@gazpachoking I guess your suggestion would make sense more than the current implementation, but I think if movie_queue is being refactored with regards to quality input, we can give it some more though and make it even more robust and generic than it currently is.

liiight:
@gazpachoking you wanna merge this or should I start looking into your suggestion of passing quality requirement instead?

gazpachoking:
@liiight Yeah, I guess that sounds fine. If we fix up movie_queue quality picking we can revisit this.

PR #563, opened 2015-08-05
Fix subtitles to work for current stable version of subliminal (1.0.1)

camon:
Updated to work with the current stable version of subliminal (1.0.1)

This solves this problem http://flexget.com/ticket/3047


camon:
@nikdoof thanks

PR #564, opened 2015-08-08
Twitter Output Plugin

nikdoof:
I've (finally) got around to updating my [output_twitter plugin](https://github.com/nikdoof/flexget-twitter) and submitting a pull for it to be included into the mainline. But, it has a few issues:

* It has a interactive CLI element, this is to assist in getting the correct oAuth tokens for the user, but as this is really the only interactive element in Flexget i'd like some feedback if its OK or not.
* It uses ```tweepy``` and not ```twitter``` like the ```twitterfeed``` plugin uses, yes its more dependencies, but I feel that [tweepy's development](https://github.com/tweepy/tweepy/commits/master) is more active than [twitter](https://github.com/sixohsix/twitter/commits/master) at the moment.
* Obvious conflict of naming, as we have ```twitterfeed``` and ```twitter```.

Feedback is welcome and needed :smile:

### DIFF
--- a/flexget/plugins/plugin_twitter.py
+++ b/flexget/plugins/plugin_twitter.py
@@ -1,0 +1,100 @@
+from __future__ import print_function
+import logging
+
+from flexget import options
+from flexget import plugin
+from flexget.event import event
+from flexget.utils.template import RenderError
+
+try:
+    import tweepy
+except ImportError:
+    tweepy = None
+
+log = logging.getLogger('twitter')
+
+
+def dependency_check():
+    if not tweepy:
+        raise plugin.DependencyError(issued_by='twitter', missing='tweepy',
+                                     message='twitter plugin requires the tweepy Python module.')
+
+
+class OutputTwitter:
+
+    schema = {
+        'type': 'object',
+        'properties': {
+            'active': {'type': 'boolean', 'default': True},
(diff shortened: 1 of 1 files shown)


gazpachoking:
> It has a interactive CLI element, this is to assist in getting the correct oAuth tokens for the user, but as this is really the only interactive element in Flexget i'd like some feedback if its OK or not.

I was also thinking about this problem as trakt plugin needs to move to oauth soon. I was thinking more along the lines of a cli command that took the different needed flags, and stored the auth in the database against an username. Then you'd specify the username in the config, which would be looked up against the database at runtime. Maybe even group this under a single cli command that different plugins registered to e.g.
```
$flexget auth twitter --username myname --key blahblah --secret --aoeuaoeu
Open the following URL in a browser to authenticate Flexget to access your Twitter account: http://blah
$flexget auth twitter --username myname --pin aoeua
```
Not sure if that's good or not. Hopefully it'd have nice help text when you were missing flags for current stage of auth. Does that sound better or worse to you?
> It uses tweepy and not twitter like the twitterfeed plugin uses, yes its more dependencies, but I feel that tweepy's development is more active than twitter at the moment.
> Obvious conflict of naming, as we have twitterfeed and twitter.

Worth converging that to the same dep and sorting out the names after this?

liiight:
Is this still relevant?

PR #565, opened 2015-08-08
notify_xmpp - Add support for multiple recipients

nikdoof:
As requested in [#3019](http://flexget.com/ticket/3019), add support to send to multiple recipients. This change will probably need a note on the update log for changing "recipient" to "recipients" in the config.

Also some minor PEP8 cleanup while I was committing.


gazpachoking:
Thanks! want to update the wiki?

nikdoof:
Will do.

PR #568, opened 2015-08-16
[WIP] IRC Daemon Plugin

nikdoof:
[Discussion thread on the forums](http://discuss.flexget.com/t/plugin-wip-irc-input/1497)

Adds a new daemon mode plug-in for IRC. New entries are devised from bot announcements in an IRC channel and are fed into Flexget. The idea is to have something like autodl-irssi/pywhatauto with the power of Flexget on the backend.

It supports autodl-irssi config files and manual configuration, but more functionality is unlocked if using the configuration files. 

Still WIP and is being tested.

### DIFF
--- a/flexget/plugins/daemon/irc.py
+++ b/flexget/plugins/daemon/irc.py
@@ -4,9 +4,10 @@ import re
 import threading
 import logging
 from random import choice
+from xml.etree import ElementTree as ET
+import urllib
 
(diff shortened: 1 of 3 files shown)


nikdoof:
Closing this pull request, as the IRC lib is Py2.7 only, and Flexget requires 2.6 support. 

Also a lot more work will be needed before its in a mergable format, which by then will need all these commits rolling up into one.

PR #569, opened 2015-08-18
Allow for self-signed SSL certificates in Transmission

ianstalk:
Python 2.7.9 and newer are much stricter about self-signed certificates. Add a config parameter to disable verification.

### DIFF
--- a/flexget/plugins/plugin_transmission.py
+++ b/flexget/plugins/plugin_transmission.py
@@ -77,11 +77,6 @@ class TransmissionBase(object):
         from transmissionrpc import TransmissionError
         from transmissionrpc import HTTPHandlerError
 
-        if not config['verify_ssl_certificates']:
-            import ssl
-            log.verbose('Setting unverified SSL context.')
-            ssl._create_default_https_context = ssl._create_unverified_context
-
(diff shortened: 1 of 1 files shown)


gazpachoking:
Can we (and would it be better) to subclass [this](https://bitbucket.org/blueluna/transmissionrpc/src/2d0a22304f0c74fb537221c3c8b49a603a9480eb/transmissionrpc/httphandler.py?at=default#httphandler.py-48) and pass it in to transmissionrpc?

ianstalk:
That's probably a more principled approach. I'll take a look into it.

gazpachoking:
@ianstalk Once this is set, will it then apply to every task for the length of the daemon run?

ianstalk:
I couldn't actually figure out how to do this the right way. It's
remarkably difficult to turn off SSL certificate verification, and there's
no easy way to add a self-signed cert to a trusted store.

gazpachoking:
I just meant, should we have a task_exit/abort handler that sets it to previous state after the task runs.

liiight:
Where we are with this?

liiight:
Reopen if relevant

PR #571, opened 2015-08-23
regex_extract plugin

nikdoof:
This is a new plugin to allow an entry to be updated with the results of a regex over a specified field. This was created to scratch a personal itch of having to post-process Formula 1 from RFM, so with a good enough regex I can parse the title data and make it available for using with deluge ```movedone``` and ```content_filename```:

    regex_extract:
      prefix: f1_
      regex: ^Formula\.?1\.(?P<year>(?:19|20)\d{2})\.Round(?:\.)?(?P<round>\d{1,2})\.(?P<location>[a-zA-Z\.]*?)\.(?P<event>Teds\.(?:Qualifying|Race)\.Notebook|FP[1-3]|Quali(?:fying)?(?:\.Build-Up)?|Race(?:\.Build-Up)?|(?:Drivers\.|Team\.Principals\.)PC|Drivers\.Parade)(?:\.(?P<channel>SkySports(?:F1)?(?:HD)?|BBC(?:(?:One|Two)(?:HD)?)?|Sat\.Feed|ITV(?:HD)?|RTBF(?:HD)?|SkySportHD1|ORTF))?(?:\.(?P<quality>720p|1080(?:i|p)?|DVD|Web-Rip))?(?:\.(?P<codec>[XH]26[45]|MPEG-2|XVID|DIVX|MP3))?(?:\.(?P<language>English|French|German|Portuguese))?(?:\.(?P<comment>[\w\.]+))?(?:-(?P<group>[\w\.]+))?$
      field: title
    set:
      label: sport
      content_filename: "Formual 1 - Round {{ f1_round }} - {{ f1_location|int }} - {{ f1_event|replace('.', ' ') }}"
      movedone: "/srv/media/Sport/Motorsports/Formula 1/{{ f1_year }}/Round {{ f1_round|int }} - {{ f1_location|replace('.', ' ' }}/"

### DIFF
--- a/flexget/plugins/modify/regex_extract.py
+++ b/flexget/plugins/modify/regex_extract.py
@@ -1,0 +1,87 @@
+from __future__ import unicode_literals, division, absolute_import
+import logging
+import re
+
+from flexget import plugin
+from flexget.event import event
+from flexget.utils.tools import ReList
+from flexget.config_schema import one_or_more
+
+log = logging.getLogger('regex_extract')
+
+
+class RegexExtract(object):
+    """
+    Updates an entry with the values of regex matched named groups
+
+    Usage:
+
+      regex_extract:
+        field: <string>
+        regex:
+          - <regex>
+        [prefix]: <string>
+
+
+    Example:
+
+      regex_extract:
+        prefix: f1_
+        field: title
+        regex:
+          - Formula\.?1.(?P<location>*?)
+
+    """
+
+    schema = {
+        'type': 'object',
+        'properties': {
+            'prefix': {'type': 'string'},
+            'field': {'type': 'string'},
+            'regex': one_or_more({'type': 'string', 'format': 'regex'}),
+        },
+    }
+
+    def on_task_start(self, task, config):
+        regex = config.get('regex')
+        if isinstance(regex, basestring):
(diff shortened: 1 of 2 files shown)


gazpachoking:
This seems pretty similar to [manipulate](http://flexget.com/wiki/Plugins/manipulate) plugin extract mode, can that already fulfil this purpose?

nikdoof:
Not really, as manipulate doesn't work with groups. If you feel it'll be better to merge this into manipulate as another option then i'll happily do so, as I understand my usage is quite the edge case :)

gazpachoking:
Oh, I see, it's creating new fields based on the regex capture group names? Hmm, guess keeping it separate sounds fine.

nikdoof:
Yeah, sorry I really should of put a better description.

It makes use of regex groups to match and extract the values and apply them straight to the Entry object values, the ```prefix``` field is in there if people want to add a prefix onto the groups.

Like in my example all my values are related to f1 so I prefix them as such, my regex extracts all the individual values out a complicated title then I use these values in the entry in ```content_filename``` and ```movedone``` on the ```deluge``` plugin.

gazpachoking:
Cool, want to add it to the wiki?

PR #572, opened 2015-08-23
Pin guessit to < 0.10.4, as it drops Py2.6 support.

nikdoof:
As title, recent test breakages are due to GuessIt dropping Python 2.6 support in a point release.


gazpachoking:
Oh shoot. I came to this conclusion on my own before seeing you already solved it. Sorry about that.

PR #573, opened 2015-08-23
A pipe now splits all show type commands

sean797:
This allow me to nicely pipe the output through other programs when some text in columns are too big - often the case with "flexget plugins".

### DIFF
--- a/flexget/plugins/cli/backlog.py
+++ b/flexget/plugins/cli/backlog.py
@@ -12,13 +12,13 @@ def do_cli(manager, options):
         num = clear_entries(options.task)
         console('%s entries cleared from backlog.' % num)
     else:
-        cols = '{:<65.64}{:<15.15}'
-        console(cols.format('Title', 'Task'))
+        cols = '{:<65.64}{:<1.1}{:<15.15}'
+        console(cols.format('Title', '|', 'Task'))
         console('-' * 80)
         with Session() as session:
             entries = get_entries(options.task, session=session)
             for entry in entries:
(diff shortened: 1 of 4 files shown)


nikdoof:
I don't think this solves your use case. I still have cutting on some commands.

```The Daily Show / 2014-09-10 / Jeremy Renner [720p / x264 / WEBRi |tv   ```

If you want a parsable output then really it should be a extra flag/option on the CLI, then you can forgo trying to format any of the data and essentially spit out a pipe delimited output, or you could switch all the output to use something like ```prettytable``` or ```tabulate``` to do the formatting and sizing for you.

sean797:
What command did you give to get that output? 
I disagree about it should be a extra flag/option. Having a table with pipe's between columns cause no issues, only makes the output easier to read.

The reason of this commit is I'm writing a bash completion script and enables me to tab complete commands like "flexget doc" with the output of "flexget plugins | awk -F '|' '/--doc/{print $1}'". If you have a better way of doing it let me know?

gazpachoking:
I think a better api is a nicer solution than trying to change cli options to be machine readable. It's something we are working on in the rest_api branch right now.

paranoidi:
It's pretty common in cli apps to provide machine parseable output with argument. At least sometimes referred porcelain or so.

sean797:
I've created a new pull request with --porcelain option on some commands. Using an api to get the data I need for bash completion purposes wouldn't work that well as it will require the daemon be running, some people use cron and don't run Flexget as a service.

gazpachoking:
#574 merged

PR #575, opened 2015-08-26
Fix subtitles subliminal http://flexget.com/ticket/3061

camon:


camon:
Failed checks why?!

gazpachoking:
> Failed checks why?!

Guessit (one of our deps) dropped 2.6 support. We need to update some of our requirements to fix that. Probably nothing to do with your changes.

camon:
@gazpachoking ok thanks

gazpachoking:
Thanks!

PR #576, opened 2015-08-27
Bugfix for 'cannot import name path' with path.py 8.0

thalamus:
The update to path.py removes the lowercase path alias which flexget relies on
Simply change it to use 'Path' instead of 'path.

http://flexget.com/ticket/3064

### DIFF
--- a/flexget/plugin.py
+++ b/flexget/plugin.py
@@ -11,7 +11,7 @@ import time
 import warnings
 from itertools import ifilter
 
-from path import path
+from path import Path
 from requests import RequestException
 
 from flexget import plugins as plugins_pkg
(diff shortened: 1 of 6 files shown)


gazpachoking:
Awesome, thanks!

gazpachoking:
Maybe we should also update our requirements to say path.py >= 6.2 (the version where the upper case was introduced.)

thalamus:
Yeah, I've just done a commit for that, I'll open a pull request for it :)

thalamus:
There's quite a few unhappy people on that path bug: https://github.com/jaraco/path.py/issues/102
Looks like the change has broken a lot of code and it seems to have caused a bit of a fuss.

antivirtel:
You can still commit to this Branch, you don't need to open a new one!

PR #577, opened 2015-08-27
Update path.Path

antivirtel:
According to the history: http://pythonhosted.org/path.py/history.html


gazpachoking:
We need to update the usage in addition to the import. :wink:

thalamus:
Already done: https://github.com/Flexget/Flexget/pull/576

PR #578, opened 2015-08-28
Make Flexget dependent on Path 6.2 or higher

thalamus:
Fallout from http://flexget.com/ticket/3064


gazpachoking:
You are quick! Thanks!

thalamus:
No problem. Flexget has saved me a ton of time and effort in the last couple of years since I started using it, so it's nice to give something back.

Incidentally, Path 8.1 restores the 'path' alias with a deprecated warning.

PR #579, opened 2015-09-02
Fix Plex Token Authentication to check local/public ip address. FIX #…

crawln45:
Fix Plex Token Authentication to check both local/public ip address. FIX# 2794, 3063


planeturban:
Looks ok.

PR #580, opened 2015-09-03
Added bash completion support

sean797:
Only installing bash completion script if platform is Linux, bash completion is installed & installation is run as root


gazpachoking:
So... I think bash completion would be really cool, what I am worried about is maintenance though. The best would be if we could auto-generate this based on our argparse instance. Otherwise maybe specifying the completion info would be an optional part of registering a cli command? That way the completion info would at least be stored with the cli handler, and it would theoretically be easier to keep them in sync.

sean797:
Yes maintenance could be tough but i don't think the cli should change too often? The only problem with that is it would make it very slow. For example add this to your setup and try completing "flexget history --task". It's slow because its running "flexget history --limit 100000 | sed -n '/Task    : /s/ Task    : //p". We could use the output of --help on each command but it would be just as slow. 

We could add comments to the argparse sections telling people to update this script as well?

extras/bash_completion/flexget
```diff
@@ -0,0 +1,272 @@
...
+_flexget_common_opts="-h --help"
+_flexget_execute_opts="--tasks --learn --no-cache --template -v -s --try-regexp --stop-waiting --disable-tracking --cli-config --dl-path --dump --dump-config --now --discover-now --tail-reset"
+
+_flexget_execute()
+{
+  local opts="${_flexget_execute_opts} ${_flexget_common_opts}"
+  COMPREPLY=($(compgen -W "${opts}" -- ${1}))
+}
+
+_fexget_daemon()
```
liiight: Typo:
'fexget' instead of 'flexget'

liiight:
this is ready, right?

sean797:
I believe so.

liiight:
Thanks for this, seems very useful. Care to update the wiki?

stevezau:
@sean797 can you please take a look at this PR and maybe open a new one when its fixed... I had to revert it as it broke the package process. No one could install flexget due to issue #829 

It seems the files were not included when paver sdist ran.

y:
> So... I think bash completion would be really cool, what I am worried about is maintenance though. The best would be if we could auto-generate this based on our argparse instance.

https://argcomplete.readthedocs.io/

liiight:
That one is actually on my backlog...

PR #584, opened 2015-09-13
Smart imdb plugin

liiight:
This plugin enables generating entries based on an entity, an entity being a person, character or company. It's based on IMDBpy which is required (pip install imdbpy).

### DIFF
--- a/flexget/plugins/input/smart_imdb.py
+++ b/flexget/plugins/input/smart_imdb.py
@@ -2,11 +2,10 @@ from __future__ import unicode_literals, division, absolute_import
 
 import logging
 import re
-from imdb._exceptions import IMDbDataAccessError
+from jsonschema.compat import str_types
 from flexget import plugin
 from flexget.event import event
 from flexget.entry import Entry
-
 
 from imdb import IMDb
 from flexget.config_schema import format_checker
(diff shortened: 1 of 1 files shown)


flexget/plugins/input/smart_imdb.py
```diff
@@ -0,0 +1,478 @@
+from __future__ import unicode_literals, division, absolute_import
+
+import logging
+import re
+from jsonschema.compat import str_types
+from flexget import plugin
+from flexget.event import event
+from flexget.entry import Entry
+
+from imdb import IMDb
```
cvium: This will throw an exception if imdbpy is not installed? Should be handled with a try-catch block imo.

liiight: I updated based on this feedback, but now i'm thinking i should have done the whole thing different, and split this into two plugins. one for the input and one for the filter... 
what do you think?

cvium: No idea. I'm not quite sure what the plugin does internally and I don't have time to taker a more thorough look at it :stuck_out_tongue:

liiight: ok then, then i'll guess i'll go with hunch and split it then :)

PR #585, opened 2015-09-14
Subliminal should fail if no subs were downloaded

ratoaq2:
subliminal.download_best_subtitles might return dictionaries with keys but empty values (None or empty lists). In this case, no subtitle was downloaded and the entry should fail.


cvium:
Thanks.

PR #586, opened 2015-09-16
Dynamic IMDB plugin

liiight:
Dynamically generate entries based on an IMDB person, character or company.
The user can also specifiy required content types(s), and job type(s).


gazpachoking:
As discussed in irc, I think we should tighten up the schema for the id, something along the lines:
```
pattern: (nm|co|ch)\d{7}
error_pattern: Get the id from the url of the person/company you want to use, e.g. http://imdb.com/person/<id here>/blah
```

liiight:
Well I still need to determine the entity type via the ID. Each entity uses a different method to get the relevant object.

gazpachoking:
> Well I still need to determine the entity type via the ID. Each entity uses a different method to get the relevant object

Ahh, yeah, I was looking at it out of context. Could probably just switch to prefix checking though, i.e. .`if imdb_id.startswith('nm'): ...`

liiight:
I suppose i could switch to that but i still need to get the id itself from the string. Isn't it better to use regex for that than some convoluted string manipulation?

gazpachoking:
> Isn't it better to use regex for that than some convoluted string manipulation?

``` python
        if imdb_id.startswith('nm'):
            person = self.ia.get_person(imdb_id[2:])
            log.info('Starting to retrieve items for person: %s' % person)
            return imdb_entity_type, person
        elif imdb_id.startswith('co'):
            company = self.ia.get_company(imdb_id[2:])
            log.info('Starting to retrieve items for company: %s' % company)
            return imdb_entity_type, company
         elif imdb_id.startswith('ch'):
            character = self.ia.get_character(imdb_id[2:])
            log.info('Starting to retrieve items for Character: %s' % character)
            return imdb_entity_type, character
```
Doesn't seem that convoluted to me. Seems much simpler in fact, no loop, no regex. You can send another PR if you agree.

Thanks for this btw, seems a neat way to automatically seed a watchlist or movie queue.

liiight:
I guess it does makes more sense with strings, ill get on it. 

Np for the plugin, thanks for this amazing platform  :grinning:

PR #587, opened 2015-09-16
Bug fix for multiple token support in Pushalot plugin

liiight:
Multiple tokens were not configured properly in schema

### DIFF
--- a/flexget/plugins/output/pushalot.py
+++ b/flexget/plugins/output/pushalot.py
@@ -31,12 +31,15 @@ class OutputPushalot(object):
     Configuration parameters are also supported from entries (eg. through set).
     """
     default_body = "{% if series_name is defined %}{{tvdb_series_name|d(series_name)}} " \
-                   "{{series_id}} {{tvdb_ep_name|d('')}}{% elif imdb_name is defined %}{{imdb_name}} "\
+                   "{{series_id}} {{tvdb_ep_name|d('')}}{% elif imdb_name is defined %}{{imdb_name}} " \
                    "{{imdb_year}}{% else %}{{title}}{% endif %}"
     schema = {
         'type': 'object',
         'properties': {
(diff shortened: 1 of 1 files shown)


flexget/plugins/output/pushalot.py
```diff
@@ -31,12 +31,15 @@ class OutputPushalot(object):
     Configuration parameters are also supported from entries (eg. through set).
     """
     default_body = "{% if series_name is defined %}{{tvdb_series_name|d(series_name)}} " \
-                   "{{series_id}} {{tvdb_ep_name|d('')}}{% elif imdb_name is defined %}{{imdb_name}} "\
+                   "{{series_id}} {{tvdb_ep_name|d('')}}{% elif imdb_name is defined %}{{imdb_name}} " \
```
gazpachoking: Can we just ditch the `\` here? Default python string concatenation or whatever it's called should take care of this, no?

flexget/plugins/output/pushalot.py
```diff
@@ -31,12 +31,15 @@ class OutputPushalot(object):
...
                    "{{imdb_year}}{% else %}{{title}}{% endif %}"
     schema = {
         'type': 'object',
         'properties': {
-            'token': one_or_more({'type': 'string'}),
+            'token': {'oneof': [
+                {'type': 'string'},
+                {'type': 'array', 'items': 'string'}
+            ]},
```
gazpachoking: Doesn't `one_or_more` produce the same schema?

liiight: I wasn't actually aware of that, so yeah, keeping it makes perfect sense. I upgraded the schema in the lastest commit to have the ability to use just a string without an explicit token property. Also tested it all using 1 & 2 tokens and everything is peachy.

gazpachoking:
Thanks!

PR #588, opened 2015-09-17
Let kat handle correctly 404 when no result found

X-dark:
When no result is found, kat return an empty page with a 404 error. This should not be catched as a warning.

### DIFF
--- a/flexget/plugins/search_kat.py
+++ b/flexget/plugins/search_kat.py
@@ -59,12 +59,15 @@ class SearchKAT(object):
 
                 log.debug('requesting: %s' % url)
                 try:
-                    r = task.requests.get(url, params=params)
+                    r = task.requests.get(url, params=params, raise_status=False)
                 except RequestException as e:
                     log.warning('Search resulted in: %s' % e)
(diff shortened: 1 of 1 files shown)


gazpachoking:
I'm not sure this will solve the issue, as r.content will still have something when the no results 404 is returned.

cvium:
I agree. It's an old issue and I have found no graceful way of handling it. We cannot properly differentiate between a proper 404 and the 404 they send back on "no results".

X-dark:
This is solving the issue. I have tested my change before making the pull request.
@gazpachoking what you missed in your snippet on IRC and when researching the issue is the ```rss=1``` parameter added to the URL.
https://kat.cr/usearch/azeaeaez%20verified%3A1/?category=tv&rss=1 is returning a 404 with no content.
https://kat.cr/xsearch/azeaeaez%20verified%3A1/?category=tv&rss=1 is returning a 404 page.

gazpachoking:
Aha, you are correct, I was looking at the regular web results. Thanks!

PR #589, opened 2015-09-18
sftp_upload plugin

ianstalk:
Plugin to upload files to an SFTP server.


gazpachoking:
Thanks, want to add it to the wiki?

PR #592, opened 2015-09-26
Add hostname to timeout exception message.

tobinjt:
This will make it easier to debug which host is the problem.


gazpachoking:
Good idea.

PR #597, opened 2015-10-01
Fix for pushover plugin 

liiight:
Added try catch block for request


gazpachoking:
Should we just catch all RequestsExceptions instead of just ConnectionErrors?

liiight:
Sure we should. I was just checking if you're paying attention...  :stuck_out_tongue_winking_eye: 
I'll change it

PR #598, opened 2015-10-03
Add more logging when setting headers.

tobinjt:


flexget/utils/requests.py
```diff
@@ -145,13 +145,23 @@ def request(self, method, url, *args, **kwargs):
...
+            msg = str(e)
+            log.debug('Exception fetching %s: %s' % (url, msg))
+            # torcache.net answers HTTP GET requests but doesn't properly answer HTTP HEAD requests - it outputs an
+            # empty string which causes a ProtocolError to be raised, which eventually turns into a ConnectionError
+            # exception.  urlrewriting unconditionally does a HEAD request for all URLs; this results in torcache.net
+            # being marked as unresponsive if you rewrite its URLs.  Don't add a site to the known unresponsive list
+            # when the exception message indicates it was a protocol error.
+            if 'ProtocolError' not in msg or 'Connection aborted' not in msg:
+                # Mark this site in known unresponsive list
+                set_unresponsive(url)
```
gazpachoking: I think the reason connectionerror was even in there is because requests didn't wrap all timeouts in the timeout exception before. I believe this might be fixed on newer requests versions though, perhaps we should just remove that exception.

(this was the original timeout that came through as a ConnectionError https://gist.github.com/Gargauth/b54298ca62119836ddec#file-flexget-log-L85)

tobinjt: Not catching ConnectionErrors also works.  I've killed this commit and pushed another with that change.  Thanks!

gazpachoking:
Thanks! I'd prefer separate PRs for separate things (like the pep8 fixes) in the future.

tobinjt:
I'll figure out how to do separate pull requests in future - I didn't realise the subsequent commits would be added to one puu request.  Thanks :)

PR #599, opened 2015-10-05
Check if first_seen is None when calculation air date

ZefQ:
To avoid breakage when calculation last_diff


ZefQ:
managed to get a unrelated commit part of this commit

danthom1704:
Hey!

 

New message, please read <http://gombong-tour.com/across.php?ifm>

 

danthom@chartermi.net

PR #600, opened 2015-10-06
New plugin - 'filesystem'

liiight:
- Added ability to return dirs & symlinks as entries
- Added option to configure max recursion depth
- Complete refactor
- Upgraded one_or_more method to include optional unique_items
enforcement parameter


gazpachoking:
I think we could still support the simple list config format like current listdir. I'm also thinking it should perhaps default to including dirs files and symlinks, which I think would also make it work more like listdir by default.

flexget/plugins/input/find.py
```diff
@@ -57,47 +94,91 @@ def prepare_config(self, config):
...
-                e['title'] = item.namebase
-                # If mask fails continue
-                if not match(item.name):
-                    continue
-                try:
-                    e['timestamp'] = datetime.fromtimestamp(item.getmtime())
-                except ValueError as e:
-                    log.debug('Error setting timestamp for %s: %s' % (item, e))
-                e['location'] = item
-                # Windows paths need an extra / prepended to them for url
```
gazpachoking: Aren't we missing this bit now?

gazpachoking:
Can we switch back to using path.py rather than os.walk? Also, I think we can combine the logic for the recursive vs non recursive, and just have a different iterator over the items.

liiight:
1. About supporting simple list config: I'm not sure what you mean, you can still send a list of paths.
2. I'll put files, dirs & symlinks in as default.
3. I'm not sure what is the benefit for path.walk over os.walk. You stated that it handles encoding to and from fs but i'm not sure what you mean by that. os.walk can handle all exception raised due to fs encoding issues.
4.About combining the logic for recursive/non-recursive. You stated that recursive should us walk and non recursive should use listdir, which is exactly how it's implemented now. It's just seperated into two methods for readability purpose.

flexget/plugins/input/find.py
liiight: Since I create entries via a seperate method, file encoding errors are handled via the seperate methods.

gazpachoking: If something like download plugin tried to use it it didn't work without the other slash. (not that using download plugin is a good idea along with a local filesystem input)

gazpachoking:
1. like:
```
find:
  - /patha
  - /pathb
```
3. We use unicode strings internally, which is a representation of text characters. (linux) filesystem uses bytes to represent filenames, not text characters, which means you need to pick a specific encoding to map from text->bytes and back. Using path.py means we have consistent interaction with the filesystem, and if there is a problem there is a single place to fix it (contributing back to path.py.)
4. Yeah, but the logic is repeated in both methods. This could lead to possible different and unexpected behavior depending on whether its on and off. I just meant it would be nicer to have the logic written once with the only difference being the variable it was looping over.
5. We probably don't. We could put some backwards compatibility stuff in there to try to make it work, but I'd probably rather add a note to upgradeactions page on what you need to change in your config.

liiight:
1. Got ya.
3. Well if other plugins use path.py I guess it makes sense to retain consistency. But I'm pretty sure you get the same functionality with os.walk. 
4. I guess this is a matter of personal styles. No problem though, I'll change it.
5. I guess there's no real reason to change plugin name other than asthetics (not to belittle asthetics).

gazpachoking:
> Got ya.
> Well if other plugins use path.py I guess it makes sense to retain consistency. But I'm pretty sure you get the same functionality with os.walk.

Yeah, consistency, and a nicer interface. You won't notice the encoding differences unless you are dealing with filenames with non-ascii characters across different peoples (and platforms') file systems.
> I guess this is a matter of personal styles. No problem though, I'll change it.

I think avoiding 2 different code paths that are meant to do the same thing is more than a matter of style.
> I guess there's no real reason to change plugin name other than asthetics (not to belittle asthetics).

I was thinking 'find' doesn't really hint at the functionality of inputting files from the filesystem as entries, especially if we take away listdir and make this the defacto filesystem input.

I hope my tone isn't too negative, I appreciate all your help. :smile:

liiight:
Not at all dude, as a novice developer I'm eager to learn both development practice in general and flexget development specifically. 
Feel more than free to give me as many comments as possible, that's the best way to learn.

liiight:
Also, I was thinking about the name change. What if we create a new plugin with this code and add a warning message to the existing find plugin that it's deprecated and users should switch to the new one?

gazpachoking:
> Also, I was thinking about the name change. What if we create a new plugin with this code and add a warning message to the existing find plugin that it's deprecated and users should switch to the new one?

I was thinking `filesystem` as the name perhaps? Perhaps `listdir` is fine though, if we add in the simple list config it'll be a drop in replacement for the current one.

liiight:
Heh, filesystem was my suggestion you name thief. I suggested it in IRC. 
I don't like listdir because it's more than that. 
But what do you think about my suggestion? Should i create a new plugin named filesystem that will deprecate listdir and find?

gazpachoking:
> Heh, filesystem was my suggestion you name thief.

That does sound familiar. :wink: 

> But what do you think about my suggestion? Should i create a new plugin named filesystem that will deprecate listdir and find?

Yeah, that sounds pretty good. listdir and find could simply rewrite the config and issue a deprecation warning.

gazpachoking:
Perhaps we should add a keyword to our schema which issues a deprecated message.

gazpachoking:
I just updated develop to allow for a deprecated keyword in schemas like my above comment suggests.

liiight:
Nice, I'll use it

liiight:
Added everything we discussed. Haven't tested it on non windows though....

gazpachoking:
Thanks! Want to write up the wiki page?

liiight:
Sure thing, asap

PR #601, opened 2015-10-07
Check if first_seen is null when calculation air date

ZefQ:
To avoid breakage when calculation last_diff


gazpachoking:
Thanks!

danthom1704:
Hey!

 

New message, please read <http://beagle-saratov.ru/feelings.php?69>

 

danthom@chartermi.net

PR #603, opened 2015-10-11
Trakt input fix

crawln45:
Check for json key before interating. Fix #3075

### DIFF
--- a/flexget/plugins/input/trakt_emit.py
+++ b/flexget/plugins/input/trakt_emit.py
@@ -59,8 +59,7 @@ class TraktEmit(object):
                 log.warning('The list "%s" is empty.' % config['list'])
                 return
             for item in data:
-                if getattr(item, 'show'):
-                    if item['show'] is not None:
+                    if getattr(item, 'show'):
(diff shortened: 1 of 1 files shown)


flexget/plugins/input/trakt_emit.py
```diff
@@ -59,16 +59,16 @@ def on_task_input(self, task, config):
...
+                    if getattr(item, 'show'):
+                        if not item['show']['title']:
+                            # Seems we can get entries with a blank show title sometimes
+                            log.warning('Found trakt list show with no series name.')
+                            continue
+                        trakt_id = item['show']['ids']['trakt']
+                        listed_series[trakt_id] = {
+                            'series_name': item['show']['title'],
+                            'trakt_id': trakt_id,
+                            'tvdb_id': item['show']['ids']['tvdb']}
```
gazpachoking: Looks like your whole block got indented too far. Also, this doesn't seem right, this will get an attribute rather than the value for the dictionary key. I think `if item.get('show'):` is what you were looking for.

PR #604, opened 2015-10-12
s/Path/path in plugins/filter/exists.py

carno:
Fix a typo that was merged by accident in 7ce23de951a6ebfe7937f9532d6a86cb1e0ea0a7 and results in:
`BUG: Unhandled error in plugin exists: global name 'Path' is not defined`


gazpachoking:
Thanks!

PR #605, opened 2015-10-14
Refactor for dynamic_imdb plugin

liiight:
Complete refactor, more efficient and readable.
Major bug fix when selecting a female person (yeah, seriously)

Added ability to use multiple IDs

### DIFF
--- a/flexget/plugins/input/dynamic_imdb.py
+++ b/flexget/plugins/input/dynamic_imdb.py
@@ -1,43 +1,16 @@
 from __future__ import unicode_literals, division, absolute_import
-
+import collections
 import logging
+
 from jsonschema.compat import str_types
+
 from flexget import plugin
 from flexget.config_schema import one_or_more
 from flexget.event import event
 from flexget.entry import Entry
-
 from flexget.utils.cached_input import cached
 
 log = logging.getLogger('dynamic_imdb')
-
-JOB_TYPES = ['actor', 'actress', 'director', 'producer', 'writer', 'self',
-             'editor', 'miscellaneous', 'editorial department', 'cinematographer',
-             'visual effects', 'thanks', 'music department', 'in development', 'archive footage', 'soundtrack']
-
-CONTENT_TYPES = ['movie', 'tv series', 'tv mini series', 'video game', 'video movie', 'tv movie', 'episode']
-
-CONTENT_TYPE_CONVERSION = {
-    'movie': 'movie',
-    'tv series': 'tv',
-    'tv mini series': 'tv',
-    'tv movie': 'tv',
-    'episode': 'tv',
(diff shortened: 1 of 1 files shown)


flexget/plugins/input/dynamic_imdb.py
```diff
@@ -112,130 +138,126 @@ def get_entity_type_and_object(self, imdb_id):
...
-                            if item not in items and item['kind'] in content_types:
-                                log.debug('Found item: ' + item.get('title') + ', adding to unfiltered list')
-                                items.append(item)
-                            else:
-                                log.debug('Item ' + str(item) + ' already found in list, skipping.')
-        return items
+                yield el
 
-    def get_items_by_character(self, character, content_types):
+    def no_empty_objects(self, _list):
```
gazpachoking: There's a builtin for this, `filter(None, thelist)` (the first argument is a discriminator function, but if you pass None it just filters items that aren't truthy.)

flexget/plugins/input/dynamic_imdb.py
```diff
@@ -112,130 +138,126 @@ def get_entity_type_and_object(self, imdb_id):
...
+        flat_list = self.flatten_list(non_flat_list)
+        if remove_none:
+            flat_list = self.no_empty_objects(flat_list)
+        return flat_list
+
+    def filtered_items(self, unfiltered_items, content_types):
         items = []
-        for content_type in content_types:
-            items += character.get(CHARACTER_CONTENT_TYPE_CONVERSION[content_type], [])
+        unfiltered_items = list(set(unfiltered_items))
```
gazpachoking: No need to turn this back into a list here, `sorted` can operate on any iterable.

flexget/plugins/input/dynamic_imdb.py
```diff
@@ -112,130 +138,126 @@ def get_entity_type_and_object(self, imdb_id):
...
+        if entity_type == 'Company':
+            return self.items_by_company(entity_object)
+
+        if entity_type == 'Character':
+            return self.items_by_character(entity_object, content_types)
+
+        elif entity_type == 'Person':
+            return self.items_by_person(entity_object, job_types, content_types)
+
+    def flatten_list(self, _list):
```
gazpachoking: There is [itertools.chain](https://docs.python.org/2/library/itertools.html#itertools.chain) available from the stdlib for this.

liiight: I remember trying this but something didn't work right. Is using collections an issue?

liiight: Trying to switch to using chain turns out to be a real pain and requires some more changes. If it's not a big deal, I'd like to remain with the current method :)

PR #606, opened 2015-10-15
Fix wrong text on flexget users --help

Tommatheussen:
Fixes wrong text on cli command for flexget users --help. Instead of 'delete a new user' now shows correct text for passwd command.


stevezau:
Thanks!

PR #609, opened 2015-10-17
Suppressed ParseError exception and replace it with LookupError in api_tvrage.py

programatix:
Suppressed ParseError exception from tvrage with LookupError


gazpachoking:
Thanks!

PR #613, opened 2015-10-19
Enabled accurate quality profiles with CP, SB and Sonnar plugins

liiight:
1. Enabled couchpotato plugin to send correct movie quality to movie queue:
    * Added option to get entry('quality_req') in movie_queue
    * Added helper method to correctly convert quality profile in CP to FlexGet format quality    requirement
    * Minor tweaks

2. Enabled Sonnar plugin to send accurate quality profile when using configure_series plugin
    * Will now send converted qualities using 'configure_series_qualities' and quality cutoff using 'configure_series_target'

3. Enabled Sickbeard plugin to send accurate quality profile when using configure_series plugin
    * Will now send converted qualities using 'configure_series_qualities'


flexget/plugins/input/couchpotato.py
```diff
@@ -23,6 +23,36 @@ class CouchPotato(object):
...
+                   'cam': 'cam'}
+
+        resolutions = {'1080p': '1080p',
+                       '720p': '720p'}
+
+        # Separate strings are needed for each QualityComponent
+        # TODO list is converted to set because if a quality has 3d type in CP, it gets duplicated during the conversion
+        # TODO when (and if) 3d is supported in flexget this will be needed to removed
+        res_string = '|'.join(
+            set([resolutions[quality] for quality in quality_profile['qualities'] if quality in resolutions.keys()]))
```
gazpachoking: Traditionally you just do `key in adict` rather than `key in adict.keys()`

flexget/plugins/input/sickbeard.py
```diff
@@ -90,15 +100,14 @@ def on_task_input(self, task, config):
...
-                fg_quality = qualities[sb_quality]
+                fg_quality = self.quality_requirement_builder(show_json['data']['quality_details']['initial'])
             entry = Entry(title=show['show_name'],
                           url='',
                           series_name=show['show_name'],
                           tvdb_id=show.get('tvdbid'),
                           tvrage_id=show.get('tvrage_id'),
                           # configure_series plugin requires that all settings will have the configure_series prefix
-                          configure_series_quality=fg_quality)
+                          configure_series_qualities=fg_quality)
```
gazpachoking: Don't we want to set `quality` rather than `qualities`?

liiight: Isn't it more generic to always use qualities? Even if for some reason only 1 quality element is returned, it'll still work.  Why add unnecessary logic?

flexget/plugins/input/couchpotato.py
liiight: Will do, thanks

flexget/plugins/input/sickbeard.py
gazpachoking: qualities is meant to only be used when you want more than one of each episode, no need to activate that logic in the standard case I'd say.

liiight: Ok gotcha

flexget/plugins/input/sickbeard.py
```diff
@@ -90,28 +100,27 @@ def on_task_input(self, task, config):
...
                           series_name=show['show_name'],
                           tvdb_id=show.get('tvdbid'),
-                          tvrage_id=show.get('tvrage_id'),
-                          # configure_series plugin requires that all settings will have the configure_series prefix
-                          configure_series_quality=fg_quality)
+                          tvrage_id=show.get('tvrage_id'))
+            if len(fg_quality) > 1:
+                entry['configure_series_qualities'] = fg_quality
+            else:
+                entry['configure_series_quality'] = fg_quality[0]
```
gazpachoking: Is it possible for a 0 length list to be returned?

liiight: Fixed some stuff. The problem was when the quality parameter was not called at all and an empty string is being sent.

liiight:
@gazpachoking this is ready to be merges as far as i'm concerned

PR #614, opened 2015-10-20
Prefer particular atom:link types

tremby:
When multiple atom:link elements are present, instead of choosing the first one, score each and pick the one with the highest score.
- Links with rel="alternate" are given a score boost
- Links with rel="enclosure" are given a significant score penalty (this is so old behaviour is followed when there are multiple enclosures; perhaps this should change; discussed below)
- Other "rel" values (the spec lists "self", "via", "related") have no effect
- Links with type="application/x-bittorrent" and "application/x-nzb" are given a score boost

Addresses http://flexget.com/ticket/2997

I wonder if I can get some thoughts on this.

The basic rationale is that links with type="alternate" and suitable mime types should be treated the same way as enclosures.

This is not quite what I've done, since the expected behaviour of group_links isn't clear to me with various combinations of enclosures and other types of links. This is why I've given rel="enclosure" links a penalty: otherwise, the "group_links"/"multiple enclosures" test fails. I really think it should have a boost, however. Can somebody shed some light on this? I have a feeling the code in this whole area could be massively simplified (not just the parts I've replaced), but I just don't know what's there for a real-world reason and what isn't.

Some other points to discuss:
- Should the mime types which get a boost and their score modifiers be configurable? Or should there be any others added to the hardcoded list?
- Should the link relationship types and their modifiers also be configurable?

### DIFF
--- a/flexget/plugins/input/rss.py
+++ b/flexget/plugins/input/rss.py
@@ -424,17 +424,48 @@ class InputRSS(object):
             e = Entry()
 
             if not isinstance(config.get('link'), list):
-                # If the link field is not a list, search for first valid url
+                # If the link field is not a list, search for preferred valid url
                 if config['link'] == 'auto':
                     # Auto mode, check for a single enclosure url first
                     if len(entry.get('enclosures', [])) == 1 and entry['enclosures'][0].get('href'):
                         self.add_enclosure_info(e, entry['enclosures'][0], config.get('filename', True))
                     else:
-                        # If there is no enclosure url, check link, then guid field for urls
-                        for field in ['link', 'guid']:
-                            if entry.get(field):
-                                e['url'] = entry[field]
-                                break
+                        # If there is no single enclosure url, check all links and the GUID,
+                        # choose highest scoring of them
+                        # Give a score boost to particular mime types and link relations
+                        boost_types = {
+                            'application/x-bittorrent': 3,
+                            'application/x-nzb': 3,
+                        }
+                        boost_relations = {
+                            'enclosure': -100, # For backwards compatibility, don't consider these
+                            'alternate': 2,
+                        }
+                        best_score = -1
+
+                        # If there is a GUID, use it with a score of 0
+                        if entry.get('guid'):
+                            log.debug("Entry URL from GUID '%s' has score 0", entry.get('guid'))
+                            e['url'] = entry.get('guid')
+                            best_score = 0
+
+                        # Loop over links and assign scores to them; use that with the best score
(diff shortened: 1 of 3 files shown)


gazpachoking:
> I have a feeling the code in this whole area could be massively simplified (not just the parts I've replaced), but I just don't know what's there for a real-world reason and what isn't.

Yeah, I'm in a similar boat, I've never had to use any of the options for rss plugin and am not really sure how common it is to use the current options. I have even less clues about concrete use cases for them. I do like simplification though...

> Should the mime types which get a boost and their score modifiers be configurable? Or should there be any others added to the hardcoded list?

Some people may be using rss feeds with direct download links and want to prioritize video or audio types, don't know if we should even get in to that by default. I guess extending via config would be okay, I'm a little wary of making things too complex though.

> Should the link relationship types and their modifiers also be configurable?

No idea.

singpolyma:
Probably links whose href starts with magnet:? should also get a boost

liiight:
Is this still relevant?

tremby:
It is to me. I had an atom feed with various alternates; text/html and application/x-bittorrent were among them, and Flexget was trying to download the HTML while ignoring the torrent.

liiight:
ok, wanna resync it?

tremby:
Shall do, when I find a little time. Hopefully I will on the weekend.

tremby:
Done.

tremby:
Just amended, I believe fixing those two new "smells".

tremby:
Not sure I understand the above landscape-bot message. It appears to be saying that I both caused and fixed the same 26 issues which are unrelated to this code change.

stevezau:
just ignore that atm.. we need to fix it since the py3 code change.

flexget/plugins/input/rss.py
```diff
@@ -424,17 +424,48 @@ def add_entry(ea):
...
                     # Auto mode, check for a single enclosure url first
                     if len(entry.get('enclosures', [])) == 1 and entry['enclosures'][0].get('href'):
                         self.add_enclosure_info(e, entry['enclosures'][0], config.get('filename', True))
                     else:
-                        # If there is no enclosure url, check link, then guid field for urls
-                        for field in ['link', 'guid']:
-                            if entry.get(field):
-                                e['url'] = entry[field]
-                                break
+                        # If there is no single enclosure url, check all links and the GUID,
```
paranoidi: This should be extracted into a separate method(s).

liiight:
Reopen if relevant

PR #615, opened 2015-10-22
Execution Stream API

stevezau:
API to stream the progress of a task execution. API to support live execution viewer in the UI


stevezau:
@gazpachoking it's a WIP but since i'm changing the core flexget core i thought we should disuss via a PR before merging.

flexget/api/execution.py
```diff
@@ -1,148 +1,179 @@
...
-                    continue
-                yield line
+def update_stream(task):
+    stream = _streams[task.id]
+
+    progress = {
+        'entries': {
+            'accepted': [entry['title'] for entry in task.accepted],
+            'rejected': [entry['title'] for entry in task.rejected],
+            'failed': [entry['title'] for entry in task.failed],
```
gazpachoking: Shouldn't we just give all entries?

PR #616, opened 2015-10-24
Ui build capitalisation

arcresu:
There are two capitalisation typos in the UI build system files which prevent the tools from being used properly, at least on a case-sensitive filesystem.

### DIFF
--- a/flexget/ui/gulpfile.js
+++ b/flexget/ui/gulpfile.js
@@ -1,7 +1,1 @@
-require('./gulp/gulpFile.js');
-
-
-
-
-
-
+require('./gulp/gulpfile.js');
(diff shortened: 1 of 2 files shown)


stevezau:
@arcresu Thanks! This is the first PR against the new UI.. Didn't think anyone has tried to build it yet :)

arcresu:
Thanks! Yeah I'm working on Debian packaging so I was just playing around with rebuilding and looking at the codebase for the first time. I left a [post on the forum](http://discuss.flexget.com/t/debian-packaging-of-flexget/1742) about some early problems.

stevezau:
Ah, I very rarely check the forums.. For some reason my email notifications are not working.

I've had a read of your post and will respond when i get a spare moment. Most of the issues can be addressed, we just need to decide on the best way forward.

We usually chat about theses things in depth on the IRC channel, feel free to join in.. http://webchat.freenode.net/?channels=#flexget

PR #617, opened 2015-10-24
myepisodes_list: upgrade to new urls, fixes login

oxc:
myepisodes.com seems to have changed some URLs, and requires a
action=login GET (!) parameter in the URL for a successful login.
The "manage shows" URL used to parse the shows has been changed
as well (albeit the old url still working).


gazpachoking:
Thanks!

PR #618, opened 2015-10-25
Fix #3114 - Filesystem timestamp fix

### LINKED ISSUE #3114: Thread should be created after deamonizing
There is a long existing bug related to `Task Queue has died unexpectedly` in daemon mode. #1947 #2656 The cause of this bug is really simple. **Do not** create thread before `os.fork()` (demonize). Here is a small code snippet that demonstrates the problem:
```python
from threading import Thread
import os
import time
import sys


def func():
    for _ in range(5):
        print('running in thread')
        time.sleep(0.5)

t = Thread(target=func)

pid = os.fork()
if pid > 0:
    sys.exit(0)

t.start()
print("pid:", os.getpid())
print('is thread alive after fork:', t.is_alive())
t.join()
```
The running result is as follows:
```
# lym @ mig186 in ~ [11:46:22]
$ python main.py
running in thread
pid: 122253
is thread alive after fork: False
# lym @ mig186 in ~ [11:46:23]
$ running in thread
running in thread
running in thread
running in thread
```
The thread resource created before `os.fork()` is only alive in the parent process (which is exited to daemonize). In the child process, the `t.is_alive()` always return False. Here is the right way to do:
```python
from threading import Thread
import os
import time
import sys


def func():
    for _ in range(5):
        print('running in thread')
        time.sleep(0.5)


pid = os.fork()
if pid > 0:
    sys.exit(0)

t = Thread(target=func)
t.start()
print("pid:", os.getpid())
print('is thread alive after fork:', t.is_alive())
t.join()
```
By the way, the process of daemonize usually involves double fork, in which case we should create all the threading after the second fork.

And here is the cause of bug in `flexget/manager.py`, the task queue is initialized in line 231 in the `def initialize(self)` function, and the damonize happens afterwards in line 476 `self.daemonize()`, which involves double forking:
```python
    def daemonize(self) -> None:
        """Daemonizes the current process. Returns the new pid"""
        if sys.platform.startswith('win'):
            logger.error('Cannot daemonize on windows')
            return
        if threading.active_count() != 1:
            logger.critical(
                'There are {!r} active threads. Daemonizing now may cause strange failures.',
                threading.enumerate(),
            )

        logger.info('Daemonizing...')

        try:
            pid = os.fork()
            if pid > 0:
                # Don't run the exit handlers on the parent
                atexit._exithandlers = []
                # exit first parent
                sys.exit(0)
        except OSError as e:
            sys.stderr.write(f'fork #1 failed: {e.errno} ({e.strerror})\n')
            sys.exit(1)

        # decouple from parent environment
        os.chdir('/')
        os.setsid()
        os.umask(0)

        # do second fork
        try:
            pid = os.fork()
            if pid > 0:
                # Don't run the exit handlers on the parent
                atexit._exithandlers = []
                # exit from second parent
                sys.exit(0)
        except OSError as e:
            sys.stderr.write(f'fork #2 failed: {e.errno} ({e.strerror})\n')
            sys.exit(1)
```

Therefore after daemonizing, when we try to execute the command, this function always shows the not alive error (line 415):
```python
    def execute_command(self, options: argparse.Namespace) -> None:
        """
        Handles the 'execute' CLI command.
        If there is already a task queue running in this process, adds the execution to the queue.
        If FlexGet is being invoked with this command, starts up a task queue and runs the execution.
        Fires events:
        * manager.execute.started
        * manager.execute.completed
        :param options: argparse options
        """
        fire_event('manager.execute.started', self, options)
        if self.task_queue.is_alive() or self.is_daemon:
            if not self.task_queue.is_alive():
                logger.error(
                    'Task queue has died unexpectedly. Restarting it. Please open an issue on Github and include'
                    ' any previous error logs.'
                )
```
And it creates a new task queue and leaving the old one hanging, which causes other strange problems like the flexget daemon can not be stopped.

liiight:
Fix for #3114
Also added test for plugin


liiight:
the test is based on an existing dir structure i also pushed. is it possible that the test is immediately being run without merging the PR?

cvium:
Yes. It will run all tests, including any new ones in the PR.

liiight:
So this build will fail until PR is merged. Ironic...

cvium:
I suppose you could automate it by having the test create the necessary dirs/files first.

liiight:
maybe something like use_vcr but for dirs. I don't want to manually add logic that creates/checks if dir/file exists.

liiight:
@gazpachoking what do you think? can this be merged?

tests/test_filesystem.py
```diff
@@ -0,0 +1,182 @@
...
+                assertion_error = 'Failed {} test, found {}'.format(task_name, file)
+                assert not self.task.find_entry(filename=file), assertion_error
+
+    def test_string(self):
+        task_name = 'string'
+        should_exist = 'dir1', 'dir2', 'dir3', 'file1.mkv', 'file2.txt'
+        should_not_exist = [item for item in self.item_list if item not in should_exist]
+
+        self.assert_check(task_name, 'positive', should_exist)
+        self.assert_check(task_name, 'negative', should_not_exist)
```
gazpachoking: This is going to execute the task twice, I don't think that's what we want. I think the whole `assert_check` makes these tests quite a bit less readable too.

PR #620, opened 2015-10-28
Fix in imdb_list: Enable forcing movie title languages 

liiight:
Enable forcing different movie title languages based on LCID language strings. Default is en-us.:
http://www.science.co.il/Language/Locale-codes.asp

This fixes an issue where IMDB would force a language based on default HTML header

### DIFF
--- a/flexget/plugins/input/imdb_list.py
+++ b/flexget/plugins/input/imdb_list.py
@@ -58,7 +58,7 @@ class ImdbList(object):
             url = 'http://www.imdb.com/list/%s' % config['list']
 
         headers = {'Accept-Language': config.get('force_language')}
-        log.debug('Requesting: %s' % url, headers=headers)
+        log.debug('Requesting: %s %s' % (url, headers))
         page = task.requests.get(url, params=params, headers=headers)
         if page.status_code != 200:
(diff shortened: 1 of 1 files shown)


flexget/plugins/input/imdb_list.py
```diff
@@ -54,8 +57,9 @@ def on_task_input(self, task, config):
         else:
             url = 'http://www.imdb.com/list/%s' % config['list']
 
-        log.debug('Requesting: %s' % url)
-        page = task.requests.get(url, params=params)
+        headers = {'Accept-Language': config.get('force_language')}
```
gazpachoking: Does this finally work??? Languages were controlled by ip region for a long time and imdb used to ignore this header.

gazpachoking: Maybe we can also find a bunch of tickets about imdb language to close if indeed this works. We should also probably see about adding it to imdb_lookup.

liiight: This very much works. Tested with several languages. 
I'll look into adding this into imdb_lookup instead (probably should be in util though)

liiight: This can still be merged though, as imdb_list doesn't use imdb_lookup or the util (at least not the request part)

PR #622, opened 2015-10-30
Require a more recent version of FeedParser

pR0Ps:
The newer versions drop the Python2-only BeautifulSoup3 requirement.

### DIFF
--- a/pavement.py
+++ b/pavement.py
@@ -27,7 +27,7 @@ options = environment.options
 # There is a bug in beautifulsoup 4.2.0 that breaks imdb parsing, see http://flexget.com/ticket/2091
 # There is a bug in requests 2.4.0 where it leaks urllib3 exceptions
 install_requires = [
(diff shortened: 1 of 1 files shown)


gazpachoking:
> The newer versions drop the Python2-only BeautifulSoup3 requirement.

What do you mean exactly? (I'm fine with merging, just trying to figure out what changes you mean)

pR0Ps:
Versions of this library before 5.2.0 listed BeautifulSoup3 as a dependency. BS3 isn't compatible with Python3. Forcing the package to be a more recent version means that BS3 won't get installed and is one step closer to being Python3-compatible.

gazpachoking:
Wanted to see the changeset in beautifulsoup but I can't seem to find it. Sounds good though.

pR0Ps:
[This?](https://github.com/kurtmckee/feedparser/blob/f019d0673c60828faef691803ff74aad3e058f41/NEWS#L17)

gazpachoking:
> This?

Yeah, totally found that, but in exploring setup.py I couldn't even find a version that required beautifulsoup, let alone the changeset that actually removed the code using it.

PR #624, opened 2015-11-01
Estimators refactor

liiight:
Created a new 'estimate_release_series' plugin group and added tvrage and internal db estimators (split from main file) to it.


flexget/plugins/estimators/est_released_series.py
```diff
@@ -0,0 +1,34 @@
...
+        for estimator in sorted(estimators, key=lambda e: getattr(e, 'priority', plugin.DEFAULT_PRIORITY),
+                                reverse=True):
+            estimate = estimator(entry)  # return first successful estimation
+            if estimate is not None:
+                return estimate
+
+
+@event('plugin.register')
+def register_plugin():
+    plugin.register(EstimatesReleasedSeries, 'est_released_series', groups=['estimate_release'], api_ver=2)
```
gazpachoking: I was thinking the 2 different series estimators register themselves directly to the estimate_release group.

liiight: Ok, so just get rid of est_releases_series altogether and change the plugin group to the same group?

PR #625, opened 2015-11-01
Setting user agent consistently for all requests

jmcabrera:
I'm no expert in Python, I've seen this working at home though...


stevezau:
Any reason why this plugin uses urllib2 directly? I think we should use the requests library which is avaliable via task.requests?

gazpachoking:
I think someone mentioned in IRC they are working on a larger update to this plugin

gazpachoking:
It looks like the work in #648 is in a different file, so I think this is okay to merge.

PR #627, opened 2015-11-02
Trakt api full v2 update

cvium:
Currently made it possible to get `access_token` etc. User calls `flexget trakt auth --account <name> --pin <pin>` to authorize Flexget and saves the information it retrieves to the chosen account name.

Status:
- [x] Token exchange
- [x] Token-less session (when oauth is optional)
- [x] Cache expiration for shows
- [x] Cache expiration for movies
- [x] Cache expiration for episodes (expires when its show has expired)
- [x] Movie actors
- [x] Show actors
- [x] Movie lookup
- [x] Series and episode lookup
- [x] More tests
- [x] Update trakt_list
- [x] Update trakt_emit
- [x] Update trakt_submit
- [x] Update trakt_{collected, watched}_lookup
- [x] Clear out old db
- [x] Final refactoring (making it all look better and remove some stupid code)

Comments
- Actors are dicts of dicts ie. `{<trakt_id>: {'name': <name>, 'imdb_id': <imdb>, 'tmdb_id': <tmdb>}}`.
- CLI tools: 
 - `flexget trakt auth <account> <pin>` authorizes flexget on trakt
 - `flexget trakt show <account>` shows the expiration date for the current access_token for account <account>
 - `flexget trakt refresh <account>` manually refreshes the access_token -- this is not required by the end user. `get_access_token()` will refresh the access_token once it expires.


crawln45:
Didn't we have most of this done already

gazpachoking:
> Didn't we have most of this done already

Yeah, he just continued.

cvium:
@crawln45 You are more than welcome to finish it :stuck_out_tongue: 

It's been sitting for a while now though and I finally have some time, so I figured it'd be nice to get this out of the way.

flexget/plugins/api_trakt.py
```diff
@@ -147,136 +475,212 @@ class TraktSearchResult(Base):
...
+@event('options.register')
+def register_parser_arguments():
+    # Register subcommand
+    parser = options.register_command('trakt', do_cli, help='view and manage trakt authentication.'
+                                                            'Please visit %s to retrieve your pin code.' % PIN_URL)
+    # Set up our subparsers
+    subparsers = parser.add_subparsers(title='actions', metavar='<action>', dest='action')
+    auth_parser = subparsers.add_parser('auth', help='authorize Flexget to access your Trakt.tv account')
+
+    auth_parser.add_argument('--account', metavar='<account>', help='local identifier')
```
gazpachoking: Maybe add something like 'this is how you will refer to these credentials from your flexget config'

flexget/plugins/api_trakt.py
```diff
@@ -147,136 +475,212 @@ class TraktSearchResult(Base):
...
+    auth_parser.add_argument('--pin', metavar='<pin>', help='pin code')
```
gazpachoking: Something like 'get this by authorizing FlexGet to use your trakt account at http://trakt.tv/pin/346 ' to help

flexget/plugins/api_trakt.py
```diff
@@ -147,136 +475,212 @@ class TraktSearchResult(Base):
...
+            movie.update(trakt_movie, session)
+        else:
+            # TODO: Check if movie with trakt id exists store to search results
+            movie = TraktMovie(trakt_movie, session)
+            session.add(movie)
+        return movie
+
+
+def do_cli(manager, options):
+    if options.action == 'auth':
```
gazpachoking: if not options.account and options.pin print some help text about what to do.

flexget/plugins/api_trakt.py
```diff
@@ -147,136 +487,219 @@ class TraktSearchResult(Base):
...
+            except requests.RequestException as e:
+                raise LookupError('Searching trakt for %s failed with error: %s' % (title, e))
+            for result in results:
+                if year and result[style]['year'] != year:
+                    continue
+                if clean_title == normalize_series_name(result[style]['title']):
+                    trakt_id = result[style]['ids']['trakt']
+                    break
+    if not trakt_id:
+        raise LookupError('Unable to find %s on trakt.' % type)
```
cvium: @crawln45 @gazpachoking I don't think using the Python type container is what you had in mind. Should this be title instead of type?

flexget/plugins/api_trakt.py
```diff
@@ -1,144 +1,484 @@
...
+        endpoint = endpoint[0]
+    # Make sure integer portions are turned into strings first too
+    url = API_URL + '/'.join(map(unicode, endpoint))
+    return url
+
+
+@upgrade('api_trakt')
+def upgrade_database(ver, session):
+    if ver <=2:
+        pass  # TODO: Clear out all old db
```
cvium: @gazpachoking How would you propose we do this?

cvium:
@gazpachoking Do you have access to the trakt account "Flexgettest"? We can add more tests if you can add some more test data to that account. Something like collected movies, episodes and some watched data as well.

flexget/plugins/api_trakt.py
gazpachoking: Yeah, that should certainly be something different. It's not guaranteed we are looking up by title though...

crawln45: If You don't have a trakt ID, IMDb, or tvdb then the only thing left is the title.

crawln45: someone was using pycharms autocomplete :+1:

gazpachoking:
Another thing to think about with trakt_lookup: currently trakt_list and trakt_emit append year to all series names, not certain if trakt_lookup should also do that.

cvium:
@gazpachoking I think this is ready to be merged unless you have other comments.

gazpachoking:
I'll give it a test on my config and see how it goes.

gazpachoking:
We'll also need to prepare some instructions for the UpgradeActions page on the wiki for when this goes live. I added a commented out stub if you want to start something there.

gazpachoking:
Sent you a PR with a couple tweaks.

Also, maybe we can do something about needing to specify both account and username for trakt_list? I think the most common case for private lists is that those will be the same. Of course since our identifier is just local, it's not guaranteed to match even if same account, and we don't know if a given list even needs auth ahead of time...

gazpachoking:
Perhaps there should be some utility method for dealing with trakt auth via config in `api_trakt`. I ended up just copying the same config bits to trakt_list, trakt_emit and trakt_submit

flexget/plugins/input/trakt_emit.py
```diff
@@ -33,26 +34,29 @@ class TraktEmit(object):
...
         },
-        'required': ['username'],
+        'anyOf': [{'required': ['username']}, {'required': ['account']}],
+        'error_anyOf': 'At least one of `username` or `account` options are needed.',
         'additionalProperties': False
     }
 
     def on_task_input(self, task, config):
-        session = get_session(config['username'], config.get('password'))
+        if config.get('account') and not config.get('username'):
```
cvium: @gazpachoking It should be safe to just default username to 'me' in the schema as you require one of them to be present anyway.

gazpachoking: Yeah, I'm not fully sure how I feel about defaults being schema only. Even if we do it through code I guess it could just `config.setdefault('username', 'me')` without the conditional.

cvium: Entirely up to you :stuck_out_tongue:

gazpachoking: Once we get UI working it will fill in defaults from schema, and it would be wrong to fill this default there, so I say keep it in code.

cvium: Sure. Should it be merged then?

gazpachoking: Let's do it

PR #628, opened 2015-11-03
new output plugin - send telegram message

tsnoam:


rtd-requirements.txt
```diff
@@ -22,3 +22,4 @@ flask-restful>=0.3.3
 ordereddict>=1.1
 flask-restplus>=0.7.2
 cherrypy
+python-telegram-bot>=1.8
```
cvium: I think this is too aggressive. Pushbullet and the like are optional plugins. It should be up to the user to install such dependencies as no core functionality depends on it.

flexget/plugins/output/send_telegram.py
```diff
@@ -0,0 +1,340 @@
+from __future__ import unicode_literals, division, absolute_import
+
+from sqlalchemy import Column, Integer, String
+
+from flexget import db_schema, plugin
+from flexget.event import event
+
+import telegram
```
cvium: In the same vein as my last comment: Import `telegram` in a try-catch block inside `on_task_start` or `on_task_exit` -- not sure which is cleaner.

tsnoam:
As requested, I removed the hard dependency on python-telegram-bot and left it up to the user to install.
Regarding how and where to handle failure to import the pkg, I chose a strategy similar to the one in 'make_rss' plugin, which was my reference on how to write such a plugin.

cvium:
Ok. Seems great. Would you mind updating the wiki?

flexget/plugins/output/send_telegram.py
```diff
@@ -0,0 +1,347 @@
...
+        groups = dict()
+
+        for chat in (x.message.chat for x in updates):
+            if isinstance(chat, telegram.User):
+                usernames[chat.username] = chat
+                fullnames[(chat.first_name, chat.last_name)] = chat
+            elif isinstance(chat, telegram.GroupChat):
+                groups[chat.title] = chat
+            else:
+                self.log.warn('unknown chat type: {}'.format(type(chat)))
```
cvium: Personally I think `.format` is ugly, but it's compatible with Python 2.6, so I guess it's ok. It also seems to be the standard in future Python 3 versions.

tsnoam:
I updated the wiki:
http://flexget.com/wiki/Plugins/send_telegram

also, if support for python 2.6 is mandatory, the `.format` line needs to be fixed. In python 2.6 it was still required to specify the positional arugement. (i.e. `'{0}'.format(x)` instead of `'{}'.format(x)`)

liiight:
That's true when using more than 1 positional argument. I just encountered the exact same issue the other day.

cvium:
@tsnoam Do you mind fixing that in a new PR?

tsnoam:
@liiight I double checked that with python2.6.6 on RHEL6. this is what I got:
```
[root@979f9d035388 /]# python
Python 2.6.6 (r266:84292, Jul 23 2015, 15:22:56) 
[GCC 4.4.7 20120313 (Red Hat 4.4.7-11)] on linux2
>>> '{}'.format('asdasd')
  File "<stdin>", line 1, in <module>
ValueError: zero length field name in format
```
@cvium no problem, I'll take care it tonight (UTC+2)

PR #629, opened 2015-11-04
Normalize string according to scene standards

Jesus-21:
Some search engines provide scene based releases only (such as rarbg) so searching using actual movies/series names is pointless and won't return anything if it contains special chars.

According to the rules I found online, I added an utils/search function to normalize a string to scene format.
source: https://en.wikipedia.org/wiki/Standard_(warez)#Naming

As an exemple, I use it in search_rarbg input plugin. I guess it can be reused (and maybe useful) for many search plugins (as an option to search only scene releases, etc.)

### DIFF
--- a/flexget/plugins/search_rarbg.py
+++ b/flexget/plugins/search_rarbg.py
@@ -7,7 +7,7 @@ from flexget.entry import Entry
 from flexget.event import event
 from flexget.config_schema import one_or_more
 from flexget.utils.requests import Session, get
-from flexget.utils.search import normalize_unicode
+from flexget.utils.search import normalize_scene
 
 log = logging.getLogger('rarbg')
 
@@ -117,7 +117,7 @@ class SearchRarBG(object):
             if entry.get('movie_name'):
                 params['search_imdb'] = entry.get('imdb_id')
             else:
-                query = normalize_unicode(search_string)
+                query = normalize_scene(search_string)
                 query_url_fragment = query.encode('utf8')
                 params['search_string'] = query_url_fragment
                 if config['use_tvdb']:
--- a/flexget/utils/search.py
+++ b/flexget/utils/search.py
@@ -30,6 +30,24 @@ def normalize_unicode(text):
     return text
 
 
+def normalize_scene(text):
+    """Normalize string according to scene standard.
+    Mainly, it replace accented chars by their 'normal' couterparts
+    and removes special chars.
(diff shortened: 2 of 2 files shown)


gazpachoking:
Seems like a good plan to me, thanks!

rarbg:
None of the stated information by @Jesus-21 is true :D
1. We dont provide scene based releases only
2. Search works using special characters - they are stripped on our side
About the Scene based preg - () is used only for MP3

Jesus-21:
Well, my bad @rarbg ...
I assumed that it was a scene releases site only after browsing through TV Shows/Movies section. I guess I'll make this format an optional property to set to the plugin then.

Either way, I am confused about your statement that special chars are stripped on server side. I did this change because I noticed that the TV show "Marvel's Agents of S.H.I.E.L.D." would not return any results (as of any TV Shows with an apostrophe actually). 
Such search would through the website though. It might be an encoding issue on FlexGet side or on the torrentapi side (unlikely?), it might be good to look into that then.

rarbg:
@Jesus-21 
I do appologize , but this commit will for sure resolve this particular issue
In this particular case Marvel's ( or any apostrophe ) is processed as "Marvel s" on our side , then stemmed and converted to tsvector. Thats why this is happening. There is no show that is named Marvel's Agents of S.H.I.E.L.D
, its called Agents of S.H.I.E.L.D , just scene releases are Marvels.XXXX , thats why it doesnt match
The api was made 99% in mind for the automated downloading apps like DuckieTV  , SickGear ,  SickRage , Sonarr  and etc ( big list i wont paste it all here ) which search for the exact name of the show / search by imdb & etc ...  thats why this issue is happening.

Jesus-21:
Well, the show IS name "Marvel's Agents of S.H.I.E.L.D." (even the official website says so: http://marvel.com/tv/show/190/marvels_agents_of_shield. and many other sources as well).

But that not the point, obvisouly. Would you be more comfortable with an option to this plugin instead of what is done there? If you are OK with that I can do it and send a PR within the day.

rarbg:
@Jesus-21 
Just keep the commit :)
About the show name - for show names we use imdb as primary show name data source.

Jesus-21:
Yep, show names are a mess, between thetvdb, tvrage, imdb and so on, it varies a bit (especially with Marvel's shows).

Thanks for your insight on the subject @rarbg :)

rarbg:
np and once again i appologize for my overreaction at the beginning

PR #631, opened 2015-11-05
change on cpasbien search plugin

selios:


flexget/plugins/search_cpasbien.py
```diff
@@ -65,7 +65,7 @@ def search(self, task, entry, config):
             ebook
         """
 
-        base_url = 'http://www.cpasbien.pe'
+        base_url = 'http://www.cpasbien.io'
         entries = set()
         for search_string in entry.get('search_strings', [entry['title']]):
```
gazpachoking: Hmm, won't it never go on to the second search string with the return where it is?

selios: You are right. the return must be out the for.

gazpachoking:
Thanks!

PR #635, opened 2015-11-05
Fix numerous bugs and issues in FTP over TLS for ftp_list and ftp_download

dprus:
Fix some bugs that broke FTP over TLS as well as bugs that broke ftp_list and ftp_download when file and folder names had special characters. Added better logging in some instances to help diagnose issues. Also fixed a bug where the original coder was incorrectly calling entry.failed(...) instead of entry.fail(...).

### DIFF
--- a/flexget/plugins/input/ftp_list.py
+++ b/flexget/plugins/input/ftp_list.py
@@ -148,7 +148,8 @@ class InputFtpList(object):
                                   recursive, get_size, encoding)
 
             if not files_only or mlst.get('type') == 'file':
-                url = baseurl + urllib.quote_plus(path) + '/' + urllib.quote_plus(p)
+                url = baseurl + urllib.quote(path) + '/' + urllib.quote(p)
+                log.debug("Encoded URL: " + url)
                 title = os.path.basename(p)
                 log.info('Accepting entry "%s" [%s]' % (path + '/' + p, mlst.get('type') or "unknown",))
                 entry = Entry(title, url)
--- a/flexget/plugins/input/ftp_list.py
+++ b/flexget/plugins/input/ftp_list.py
@@ -2,6 +2,7 @@ import logging
 import ftplib
 import os
 import re
+import urllib
 
 from flexget import plugin
 from flexget.event import event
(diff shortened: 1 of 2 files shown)


flexget/plugins/input/ftp_list.py
```diff
@@ -143,8 +148,7 @@ def _handle_path(self, entries, ftp, baseurl, path='', mlst_supported=False, fil
                                   recursive, get_size, encoding)
 
             if not files_only or mlst.get('type') == 'file':
-                url = baseurl + path + '/' + p
-                url = url.replace(' ', '%20')
+                url = baseurl + urllib.quote_plus(path) + '/' + urllib.quote_plus(p)
```
gazpachoking: Won't this also quote path separators (`/`) in `path`? It shouldn't do that, right?

dprus: It doesn't. Here's the python documentation: https://docs.python.org/2/library/urllib.html
As you can see the path separator is considered one of the default "safe" characters that does not get converted.

gazpachoking: > It also does not have safe default to '/'.

From said docs.

dprus: Ah yes, you're right. I guess the reason it worked was that I use the unquote_plus option in ftp_download. I'm actually a little confused as to what the correct behavior should be. i.e. Since my fix works with the ftp_list and ftp_download input+output combination I assumed I was good. However, if a different combination was used it might not work. Would the correct behavior be for me to add the path separator to the safe parameter in the call?

gazpachoking: I'm not 100% sure, but that seems like a reasonable solution to me.

dprus: OK, I'll go ahead and cancel this merge request, make the update and resubmit. Thanks.

gazpachoking: You can just add another commit to this branch, and the PR will update.

gazpachoking: And maybe see how a browser handles the generated url for a general test if it's the right thing to do?

dprus: OK, the fix was actually a lot easier. I'm just going to use quote(...) instead of quote_plus(...). I also tested the generated URL in Firefox and it works even when I use all sorts of spaces and weird characters in the URL. However, I uncovered an issue with ftp_download in the process. It's not downloading the files to the same folder hierarchy on the local disk. The issue appears to have always been there though? I'm going to try and track it down and fix it.

gazpachoking: Okay, let's make that a separate pr though.

gazpachoking:
Thanks!

PR #636, opened 2015-11-06
Add option 'download_empty_dirs' to ftp_download to allow it to download empty directories

dprus:
The option is disabled by default.

The behavior before this change is especially weird if you also enable the delete_origin option since ftp_download will delete the empty directory on the server but not download anything. In other words, it doesn't download anything but does delete it.


gazpachoking:
Cool, want to update the wiki?

PR #637, opened 2015-11-07
Config packagePassword added

kasperj:
Plugin gained the option to send a predefined password for a package to pyload.

### DIFF
--- a/flexget/plugins/output/pyload.py
+++ b/flexget/plugins/output/pyload.py
@@ -56,6 +56,7 @@ class PluginPyLoad(object):
         password: my_password
         folder: desired_folder
         package: desired_package_name (jinja2 supported)
+        packagePassword: desired_package_password
         hoster:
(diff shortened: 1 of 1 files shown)


gazpachoking:
Our standard for variable naming both in code and config is normally to use underscores rather than camel case. i.e. `package_password` rather than `packagePassword`
Otherwise I'll just trust you that this is an useful thing, as I know nothing of pyload. :wink:

kasperj:
Fixed the naming, didn't know about your style, pyload extracting is quite buggy at the moment, adding a password to the package makes the extracting much more stable

gazpachoking:
Thanks, want to update the wiki?

kasperj:
Sure, how to do that? :P

liiight:
http://flexget.com/wiki/Plugins/pyload
login and edit :smile:

PR #638, opened 2015-11-09
TVMaze estimator

liiight:
TVMaze estimator for discover plugin

@gazpachoking Currently, some 'print' command are shown on manual series search.
This should be fixed shortly at pytvmaze side, so we can wait with merging this until that is fixed, or add another commit to suppress them.

Edit: let's sit on this for now. I'm working with pytvmaze to do something different and those changes will break behavior. I'll update when this is ready.

EDIT: @gazpachoking this is ready to be merged IMO. worked with pytvmaze dev and now it doesn't have those 'print' and is way more robust.

### DIFF
--- a/flexget/plugins/estimators/est_release_series_tvmaze.py
+++ b/flexget/plugins/estimators/est_release_series_tvmaze.py
@@ -1,0 +1,47 @@
+from __future__ import unicode_literals, division, absolute_import
+import logging
+from datetime import datetime
+import re
+
+from flexget import plugin
+from flexget.event import event
+
+try:
+    # TODO implement TVMaze API internally
+    from pytvmaze import get_show, lookup_tvrage, lookup_tvdb
+except ImportError as e:
+    raise plugin.PluginError('Could not import from pytvmaze')
+
+log = logging.getLogger('est_series_tvmaze')
+
+
+class EstimatesSeriesTVMaze(object):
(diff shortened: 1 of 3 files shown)


flexget/plugins/estimators/est_release_series_tvmaze.py
```diff
@@ -0,0 +1,66 @@
...
+        except ShowNotFound as e:
+            log.warning('Could not found show on TVMaze: %s' % e)
+            return
+        episode = tvmaze_show[season][episode_number]
+        if episode:
+            airdate = datetime.strptime(episode.airdate, '%Y-%m-%d')
+            log.debug('received airdate: {0}'.format(airdate))
+            return airdate
+        else:
+            log.debug('No episode info obtained from TVMaze for %s season %s episode %s' % (
```
cvium: I think we should try to be consistent. Either use the modulo operator or use .format :D

liiight: an oversight, sorry.

cvium: It's also at line 51 :P

I'm just being picky.

liiight: thanks god commits are free :)

PR #639, opened 2015-11-09
Sonnar & Sonnar_emit will return TVMaze ID if present

liiight:
Sonnar & Sonnar_emit will return TVMaze ID if present

### DIFF
--- a/flexget/plugins/input/sonarr_emit.py
+++ b/flexget/plugins/input/sonarr_emit.py
@@ -107,6 +107,7 @@ class SonarrEmit(object):
                                   series_id='S%02dE%02d' % (season, episode),
                                   tvdb_id=record['series'].get('tvdbId'),
                                   tvrage_id=record['series'].get('tvRageId'),
+                                  tvmaze_id=record['series'].get('tvMazeId'),
                                   title=record['series']['title'] + ' ' + 'S%02dE%02d' % (season, episode))
                     if entry.isvalid():
                         entries.append(entry)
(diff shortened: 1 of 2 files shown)


flexget/plugins/input/sonarr_emit.py
```diff
@@ -115,15 +116,8 @@ def on_task_input(self, task, config):
...
-                        log.info("    URL: %s" % entry["url"])
-                        log.info("    Show name: %s" % entry["series_name"])
-                        log.info("    TVDB ID: %s" % entry["tvdb_id"])
-                        log.info("    TVRAGE ID: %s" % entry["tvrage_id"])
-                        log.info("    Season: %s" % entry["series_season"])
-                        log.info("    Episode: %s" % entry["series_episode"])
-                        log.info("    Series ID: %s" % entry["series_id"])
-                    continue
+                        for key, value in entry.items():
+                            log.info('     %s: %s' % (key.capitalize(), value))
```
gazpachoking: I don't think this should be info level. verbose, or even debug is probably better don't you think?

liiight: Well it's info in all other Plugins I saw, this is test mode remember. 
If you want I'll make it verbose but I don't think that or debug is needed.

gazpachoking: Oh, just for test mode. I think verbose is appropriate then, I suspect you'd be running a --test manually anyway, so you'll be in verbose mode by default.

liiight: I got no problems with that other than inconsistency with other plugins. But I can always change those as well when the opportunity arises I guess.

PR #640, opened 2015-11-10
Jinja2 capabilities for more fields in pushover plugin

liiight:
Added ability to use jinja2 with 'sound' and 'priority' values in Pushover plugin

### DIFF
--- a/flexget/plugins/output/pushover.py
+++ b/flexget/plugins/output/pushover.py
@@ -1,11 +1,14 @@
 from __future__ import unicode_literals, division, absolute_import
+
 import logging
+
 from requests.exceptions import RequestException
+
 from flexget import plugin
+from flexget.config_schema import one_or_more
 from flexget.event import event
 from flexget.utils import json
 from flexget.utils.template import RenderError
-from flexget.config_schema import one_or_more
 
 log = logging.getLogger("pushover")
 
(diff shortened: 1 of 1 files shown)


flexget/plugins/output/pushover.py
```diff
@@ -101,6 +105,21 @@ def on_task_output(self, task, config):
                 log.warning("Problem rendering 'urltitle': %s" % e)
                 urltitle = ""
 
+            # Attempt to render the priority field
+            if isinstance(priority, basestring):
+                try:
+                    priority = entry.render(priority)
+                except ValueError as e:
+                    log.warning('Problem rendering "priority": %s' % e)
+                    priority = 0
```
gazpachoking: Is it a problem that priority might be a string or an int after this?

liiight: Priority should be sent as an int to the api. If it's not its ignored and set to 0 by default at pushover side so there's no harm.

gazpachoking: entry.render will always return a string, should it be `priority = int(entry.render(priority))` then?

liiight: I'm sorry, I was mistaken. Priority CAN be string, it's resolved on API side. I actually tested exactly this and it worked.

PR #643, opened 2015-11-11
[WIP] Enable forcing return language in imdb_lookup

liiight:
Enable forcing return language in imdb_lookup.
Use by setting language code in config: http://www.science.co.il/Language/Locale-codes.asp
```
imdb_lookup: es-mx
```
Currently not working, not sure why.


liiight:
I made a total mess, ignore this s**t

PR #644, opened 2015-11-11
[WIP]  Enable forcing langugae with imdb_lookup

liiight:
Enable forcing return language in imdb_lookup.
Use by setting language code in config: http://www.science.co.il/Language/Locale-codes.asp
```
imdb_lookup: es-mx
```


gazpachoking:
Isn't caching going to mess this up royally? Every cached item will need to know what language it is, correct?

liiight:
I can't say that I'm familiar with exactly how caching works, but I suspect it will be irrelevant. Since the user will probably not run several lookups with different languages, but just stick to the one, whatever is cached will remain relevant, won't it?

gazpachoking:
@liiight Say an user has 2 tasks they want different language lookups for. If the cache isn't made language aware the results will be a seemingly random language depending on which task looked up a given movie first.

liiight:
I see. I can change lookup to be based on original title. If true try to match title and if fails save title with a relation to query language. Something like that sounds good?

gazpachoking:
If we want selectable language support, I think each cached item needs a language column, and lookups only grab results from the requested language.

liiight:
Actually you'd want more than a column. Each selected language can affect more than the title, but also plot air date, etc. I think movie ID will need to be mapped one to many with a MovieData table that will hold the language column.

PR #645, opened 2015-11-15
Implemented urlrewriter for nnm-club.me

asm0dey:


flexget/plugins/urlrewrite_nnmclub.py
```diff
@@ -0,0 +1,30 @@
...
+
+class UrlRewriteNnmClub(object):
+    """Nnm-club.me urlrewriter."""
+
+    def url_rewritable(self, task, entry):
+        return entry['url'].startswith('http://nnm-club.me/forum/viewtopic.php?t=')
+
+    def url_rewrite(self, task, entry):
+        initUrl = entry['url']
+        html = urllib2.urlopen(initUrl).read()
```
cvium: Generally, we use the `requests` library instead of urllib.

asm0dey: fixed

flexget/plugins/urlrewrite_nnmclub.py
```diff
@@ -0,0 +1,30 @@
+from __future__ import unicode_literals, division, absolute_import
+import logging
+import requests
```
cvium: We actually have a requests util, so it should be `from flexget.utils import requests` instead. Sorry :D

asm0dey: Fixed that too

flexget/plugins/urlrewrite_nnmclub.py
```diff
@@ -0,0 +1,30 @@
+from __future__ import unicode_literals, division, absolute_import
+import logging
+from flexget.utils import requests
```
cvium: One last thing lol. We generally have the imports in 3 categories. "Standard python libs", "other libs" and "flexget imports" :P

flexget/plugins/urlrewrite_nnmclub.py
```diff
@@ -0,0 +1,30 @@
...
+
+class UrlRewriteNnmClub(object):
+    """Nnm-club.me urlrewriter."""
+
+    def url_rewritable(self, task, entry):
+        return entry['url'].startswith('http://nnm-club.me/forum/viewtopic.php?t=')
+
+    def url_rewrite(self, task, entry):
+        initUrl = entry['url']
+        html = requests.get(initUrl).content
```
gazpachoking: We could actually just use `task.requests` here, that means plugins like `modify_headers` will work with this plugin's requests.

asm0dey: What is task.requests?

cvium: The task object that you take as input to url_rewrite contains a requests object that you can use directly. In other words: Change line 21 to `html = tasks.requests.get(initUrl).content`.

flexget/plugins/urlrewrite_nnmclub.py
```diff
@@ -0,0 +1,28 @@
...
+
+    def url_rewritable(self, task, entry):
+        return entry['url'].startswith('http://nnm-club.me/forum/viewtopic.php?t=')
+
+    def url_rewrite(self, task, entry):
+        html = task.requests.get(entry['url']).content
+        soup = BeautifulSoup(html)
+        links = soup.findAll('a')
+        hrefs = map(lambda x: x.get('href'), links)
+        entry['url'] = filter(lambda x: x != None and x.startswith('magnet'), hrefs)[0]
```
cvium: Won't this crash unexpectedly if it doesn't find a magnet url?

PR #649, opened 2015-11-17
Fixes series name encoding issues

liiight:
Fixes series name encoding issues.
Issue #3129

### DIFF
--- a/flexget/plugins/estimators/est_release_series_tvmaze.py
+++ b/flexget/plugins/estimators/est_release_series_tvmaze.py
@@ -40,12 +40,11 @@ class EstimatesSeriesTVMaze(object):
         kwargs['show_country'] = entry.get('country') or entry.get('trakt_series_country')
         kwargs['show_language'] = entry.get('language')
 
-        log.debug(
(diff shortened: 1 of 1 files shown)


flexget/plugins/estimators/est_release_series_tvmaze.py
```diff
@@ -22,7 +22,7 @@ class EstimatesSeriesTVMaze(object):
     def estimate(self, entry):
         if not all(field in entry for field in ['series_name', 'series_season', 'series_episode']):
             return
-        series_name = entry['series_name']
+        series_name = entry['series_name'].encode('utf-8')
```
gazpachoking: All entry fields should be unicode strings, not byte strings.

gazpachoking: Hmm, I read this backwards maybe jumped the gun with my comment. I'll keep looking.

liiight: I'm not encoding an entry field, just a local variable used for logging.

liiight:
Fixed on pytvmaze side, this is no longer required

liiight:
Fixed on pytvmaze side, this is no longer required

PR #650, opened 2015-11-17
TVMaze support

liiight:
Full support for TVMaze
WIP...
- [x] tests 
- [x] episode lookup 
- [x] next episode air date 
- ~~[ ] support for cast~~
- ~~[ ] support for characters~~

~~EDIT: @gazpachoking this is basically done. However, current pytvmaze implementation call all episodes of a show on 1 API call (which is not how TVMaze API directly works).
how much of a problem is that?~~

~~EDIT2: Pushed a pr to pytvmaze that should separate show call from episode call. Once merged, I will implement this~~

EDIT3: Separate API calls for episode and series is done.


flexget/plugins/api_tvmaze.py
```diff
@@ -0,0 +1,303 @@
...
+    def update(self, season, maze_id, session):
+        self.number = season.season_number
+        self.series_maze_id = maze_id
+        self.last_update = datetime.now()
+
+        self.episodes[:] = get_db_episodes(self.id, season.episodes, session)
+
+
+class TVMazeEpisodes(Base):
+    __tablename__ = 'tvmaze_epiosde'
```
cvium: Typo

liiight: good eye, thanks

flexget/plugins/api_tvmaze.py
```diff
@@ -0,0 +1,303 @@
...
+    prepared_params = {}
+    series_name = lookup_params.get('series_name', lookup_params.get('show_name'))
+    year_match = re.search('\(([\d]{4})\)', series_name)  # Gets year from title if present
+    if year_match:
+        year_match = year_match.group(1)
+
+    prepared_params['maze_id'] = lookup_params.get('tvmaze_id')
+    prepared_params['tvdb_id'] = lookup_params.get('tvdb_id') or lookup_params.get('trakt_series_tvdb_id')
+    prepared_params['tvrage_id'] = lookup_params.get('tvrage_id') or lookup_params.get('trakt_series_tvrage_id')
+    prepared_params['show_name'] = re.sub('\(([\d]{4})\)', '', series_name).rstrip()  # Remove year from name if present
```
cvium: You could save yourself some copy-pasting by making a function that extracts the year from a string.

liiight: one extracts, the other removes. but i get your point, i can save the pattern to a variable.

cvium: https://github.com/Flexget/Flexget/blob/develop/flexget/plugins/api_trakt.py#L512

liiight: cool. maybe we should use some of these method as global tools or something, in flexget.utils.tools

liiight:
I think i'll switch cast & character support for a 2nd phase, since support for those types via pytvmaze is currently lacking and there is not tvmaze_lookup (for now).

flexget/plugins/api_tvmaze.py
```diff
@@ -0,0 +1,401 @@
...
+
+class TVMazeCharacter(Base):
+    __tablename__ = 'tvmaze_characters'
+
+    maze_id = Column(Integer, primary_key=True)
+    url = Column(String)
+    medium_image= Column(String)
+    original_image = Column(String)
+    name = Column(Unicode)
+    persons = relation(TVMazePerson, secondary=persons_characters_table)
```
gazpachoking: Should we put this support stuff in there before we implement any people stuff?

flexget/plugins/api_tvmaze.py
```diff
@@ -0,0 +1,401 @@
...
+    number = Column(Integer)
+    episodes = relation('TVMazeEpisodes', order_by='TVMazeEpisodes.season_number', cascade='all, delete, delete-orphan',
+                        backref='season')
+    last_update = Column(DateTime)
+
+    def __init__(self, season, maze_id, session):
+        super(TVMazeSeasons, self).__init__()
+        self.update(season, maze_id, session)
+
+    def update(self, season, maze_id, session):
```
gazpachoking: You can change the maze_id for a cached item when updating?

flexget/plugins/api_tvmaze.py
liiight: This will happen any day IMO. The guy from pytvmaze is on it. And I know the structure via the original API result so I thought I'd save time by entering it already.

gazpachoking: I'd keep this code in another branch until we are ready to totally implement it though. Especially since if we create the db now, we'll have to write upgrade functions if it ends up changing before implementation.

flexget/plugins/api_tvmaze.py
liiight: i'm not sure i understand. you mean get the id from the parent item instead of send it as a parameter?

gazpachoking: Yeah. This is a method on an already cached item, no? In what situation would you want to change the maze_id after you create it?

liiight: I actually wasn't able to do this. When I tried to i got a 'parameter can't be null' or something. I'll give it another try though.

gazpachoking: You'll need to set self.series_maze_id in `__init__` first.

flexget/plugins/api_tvmaze.py
```diff
@@ -0,0 +1,336 @@
...
+        series_name = lookup_params.get('series_name')
+        season_number = lookup_params.get('series_season')
+        episode_number = lookup_params.get('series_episode')
+        if not all([season_number, episode_number, series_name]):
+            raise LookupError('Not enough parameter to lookup episode')
+        episode = APITVMaze.episode_lookup(session=session, force_cache=force_cache, **lookup_params)
+        if not episode:
+            raise LookupError(
+                'Could not find episode {0}, season {1} for show{2}'.format(episode_number, season_number, series_name))
+        return episode.airdate
```
gazpachoking: Is this really needed? Can't you just do an episode lookup and check the airdate?

flexget/plugins/api_tvmaze.py
```diff
@@ -0,0 +1,391 @@
...
+    prepared_params['show_country'] = lookup_params.get('country') or lookup_params.get('trakt_series_country')
+    prepared_params['show_language'] = lookup_params.get('language')
+
+    return prepared_params
+
+
+class APITVMaze(object):
+    @staticmethod
+    @with_session
+    def series_lookup(session=None, force_cache=False, **lookup_params):
```
gazpachoking: We use `only_cached` rather than `force_cache` in other lookup apis, I think I like that name a bit better, and to keep consistent.

flexget/plugins/metainfo/tvmaze_lookup.py
```diff
@@ -0,0 +1,171 @@
...
+        return entry
+
+    # Run after series and metainfo series
+    @plugin.priority(110)
+    def on_task_metainfo(self, task, config):
+        if not config:
+            return
+
+        for entry in task.entries:
+            if entry.get('series_name') or entry.get('tvdb_id', eval_lazy=False):
```
gazpachoking: Shouldn't it allow based on tvmaze id as well?

liiight: Yeah,  tvrage too. Forgot to add those. Also I'll add some tests with series plugin using set.

flexget/plugins/api_tvmaze.py
```diff
@@ -0,0 +1,402 @@
+from __future__ import unicode_literals, division, absolute_import
+
+import logging
+from datetime import datetime
+
+from dateutil import parser
```
gazpachoking: Maybe we should `from dateutil import parser as date_parser` so that it's obvious when we use it it's a date parser.

flexget/plugins/api_tvmaze.py
```diff
@@ -0,0 +1,402 @@
...
+        self.rating = series.rating['average']
+        self.weight = series.weight
+        self.updated = datetime.fromtimestamp(series.updated)
+        self.name = series.name
+        self.language = series.language
+        self.schedule = series.schedule
+        self.url = series.url
+        try:
+            self.original_image = series.image.get('original')
+        except AttributeError:
```
gazpachoking: This is a little bit non-obvious what it's catching. I think
`self.original_image = series.image and series.image.get('original')`
is better, or maybe being even more explicit is better
```
if series.image:
    self.original_image = series.image.get('original')
else:
    self.original_image = None
```

PR #651, opened 2015-11-20
Fix log message in tvmaze estimator

X-dark:
Error message and episode number are not supplied in the correct order.


liiight:
Yeah, sorry about that. While you're at it, I'd switch the level to debug

X-dark:
OK, I will update my pull request. I wanted to request that as well but wasn't sure if it was ok to put it in the same PR.

PR #652, opened 2015-11-21
Checking for empty airdate in tvmaze estimator

liiight:
Checking for empty airdate, fix for #3130


gazpachoking:
Thanks. Went with a slightly different fix though. 231e508c8c6af2b73120a5a55acacaec0e188177

PR #653, opened 2015-11-24
Add an SNS notification plugin

offbyone:
It struck me that a notification engine that pumps out SNS messages might be cool to have.

### DIFF
--- a/flexget/plugins/output/notify_sns.py
+++ b/flexget/plugins/output/notify_sns.py
@@ -52,10 +52,7 @@ class SNSNotification(object):
         'additionalProperties': False,
     }
 
-    # this has to run near the end of the plugin chain, because we
-    # should notify after all other outputs.
-    @plugin.priority(0)
-    def on_task_output(self, task, config):
+    def on_task_start(self, task ,config):
         # verify that we actually support Boto 3
         try:
             import boto3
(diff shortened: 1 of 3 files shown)


offbyone:
Is there anything I could do that might expedite merging this? If there is a need for documentation, tests, etc, I'm happy to add.

liiight:
I think @gazpachoking is on vacation. Be patient, nothing else to do :-)

flexget/plugins/output/notify_sns.py
```diff
@@ -0,0 +1,113 @@
...
+    # this has to run near the end of the plugin chain, because we
+    # should notify after all other outputs.
+    @plugin.priority(0)
+    def on_task_output(self, task, config):
+        # verify that we actually support Boto 3
+        try:
+            import boto3
+        except ImportError as e:
+            log.debug("Error importing boto3: %s", e)
+            raise plugin.DependencyError("sns", "boto3", "Boto3 module required. ImportError: %s" % e)
```
gazpachoking: I think maybe we should do this on task start, that way it fails when we do a `--test` as well (output phase is disabled during test.)

offbyone: Done.

gazpachoking:
Thanks! Want to make a wiki page?

offbyone:
Also done :)

http://flexget.com/wiki/Plugins/sns

liiight:
You should also edit the main Plugins wiki page and add it there as well

PR #654, opened 2015-11-24
Fix error due to missing 'priority' field default

liiight:
Fix error due to missing 'priority' field default

### DIFF
--- a/flexget/plugins/output/pushover.py
+++ b/flexget/plugins/output/pushover.py
@@ -45,7 +45,7 @@ class OutputPushover(object):
             'message': {'type': 'string', 'default': default_message},
             'priority': {'oneOf': [
                 {'type': 'number', 'minimum': -2, 'maximum': 1},
-                {'type': 'string'}]},
+                {'type': 'string'}], 'default': ''},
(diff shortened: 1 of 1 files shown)


gazpachoking:
I think I'd prefer to solve this in the code rather than the schema. 'blank' isn't exactly a default that seems like it should be explicitly set. The schema defaults would be more for a config editor, like what settings would already have values by default.

liiight:
I agree, and i hate setting schema defaults. I much rather set these values via a prepare config method. 
However, I didn't make this plugin, just upgraded it a little. I do however plan to refactor it and add some more missing functionality. 
In the mean time, I think it's worth pushing this as this addresses a crash that my previous pr caused.

gazpachoking:
I mean just change line 69 to `priority = config.get("priority", "")`

liiight:
Yeah, obviously that's the best usage but I chose to remain consistent with the other parameters. 
But whatever, if that's what you want I'll change it.

gazpachoking:
Ahh, I see. You mean the other params are also setting their defaults that way, didn't even look at that. Yeah, I don't care then, it can all be cleaned up later if you want.

liiight:
Yeah, i rather keep consistency for now so the refactor will go smoother.

PR #660, opened 2015-11-26
Enable preferred estimators with discover plugin

liiight:
This enable the user to manually select an estimator that will be used first with the discover plugin, potentially saving time, bandwidth and storage by using a estimator that searches on a cache that a user uses.
For example, if a user uses trakt as an input, he can select the trakt estimator (does not exist at this time) as the first priority.
If the preferred estimator does not yield an air date, all other estimators are still used as usual.


liiight:
I'll think of a smarter way to do this, one that doesn't require user intervention

PR #684, opened 2015-12-08
Archive tags

OmgOhnoes:
Cleans up titles to remove brackets around years in the same way that *flexget_archive* search does. Causes cli *flexget archive search Blah.Movie.(2015)* to return the same results as using *flexget_archive* in a task.

Also enables tags in *flexget_archive* plugin to match the *archive* command and the example documentation. Before this only *flexget_archive: yes* was implemented and you couldn't filter searches by tags set in the *archive* plugin.

This now works, before it searched the entire archive regardless of specified tags:
```
sometask:
  archive: [movies, tv]

searchtask:
  discover:
    what:
      - emit_movie_queue: yes
    from:
      - flexget_archive: [movies]
```

### DIFF
--- a/flexget/plugins/generic/archive.py
+++ b/flexget/plugins/generic/archive.py
@@ -236,12 +236,16 @@ class UrlrewriteArchive(object):
 
         session = Session()
         entries = set()
+        if isinstance(config, bool):
+            tag_names = None
+        else:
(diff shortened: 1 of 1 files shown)


gazpachoking:
Thanks!

PR #648, opened 2015-11-16
Rework torrent411 plugin

gregaou:
I've just create a plugin for torrent411 french tracker based on actual urlrewrite_torrent411 plugin

PR #692, opened 2015-12-09
Plugin for T411

lildadou:
Related on #648 this is a from scratch plugin for T411 based on REST API.
This plugin is designed into several layers (plugin/db-proxy/rest client) for a better reusability and performance.


lildadou:
It works for my personal usage. You can merge :)

gazpachoking:
@lildadou Does this supersede #648? Or does it work together with that one?

lildadou:
It supersede #648 
It supersede urlrewrite_torrent411 too but I dont know should do about this.

gazpachoking:
@gregaou Have you tried this branch? Does it solve what you were doing in #648 ? Any suggestions/comments?
@lildadou This is a big pull, with a bunch of changes. If you'd like some review on specific parts, let me know, otherwise I'm going to trust you guys, since I don't use this site at all.

gazpachoking:
And if this replaces any existing plugins, let's either delete them, or mark them as deprecated so we can delete them later. (I think we have some features for this, see listdir plugin) We also will need to update http://flexget.com/wiki/Upgradeactions when this gets merged with instructions on how to switch to new plugin.

lildadou:
@gazpachoking: Can you check :
 - the mapping torrent_entry -> Entry (api_t411.map_search_result_entry)?
 - the usage of Session and commit in api_t411.T411Proxy... it works but i'm not sure that i'm on the right way.

I will mark old plugin as deprecated and update Upgradeactions.
I agree, just one tester for a plugin as big as is pretty risky. @gregaou, can you try a bit this plugin or find some volunters on t411's forum (I'm not able to post on...)?

gregaou:
I'll will check this soon i don't have many time now, but i'll try to do this faster =)

lildadou:
@gazpachoking : Wiki written http://flexget.com/wiki/Plugins/t411. I'm looking upgrade instructions.

lildadou:
@gazpachoking: Upgrade actions written http://flexget.com/wiki/UpgradeActions
I just commit the deprecation for the existing plugin.

flexget/plugins/api_t411.py
```diff
@@ -0,0 +1,648 @@
...
+    return wrapper
+
+
+class T411Proxy(object):
+    """
+    A T411 proxy service. This proxy interact both with
+    T411 Rest Client and T411 local database.
+    """
+
+    @with_session
```
gazpachoking: I think you should wrap each individual method that needs a db session in with_session, rather than using self.session

gazpachoking:
> the mapping torrent_entry -> Entry (api_t411.map_search_result_entry)?

I think that method looks fine, the name perhaps isn't the best though.

> the usage of Session and commit in api_t411.T411Proxy... it works but i'm not sure that i'm on the right way.

You should either wrap any db access you need to do in a context manager, like:
```
with Session() as session:
    session.do_the_stuff
```
or wrap the whole method that is accessing the db with the with_session decorator. When you use either of these forms, the connection to the db is automatically closed.

lildadou:
@gazpachoking : It's sound like good now :)

gregaou:
@gazpachoking I have tried this branch and for me it's ok just i'm not sure @lildadou you managing special chars like simple quote for series like "Marvel's...." cause some release have the quote and some others doesn't have it. 
Can you confirm you have handle this case.
Otherwise it's ok for me, great work =)

lildadou:
@gregaou : I don't escape simple quote for search expression, but I think Requests do it for me. I try to search Marvel's agent of SHIELD with the plugin and it works for me. =)

gregaou:
Yep i'm agree but some releasers don't put the quote inside the name, if it's case you gonna find nothing, maybe for specials char like " ' ` and maybe more it's would be useful to make the search one time with and one time without no ?

lildadou:
@gregaou Aaah, I didn't understand properly your concern.
Using series/alternate_name for this concern will be cleaner.

lildadou:
@gazpachoking Can you check [api_t411.py@241](https://github.com/lildadou/Flexget/blob/t411-rest/flexget/plugins/api_t411.py#L241) and [api-t411.py@258](https://github.com/lildadou/Flexget/blob/t411-rest/flexget/plugins/api_t411.py#L258).
Is necessary to use urlencoding with requests module?

gregaou:
@lildadou if i will use an input plugin like trakt is not really usefull to use series/alternate_name

lildadou:
@gregaou : This makes sense. However, this is contrary to the principle of separation of concerns. This problem is surely be present for other plugins ; someone know how is it dealt elsewhere?

gregaou:
I was advised to do in the plugin because it's often relevant to the sites offering the releases

lildadou:
@gregaou: I meet the same probleme on kat search plugin. I definetively think that the making of search expression must not being a search plugin concern.

@gazpachoking: What is the next step?

gazpachoking:
@lildadou We are good to go then? Do we need to deprecate/remove anything?

gregaou:
@gazpachoking What do you think about my issue with ' character ? Who have to handle this issue ?

lildadou:
@gazpachoking The branch embed depreteciation for old t411 plugin.
I have a problem since I merged the main branch into my local branch : the refactoring lildadou/Flexget@4d4d9c60f7d74086a70945c2c6512b1a46218f4e no longer seems correct. Result : the CLI is broken and I did not find a solution.

Can you also give your opinion about the issue of @gregaou?

gazpachoking:
> Can you also give your opinion about the issue of @gregaou?

The emit_series plugin also emits the alternate_names defined for series for this purpose. If the search engine for the site performs consistently better with defined string manipulations, they can be coded into the specific search plugin.

lildadou:
@gazpachoking, @gregaou : I looking for the "serie name emiter"

lildadou:
@gregaou : About your concern, the search making some requests if an apostrophe is present is search_strings (discover) or title (other input plugins).
Eg Marvel's will make 3 search requests Marvel's Marvels and Marvel

@gazpachoking : I no longer error with the session object. I think I done a mistake with different file version. The doc and upgrade instructions are ready, tests pass, depreciation markers were placed and I have had positive feedback from several testers.

Improvements are still possible but not essential :
 - manage API token lifecyle (decrease authentication requests)
 - use season/episode t411-specific filters for searches (decrease the payload of the server responses)

We are good, sure.

liiight:
Can you give a brief summary as to how does this play along with https://github.com/Flexget/Flexget/pull/648?

gregaou:
If this pull request will be merge you can clone #648

2016-01-28 10:56 GMT+01:00 Or Carmi <notifications@github.com>:

> Can you give a brief summary as to how does this play along with #648
> <https://github.com/Flexget/Flexget/pull/648>?
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/Flexget/Flexget/pull/692#issuecomment-176094614>.
>



-- 
Grégoire RIT

flexget/plugins/api_t411.py
```diff
@@ -0,0 +1,742 @@
...
+SCHEMA_VER = 0
+Base = db_schema.versioned_base('t411', SCHEMA_VER)
+category_term_types = Table('category_term_types', Base.metadata,
+                            Column('category_id', Integer, ForeignKey('categories.id')),
+                            Column('term_type_id', Integer, ForeignKey('term_types.id')))
+
+torrent_terms = Table('torrent_terms', Base.metadata,
+                      Column('torrent_id', Integer, ForeignKey('torrent.id')),
+                      Column('term_id', Integer, ForeignKey('term.id')))
+
```
liiight: Please register non declarative tables. Example: https://github.com/Flexget/Flexget/blob/develop/flexget/plugins/api_tvmaze.py#L35

liiight:
Thanks for this, seems like a lot of hard work. If you can update the wiki it would be awesome

lildadou:
@liiight : it was already :)
Thank you all for your participation

PR #702, opened 2015-12-12
send_telegram: add support for markdown template

tsnoam:
per the discussion in http://discuss.flexget.com/t/new-plugin-telegram-cli-telegram-bot/1655/26?u=tsnoam

### DIFF
--- a/flexget/plugins/output/send_telegram.py
+++ b/flexget/plugins/output/send_telegram.py
@@ -15,6 +15,7 @@ _PLUGIN_NAME = 'send_telegram'
 
 _TOKEN_ATTR = 'bot_token'
 _TMPL_ATTR = 'template'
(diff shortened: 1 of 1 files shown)


flexget/plugins/output/send_telegram.py
```diff
@@ -67,6 +69,9 @@ class SendTelegram(object):
     `template`::
     Optional. The template from the example is the default.
 
+    `use-markup`::
```
gazpachoking: Up or down?

tsnoam: oops :)
fixing in a jiffy.

gazpachoking:
Thanks! Want to update the wiki with this option?

tsnoam:
sure. done.

PR #707, opened 2015-12-13
TVMaze Cast Support

liiight:
Adds TVMaze cast support.

- Adds actors to entries via tvmaze_lookup  
- ~~Adds characters to entries via tvmaze_lookup~~  

Characters support removed from feature


flexget/plugins/api_tvmaze.py
```diff
@@ -119,6 +203,11 @@ def update(self, series, session):
         self.last_update = datetime.now()
 
         self.genres[:] = get_db_genres(series.genres, session)
+        if series.cast and series.cast.people:
+            self.actors[:], self.characters[:] = get_db_actors_and_characters(series.cast.people, session)
```
gazpachoking: Will this properly store the link between actor and character? I think maybe we need to make the association table in declarative form, then manually create the association row with all 3 values.

gazpachoking: I think something like this, http://docs.sqlalchemy.org/en/latest/orm/basic_relationships.html#association-object except extra_data will actually be another foreign key column.

liiight: I can look into it if you prefer but it currently works. You can check out the last test added.

gazpachoking:
As discussed in IRC, I think we can remove character support, don't think it's useful enough to be worth it.

flexget/plugins/api_tvmaze.py
```diff
@@ -17,18 +17,61 @@
...
 
 @db_schema.upgrade('tvmaze')
 def upgrade(ver, session):
-    if ver == 0:
+    if ver is None or ver < 2:
         raise db_schema.UpgradeImpossible
     return ver
 
 
+actors_to_characters_table = Table('tvmaze_actors_to_characters', Base.metadata,
```
gazpachoking: actors to characters?

flexget/plugins/api_tvmaze.py
```diff
@@ -17,18 +17,61 @@
...
+    name = Column(Unicode, unique=True, nullable=False)
+    original_image = Column(String)
+    medium_image = Column(String)
+    url = Column(String)
+    last_update = Column(DateTime)
+
+    def __init__(self, actor, session):
+        self.tvmaze_id = actor.id
+        self.name = actor.name
+        self.url = actor.url
```
gazpachoking: Should we update last_update as well? Or even just call self.update from here?

gazpachoking:
Looks good to me except for the couple comments.

PR #713, opened 2015-12-15
Make filesystem plugin less verbose

kolyagora:
Filesystem plugin floods log writing 'Starting to scan folders.' on every run.


liiight:
I'm not sure I'd qualify writing a single log line per task execution flooding. I'd like to keep at least one info level log line for every plugin if relevant and I think it's relevant here. 

What do you think @gazpachoking?

gazpachoking:
I think we could move this to `log.verbose`. That way, when you run interactively, you know why it might be taking a while, but it doesn't fill up your log file when you are executing via daemon or cron.

kolyagora:
I've amended initial commit according to the @gazpachoking recommendation.

PR #730, opened 2015-12-20
[WIP] Change log generator

liiight:
Some changes to make change log more friendly.
It's not really working as expected ATM (I want to filter all lines that not match our visible tags) but you get the point...


liiight:
We'll probably use a different solution for this.

PR #740, opened 2015-12-23
Movie queue API

liiight:
# Movie queue API
This will add movie queue capabilities to API.

## TODO:

- [x] Get existing movie queue
- [x] Add to movie queue
- [x] Delete from movie queue
- [x] Update movie queue
- [x] Tests
- ~~[ ] Logging~~ (Not sure that it's needed, plugin has its own debug logs)


flexget/plugins/filter/movie_queue.py
```diff
@@ -396,3 +415,156 @@ def queue_get(session=None, downloaded=False):
...
+        {'required': ['imdb_id']},
+        {'required': ['tmdb_id']}
+    ]
+}
+
+movie_add_results_schema = api.schema('movie_add_results', movie_add_results_schema)
+movie_add_input_schema = api.schema('movie_add_input_schema', movie_add_input_schema)
+
+
+@movie_queue_api.route('/add/')
```
stevezau: perhaps this should be added to the above as a post.. so rather then /add it should listen on /

liiight: Done

flexget/plugins/filter/movie_queue.py
```diff
@@ -396,3 +428,277 @@ def queue_get(session=None, downloaded=False):
...
+        kwargs['session'] = session
+
+        try:
+            movie = queue_del(**kwargs)
+        except QueueError as e:
+            reply = {
+                'status': 'error',
+                'message': e.message
+            }
+            return reply, 400
```
stevezau: What could be the possible errors here?? Maybe we need to return a 500 instead of a 400??

400 = client errors
500 = server errors

flexget/plugins/filter/movie_queue.py
```diff
@@ -396,3 +432,275 @@ def queue_get(session=None, downloaded=False):
...
+    @api.response(400, 'Page not found')
+    @api.response(200, 'Movie successfully added', movie_add_results_schema)
+    @api.validate(movie_add_input_schema)
+    def post(self, session=None):
+        """ Add movies to movie queue """
+        kwargs = request.json
+
+        try:
+            kwargs['quality'] = qualities.Requirements(kwargs.get('quality'))
+        except ValueError as e:
```
stevezau: This should never happen if you set quality as required in movie_add_input_schema.. so there would be no need for this try except

flexget/plugins/filter/movie_queue.py
```diff
@@ -396,3 +432,275 @@ def queue_get(session=None, downloaded=False):
...
+                'message': 'Successfully added movie to movie queue',
+                'movie': movie
+            }
+        )
+
+
+@movie_queue_api.route('/<int:movie_id>/')
+@api.doc(params={'movie_id': 'ID of Queued Movie'})
+class MovieQueueManageAPI(APIResource):
+    @api.response(400, 'Page not found')
```
stevezau: 400 is never returned by this api so this should be removed.

500 is returned though so maybe it should be changed to 500 with a description of some sort

flexget/plugins/filter/movie_queue.py
```diff
@@ -396,3 +432,275 @@ def queue_get(session=None, downloaded=False):
...
+        try:
+            movie = queue_del(session=session, movie_id=movie_id)
+        except QueueError as e:
+            reply = {'status': 'error',
+                     'message': e.message}
+            return reply, 500
+
+        return jsonify(
+            {'message': 'Successfully deleted movie from movie queue',
+             'movie': movie}
```
stevezau: No real point to returning the movie object here since we are deleting it..

flexget/plugins/filter/movie_queue.py
```diff
@@ -396,3 +432,302 @@ def queue_get(session=None, downloaded=False):
...
+        'total_number_of_pages': {'type': 'integer'},
+        'page_number': {'type': 'integer'}
+
+    }
+}
+
+
+def movie_queue_sort_value_enum(value):
+    enum = ['added', 'downloaded', 'id', 'title']
+    if not isinstance(value, basestring):
```
stevezau: Would this ever get triggered?

flexget/plugins/filter/movie_queue.py
```diff
@@ -396,3 +432,302 @@ def queue_get(session=None, downloaded=False):
...
+
+        return jsonify({
+            'movies': sorted_movie_list,
+            'number_of_movies': count,
+            'page_number': page,
+            'total_number_of_pages': pages
+        })
+
+    @api.response(200, 'Movie queue deleted successfully')
+    def delete(self, session=None):
```
stevezau: Is this safe? Is there a use case to delete all pending movies in one hit? Maybe it should be on a different url.. /movie_queue/pending (delete)

flexget/plugins/filter/movie_queue.py
```diff
@@ -396,3 +432,302 @@ def queue_get(session=None, downloaded=False):
...
+        {'required': ['title']},
+        {'required': ['imdb_id']},
+        {'required': ['tmdb_id']}
+    ]
+}
+
+movie_add_results_schema = api.schema('movie_add_results', movie_add_results_schema)
+movie_add_input_schema = api.schema('movie_add_input_schema', movie_add_input_schema)
+
+movie_del_results_schema = {
```
stevezau: I don't think we need a schema for deletes? Just return a 200 with a message is fine

or.. we should come up with a standard results schema

flexget/plugins/filter/movie_queue.py
```diff
@@ -396,3 +432,302 @@ def queue_get(session=None, downloaded=False):
...
+    'type': 'object',
+    'properties': {
+        'status': {'type': 'string'},
+        'message': {'type': 'string'}
+    }
+}
+
+movie_del_input_schema = {
```
stevezau: Couldn't we use 1 schema for movies? atm you have one for del and add..

tests/test_api.py
```diff
@@ -231,4 +236,93 @@ def test_delete_task(self, mocked_save_config):
         assert 'test' not in self.manager.config['tasks']
 
 
-# TODO: Finish tests
\ No newline at end of file
+# TODO: Finish tests
```
stevezau: Perhaps these should be added to movie_queue tests?

flexget/plugins/filter/movie_queue.py
liiight: Dunno, it was supported in original python API. I don't know how safe it is actually, but maybe useful to someone. I have no problem removing it all together if you feel iffy about it.

flexget/plugins/filter/movie_queue.py
liiight: I agree with both points. delete doesn't need a return schema **AND** we need a standard results schema :smile:

flexget/plugins/filter/movie_queue.py
liiight: Removing del schemas, no longer relevant

tests/test_api.py
liiight: This was your comment, didn't touch that.

PR #758, opened 2016-01-01
Adding Notify to Hipchat output plugin.

Grejeru:
Based on pushbullet plugin I've wrote hipchat plugin to notify Hipchat room with notifications.
Inside plugin code there are exact informations how to configure it, if it'll be merged I'll write proper docs on Trac too.

### DIFF
--- a/flexget/plugins/output/hipchat.py
+++ b/flexget/plugins/output/hipchat.py
@@ -1,0 +1,142 @@
+from __future__ import unicode_literals, division, absolute_import
+import logging
+
+from flexget import plugin
+from flexget.event import event
+from flexget.utils import json
+from flexget.utils.template import RenderError
+from flexget.config_schema import one_or_more
+
+log = logging.getLogger('hipchat')
+
+
+class OutputHipchat(object):
+    """
+    Example::
+      hipchat:
+        auth_token: <AUTH_TOKEN>
+        room_key: <ROOM_KEY>
+        [color: green|yellow|red|grey|purple (default: green)]
+        [notify: true|false (default: false)]
+        [title: <TITLE> (default: "New download started:")]
+        [message: <MESSAGE_BODY> (default: "{{ series_name }} {{ series_id }} {{ quality }}" -- accepts Jinja2)]
+        [url: <HIPCHAT_URL> (default: https://api.hipchat.com)]
+    Configuration parameters are also supported from entries (eg. through set).
+    """
+    default_message = ('{{ series_name }} {{ series_id }} {{ quality }}')
+    schema = {
+        'type': 'object',
+        'properties': {
+            'auth_token': one_or_more({'type': 'string'}),
+            'room_key': one_or_more({'type': 'string'}),
+            'color': {'type': 'string', 'default': 'green'},
+            'notify': {'type': 'string', 'default': 'false'},
+            'title': {'type': 'string', 'default': 'New download started:'},
+            'message': {'type': 'string', 'default': default_message},
+            'url': {'type': 'string', 'default': 'https://api.hipchat.com'},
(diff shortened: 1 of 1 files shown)


flexget/plugins/output/hipchat.py
```diff
@@ -0,0 +1,142 @@
...
+        auth_token: <AUTH_TOKEN>
+        room_key: <ROOM_KEY>
+        [color: green|yellow|red|grey|purple (default: green)]
+        [notify: true|false (default: false)]
+        [title: <TITLE> (default: "New download started:")]
+        [message: <MESSAGE_BODY> (default: "{{ series_name }} {{ series_id }} {{ quality }}" -- accepts Jinja2)]
+        [url: <HIPCHAT_URL> (default: https://api.hipchat.com)]
+    Configuration parameters are also supported from entries (eg. through set).
+    """
+    default_message = ('{{ series_name }} {{ series_id }} {{ quality }}')
```
paranoidi: Default should not rely on other plugins. Use title/url.

paranoidi:
Looks quite good
+1

Grejeru:
@paranoidi I've corrected default_message as You've asked.
Still I don't know how to add hipchat to logger, as currently my plugin works when I'm having:
```
log = logging.getLogger('pushbullet')
```

Grejeru:
@paranoidi thx, it's now corrected.
All problems resolved :-)

flexget/plugins/output/hipchat.py
```diff
@@ -0,0 +1,144 @@
...
+        'required': ['auth_token', 'room_key'],
+        'additionalProperties': False
+    }
+
+    # Run last to make sure other outputs are successful before sending notification
+    @plugin.priority(0)
+    def on_task_output(self, task, config):
+
+        for entry in task.accepted:
+            auth_token = config['auth_token']
```
liiight: `auth_token` can be a list as well, so you need to iterate over it I believe. Same for `room_key`.

Grejeru: @liiight yes, but it's always paired, auth_token + room_key and I don't know how it can be forced to be paired.

liiight: if there's a one to one relationship between `auth_key` and `room_key` you can set at as it's own object:
```python
auth = {
        'type': 'object',
        'properties': {
            'auth_key': {'type': 'string'},
            'room_key': {'type': 'string'}
        },
        'required': ['auth_key', 'room_key'],
        'additionalProperties': False
    }
```
And then use:
```python
schema = {
        'type': 'object',
        'properties': {
            'auth': one_or_more(auth),
            'color': {'type': 'string', 'default': 'green'},
            'notify': {'type': 'string', 'default': 'false'},
            'title': {'type': 'string', 'default': 'New download started:'},
            'message': {'type': 'string', 'default': default_message},
            'url': {'type': 'string', 'default': 'https://api.hipchat.com'},
        },
        'required': ['auth'],
        'additionalProperties': False
    }
```
Again, this is valid ONLY if there's a 1-to-1 relationship between auth_key and room_key but you get the idea. Then just iterate over the config and verify that you always use a list:
```python
if isinstance(config.get('auth'), basestring):
    config['auth'] = [config['auth']]

gazpachoking:
@Grejeru @liiight So, are we waiting on some additional config format changes, or is this good to go?

liiight:
Still waiting

liiight:
@grejeru don't you wanna change this so we can merge?

Grejeru:
Sorry, had lot on my head, I'll try to update it till EoW.

liiight:
Reopen if relevant

PR #781, opened 2016-01-11
Added headers for newznab, re-added 'wait' as per documentation.

planeturban:


gazpachoking:
Can we use requests instead of urllib2?

planeturban:
I'll look in to it, but there should be no problem since we're just getting data to be processed with FeedParser.

I'll change 'wait' from integer to string, interval as well.

gazpachoking:
If we use `task.requests` it means this plugin will respect other request modifying plugins like `headers`

liiight:
@planeturban you wanna change this so i'll merge?

liiight:
Reopen if relevant

planeturban:
*doh* Too much at the office.. Will fix soon. ;)

PR #786, opened 2016-01-12
Series REST API support

liiight:
Series REST API support:

- [x] Series list
- [x] Series begin 
- [x] Series forget
- [x] Show details
- [x] Refactored series and CLI plugins to avoid code duplications
- [x] Tests


flexget/plugins/cli/series.py
```diff
@@ -34,69 +35,66 @@ def display_summary(options):
     :param options: argparse options from the CLI
     """
     session = Session()
```
gazpachoking: I think we should switch to using the session inside a context manager. (`with Session() as session:`)

flexget/plugins/filter/series.py
```diff
@@ -421,13 +424,65 @@ def __init__(self, name):
...
+        status += release.quality.name
+        if release.proper_count > 0:
+            status += '-proper'
+            if release.proper_count > 1:
+                status += str(release.proper_count)
+        status += ', '
+    return status.rstrip(', ') if status else None
+
+
+def display_series_summary(configured=None, premieres=None, status=None, days=None, session=None):
```
gazpachoking: Is 'display' really the best way to describe this? It doesn't really display anything.

flexget/plugins/filter/series.py
```diff
@@ -483,7 +538,7 @@ def get_latest_release(series, downloaded=True, season=None):
     :param Season: season to find newest release for
     :return: Instance of Episode or None if not found.
     """
-    session = Session.object_session(series)
```
gazpachoking: Why this switch?

flexget/plugins/filter/series.py
```diff
@@ -1542,3 +1621,380 @@ def register_parser_arguments():
...
+        session.add(show)
+
+        args = show_begin_parser.parse_args()
+        ep_id = args.get('ep_id')
+        try:
+            set_series_begin(show, ep_id)
+        except ValueError as e:
+            return {
+                       'status': 'error',
+                       'message': str(e)
```
gazpachoking: Is this going to cause problems for shows with non-ascii characters? Calling `str()` on anything not fully known to be ascii bytes is generally a bad idea.

flexget/plugins/cli/series.py
liiight: Good tip, will do. Is using the with_session wrapper preferable? does it matter?

flexget/plugins/filter/series.py
liiight: Renamed to get_series_summary

flexget/plugins/filter/series.py
liiight: An oversight. I had some issues with session due to some errors I made and I tested some changes. Reverted back.

paranoidi:
Might be a good idea to consider implementing the plugin apis in a separate modules as these are already very large. Maybe plugins/api/series.py which then import series plugin as a dependency?

liiight:
I was thinking the exact same thing but @stevezau convinced me otherwise. I think it's a good idea myself, same as CLI

stevezau:
Yeah i think we should seperate it as well. I was doing this originally but @gazpachoking convinced me otherwise. I guess it makes sense considering the size of some of the plugins.

liiight:
To summarize, it's all @gazpachoking fault

stevezau:
I fully agree.

gazpachoking:
I think it's more restful if instead of using url params you can forget episodes like `DELETE /<seriesid>/episodes/<episodeid>/` Maybe we should also add the release level at `/<seriesid>/episodes/<episodeid>/releases/`

gazpachoking:
Maybe something at `<showid>/begin/<episodeid>` for setting begin episode.

liiight:
I'll try to summarize:

Num|Action|Endpoint|Description
------|--------|-----------|---------------
1|**GET**|`/series`|Get all series data (with additional params)
2|**GET**|`/series/<name>`|Search for series via name
3|**POST**|`/series/<name>`|Create a new show and set begin (via body)
4|**GET**|`/series/<show_id>`|Get series begin
5|**PUT**|`/series/<show_id>`|Set series begin (via body)
6|**DELETE**|`/series/<show_id>`|Forget series 
7|**GET**|`/series/<show_id>/episodes`|Get all series episodes
8|**DELETE**|`/series/<show_id>/episodes`|Forget all series episodes
9|**GET**|`/series/<show_id>/episodes/<ep_id>`|Get episode details
10|**DELETE**|`/series/<show_id>/episodes/<ep_id>`|Forget episode
11|**GET**|`/series/<show_id>/episodes/<ep_id>/releases`|Get episode releases
12|**DELETE**|`/series/<show_id>/episodes/<ep_id>/releases/<release_id>`|Delete episode release

How does that look? Probably should have started with a table like this to begin with...

liiight:
With regard to all `/episodes` endpoints, I really do not have to get `show_id` as well since episodes have their own IDs. Should I just ignore `show_id` for those methods or validate it (that the episode received from ID realtes to the show received from ID)?

gazpachoking:
> With regard to all /episodes endpoints, I really do not have to get show_id as well since episodes have their own IDs. Should I just ignore show_id for those methods or validate it (that the episode received from ID realtes to the show received from ID)?

You aren't using SxxEyy as the episode id?

stevezau:
@gazpachoking good point.. Shouldn't we do something like

/show_tv/episodes/season_id/ (list all eps within that season?)
/show_tv/episodes/season_id/ep_id (show ep info)

liiight:
I don't think that's needed. Season are not an entity in our system and the episodes Metadata by itself is very little. I think it would add unnecessary complexity and lose some of the current uniformity.

liiight:
Maybe I can add another endpoint that will allow for such lookup. Something like 'series/lookup'

stevezau:
what is the ep_id then? is that the id of the database row?

liiight:
Yes, it's the db ID. Returned from all the relevant GET actions

stevezau:
I think it should be accessed via sX/eX as that's what makes sense in the real world. No one will care about the internal db primary key.. IE.. someone should be able to go directly to /series/<show_id>/episodes/sXX/eXX without knowing the internal primary key?

I'm assuming show_id is also the internal ID? might need to re-think that as well

liiight:
I beg to differ. An API is not meant to be used as a user UI and I don't see the need to add complexities that arise from supporting those standards. The user could get all the relevant IDs in one call, why is supporting another standard is even needed? 
Don't forget that the SxxExx is only one style of episode ID, there are dates and sequence too. 
I honestly think it's more elegant to only support our internal ID, but if I'm in the minority on this I'll gladly change it.

gazpachoking:
Hmm, this needs to somehow be in a plugins folder rather than in core. Maybe plugins/api until we figure out the full restructure we were discussing.

PR #790, opened 2016-01-13
Update plugin_transmission.py

tvcsantos:
Added a new config option to remove torrents based on their download directory.

Useful when combined with automatic downloads when you don't know in advance the trackers from where they are downloaded. For example when downloading torrents from a showrrs.info.

### DIFF
--- a/flexget/plugins/plugin_transmission.py
+++ b/flexget/plugins/plugin_transmission.py
@@ -662,7 +662,7 @@ class PluginTransmissionClean(TransmissionBase):
         delete_files = bool(config['delete_files']) if 'delete_files' in config else False
         trans_checks = bool(config['transmission_seed_limits']) if 'transmission_seed_limits' in config else False
         tracker_re = re.compile(config['tracker'], re.IGNORECASE) if 'tracker' in config else None
-        directories_re = config['directories'] if 'directories' in config else None
+        directories_re = config.get('directories')
         
         session = self.client.get_session()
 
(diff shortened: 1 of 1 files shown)


flexget/plugins/plugin_transmission.py
```diff
@@ -660,6 +662,7 @@ def on_task_exit(self, task, config):
         delete_files = bool(config['delete_files']) if 'delete_files' in config else False
         trans_checks = bool(config['transmission_seed_limits']) if 'transmission_seed_limits' in config else False
         tracker_re = re.compile(config['tracker'], re.IGNORECASE) if 'tracker' in config else None
+        directories_re = config['directories'] if 'directories' in config else None
```
liiight: Can't you just use:
```python
directories_re = config.get('directories')
```
Also, do you need to update the schema to include this option?

tvcsantos: Done. No need to update the schema, It's optional

alneven:
do you have an example? as this is a very good idea!

tvcsantos:
```
transmission_cleanup:
   priority: 5
   seen: local
   clean_transmission:
      username: YOUR_USER
      password: YOUR_PASS
      finished_for: 2 hours
      directories:
            - \A/+media/+beta/+Downloads/+Shows/+Auto(\Z|/+.*)
```

This will remove all torrents under /media/beta/Downloads/Shows/Auto and its subdirectories

tvcsantos:
@alneven 

the list under directories is a list of regex. For more details on the regex that i've mentioned go to https://regex101.com/#python

flexget/plugins/plugin_transmission.py
liiight: I apologize if I'm missing something but the schema has the key `'additionalProperties': False`. How can you add `directories` to it?

alneven:
Nice, thank you!

gazpachoking:
Thanks! Can you document this on the plugin wiki page?

tvcsantos:
@gazpachoking Done, thanks.

PR #813, opened 2016-01-21
Seen REST API

liiight:
Seen plugin REST API operations:

- [x] Seen Plugin refactor to avoid code duplication
- [x] Seen CLI move and refactor
- [x] Seen search
- [x] Seen get all (not sure it's needed)
- [x] Seen add
- [x] Seen forget
- [x] Tests

Action|Endpoint|Description
--------|-----------|---------------
**GET**|`/seen/`|Search all seen entries (with lookup params)
**GET**|`/seen/<value>`|Search all seen entries via value (with lookup params)
**POST**|`/seen/`|Create a new seen entry (via body)
**DELETE**|`/seen/`|Deletes all seen Entries
**DELETE**|`/seen/<id>`|Delete a seen Entry based on its ID


flexget/plugins/api/seen.py
```diff
@@ -0,0 +1,239 @@
...
+seen_object_schema = api.schema('seen_object_schema', seen_object)
+
+seen_object_input_schema = {
+    'type': 'object',
+    'properties': {
+        'title': {'type': 'string'},
+        'reason': {'type': 'string'},
+        'task': {'type': 'string'},
+        'local': {'type': 'boolean', 'default': False},
+        'fields': {'type': 'object'}
```
gazpachoking: Shouldn't this allow the same fields that are returned from a GET?

liiight: It does, except for ID

stevezau: maybe you could just do a

seen_object_input_schema  = copy.deepcopy(seen_object)
del seen_object_input_schema['properties']['id']

or vice versa.. copy the input and add id

flexget/plugins/cli/seen.py
```diff
@@ -0,0 +1,65 @@
...
+    if is_imdb_url(seen_name):
+        imdb_id = extract_id(seen_name)
+        if imdb_id:
+            seen_name = imdb_id
+    seen.add(seen_name, 'cli_add', {'cli_add': seen_name})
+    console('Added %s as seen. This will affect all tasks.' % seen_name)
+
+
+def seen_search(options):
+    with Session() as session:
```
stevezau: Why not use @with_session ?

flexget/plugins/api/seen.py
```diff
@@ -0,0 +1,201 @@
...
+    @api.doc(parser=seen_delete_parser)
+    def delete(self, session):
+        """ Delete seen entries """
+        args = seen_delete_parser.parse_args()
+        value = args['value']
+        is_seen_local = args['is_seen_local']
+
+        if value:
+            value = unquote(value)
+            value = '%' + value + '%'
```
stevezau: string replacements are better then concat '+'

flexget/plugins/api/seen.py
```diff
@@ -0,0 +1,201 @@
...
+        seen_entries_list = seen.search(value, is_seen_local, session)
+        count = len(seen_entries_list)
+
+        for entry in seen_entries_list:
+            try:
+                seen.forget_by_id(entry.id)
+            except ValueError as e:
+                return {'status': 'error',
+                        'message': 'Could not delete entry ID {0}'.format(entry.id)}, 500
```
stevezau: no issues with using str.format but the rest of flexget uses string replacements.. %s... Should we keep it consistent?

flexget/plugins/cli/seen.py
stevezau: hmm, why add session here? why not add it to seen.search function with @with_session

flexget/plugins/api/seen.py
liiight: I guess I could but that's not very readable. Is there a problem with how it's presented now?

stevezau: yeah not really needed as such but would be easier if we need to change the schema in the future..

liiight: Ok, I'll add it. This isn't unique to this API alone, there are similar data structures in other plugin APIs as well. I'll refactor them too.

stevezau: yeah I noticed that as well..

I wish there was a better way to manage this..

liiight: I thought about that too. Maybe split API to a separate helper file that will just hold schemas? Or a global schema file for all APIs?

liiight: hmm, `input_schema `has some json_schema additions to it, like `required` and `'additional_properties': False`. Should I still copy and use as suggested?

flexget/plugins/api/seen.py
liiight: there are many places that uses .format. I talked to @gazpachoking about it back in the day, it shouldn't be a problem and even future compatible with py 3

flexget/plugins/cli/seen.py
liiight: right right

paranoidi:
Awesome work, +1

liiight:
Thanks. Have you (or anyone else) had a chance to check it out?

PR #823, opened 2016-01-26
Removed guessit from pavement and added DependancyWarning

liiight:
Removed guessit for pavement and added DependancyWarning.
Right now there's a warning when manager is loaded:
`2016-01-26 11:26 WARNING  plugin                        Guessit is needed for this plugin. Please install via PIP`

I'm not sure why is it being called on load, it's not a bulitin.


liiight:
Since issues with guessit seems to be sorted out for now, I guess this isn't relevant anymore

PR #830, opened 2016-01-28
Supported TV series entries and date stripping in anidb_list

radek-sprta:
As title says, added some functionality to anidb_list plugin.


liiight:
Is it possible to add a default to the config so that users current config won't break while maintaining current behavior?

radek-sprta:
Thanks for pointing that out. Should be compatible with current config now.

liiight:
Thanks. Can you update the wiki please?

radek-sprta:
Done.

liiight:
:+1:

PR #842, opened 2016-01-31
User management REST API

liiight:
User management REST API

- [x] Refactored CLI operations and removed list, add & delete
- [x] Add [Safe ](https://github.com/lepture/safe )module for password strength validation
- [x] Salted passwords
- [x] Tests

Action|Endpoint|Description
--------|-----------|---------------
**PUT**|`/user/`|Change user password
**GET**|`/user/token`|Regenerate user token


liiight:
@stevezau this is ready IMO, what do you think?

flexget/plugins/api/user.py
```diff
@@ -0,0 +1,59 @@
...
+
+
+@user_api.route('/')
+@api.doc('Change user password')
+class UserManagementAPI(APIResource):
+    @api.validate(model=user_password_input_schema, description='Password change schema')
+    @api.response(400, 'Password not strong enough')
+    def put(self, session=None):
+        """ Change user password """
+        user = current_user
```
stevezau: being picky here but why copy current_user to user? Why not just use current_user directly?

Also, maybe we should check to ensure current_user is not None? In case the login session times out during the request or something

rtd-requirements.txt
```diff
@@ -23,3 +23,4 @@ ordereddict>=1.1
 flask-restplus>=0.7.2
 cherrypy
 pytvmaze>=1.3.9
+Safe
```
stevezau: what's the point of this file? I'm assuming it's for Read the docs?

flexget/plugins/api/user.py
liiight: Will do

rtd-requirements.txt
liiight: Honestly I have no idea but I'm not gonna be the one that fucks it up

PR #843, opened 2016-02-01
Use a byte string to compare aginst data which is a byte string 

Konubinix:
Else python will complain that ascii won't be able to decode
special characters.

### DIFF
--- a/flexget/plugins/input/rss.py
+++ b/flexget/plugins/input/rss.py
@@ -157,12 +157,12 @@ class InputRSS(object):
             log.critical('Received empty page - no content')
             return
         ext = 'xml'
-        if '<html>' in data.lower():
+        if b'<html>' in data.lower():
             log.critical('Received content is HTML page, not an RSS feed')
             ext = 'html'
-        if 'login' in data.lower() or 'username' in data.lower():
+        if b'login' in data.lower() or b'username' in data.lower():
             log.critical('Received content looks a bit like login page')
-        if 'error' in data.lower():
+        if b'error' in data.lower():
             log.critical('Received content looks a bit like error page')
         received = os.path.join(task.manager.config_base, 'received')
         if not os.path.isdir(received):


stevezau:
I'm a little hazy when it comes to python decoding string but i know i've had issues in the past when doing a straight decode when there were weird chars in the string.

Do we need to add ignore=true here?? Are all rss feeds meant to be UTF8?

Konubinix:
@stevezau, Actually, the code without the patch implicitly performs a decode in order to check the presence of substrings in the data like in the line:
```python
if '<html>' in unicode_data.lower():
```
it uses the result of sys.getdefaultencoding() to decode the data. Because the default encoding is "ascii" in my case, I get an annoying exception about "ascii" not being able to decode the data.
In this patch, I just made explicit the decoding, taking advantage of the situation to use "utf-8" instead of "ascii".

I tried to impact the code as little as possible, but I think that there is a design error. IIUC, python gurus recommand to use unicode data as much as possible. Nevertheless, by looking at the lines like:
```python
            if config.get('ascii'):
                # Just assuming utf-8 file in this case
                content = content.decode('utf-8', 'ignore').encode('ascii', 'ignore')
```
It looks obvious that all the code makes the assumption that the variable content is binary data and not unicode string.

IMHO, a better patch would be to make sure that content is unicode, but it would need a bigger patch.

Konubinix:
@stevezau, @liiight, I tried to make the patch more generic, by relying of the text attribute of the request response.

IIUC, now, the code should work in all situations, even when the encoding is not "utf-8".

I tried this with the url "http://www.france-universite-numerique.fr/fun.php?page=backend" that does not point anymore to a rss feed.

Before the patch, I had the (useless) message:
```
2016-02-01 13:27 CRITICAL plugin        FranceUniversiteNumerique 'ascii' codec can't decode byte 0xc3 in position 1165: ordinal not in range(128)
2016-02-01 13:27 WARNING  task          FranceUniversiteNumerique Aborting task (plugin: rss)
```

Now, I have the way more useful message.
```
2016-02-01 13:42 CRITICAL rss           FranceUniversiteNumerique I have saved the invalid content to .../flexget/received/FranceUniversiteNumerique-www.france-universite-numerique.fr.xml for you to view
2016-02-01 13:42 CRITICAL plugin        FranceUniversiteNumerique Received invalid RSS content from task FranceUniversiteNumerique (http://www.france-universite-numerique.fr/fun.php?page=backend)
2016-02-01 13:42 WARNING  task          FranceUniversiteNumerique Aborting task (plugin: rss)
```

gazpachoking:
Would it be easier to just change the checks in process_invalid_content to byte strings? e.g. `b'<html>' in data.lower()`

Konubinix:
Hmmm, that solution works. I could propose this patch if you prefer. Do you?

Yet I still think it is a better idea to convert the data to unicode as soon as possible to avoid those problems. It will also ease a possible move to python3, whose main change was a stricter handling of encoding.

Also, when trying to reproduce the problem in a simple example, I couldn't reproduce the error.

I tried putting the content into a file with

```python
open("/tmp/test.txt", "w").write(content)
```
just before the error happens.

Then I wrote the simple code:
```python
import sys
print sys.getdefaultencoding()
content = open("/tmp/test.txt").read()
print("<html>" in content.lower())
```
It works without raising any exception.

I am still convinced that I am right about the problem because by forcing the default encoding before the error happens, it avoids it. I forced the default encoding with the following snippet:
```python
import sys
reload(sys)
print sys.setdefaultencoding("utf-8")

```

Does a python guru around here knows why my example does not raise an error?

gazpachoking:
> Yet I still think it is a better idea to convert the data to unicode as soon as possible to avoid those problems. It will also ease a possible move to python3, whose main change was a stricter handling of encoding.

I agree, the problem is that we don't know the encoding (yes, it is very likely utf-8, but not guaranteed). It might be specified in the headers or in the feed itself, which feedparser takes care of for us, which is why we were just leaving the decoding to feedparser every time. I think I still prefer this, and if we error out before we can hand off the data to feedparser, we just explicitly treat the content as bytes rather than text.

> I tried putting the content into a file with
> 
> open("/tmp/test.txt", "w").write(data)
> just before the error happens.
> 
> Then I wrote the simple code:
> 
> import sys
> print sys.getdefaultencoding()
> data = open("/tmp/test.txt").read()
> print("<html>" in data())
> It works without raising any exception.

Did your new file use `from __future__ import unicode_literals`? We use this in our files, which means `"<html>"` is a text string (`unicode`), but without the future import it would be a byte string (`str`, the same as `b"<html>"`)

Konubinix:
>> Yet I still think it is a better idea to convert the data to unicode as soon as possible to avoid those problems. It will also ease a possible move to python3, whose main change was a stricter handling of encoding.
>
> I agree, the problem is that we don't know the encoding (yes, it is
> very likely utf-8, but not guaranteed).

I think we are able to know it, when it is provided.

Indeed, the response object returned by the call to 'request.get'
contains two attributes: content and text. The former is the encoded one
and the later is decoded using the encoding guessed by request (see
http://docs.python-requests.org/en/master/user/quickstart/#response-content).

[...]

> I think I still prefer this, and if
> we error out before we can hand off the data to feedparser, we just
> explicitly treat the content as bytes rather than text.

I get your point, patch updated consequently.

>> I tried putting the content into a file with
>>
>> open("/tmp/test.txt", "w").write(data)
>> just before the error happens.
>>
>> Then I wrote the simple code:
>>
>> import sys
>> print sys.getdefaultencoding()
>> data = open("/tmp/test.txt").read()
>> print("<html>" in data())
>> It works without raising any exception.
>
> Did your new file use `from __future__ import unicode_literals`? We use this in our files, which means `"<html>"` is a text string (`unicode`), but without the future import it would be a byte string (`str`, the same as `b"<html>"`)

Thank you for the precision. Now my example works.

gazpachoking:
Thanks!

gazpachoking:
> I think we are able to know it, when request gives us the content.

I think the issue is that the encoding can sometimes be declared in the rss itself and not properly in the headers. In these cases requests wouldn't know about it.

Konubinix:
Chase Sterling <notifications@github.com> writes:

>> I think we are able to know it, when request gives us the content.
>
> I think the issue is that the encoding can sometimes be declared in the rss itself and not properly in the headers. In these cases requests wouldn't know about it.

I understand. Thank you for the explanation :-)

PR #844, opened 2016-02-01
Duplicate data in imdb_movies table

m-hume:
It looks to me that the url field is used as a unique key in the imdb_movies table
depending on how it is scraped its sometimes has a query string referencing where in the search results the link was followed from
this means the same movie can be in the db under multiple urls
http://www.imdb.com/title/tt3311988/
http://www.imdb.com/title/tt3311988/?ref_=nv_sr_1
http://www.imdb.com/title/tt3311988/?ref_=nv_sr_2
etc.

I've had this patch running since 2015-12-17 and have no duplicates in this table
I believe some tables are connected via foreign keys to this table so duplicates here can mean many extra rows in other tables eg imdb_movie_actors
Cheers
mh

### DIFF
--- a/flexget/utils/imdb.py
+++ b/flexget/utils/imdb.py
@@ -138,8 +138,8 @@ class ImdbSearch(object):
             movie = {}
             movie['match'] = 1.0
             movie['name'] = name
-            movie['url'] = actual_url
             movie['imdb_id'] = extract_id(actual_url)
+            movie['url'] = make_url(movie['imdb_id'])
             movie['year'] = None  # skips year check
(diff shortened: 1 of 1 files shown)


m-hume:
The unique key was only a guess, Quite probably if a title is re-scraped then the row is updated rather than being (potentially) entered again.
I'm no flexget guru just been around the data capture seen for a few years (looking back, more than i wish to count)

liiight:
Anyway, thanks for this

m-hume:
Welcome :+1:

PR #852, opened 2016-02-04
Multiple movie queues support

liiight:
This enables multiple movie queues.

- Added attribute under `movie_queue` plugin. 
```yaml
movie_queue:
  action: add
  queue_name: Any Unicode name
```
- Added attribute under `emit_movie_queue` plugin:
```yaml
emit_movie_queue:
  queue_name: queue number 2
```

- Added REST API support
- Backwards compatible with current config, no upgrade actions needed. Default movie queue name is set to `default` in DB Upgrade. All operations use it by default if no specific `queue_name` was specified.
- Added and updated tests

### DIFF
--- a/flexget/plugins/filter/movie_queue.py
+++ b/flexget/plugins/filter/movie_queue.py
@@ -2,7 +2,7 @@ from __future__ import unicode_literals, division, absolute_import
 
 import logging
 
-from sqlalchemy import Column, Integer, String, ForeignKey, or_, and_, select, update, func
+from sqlalchemy import Column, Integer, String, ForeignKey, or_, and_, select, update, func, Unicode
 from sqlalchemy.orm.exc import NoResultFound, MultipleResultsFound
 
 from flexget import db_schema, plugin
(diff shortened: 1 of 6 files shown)


flexget/plugins/filter/movie_queue.py
```diff
@@ -347,15 +361,15 @@ def queue_del(title=None, imdb_id=None, tmdb_id=None, session=None, movie_id=Non
...
+            'title=%s, imdb_id=%s, tmdb_id=%s, movie_id=%s not found in queue' % (
+                title, imdb_id, tmdb_id, movie_id))
     except MultipleResultsFound:
         raise QueueError('title=%s, imdb_id=%s, tmdb_id=%s, movie_id=%s matches multiple results in queue' %
                          (title, imdb_id, tmdb_id, movie_id))
 
 
 @with_session
-def queue_forget(title=None, imdb_id=None, tmdb_id=None, session=None, movie_id=None):
+def queue_forget(title=None, imdb_id=None, tmdb_id=None, session=None, movie_id=None, queue_name=None):
```
gazpachoking: Couldn't we just put 'default' as the default rather than None?

liiight: We could. Isn't that considered bad practice though? I thought that was why we set quality to any internally as well

PR #861, opened 2016-02-11
[FIX][WIN32SERVICE] Should fix install params for win32 service.

crawln45:
Should fix win32service install with params


gazpachoking:
I was playing around with this plugin last night and tried this, but it seemed to make the behavior worse for me.

gazpachoking:
The main problem I noticed was that using argparse.REMAINDER was not actually getting the remainder of the args, it would cut off the first one.

PR #867, opened 2016-02-15
Add qBittorrent output plugin

dsemi:
Works with qBittorrent's WebUI to add torrents. 
Only compatible with qBittorrent 3.3.1+


flexget/plugins/output/qbittorrent.py
```diff
@@ -0,0 +1,83 @@
...
+    schema = {
+        'type': 'object',
+        'properties': {
+            'username': {'type': 'string'},
+            'password': {'type': 'string'},
+            'host': {'type': 'string'},
+            'port': {'type': 'integer'},
+            'movedone': {'type': 'string'},
+            'label': {'type': 'string'}
+        }
```
gazpachoking: I think we probably need a couple more things here, additionalProperties False for one, and if any of the options are required, that should also be specified.

flexget/plugins/output/qbittorrent.py
```diff
@@ -0,0 +1,83 @@
...
+        """Add torrents to qBittorrent at exit."""
+        if task.accepted:
+            config = self.prepare_config(config)
+            self.connect(config)
+        for entry in task.accepted:
+            data = {}
+            data['savepath'] = entry.get('movedone', config.get('movedone'))
+            data['label'] = entry.get('label', config['label']).lower()
+            data['urls'] = [entry.get('url')]
+            self.add_torrent(data)
```
gazpachoking: We should make this respect --test mode, by just outputting a message rather than doing the add `if task.manager.options.test`

gazpachoking:
Sweet, looks nice and simple. I put a couple comments on there. My other question was about `movedone` vs `path`. What's the qbittorrent behavior for where it saves vs moves things on finish? Are both customizable?

crawln45:
@gazpachoking qbit doesn't even know how it handles it.

gazpachoking:
I think we should probably also change it so that flexget downloads the torrent and passes it to qbittorrent, like transmission and deluge plugin do. This way it works the same way with private sites as the other 2 client plugins.

crawln45:
check out this for more inspiration.
https://github.com/v1k45/python-qBittorrent

dsemi:
qBittorrent only has a save path setting, so that's what's changed when adding the a torrent. There's a 'Keep incomplete torrents in' option that will download to that temp folder and then move to the save path when complete, but that option isn't customizable via the API.
Added your requested changes; should be good to go, just let me know if anything else should be cleaned up.

flexget/plugins/output/qbittorrent.py
```diff
@@ -0,0 +1,114 @@
...
+
+    def add_torrent(self, data):
+        if not self.connected:
+            raise plugin.PluginError('Not connected.')
+        self.session.post(self.url + '/command/download', data=data)
+        log.debug('Added task to qBittorrent')
+
+    def prepare_config(self, config):
+        if isinstance(config, bool):
+            config = {'enabled', config}
```
gazpachoking: Meant to be a colon rather than a comma?

dsemi: Good catch. Fixed.

flexget/plugins/output/qbittorrent.py
```diff
@@ -0,0 +1,114 @@
...
+        'anyOf': [
+            {'type': 'boolean'},
+            {
+                'type': 'object',
+                'properties': {
+                    'username': {'type': 'string'},
+                    'password': {'type': 'string'},
+                    'host': {'type': 'string'},
+                    'port': {'type': 'integer'},
+                    'movedone': {'type': 'string'},
```
gazpachoking: I'm thinking we should change this option to 'path' rather than movedone. Transmission works similarly, and that's the option name in that plugin.

gazpachoking:
Other than the 'movedone' thing, looks good to me.

crawln45:
Just as a future thought. You can look at https://github.com/v1k45/python-qBittorrent for more inspiration. I thought it was intresting.

dsemi:
@gazpachoking Done

@crawln45 I know it exists. We only use one API call; it's a bit of overkill to pull in the library for a single post request.

crawln45:
@dsemi That was just for referencing the api. I found it a bit more informative than the webapi docs that are supplied in github for qbit.

gazpachoking:
Cool thanks! Want to add a wiki page?

dsemi:
Sure. I'll be able to get to it next week sometime. Is the wiki updated [here](https://github.com/Flexget/wiki)? The description says that's just a backup of the wiki.

gazpachoking:
The wiki is at http://flexget.com. We can add a page at http://flexget.com/wiki/Plugins/qbittorrent and link to it from http://flexget.com/wiki/Plugins#a3rdpartysoftwareoutput

barhom:
http://flexget.com/wiki/Plugins/qbittorrent is really needed.

I tried using this plugin right now and was able to add a torrent to qbittorrent. What is not working is setting the Download directory and setting the category so that it shows up in the webui correctly.

dsemi:
Ah forgot about this. Just added the wiki page for qBittorrent and linked in 3rd party output plugins list.
http://flexget.com/wiki/Plugins/qbittorrent

barhom:
@dsemi Have you tried "path" and "label" using qbittorrent-nox?
I cannot get those two to be applied. The torrent does get added though.

barhom:
I have tried 'path' again now and that works.

I could not get "label" to work though. Is it maybe because "label" (qbittorrent desktop) and "category" (web-ui) is not the same?

barhom:
New PR on https://github.com/Flexget/Flexget/pull/984

This is tested to work with the 'label' setting.

dsemi:
They haven't been updating their web api documentation, but the python bindings written [here](https://github.com/v1k45/python-qBittorrent/blob/master/qbittorrent/client.py#L306) use 'label'

barhom:
My guess is that 'label' is what it supposed to be what qbittorrent want to use for the webapi.

However, current master of qbittorrent accepts only 'category' in the POST FORM for setting the 'label' to a new torrent.

Ill make an issue on qbittorrent issue tracker to see if the fix can be quickly merged into qbittorrent.

dsemi:
Ah you're testing against the master. Looks like they changed the name in [this commit](https://github.com/qbittorrent/qBittorrent/commit/dd3466322455ad5779a01a80bbfea2e229cd5891). Either they should change the name back (preferable since they really shouldn't be changing the api names between versions, bad practice to break all existing compatibility) or we have detect the running version of qBittorrent and use the correct key of 'label' or 'category' based on that.

liiight:
I think the correct behavior is to match the one in the master branch. If the plugin/PR that was opened does not match that, then it should...

PR #885, opened 2016-02-25
Ui development

Tommatheussen:
Hello all,

I wanted to get into developing the Flexget Web-UI a couple of months ago, but was caught up in a lot off work/other projects then.
Right now I got some more free time on my hands and was able to have a look at it.

Right now I added a profile page, on which you are able to change the password of the user. 
It needs some work though( gotta look into Angular-material some more, cause I use bootstrap at work ).
I also changed/updated the modal service (it was not used anymore, is that correct?), I changed it to be more generic, so it reduces code duplication, and provides and easier way to open a dialog and handle the user's reaction. Right now the Dialog is used in the reload/shutdown functions.

I'm gonna try to be as active as possible, and provide regular updates to this branch.
Is the TODO list on [Web-UI](http://flexget.com/wiki/Web-UI) still up-to date?

Just for some further information, I could try to build some API functions if needed, but I don't have a lot off experience with Python ;)


stevezau:
@Tommatheussen yey someone to help me with the ui development! I'm in the same boat, when i have free time i try commit as much as i can.

I will review this ASAP but please do take a look at angular material as we are using it heavily.

I am also going to write a page on how to get started with the flexget ui development.

stevezau:
@Tommatheussen i think we should remove the dialog stuff. I created that early on when angular material was at ver 0.11, it seems to be more functional now.

stevezau:
@Tommatheussen , do you think you can submit the changes to the webui branch? you should have access now.

Also, the code looks ok to me. I didn't have a look at the profile but i was thinking about using a dialog or a toggle right side nav for profile.. 

https://material.angularjs.org/latest/demo/sidenav

thoughts?

Tommatheussen:
Closing this as we have a seperate branch for this now, anything usefull will be moved there.

PR #889, opened 2016-02-26
Add Slack output plugin

PhilRW:
This is my first attempt at contributing to this project. I wanted a way to notify my private Slack channel when a new download started. I based most of the code off of the Pushbullet plugin (thanks to whoever wrote that!).

Please let me know if I need to fix anything up -- and thanks!

### DIFF
--- a/flexget/plugins/output/slack.py
+++ b/flexget/plugins/output/slack.py
@@ -16,10 +16,18 @@ class OutputSlack(object):
 
       slack:
         webhook-url: <string>
-        [text: <string>] (default: "{{task}} - Download started: {{series_name}} {{series_id}}" -- accepts Jinja2)
+        [text: <string>] (default: "{{task}} - Download started:
+                                    {% if series_name is defined %}
+                                    {{tvdb_series_name|d(series_name)}} {{series_id}} {{tvdb_ep_name|d('')}}
+                                    {% elif imdb_name is defined %}
+                                    {{imdb_name}} {{imdb_year}}
+                                    {% else %}
+                                    {{title}}
+                                    {% endif %}"
         [channel: <string>] (override channel, use "@username" or "#channel")
         [username: <string>] (override username)
-        [icon-emoji: <string>] (override emoji icon, do not include the colons: e.g., use "tv" instead of ":tv:")
+        [icon-emoji: <string>] (override emoji icon, do not include the colons:
+                                e.g., use "tv" instead of ":tv:")
 
     Configuration parameters are also supported from entries (e.g., through set).
     """
--- a/flexget/plugins/output/slack.py
+++ b/flexget/plugins/output/slack.py
@@ -1,0 +1,120 @@
+from __future__ import unicode_literals, division, absolute_import
+import logging
+
+from flexget import plugin
+from flexget.event import event
+from flexget.utils import json
+from flexget.utils.template import RenderError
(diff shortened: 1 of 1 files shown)


flexget/plugins/output/slack.py
```diff
@@ -0,0 +1,120 @@
...
+log = logging.getLogger('slack')
+
+
+class OutputSlack(object):
+    """
+    Example:
+
+      slack:
+        webhook-url: <string>
+        [text: <string>] (default: "{{task}} - Download started: {{series_name}} {{series_id}}" -- accepts Jinja2)
```
liiight: This will crash if those value are not present. Better to remove them or use some if condition.

flexget/plugins/output/slack.py
```diff
@@ -0,0 +1,120 @@
...
+        },
+        'required': ['webhook-url'],
+        'additionalProperties': False
+    }
+
+    @plugin.priority(0)
+    def on_task_output(self, task, config):
+
+        for entry in task.accepted:
+            webhook_url = config.get('webhook-url')
```
liiight: I'm not familiar with slack api but wouldn't it be better to set a base url to use instead of passing the entire url from config?

flexget/plugins/output/slack.py
```diff
@@ -0,0 +1,120 @@
...
+        headers = {
+            'Content-Type': 'application/json',
+            'Accept': 'application/json',
+            'User-Agent': 'Flexget'
+        }
+        try:
+            response = task.requests.post(webhook_url, headers=headers,
+                                          data=json.dumps(data),
+                                          raise_status=False)
+        except Exception as e:
```
liiight: Would it be possible to catch a narrower exception? It's almost never a good idea to catch everything like this

flexget/plugins/output/slack.py
```diff
@@ -0,0 +1,120 @@
...
+            return
+
+        response_code = response.status_code
+
+        if response_code == 200:
+            log.debug('Slack notification sent')
+        elif response_code == 500:
+            log.warning('Slack notification failed: server problem')
+        elif response_code >= 400:
+            log.error('Slack API error {}: {}'.format(response_code, response.content))
```
liiight: Please use index numbers inside string format params. This is not compatible with python 2.6. Not a huge deal but easily solvable

flexget/plugins/output/slack.py
liiight: Maybe I'm picky, but default in the doc doesn't actual match real default

flexget/plugins/output/slack.py
PhilRW: I thought about that, too.... From what I can tell, the first random string in the URL is the team, the 2nd is the service, and the 3rd is the key, which gets regenerated whenever the user clicks "regenerate" in the Slack configuration. I could break this out into three separate values and make them all required and then hard-code the rest of the URL, but I would have to add a hefty bit of explanation because the Slack website itself doesn't supply these individually, it simply gives the entire URL to copy and paste.

liiight: Cool, if you know it fits better it's all good.

liiight:
Merged. You wanna add a wiki page?

PhilRW:
Sure! I'll start working on one now.

PR #893, opened 2016-02-27
Release group for series, fix for filesystem, improve search for old t411 plugin

Marmau:
Hi,

It's my first pull request ever, and it's a little messy...
So,
**series**
I added "release_group" entry in series. It is filled by guessit parser, and can be use in content_filename.

**filesystem**
Currently, filter "file" returns regular files and symlinks. I added a condition to differenciate filters "file" and "symlink"

**Old t411**
Search engine is not relevant with episode and season in request :
"My Serie S05E01" will return many of result of "My Serie" but not for "S05E01"...
Fortunately, there is season and episode filter. I made some changes to use these filters.


paranoidi:
Thanks!

PR #897, opened 2016-02-28
[WIP] Added device authorization and deprecated PIN auth

cvium:
http://docs.trakt.apiary.io/#reference/authentication-devices

### DIFF
--- a/flexget/plugins/api_trakt.py
+++ b/flexget/plugins/api_trakt.py
@@ -58,39 +58,31 @@ def token_created_date(created):
     return datetime.fromtimestamp(created)
 
 
-def device_auth(account, session):
+def device_auth():
     data = {'client_id': CLIENT_ID}
     r = requests.post(get_api_url('oauth/device/code'), data=data).json()
     device_code = r['device_code']
     user_code = r['user_code']
-    verification_url = r['verification_url']
     expires_in = r['expires_in']
     interval = r['interval']
 
     console('Please visit {0} and authorize Flexget. Your user code is {1}. Your code expires in '
-            '{2} minutes.'.format(verification_url, user_code, expires_in/60.0))
+            '{2} minutes.'.format(r['verification_url'], user_code, expires_in / 60.0))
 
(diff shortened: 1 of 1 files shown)


flexget/plugins/api_trakt.py
```diff
@@ -80,10 +81,59 @@ def get_access_token(account, token=None, refresh=False, re_auth=False):
...
+                    expires_in = r['expires_in']
+                    interval = r['interval']
+
+                    console('Please visit {0} and authorize Flexget. Your user code is {1}. Your code expires in '
+                            '{2} minutes.'.format(verification_url, user_code, expires_in/60.0))
+
+                    log.debug('Polling for user authorization.')
+                    data['code'] = device_code
+                    data['client_secret'] = CLIENT_SECRET
+                    end_time = time.time() + 5
```
gazpachoking: Shouldn't this be + expire seconds?

cvium: pff, haven't you ever had some leftover debugging code lying around? :P

flexget/plugins/api_trakt.py
```diff
@@ -80,10 +128,16 @@ def get_access_token(account, token=None, refresh=False, re_auth=False):
...
             elif token:
+                # We are only in here if a pin was specified, so it's safe to use console instead of logging
+                console('PIN authorization has been deprecated.')
                 data['code'] = token
                 data['grant_type'] = 'authorization_code'
             else:
-                raise plugin.PluginError('Account %s not found in db and no pin specified.' % account)
+                log.debug('No pin specified for an unknown account %s. Attempting to authorize device.' % account)
+                try:
+                    device_auth(account, session)
```
gazpachoking: Shouldn't this return after, instead of continuing to the token submit? Also, there isn't logic inside there when adding a new auth to check if there is an existing db entry already for that account. Maybe we should have 2 helper methods that return the access and refresh tokens for pin and device auth, then keep the db storage in the same method.

PR #923, opened 2016-03-05
The big component refactor

kristenmills:
Lots of changes here but all I've really done is made everything a component and broke up some of the monolithic views into smaller components.  

I tried my best to not break anything and I don't think I did but I didn't want to merge this in until someone else could sanity check me.


stevezau:
good work!

PR #854, opened 2016-02-06
Generic list interface

gazpachoking:
Start of this idea:

http://flexget.com/wiki/Drafts/ListInterface

So far just hacking trakt_add, trakt_remove, and trakt_list into one plugin which implements a standard interface for adding removing and getting a list (add_entries, remove_entries, and list_entries)

Stuff to figure:
- [x] there are some lists that work with trakt_list, but we can't add/remove from (the ratings lists.) Also, if using public lists when not authenticated. Probably throw a PluginError
- [x] Move all plugins ending in _list to this framework, and have them raise errors like above if they aren't editable.
- [x] Make imdb lists editable!? https://gist.github.com/gazpachoking/28d666bf6e60b30ce3a9
- [x] Add a bit of error handling to new imdb_list
- [ ] Perhaps we should enforce the methods of the list objects to take the session keyword, so that for plugins using the database we can control the session from an outside scope if needed. Or maybe have it passed in on init is better and get rid of all the with_sessions. EDIT: Not too sure about this anymore

PR #924, opened 2016-03-05
Add input plug-in for npo.nl watchlist

jeroenl:
This plug-in creates entries for a user's watchlist entries on npo.nl (Dutch public television). 

The entries can for example be downloaded
using https://bitbucket.org/Carpetsmoker/download-npo

### DIFF
--- a/flexget/plugins/input/npo_watchlist.py
+++ b/flexget/plugins/input/npo_watchlist.py
@@ -44,7 +44,7 @@ class NPOWatchlist(object):
             'password': {'type': 'string'},
             'remove_accepted': {'type': 'boolean', 'default': False}
         },
-        'required': [],
+        'required': ['email', 'password'],
         'additionalProperties': False
     }
     
--- a/flexget/plugins/input/npo_watchlist.py
+++ b/flexget/plugins/input/npo_watchlist.py
@@ -1,32 +1,28 @@
 from __future__ import unicode_literals, division, absolute_import
 import logging
-import urllib
-
-from jinja2 import TemplateSyntaxError
 
 from flexget import plugin
 from flexget.entry import Entry
 from flexget.event import event
-from flexget.task import Task
-from flexget.utils.search import normalize_unicode
 
 from bs4 import BeautifulSoup
+from datetime import datetime
+
 import unicodedata
-
 import requests
-
 import re
-from datetime import datetime, timedelta
 
 log = logging.getLogger('search_npo')
 search_results_regex = re.compile('<a href="([^"]+)" data-scorecard="{&quot;name&quot;:&quot;zoeken')
+
 
 class NPOWatchlist(object):
     """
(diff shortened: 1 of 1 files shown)


flexget/plugins/input/npo_watchlist.py
```diff
@@ -0,0 +1,152 @@
...
+        """
+    
+    schema = {
+        'type': 'object',
+        'properties': {
+            'email': {'type': 'string'},
+            'password': {'type': 'string'},
+            'remove_accepted': {'type': 'boolean', 'default': False}
+        },
+        'required': [],
```
gazpachoking: Are email and password supposed to be in here? This being empty is why the tests are failing.

gazpachoking:
Cool. I'm thinking this will be a good candidate for the new list plugin interface we are working on here #854. It allows the user to give a lot more explicit control of adding and removing from the list. Curious your thoughts on that branch, and whether you think it'd work nicely for your plugin.

jeroenl:
The list interface looks good, would be nice to have that a bit more generic. Shall I wait until that's stable, and then take a stab at porting it to use that?

gazpachoking:
Yeah, you can wait until we get it in. We need to write some tests still, and clear up the documentation. In terms of being more generic, what were you thinking? The idea is that they all operate on entries, and each list plugin gets to decide what bits of the entry it cares about.

jeroenl:
With more generic, I meant your proposal. Good to have a consistent way of dealing with lists, just like discover gives us a consistent way of searching.

One thing my plug-in does that I can't quite fit on this, is that it has the option of removing the entry from the watchlist if it was successfully downloaded. Is there a way to have list_remove operate in the complete phase, or at least after the entry is processed? I was struggling with that.
(Sounds almost like what the series plug-ins do in terms of tracking, but that seems way overkill).

Or should I keep the 'remove_accepted' option in, even while using the list interface?

gazpachoking:
Yeah, we built that logic into the list_match plugin, it has an option to remove when an item from the list is accepted (and output successfully.) Your case is a little different, because the way the other plugins work the entries created by the list plugin aren't the same ones that are being considered for download. Yours would end up looking a little more redundant I guess in the list system:
```yaml
npo_watchlist: opts
list_match:  # This accepts anything in the list, and removes it from list if entry completes succesfully
- npo_watchlist: opts
download: yes  # or whatever output plugin
```
the other option would be to do it with list remove:
```yaml
npo_watchlist: opts
accept_all: yes  # Without list match, we just accept everything
download: yes
list_remove:  # This runs on output phase, which would be the same as download, but we should probably tweak it to run with a later default priority than the 'primary' output plugins. Result would then be similar.
- npo_watchlist: opts
```

gazpachoking:
Thanks! Stay tuned for more info on the list interface. It'd be nice to get a bunch of plugins standardized on it once we roll it out.

PR #927, opened 2016-03-07
[WIP] Caching for Trakt Collection/Watched History

cvium:
- [x] Movie collection caching 
- [ ] Episode collection caching
- [ ] Movie watched caching 
- [ ] Episode watched caching
- [ ] Database pruning ie. set `collected_at` and `watched_at` columns to Null when stuff is being updated. Then call some pruning function to clean the rows where both those columns are Null. Not sure this is required...

Refactored some of the collection stuff and `trakt_lookup` now _optionally_ takes a `username`. If `username` is not specified, then the `last-activities` endpoint is used to determine whether the database cache needs to be updated.


cvium:
Closed. Complexity trade-off isn't worth it. Data changes too often to be worth saving it in a database.

PR #932, opened 2016-03-08
Added show collection and watched status

cvium:
~~OAUTH is required, so it's not as nice as I would have liked it... Alternatively, we could just iterate over the entire collection and _count_ the number of collected episodes and compare it to `aired_episodes`. Speed might be an issue though.~~

It just iterates over the whole collection/watched to get the number of collected/watched episodes. Very simple and it allows people to check the progress for public profiles, so no OAUTH is required.

### DIFF
--- a/flexget/plugins/api_trakt.py
+++ b/flexget/plugins/api_trakt.py
@@ -860,10 +860,9 @@ class ApiTrakt(object):
             if trakt_data.id in cache:
                 log.error('work')
                 series = cache[trakt_data.id]
-                num = 0
-                for s in series['seasons']:
-                    num += len(s['episodes'])
-                in_collection = num == trakt_data.aired_episodes
+                # specials are not included
+                number_of_collected_episodes = sum(len(s['episodes']) for s in series['seasons'] if s['number'] > 0)
+                in_collection = number_of_collected_episodes >= trakt_data.aired_episodes
         elif style == 'episode':
             if trakt_data.show.id in cache:
                 series = cache[trakt_data.show.id]
@@ -895,10 +894,9 @@ class ApiTrakt(object):
         if style == 'show':
             if trakt_data.id in cache:
                 series = cache[trakt_data.id]
-                num = 0
-                for s in series['seasons']:
-                    num += len(s['episodes'])
-                watched = num == trakt_data.aired_episodes
+                # specials are not included
+                number_of_watched_episodes = sum(len(s['episodes']) for s in series['seasons'] if s['number'] > 0)
+                watched = number_of_watched_episodes == trakt_data.aired_episodes
         elif style == 'episode':
             if trakt_data.show.id in cache:
                 series = cache[trakt_data.show.id]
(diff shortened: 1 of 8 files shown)


flexget/plugins/api_trakt.py
```diff
@@ -851,15 +847,24 @@ def lookup_movie(session=None, only_cached=None, **lookup_params):
...
-        cache = ApiTrakt.user_cache[ident]['collection'][style + 's']
+        cache = ApiTrakt.user_cache[ident]['collection'][style_ident]
         if style == 'show':
+            if trakt_data.id in cache:
+                log.error('work')
+                series = cache[trakt_data.id]
+                num = 0
+                for s in series['seasons']:
+                    num += len(s['episodes'])
+                in_collection = num == trakt_data.aired_episodes
```
gazpachoking: Do either of these numbers take into account specials? What should the behavior with specials be? Should you need every special for it to be marked collected? I think trakt's builtin behavior says they don't matter.

flexget/plugins/api_trakt.py
```diff
@@ -851,15 +847,24 @@ def lookup_movie(session=None, only_cached=None, **lookup_params):
...
         in_collection = False
-        cache = ApiTrakt.user_cache[ident]['collection'][style + 's']
+        cache = ApiTrakt.user_cache[ident]['collection'][style_ident]
         if style == 'show':
+            if trakt_data.id in cache:
+                log.error('work')
+                series = cache[trakt_data.id]
+                num = 0
+                for s in series['seasons']:
+                    num += len(s['episodes'])
```
gazpachoking:
```python
# Number of watched non-special episodes
num = sum(len(s['episodes']) for s in series['seasons'] if s['number'] > 0)
```
May be better depending on answer to the specials question below.

flexget/plugins/api_trakt.py
```diff
@@ -695,67 +691,70 @@ def get_trakt(style=None, title=None, year=None, trakt_id=None, trakt_slug=None,
...
         raise plugin.PluginError('Unable to get data from trakt.tv: %s' % e)
 
 
-def add_user_to_cache(ident):
-    ApiTrakt.user_cache.setdefault(ident, {}).setdefault('watched', {}).setdefault('shows', {})
-    ApiTrakt.user_cache.setdefault(ident, {}).setdefault('watched', {}).setdefault('movies', {})
-    ApiTrakt.user_cache.setdefault(ident, {}).setdefault('collection', {}).setdefault('shows', {})
-    ApiTrakt.user_cache.setdefault(ident, {}).setdefault('collection', {}).setdefault('movies', {})
+def get_user_cache(username=None, account=None):
+    identifier = str(account) + '|' + str(username)
```
gazpachoking: Do we need to call str here? If user puts non-ascii in their account name this will crash. perhaps `identifier = '{}|{}'.format(account, username or 'me')`

cvium: Shouldn't `from __future__ import unicode_literals` take care of that?

PR #943, opened 2016-03-11
fix API calls missing trailing slash to avoid unnecessary redirections

ghyster:
I'm hosting flexget ui behind an apache reverse proxy with ssl activated and some api calls don't work. When there's an api call without a trailing slash a redirection is made to an absolute url without ssl, thus cancelling the request


stevezau:
thanks!

PR #906, opened 2016-03-01
T411 domain changed from t411.in to t411.ch

poulpito:
Hi :) 
first pull request, hope i'm doing it right 

t411 change the domain  from t411.in to t411.ch 
so i update  api references to the new domain.

sorry for differents commits, i've update files online and haven't find how to modify all in one commit.

PR #944, opened 2016-03-11
Fix missing domain change from t411.in to t411.ch

Ekhozie:
Hello,
This pull request just fix one missing domain change from PR https://github.com/Flexget/Flexget/pull/906
This fixes url rewrite error on my system


stevezau:
thanks

PR #946, opened 2016-03-11
Session now retrieved from task.requests, respecting domain_delays an…

qvazzler:
…d other task config items. Also added default domain delay of 4 seconds due to increased spam filters by SCC

Planning to do this for most search plugins, at least to use the task.requests session.
This is a needed fix in order to use SCC currently.

### DIFF
--- a/flexget/plugins/search_sceneaccess.py
+++ b/flexget/plugins/search_sceneaccess.py
@@ -235,7 +235,7 @@ class SceneAccessSearch(object):
         session = task.requests
 
         if not 'sceneaccess.eu' in session.domain_limiters:
-            session.add_domain_limiter(TimedLimiter('sceneaccess.eu', '4 seconds'))
+            session.add_domain_limiter(TimedLimiter('sceneaccess.eu', '7 seconds'))
 
         if not session.cookies:
             log.debug('Logging in to %s...' % URL)
--- a/flexget/plugins/search_sceneaccess.py
+++ b/flexget/plugins/search_sceneaccess.py
@@ -15,7 +15,7 @@ from flexget.utils.requests import Session, TimedLimiter
 log = logging.getLogger('search_sceneaccess')
 
 session = None
-#session = Session()
+
 CATEGORIES = {
     'browse':
         {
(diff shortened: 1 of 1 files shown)


flexget/plugins/search_sceneaccess.py
```diff
@@ -234,6 +235,14 @@ def search(self, task, entry, config=None):
             Search for entries on SceneAccess
         """
 
+        try:
+            session = task.requests
+        except:
```
cvium: catching everything like this is ill-advised. I also don't see it ever throwing an exception. You're accessing an object member that is set on object creation.

flexget/plugins/search_sceneaccess.py
```diff
@@ -10,11 +10,11 @@
 from flexget.event import event
 from flexget.utils.soup import get_soup
 from flexget.utils.search import torrent_availability, normalize_unicode, clean_title
-from flexget.utils.requests import Session
+from flexget.utils.requests import Session, TimedLimiter
 
 log = logging.getLogger('search_sceneaccess')
 
-session = Session()
+session = None
```
cvium: Is it necessary to declare the variable in global scope? Isn't it only used in `search`?

flexget/plugins/search_sceneaccess.py
```diff
@@ -10,11 +10,11 @@
 from flexget.event import event
 from flexget.utils.soup import get_soup
 from flexget.utils.search import torrent_availability, normalize_unicode, clean_title
-from flexget.utils.requests import Session
+from flexget.utils.requests import Session, TimedLimiter
```
cvium: Is the `Session` import used anymore?

flexget/plugins/search_sceneaccess.py
qvazzler: For now, just want to replicate previous code where possible.

flexget/plugins/search_sceneaccess.py
```diff
@@ -234,6 +232,11 @@ def search(self, task, entry, config=None):
             Search for entries on SceneAccess
         """
 
+        session = task.requests
+
+        if not 'sceneaccess.eu' in session.domain_limiters:
```
cvium: pep standard prefers `if string not in bla bla` and I'm a firm believer in adhering to standards *hint hint*

qvazzler: Very well..!

cvium:
Is it ready then?

qvazzler:
Sure feels like it! @gazpachoking

PR #958, opened 2016-03-16
Create better path-safe series title

jeroenl:


gazpachoking:
So, should this plugin just implement pathscrub cleaning? http://flexget.com/wiki/Plugins/pathscrub That utility is designed for any plugin which writes to the filesystem to use, to make sure dynamic paths are safe for the specific filesystem.

jeroenl:
Ah, right! I did look for a standard way of doing this outside Flexget, but didn't expect a plugin for it.
I'll take a look, and will either integrate it or find a way to combine pathscrub with exec. Any way to use pathscrub from Jinja?

gazpachoking:
Yep, it can be used from jinja http://flexget.com/wiki/Jinja#Jinja2Filters

If you don't specify the OS in jinja, it'll base it on however the plugin is configured (which is based on your OS by default)

jeroenl:
Ok, thanks, that's a better alternative then. Thanks for the tip.

Issue #978, opened 2016-03-27
anidb_list: AttributeError: 'NoneType' object has no attribute 'find' 


avaddon:
➜  python -V
Python 2.7.10
➜  flexget -V
1.2.495

2016-03-27 21:01 VERBOSE  anidb_list    anidb Retrieving AniDB list: mywishlist
2016-03-27 21:01 VERBOSE  anidb_list    anidb Entry does not match the requested type
2016-03-27 21:01 CRITICAL task          anime_horriblesubs BUG: Unhandled error in plugin configure_series: 'NoneType' object has no attribute 'find'
Traceback (most recent call last):
  File "/Library/Python/2.7/site-packages/flexget/task.py", line 438, in __run_plugin
    return method(*args, **kwargs)
  File "/Library/Python/2.7/site-packages/flexget/event.py", line 22, in __call__
    return self.func(*args, **kwargs)
  File "/Library/Python/2.7/site-packages/flexget/plugins/plugin_configure_series.py", line 69, in on_task_start
    result = method(task, input_config)
  File "/Library/Python/2.7/site-packages/flexget/event.py", line 22, in __call__
    return self.func(*args, **kwargs)
  File "/Library/Python/2.7/site-packages/flexget/utils/cached_input.py", line 142, in wrapped_func
    response = func(*args, **kwargs)
  File "/Library/Python/2.7/site-packages/flexget/plugins/input/anidb_list.py", line 64, in on_task_input
    a = tr.find('label').find('a')
AttributeError: 'NoneType' object has no attribute 'find'

molybden123:
Traceback (most recent call last):
  File "/usr/local/lib/python2.7/dist-packages/flexget/task.py", line 477, in __run_plugin
    return method(*args, **kwargs)
  File "/usr/local/lib/python2.7/dist-packages/flexget/event.py", line 23, in __call__
    return self.func(*args, **kwargs)
  File "/usr/local/lib/python2.7/dist-packages/flexget/utils/cached_input.py", line 148, in wrapped_func
    response = func(*args, **kwargs)
  File "/usr/local/lib/python2.7/dist-packages/flexget/plugins/input/anidb_list.py", line 54, in on_task_input
    trs = soup.find_all('tr')
AttributeError: 'NoneType' object has no attribute 'find_all'

cvium:
Please open a new issue instead.

PR #980, opened 2016-03-27
Update anidb_list.py.

avaddon:
Fix for https://github.com/Flexget/Flexget/issues/978


liiight:
Can you resync PR please? we had some issues on dev branch that caused other tests to fail

PR #984, opened 2016-03-28
webui form data edited from 'label' to 'category'

barhom:
The qbittorrent-nox webui takes 'category' and not 'label' as its formdata to specify which label to add the torrent to

### DIFF
--- a/flexget/plugins/output/qbittorrent.py
+++ b/flexget/plugins/output/qbittorrent.py
@@ -95,7 +95,8 @@ class OutputQBitTorrent(object):
         for entry in task.accepted:
             data = {}
             data['savepath'] = entry.get('path', config.get('path'))
-            data['label'] = entry.get('label', config['label']).lower()
+            data['label'] = entry.get('label', config['label']).lower()  # qBittorrent v3.3.3 and earlier
+            data['category'] = entry.get('label', config['label']).lower()  # qBitTorrent v3.4 and newer
             data['urls'] = [entry.get('url')]
             if task.manager.options.test:
                 log.info('Test mode.')


dsemi:
I just tested, 'label' is working fine for me as-is. What version of qBittorrent are you using? I'm on v3.3.3 and my Web UI says 'Labels' not 'Categories' (I'm also not using qbittorrent-nox).  Changing label to category breaks labeling for me.  I'd try updating your version of qBittorrent to see if that fixes your issue, and if not then a better solution would be to add a new config option and specify on the wiki how it differs from label.

barhom:
Im using latest master branch of qbittorrent (3.4 something Alpha). Are you sure your WEBUI shows 'labels' and not 'categories'?

Here's an image from the WebUI:
http://i.imgur.com/bCBAABb.png

dsemi:
Looks like they changed the name in [this commit](https://github.com/qbittorrent/qBittorrent/commit/dd3466322455ad5779a01a80bbfea2e229cd5891). The most recent release is still v3.3.3 and that uses Labels.
In my opinion, qBittorrent shouldn't be changing names in their API because it'll break everything that used 'label' when they release v3.4.0. You can open an issue with them to change the name back or support either label or category. The other option is to check for a version number in the flexget output plugin and change the key to either 'label' or 'category' based on whether the version is above 3.4.0 or not.
Is there something in the master branch that you need that's not in v3.3.3. If not then you can just use that version and the plugin will work properly.

barhom:
@dsemi 
I edited the PR a little bit. This way we send both label and category to support both old and new version.
I doubt this will introduce any errors, what do you think?

dsemi:
Seems fine to me as long as no versions break with the extra parameters.

liiight:
can you sync the pr please?

stevezau:
update on this?

PR #988, opened 2016-03-29
[file_age] new plugin to reject/accept files based on last access time etc.

cvium:


liiight:
Maybe call it `file_stats_filter`?

paranoidi:
This is annoyingly specific for filesystem. Why not generic age plugin that works with RSS feeds etc?

flexget/plugins/filter/file_age.py
```diff
@@ -0,0 +1,67 @@
...
+          file_age:
+            stat: 'access'
+            age: '7 days'
+    """
+
+    schema = {
+        'type': 'object',
+        'properties': {
+            'stat': {'type': 'string', 'enum': ['access', 'modified', 'created']},
+            'reject': {'type': 'boolean', 'default': True},
```
paranoidi: usual convention would be "action: [accept|reject]"

liiight:
what age attribute does rss feeds have?

flexget/plugins/filter/file_age.py
```diff
@@ -0,0 +1,67 @@
...
+    def on_task_filter(self, task, config):
+        for entry in task.entries:
+            if 'location' not in entry:
+                log.debug('Entry %s is not a local file. Skipping.', entry['title'])
+                continue
+            filepath = entry['location']
+            if config['stat'] == 'created' and not sys.platform.startswith('win'):
+                log.warning('File creation time is not available in *nix OS and defaults to modification time.')
+            file_stats = {}
+            file_stats['access'] = datetime.fromtimestamp(os.path.getatime(filepath))
```
paranoidi: Filesystem plugin should probably set these as entry fields. Then the "age" plugin could work nicely like

age:
  field: access
  amount: 7 days
  action: reject

and thus be generic

liiight:
@cvium https://github.com/Flexget/Flexget/commit/642c505830cf1b90ba1d3c835740d03f7c5fcb53

flexget/plugins/filter/age.py
```diff
@@ -0,0 +1,56 @@
...
+            amount: '7 days'
+            action: 'accept'
+    """
+
+    schema = {
+        'type': 'object',
+        'properties': {
+            'field': {'type': 'string'},
+            'action': {'type': 'string', 'enum': ['accept', 'reject']},
+            'amount': {'type': 'string', 'format': 'interval'}
```
liiight: `amount` is kinda weird. I think `age` worked well.

Issue #1013, opened 2016-04-07
QBittorent Windows Paths


Zephyrus29:
When I have a path set like`path: "F:/TV Shows/{{ series_name }}/Season {{ series_season }}/"`, any accepted torrents are downloaded to that directory but all spaces are replaced with +'s like `F:\TV+Shows\Brooklyn+Nine-Nine\Season+3`

qvazzler:
What is the actual issue that you are facing? Could you share logs / config?

qvazzler:
@dsemi is it supposed to be `savepath `or `save_path `in qbittorrent.py? Looking at API here shows save_path.. https://github.com/qbittorrent/qBittorrent/wiki/WebUI-API-Documentation#download-torrent-from-url

Zephyrus29:
My output configuration is
```
 exec: 
      on_output:
        for_accepted:
          - echo %DATE% %TIME% found {{series_name}} {{series_id}} at {{url}} >> acceptedTorrents.txt
     qbittorrent:
      username: admin
      password: password
      host: 127.0.0.1
      port: 8080
      path: "F:/TV Shows/{{ series_name }}/Season {{ series_season }}/"

```
And here is the relevant section of the log files:
```
2016-04-07 18:50 VERBOSE  task          TV_download     ACCEPTED: `Brooklyn Nine Nine S03E21 720p HDTV HEVC x265 RMTeam 720P HEVC` by series plugin because target quality
2016-04-07 18:50 INFO     urlrewriter   TV_download     Entry 'Brooklyn Nine Nine S03E21 720p HDTV HEVC x265 RMTeam 720P HEVC' URL rewritten to https://torcache.net/torrent/3B1A1469C180F447B77021074DBBCCAEF62611E5.torrent (with torrentz)
2016-04-07 18:50 INFO     urlrewriter   TV_download     Entry 'The Americans 2013 S04E04 720p HDTV HEVC x265 RMTeam 720P HEVC' URL rewritten to https://torcache.net/torrent/367AB1F057DD6056A28F717FEF73BFFD8A48B189.torrent (with torrentz)
2016-04-07 18:50 INFO     download      TV_download     Downloading: Brooklyn Nine Nine S03E21 720p HDTV HEVC x265 RMTeam 720P HEVC
2016-04-07 18:50 INFO     download      TV_download     Downloading: The Americans 2013 S04E04 720p HDTV HEVC x265 RMTeam 720P HEVC
[...]

PR #1033, opened 2016-04-13
Change qBittorrent plugin to use multipart request for adding torrents

dsemi:
qBittorrent's API specifies using a multipart request to add torrents, so each piece is encoded as plain text rather than form encoded. This should fix issues like [this](https://github.com/Flexget/Flexget/issues/1013).

### DIFF
--- a/flexget/plugins/output/qbittorrent.py
+++ b/flexget/plugins/output/qbittorrent.py
@@ -61,7 +61,8 @@ class OutputQBitTorrent(object):
     def add_torrent(self, data):
         if not self.connected:
             raise plugin.PluginError('Not connected.')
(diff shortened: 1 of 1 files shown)


flexget/plugins/output/qbittorrent.py
```diff
@@ -96,7 +97,7 @@ def on_task_output(self, task, config):
             data = {}
             data['savepath'] = entry.get('path', config.get('path'))
             data['label'] = entry.get('label', config['label']).lower()
-            data['urls'] = [entry.get('url')]
+            data['urls'] = entry.get('url')
```
liiight: This isn't a list anymore?

dsemi: No.

PR #1044, opened 2016-04-17
IPTorrents - Only parse search results

ianstalk:
The IPTorrents plugin attempts to parse any link ending in ".torrent", which causes it to go through the "Recommended" and "Top" torrents section at the top of the page. This isn't desirable behavior, and causes a crash when those links don't include the attributes it expects. This change will make it specifically parse the search results table on the page.


ianstalk:
Actually it looks like there are more issues-- the search results layout has changed substantially. I'll work in getting it working again.

liiight:
Is this ready?

ianstalk:
Should be. I've been dogfooding it for the past few days, and it seems to
be working again.

PR #1052, opened 2016-04-19
DB backup on upgrade plugin

liiight:
<githubroot>This plugin purpose is to backup DB when DB upgrade is needed in order to avoid DB corruption in case of failed backup.
<br />
<br />It is used with the `--backup-db` flag.
<br />When a DB upgraded is needed it will fire the `manager.backup_db` event which is caught by the plugin.
<br />
<br />A local copy of the DB will be created with the format: `backup-<CONFIG_NAME>-<DATE>_<TIME>-<FLEXGET_VERSION>.sqlite`. 
<br />Example: `backup-config2-20160419_1217-1.2.515.dev.sqlite`
<br />
<br />When the flag is not present and a DB upgrade is initiated, a log info appears:
<br /> `DB is upgrading without a backup, consider using the --backup-db-on-upgrade flag`.
<br />
<br />In addition, a new CLI command was introduced to manually use this plugin:
<br />`flexget database backup`.
<br />
<br />
<br /></githubroot>

### DIFF
--- a/flexget/manager.py
+++ b/flexget/manager.py
@@ -197,10 +197,7 @@ class Manager(object):
         if db_schema.upgrade_required():
             log.info('Database upgrade is required. Attempting now.')
             if self.options.backup_db:
-                log.verbose('backup db on upgrade flag detected, trying to backup DB')
                 fire_event('manager.backup_db', self)
-            else:
-                log.info('DB is upgrading without a backup, consider using the `--backup-db-on-upgrade` flag')
             fire_event('manager.upgrade', self)
             if manager.db_upgraded:
                 fire_event('manager.db_upgraded', self)
--- a/flexget/manager.py
+++ b/flexget/manager.py
@@ -200,7 +200,7 @@ class Manager(object):
                 log.verbose('backup db on upgrade flag detected, trying to backup DB')
                 fire_event('manager.backup_db', self)
             else:
-                log.info('Consider using the `--backup-db-on-upgrade` flag')
+                log.info('DB is upgrading without a backup, consider using the `--backup-db-on-upgrade` flag')
             fire_event('manager.upgrade', self)
             if manager.db_upgraded:
                 fire_event('manager.db_upgraded', self)
--- a/flexget/manager.py
+++ b/flexget/manager.py
@@ -196,6 +196,11 @@ class Manager(object):
 
         if db_schema.upgrade_required():
             log.info('Database upgrade is required. Attempting now.')
+            if self.options.backup_db_on_upgrade:
+                log.verbose('backup db on upgrade flag detected, trying to backup DB')
+                fire_event('manager.backup_db', self)
+            else:
+                log.info('Consider using the `--backup-db-on-upgrade` flag')
             fire_event('manager.upgrade', self)
             if manager.db_upgraded:
                 fire_event('manager.db_upgraded', self)
--- a/flexget/manager.py
+++ b/flexget/manager.py
@@ -196,7 +196,7 @@ class Manager(object):
 
         if db_schema.upgrade_required():
             log.info('Database upgrade is required. Attempting now.')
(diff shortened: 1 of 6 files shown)


flexget/plugins/api_tvmaze.py
```diff
@@ -17,14 +17,14 @@
 
 log = logging.getLogger('api_tvmaze')
 
-DB_VERSION = 4
+DB_VERSION = 5
```
gazpachoking: Is this upgrade meant to be mixed in this pr?

flexget/manager.py
```diff
@@ -196,6 +196,11 @@ def initialize(self):
 
         if db_schema.upgrade_required():
             log.info('Database upgrade is required. Attempting now.')
+            if self.options.backup_db:
+                log.verbose('backup db on upgrade flag detected, trying to backup DB')
+                fire_event('manager.backup_db', self)
+            else:
+                log.info('DB is upgrading without a backup, consider using the `--backup-db-on-upgrade` flag')
```
gazpachoking: I think this should be moved to the backup plugin, running on the manager.upgrade event.

flexget/options.py
```diff
@@ -394,6 +395,9 @@ def _debug_tb_callback(self, *dummy):
 manager_parser.add_argument('--cron', action=CronAction, default=False, nargs=0,
                             help='use when executing FlexGet non-interactively: allows background '
                                  'maintenance to run, disables stdout and stderr output, reduces logging level')
+manager_parser.add_argument('--backup-db', action='store_true', default=False,
+                            help='in case a DB upgrade is required, a db backup will be created')
+
```
gazpachoking: This should be moved to the backup plugin as well.

flexget/plugins/api_tvmaze.py
liiight: Oops, did it for a test

flexget/options.py
liiight: Will do

flexget/plugins/cli/database.py
```diff
@@ -77,16 +79,21 @@ def reset_plugin(options):
             console('Unable to reset %s: %s' % (plugin, e.message))
 
 
+def backup(manager):
+    fire_event('manager.backup_db', manager)
```
gazpachoking: Hmm, a plugin is firing a manager event? This doesn't feel quite right to me. I think the framework may support registering this from the backup plugin, where it can run itself directly, and skip the event.

flexget/manager.py
liiight: You mean the log message that no backup was created?

gazpachoking: Yes, right now you have the backup plugin as a plugin, but have functionality mixed into the core. I think we should separate all functionality related to the backup plugin into the backup plugin.

flexget/plugins/cli/database.py
liiight: So I'll add an argument to the database subparser from the backup pkugin. Got it, although I'm not sure as to why firing that event doesn't feel right to you

liiight: Nm, misread at first. Got it

liiight: Not sure how to do this @gazpachoking, doesn't look so straight forward.

liiight:
I have no idea why it fails, it should register that option correctly AFAIK.

paranoidi:
Manual argument or command for backup is not going to save anyone who would be benefiting from these kind of functionalities.

Perhaps it should be automatic and be skipped with an argument ...

liiight:
Main fear is gobbling people's storage without their knowledge or permission

liiight:
so should i just try to save 1 backup of the db if there's space? this seem really beneficial to me

liiight:
Will probably go in a different route

PR #1064, opened 2016-04-21
Webui movie plugin

NoMaillard:
This is a start at the movies list.

It need some responsivness (thinking about just the poster with limited info)

Suggestions are welcome


flexget/ui/src/plugins/movies/movies.component.js
```diff
@@ -31,9 +31,14 @@
 
         vm.movies.forEach(function (movie) {
 
-          $http.get('http://www.omdbapi.com/?i='+ movie.movies_list_ids[0].id_value +'&plot=short&r=json')
+          $http.get('/api/trakt/movie/', {
```
Tommatheussen: You might want to move this call to the separate movie-entry controller, making the general one more readable and easier to maintain.

NoMaillard: sure thing

PR #1068, opened 2016-04-23
[FIX] Add ConnectionError except to try clause to IMDB_LIST

crawln45:
FIX# 1058


flexget/plugins/list/imdb_list.py
```diff
@@ -104,6 +104,8 @@ def items(self):
                                  (self.list_id, self.user_id))
             except HTTPError as e:
                 raise PluginError(e.args[0])
+            except ConnectionError as e:
```
liiight: If it's the same handling as HTTPError wouldn't it be better to group those together?

PR #1070, opened 2016-04-23
[FIX]Fix# 1058 Add ConnectionError handling to request

crawln45:
Fix imdb_lists crash from ConnectionError

### DIFF
--- a/flexget/plugins/list/imdb_list.py
+++ b/flexget/plugins/list/imdb_list.py
@@ -6,7 +6,7 @@ import re
 from collections import MutableSet
 from datetime import datetime
 
-from requests.exceptions import ConnectionError, HTTPError
+from requests.exceptions import RequestException
 
 from flexget import plugin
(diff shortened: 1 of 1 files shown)


flexget/plugins/list/imdb_list.py
```diff
@@ -102,7 +102,7 @@ def items(self):
             try:
                 r = self.session.get('http://www.imdb.com/list/export?list_id=%s&author_id=%s' %
                                  (self.list_id, self.user_id))
-            except HTTPError as e:
+            except [HTTPError, ConnectionError] as e:
```
stevezau: why not just use RequestException?

liiight: Also, I think that syntax requires `()`

crawln45: There no () needed. @stevezau because I'm tired.

PR #1077, opened 2016-04-24
Fix discover with limit for Python 3

jeroenl:
Add fall-back blank sort key for discover, to produce same behaviour as
in Python 2. Without this it crashes on:

`TypeError: unorderable types: NoneType() < NoneType()`


stevezau:
thanks for all the fixes!

jeroenl:
No problem! Mostly self-serving. :smile:

PR #1078, opened 2016-04-24
Py3: Fix "'list' object is not callable, in regexp plugin

jeroenl:
There was a name clash between the list of fields that had to be unquoted and the unquote  method, both called 'unquote'. Also, unquote was not imported.

### DIFF
--- a/flexget/plugins/filter/regexp.py
+++ b/flexget/plugins/filter/regexp.py
@@ -10,6 +10,11 @@ from flexget.config_schema import one_or_more
 from flexget.entry import Entry
 from flexget.event import event
 
+try:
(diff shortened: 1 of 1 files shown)


flexget/plugins/filter/regexp.py
```diff
@@ -10,6 +10,11 @@
 from flexget.entry import Entry
 from flexget.event import event
 
+try:
```
stevezau: Should be able to use 

from future.moves.urllib.parse import quote

stevezau: I mean unquote

PR #1079, opened 2016-04-24
List interface: improve logging

jeroenl:
Before the generic list interface was introduced, each entry that was added to the movie queue was logged. This helped tremendously to see what was happening in a task.

I've restored this logging in list_add/remove/accept.


liiight:
I don't agree with this since it will give the appearance that every entry on task.accepted can be operated on which is not the case. This will just lead to confusion imo

stevezau:
you are right, the list plugins (impl) should be doing the logging.

I've reverted the change.

liiight:
@jeronel if you think it should be verbose, change it in the list plugin themselves

gazpachoking:
We have the -v option for printing out entries and their state, does that work for you @jeroenl ?

PR #1081, opened 2016-04-24
Py3: Fix error handling in download plugin

jeroenl:
While trying to store an error page, the plugin returned:
`TypeError: write() argument must be str, not bytes`


flexget/plugins/output/download.py
```diff
@@ -155,9 +155,9 @@ def save_error_page(self, entry, task, page):
         received = os.path.join(task.manager.config_base, 'received', task.name)
         if not os.path.isdir(received):
             os.makedirs(received)
-        filename = os.path.join(received, '%s.error' % entry['title'].encode(sys.getfilesystemencoding(), 'replace'))
+        filename = os.path.join(received, '%s.error' % entry['title'].encode(sys.getfilesystemencoding(), 'replace').decode(sys.getfilesystemencoding()))
```
stevezau: Hmm, we are encoding then decoding?? that does not sound right?

what error did you get?

jeroenl: The error came below. For this, I noticed that the filename shows up as:
path/to/b'entry title'.error
as opposed to (in Py2)
path/to/entry title.error

Encode then decode in this case would at least drop characters that are not valid in the file system encoding, which I'm guessing at some point fixed a bug. I do wonder though if maybe we shouldn't change the way the file name is generated. Perhaps using pathscrub?

stevezau:
hmm yeah i think we can just do entry['title'] now as we enforce it to be text here https://github.com/Flexget/Flexget/blob/646d6ed86a79599ff1c9aba9107cadbccfddff77/flexget/entry.py#L189-L194

do you want to update the PR?

gazpachoking:
Maybe we should use path.py for these filename operations rather than raw text

jeroenl:
How's this?

stevezau:
hmm looks ok to me.. @gazpachoking ?

gazpachoking:
Looks fine to me.

cvium:
@jeroenl Wanna sync this so we can merge?

PR #1083, opened 2016-04-25
[Webui] series view

NoMaillard:
### what has been done

when you click on the episodes button from a show, a row gets inserted below, listing the episodes and assosiated actions. here's what it looks like : https://gyazo.com/7c3f585b654e1145aea5decc80a42004

### what needs to be done

- ~~fix the problem when the height of the list is less than the banner's height~~ (still work to be done but doesn't overlap anymore)
- ~~add a close button~~
- ~~implement the releases dialog~~
- calculate the banner's dominant color to fill the back ground, (no solution found, CORS related issue)
- ~~make it responsive ( only three columns possible yet )~~

### possible improvements

- ~~add a little triangle pointing to the show's card~~

as always, suggestions are warmly welcome


flexget/ui/src/plugins/series/components/series-show/series-show.component.js
```diff
@@ -2,83 +2,85 @@
...
-          controller: 'seriesUpdateController',
-          controllerAs: 'vm',
-          templateUrl: 'plugins/series/components/series-update/series-update.tmpl.html',
-          locals: {
-            showId: vm.show.show_id,
-            params: params
-          }
-        });
-      }
+    function loadMetadata() {
```
stevezau: is it worth creating a service for this?? maybe a series service

NoMaillard: there's already one so we could just move the call there

stevezau: yep makes sense

Tommatheussen:
Looks good at first sight, will have a closer look at it later today at home

flexget/ui/src/plugins/series/series.module.js
```diff
@@ -7,7 +7,7 @@
     seriesModule.run(function ($state, route, sideNav, toolBar) {
         route.register('series', '/series', 'series-view');
         route.register('episodes', '/series/:id/episodes', 'series-episodes-view');
```
Tommatheussen: I think this route can be removed, since there is no more routing to a different episodes page right?

flexget/ui/src/plugins/series/series.component.js
```diff
@@ -1,90 +1,120 @@
...
+        });
+    } else {
+        options.page = 1;
+        getSeriesList();
     }
+}
 
-    vm.searchTerm = "";
+vm.showEpisodes = function (show) {
+    vm.selectedShow = null;
```
Tommatheussen: I'm not sure if you planned it already, but a check to see if this show is already selected, and preventing the list from reopening (and reloading data) could be nice.

flexget/ui/src/plugins/series/series.module.js
NoMaillard: Oh yes thanks for pointing out!

flexget/ui/src/plugins/series/series.component.js
NoMaillard: I see,  easy fix

NoMaillard:
I tested the height problem with vivaldi and it works fine but on chrome canary and firefox, it's not.
i'm confused

flexget/ui/src/plugins/series/components/series-episode/series-episode.component.js
```diff
@@ -1,105 +1,79 @@
...
-          console.log(error);
-        });
-      }
+            $mdDialog.show(confirm).then(function() {
+                $http.delete('/api/series/' + vm.show.show_id + '/episodes/' + vm.episode.episode_id + '/releases', { params: { forget: true}})
+                .success(function(data) {
+                    //Remove all loaded releases from the page and set variables for the accordion
+                    vm.releases = undefined;
+                    vm.episode.episode_number_of_releases = 0;
+                    releasesOpen = false;
```
Tommatheussen: A couple of these lines can be removed I think, they where used for the old behaviour to open the releases accordion.

flexget/ui/src/plugins/series/components/series-episode/series-episode.component.js
```diff
@@ -1,105 +1,79 @@
...
 
     function seriesEpisodeController($mdDialog, $http, $stateParams, $filter){
-      var vm = this;
+        var vm = this;
 
-      //Small hacks to make the accordion work, if that ever gets in Angular Material, this can be replaced
-      var releasesOpen = false;
-      vm.toggleReleases = function() {
-        releasesOpen = !releasesOpen;
+        loadReleases();
```
Tommatheussen: I think the releases should only be loaded once a use clicks the 'show releases' button, otherwise a lot of data will be loaded which might not be needed at all.

flexget/ui/src/plugins/series/components/series-episode/series-episode.component.js
NoMaillard: oh yes that's true... will do

Tommatheussen:
In general we might have to do some cleaning up of leftover stuff, but it seems it's pretty good to merge in, @NoMaillard got any big changes coming up still?
Maybe @stevezau wants to review as well before we merge, if he has time :P

flexget/ui/src/plugins/series/components/series-episode/series-episode.component.js
NoMaillard: thing is the button is greyed out is there is no releases,
so we'd have to inform the user once they clicked on the button that there is no releases

NoMaillard:
@Tommatheussen nope just some fixing up, and there's still that height problem...

flexget/ui/src/plugins/series/components/series-episode/series-episode.component.js
Tommatheussen: Isn't the button greyed out because of the call from the episode list? I think it has an attribute episode_number_of_releases which it uses, not sure though

Tommatheussen:
I'm reviewing this right now to see if we can merge this in. Some points that could be added later on (in develop branch):

- When you click episodes again when the list is already open, might be useful if the list closes again. Not sure if this would be wanted behaviour?
- On mobile: it might be nice if instead of the text on the buttons (delete episode, delete releases, ...) we show icons, since the list gets cluttered and overlaps partially.
- Releases dialog should have a close option, other than using ESC key

Will test delete etc. actions at home on a test database. If no major issues are found there I will merge this in and open an issue as a reminder for these small points, unless you rather change them in this PR still, @NoMaillard ?

NoMaillard:
@Tommatheussen duely noted, you can merge it, the changes are minor.
I will try to work on them this weekend

Issue #1105, opened 2016-04-28
[IMDB] Movies made for TV never found by IMDB lookup.


drwyrm:
IMDB search query is leaving all "(TV Movie)" results out.

We should not set "ttype", as to allow (TV Movie) results, and filter any possible (TV Series) result.

Example movie title that is never found by imdb search: "Carny" from 2009.

Issue #1106, opened 2016-04-28
[IMDB] Geolocation makes movie titles inconsistent depending on the source IP.


drwyrm:
IMDB geolocation causes lots of issues when you are outside of USA, as the names are localized using the source IP address.

This can be "fixed" by tricking IMDB into thinking that the request is from an USA IP address. IMDB geolocation honors X-Forwarded-For http headers, so, all that is needed is setting the header to a random IP address on a block assigned to the USA.

Fixing this would make imdb lookups consistent for all users, since all searchs would be done using the same language.

liiight:
Imdb list have a `force_language` option. I started doing same thing for imdb lookup but dropped it a long time ago.

drwyrm:
From a quick test here, Accept-Language headers makes no difference regarding search results, even when using "Accept-Language: en-US" I get localized titles for most movies.

With "X-Forwarded-For: <any ipv4 from USA>" I get consistent, always english names that match releases.

PR #1108, opened 2016-04-29
Imdb search related changes

drwyrm:
The following improves imdb search results for me.


crawln45:
Please run pytests for imdb to make sure they are correct and pass. May have to delete cassettes depending on what tests call imdb search.

drwyrm:
pytest fails for the following names:

MovieParser | selected entry from imdb search page | resulting imdb id
name=Subs,year=None,quality=unknown | Subs (2007) (TV Movie) | tt0969760
name=ASDFASDFASDF,year=None,quality=unknown | Lookup failed
name=Spirited Away,year=None,quality=unknown | Spirited Away (2001) | tt0245429
name=Existence,year=2012,quality=unknown | Existence (2012) | tt2402107
name=The Matrix,year=None,quality=unknown | The Matrix (1999) | tt0133093

The only case which this commit actually changes the result is the "Subs" case, but it is intended.

flexget/utils/imdb.py
```diff
@@ -22,6 +23,7 @@
 
 # this makes most of the titles to be returned in english translation, but not all of them
 requests.headers.update({'Accept-Language': 'en-US,en;q=0.8'})
+requests.headers.update({'X-Forwarded-For': '24.110.%d.%d' % (random.randint(0, 254), random.randint(0,254))})
```
liiight: I don't know if this should be hardcoded. If IMDB decides to filter out this IP we're fucked. 
Also, some users may prefer to force the language to something different than en.

drwyrm: Current code already has harcoded header trying to enforce en-US (probably ineffective), and we would default to english anyway...

I can work on making it configurable, but if you don't mind, I'd prefer to work on a few other things before getting back to this. :)

flexget/utils/imdb.py
```diff
@@ -166,7 +168,15 @@ def search(self, name):
             movie = {}
             additional = re.findall(r'\((.*?)\)', row.text)
             if len(additional) > 0:
-                movie['year'] = additional[-1]
+                try:
+                    int(additional[-1])
```
cvium: I'm not really a fan of this; it's obscure and silly. Why not just check for type?

drwyrm:
Adding (Video) to the results found the following movies that wasn't being found before:
```
30 Days of Night: Dark Days (2010) (Video)
A Dangerous Man (2009) (Video)
Kung Fu Panda: Secrets of the Furious Five (2008) (Video)
Wrong Turn 3: Left for Dead (2009) (Video)
Wrong Turn 5: Bloodlines (2012) (Video)
Babylon 5: The Lost Tales (2007) (Video)
Wrong Turn 2: Dead End (2007) (Video)
Death Race 2 (2010) (Video)
Futurama: Into the Wild Green Yonder (2009) (Video)
Tinker Bell and the Great Fairy Rescue (2010) (Video)
Jackass 2.5 (2007) (Video)
Vacancy 2: The First Cut (2008) (Video)
Born to Race: Fast Track (2014) (Video)
Dragonheart 3: The Sorcerer's Curse (2015) (Video)
Futurama: Bender's Game (2008) (Video)
Abandoned (2010) (Video)
Behind Enemy Lines II: Axis of Evil (2006) (Video)
Stargate: Continuum (2008) (Video) aka "Stargate Continuum: The Movie"
Birdman (2014) (Video) --- This result is actually wrong, imdb title for the correct movie is "Birdman or (The Unexpected Virtue of Ignorance) (2014)"
Get Smart's Bruce and Lloyd Out of Control (2008) (Video)
Death Race: Inferno (2013) (Video) aka "Death Race 3"
Justice League: The New Frontier (2008) (Video)
Dr. Dolittle: Tail to the Chief (2008) (Video)
Beta House (2007) (Video) aka "American Pie Presents Beta House"
Open Season 2 (2008) (Video)
Batman: The Dark Knight Returns, Part 1 (2012) (Video)
```

List goes on and on... Stopped copy n pasting to reply other comment. :)

flexget/utils/imdb.py
drwyrm: Im not a fan of it either, if you know a better way to do it, please, advise. This is the google result for "python isint".

http://stackoverflow.com/questions/1265665/python-check-if-a-string-represents-an-int-without-using-try-except

Should I use .isdigit() or regex matching?

cvium: Is there any particular reason that `isinstance(some_variable, int)` does not work?

drwyrm: It's a string.
>>> isinstance('2016', int)
False

cvium: I think either `isdigit()` or regex is better. Since you want it to be a four digit year (let's not write too general code here), you can use regex to check that it is indeed four digits.

crawln45: isdigit() is also less resource intensive. since all it does is do a try int(string)

liiight:
is this ready?

drwyrm:
The current code doesn't support enforcing a language, and it's not trivial to add a config option in this case. There are tons of assumptions all around that assume titles from imdb searches(and titles in general) are in english, and properly supporting multiple languages here would need a lot more code than this bugfix is supposed to have.

liiight:
This should close #1105 and #1106 right?

PR #1114, opened 2016-04-30
Fixes #1100, proper encoding for path and content_filename based on py version.

### LINKED ISSUE #1100: Transmission 2.90+ issue
Hi to all.

Flexget version is 2.0.2.
Python version is 3.4.2.
And python3-transmissionrpc (0.11-2)

My config worked great with transmission 2.84
I have installed transmission 2.92 and now I get errors in my log
`TransmissionError: Query failed with result "download directory path is not absolute".`

In transmission's changelog I found 

> Transmission 2.90 (2016/02/28)
> All Platforms
> ...
> Require absolute paths in RPC requests
> ...

There is nothing about it in [rpc-spec](https://trac.transmissionbt.com/browser/tags/2.92/extras/rpc-spec.txt).
And nothing in [transmissionrpc 0.11 Release notes](https://bitbucket.org/blueluna/transmissionrpc/wiki/releasenotes/0.11)

But may be it's relative to [this](https://bitbucket.org/blueluna/transmissionrpc/issues/81/py3-typeerror-endswith-first-arg-must-be) and [this](https://github.com/Flexget/Flexget/issues/1066)

And there are a lot of people who complain. As [example](https://forum.transmissionbt.com/viewtopic.php?t=17884). And it seems like someone [solved the problem](https://github.com/SickRage/sickrage-issues/issues/1071)

Is there a way to fix it?

OmgOhnoes:
Fixes #1100.

There's probably a more elegant way to handle this so i'm open to ideas/comments as i'm still learning python.

Basically:

Py2 returns unicode from pathscrub which is encoded to UTF-8 and Transmission is fine with that.
Py3 returns str from pathscrub, the escaped characters do not work with setting the path (which must be absolute) OR with the content_filename.

Checking for py3 and not encoding with UTF-8 where necessary allows Cyrillic characters (and other international characters) to be passed to Transmission without error.

### DIFF
--- a/flexget/plugins/plugin_transmission.py
+++ b/flexget/plugins/plugin_transmission.py
@@ -8,7 +8,6 @@ import logging
 import base64
 import re
 import time
-import sys
 from datetime import datetime
 from datetime import timedelta
 from netrc import netrc, NetrcParseError
@@ -298,6 +297,13 @@ class PluginTransmission(TransmissionBase):
         if task.accepted:
             self.add_to_transmission(self.client, task, config)
 
+    def _decode_if_bytes(self, data):
+        """Transmission doesn't like encoded byte strings (Py3) so decode"""
+        if type(data) == bytes:
+            return data.decode('UTF-8')
+        else:
+            return data
+
     def _make_torrent_options_dict(self, config, entry):
 
         opt_dic = {}
@@ -318,11 +324,7 @@ class PluginTransmission(TransmissionBase):
         if opt_dic.get('path'):
             try:
                 path = os.path.expanduser(entry.render(opt_dic['path']))
-                # return unicode in 2.7, str in 3.5
-                if sys.version_info[0] == 3:
-                    add['download_dir'] = pathscrub(path)
-                else:
-                    add['download_dir'] = pathscrub(path).encode('UTF-8')
+                add['download_dir'] = self._decode_if_bytes((pathscrub(path).encode('UTF-8')))
             except RenderError as e:
                 log.error('Error setting path for %s: %s' % (entry['title'], e))
         if 'bandwidthpriority' in opt_dic:
@@ -521,10 +523,10 @@ class PluginTransmission(TransmissionBase):
                             if 'download_dir' not in options['add']:
                                 download_dir = cli.get_session().download_dir
                             else:
-                                if sys.version_info[0] == 3:
-                                    download_dir = options['add']['download_dir']
-                                else:
-                                    download_dir = options['add']['download_dir'].decode('UTF-8')                                
+                                download_dir = options['add']['download_dir']
+                                """If this isn't a Py3 str it needs to be decoded"""
+                                if not isinstance(download_dir, str):
+                                    download_dir = download_dir.decode('UTF-8')
 
                             # Get new filename without ext
                             file_ext = os.path.splitext(fl[r.id][main_id]['name'])[1]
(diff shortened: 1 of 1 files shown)


OmgOhnoes:
If there are no other suggestions on how to approach this should i clean up the code so it can be merged?

cvium:
I'm not a fan of checking for Python versions. @gazpachoking or @stevezau should know what to do.

OmgOhnoes:
Yeah, i'm not a fan either. I just couldn't figure out a way to standardize it. It could alternatively check for type returned by pathscrub i think.

crawln45:
check to see if adding native(path) changes anything. I don't use transmission so I'm not equipped to actually test it.

stevezau:
Yeah I think there is a more elegant solution rather then checking py versions..

If i understand correctly py2 pathscrub is fine but py3 pathscrub escapes characters? Do you have an example i can work with?

Bop4yN:
Hi. Sorry for the late reply.
Now I've updated to flexget-2.0.9.
Testing on transmission-daemon 2.84.

I have the same error. Torrent is added to transmission, but path looks like bytes.

And here is one more thing. I'am executing my own script to remove old episodes from transmission.
On python 2.7 transmission responded something like
`INFO     exec          BaibaKo         Stdout: localhost:9091/transmission/rpc/ responded: "success"\nТоррент "Тёмное дитя.s04e01.XviD.avi" с ID 116 удалён.\n
`
But now
`INFO     exec          BaibaKo         Stdout: b'localhost:9091/transmission/rpc/ responded: "success"\n\xd0\xa2\xd0\xbe\xd1\x80\xd1\x80\xd0\xb5\xd0\xbd\xd1\x82 "\xd0\xa1\xd0\xb5\xd0\xbc\xd1\x8c\xd1\x8f.s01e05.XviD.\xd0\x9c\xd0\xbe\xd1\x8f \xd0\xb2\xd0\xb7\xd1\x8f\xd0\xbb\xd0\xb0.avi" \xd1\x81 ID 78 \xd1\x83\xd0\xb4\xd0\xb0\xd0\xbb\xd1\x91\xd0\xbd.\n'
`

Is there any info I can provide?

OmgOhnoes:
I've figured out a way to do it by checking data type instead of python version. I'll update the PR tomorrow.

Bop4yN:
**OmgOhnoes**, thank you

flexget/plugins/plugin_transmission.py
```diff
@@ -298,6 +297,13 @@ def on_task_output(self, task, config):
         if task.accepted:
             self.add_to_transmission(self.client, task, config)
 
+    def _decode_if_bytes(self, data):
+        """Transmission doesn't like encoded byte strings (Py3) so decode"""
+        if type(data) == bytes:
```
cvium: Isn't `isinstance` the preferred way of checking for types since it properly supports inheritance?

OmgOhnoes: Basically i didn't have any other data to go on and i know it's an issue with this specific type in Py3. I didn't want to introduce other errors if things should change in the future with bytes.

I used isinstance later as i need to check for Py3 without checking for a Python version and it doesn't matter if it's descendants match.

If i'm wrong in my thinking just let me know and i'll change it!

stevezau:
I'll take a look at this soon.. just been flat out.

There are utils in python future that should be able to make this seamless. http://python-future.org/compatible_idioms.html

Bop4yN:
I really don't know is last commit merged to v2.0.11.
But nothing changed for me.
transmission-daemon 2.84

Everything is ok in log
`2016-05-05 11:35 INFO     download      BaibaKo         Downloading: Скорпион /Scorpion /s02e24 /WEB-DLRip XviD /Тоби или не Тоби (Финал сезона)`
`2016-05-05 11:35 INFO     transmission  BaibaKo         "Скорпион /Scorpion /s02e24 /WEB-DLRip XviD /Тоби или не Тоби (Финал сезона)" torrent added to transmission`

But there is an error in Transmission Remote GUI
`Permission denied (b'/mnt/Video/01 Serials/\xd0\xa1\xd0\xba\xd0\xbe\xd1\x80\xd0\xbf\xd0\xb8\xd0\xbe\xd0\xbd/\xd0\xa1\xd0\xba\xd`
And path is
`b'/mnt/Video/01 Serials/\xd0\xa1\xd0\xba\xd0\xbe\xd1\x80\xd0\xbf\xd0\xb8\xd0\xbe\xd0\xbd/\xd0\xa1\xd0\xba\xd0\xbe\xd1\x80\xd0\xbf\xd0\xb8\xd0\xbe\xd0\xbd - \xd0\xa1\xd0\xb5\xd0\xb7\xd0\xbe\xd0\xbd 02'/Скорпион.s02e24.XviD.Тоби или не Тоби (Финал сезона).avi`

And look at the path closer. Is it supposed to be " **b'**/my/path**'** "?
I guess there should be no single quotes and "b" at start.

crawln45:
@Bop4yN nothing has been merged yet. Reason your seeing bytes is because we encoded a string into utf-8 which turns it into a bytestring. Because we need to encode in py2.7 to convert that unicode it receives into a utf-8 string. it get's kinda complicated at this point. hehe.

PR #1121, opened 2016-05-01
Make `list_queue` operate correctly

liiight:
This fixes `list_queue` behavior by moving the list deletion part to `on_task_learn` phase.
Added `get` method for all list plugins that return matched entry or None.
Tested on all list plugins.


flexget/plugins/list/couchpotato_list.py
```diff
@@ -181,10 +181,10 @@ def movies(self):
 
     def _find_entry(self, entry):
         for cp_entry in self.movies:
-            if any(entry.get(id) is not None and entry[id] == cp_entry[id] for id in self.supported_ids):
-                return True
-            if entry.get('title').lower() == cp_entry.get('title').lower():
-                return True
+            for id in self.supported_ids:
```
stevezau: Using id will raise pep8 errors. Maybe use entry_id?

PR #1134, opened 2016-05-04
Update newpct.py

Lukeid:
For newpct and newpct1 urls.

### DIFF
--- a/flexget/plugins/urlrewrite/newpct.py
+++ b/flexget/plugins/urlrewrite/newpct.py
@@ -24,7 +24,7 @@ class UrlRewriteNewPCT(object):
     def url_rewritable(self, task, entry):
         url = entry['url']
         rewritable_regex = '^http:\/\/(www.)?newpct1?.com\/.*'
-        return re.match(rewritable_regex, url) and not url.startswith('http://www.newpct1.com/download/') and not url.startswith('http://www.newpct.com/torrents/')
+        return re.match(rewritable_regex, url) and not url.endswith('.torrent')
 
     # urlrewriter API
     def url_rewrite(self, task, entry):
@@ -38,7 +38,7 @@ class UrlRewriteNewPCT(object):
             return self.parse_newpct(url)
 
     def parse_newpct(self, url):
-        log.verbose('Newpct URL: %s' % url)
+        log.verbose('Newpct URL: %s', url)
         page = requests.get(url)
         try:
             soup = get_soup(page.text)
@@ -49,10 +49,10 @@ class UrlRewriteNewPCT(object):
         if len(torrent_ids) == 0:
             raise UrlRewritingError('Unable to locate torrent ID from url %s' % url)
         torrent_id = torrent_id_prog.search(torrent_ids[0]).group(1)
-        return 'http://www.newpct.com/torrents/%s.torrent' % ('{:0>6}'.format(torrent_id))
+        return 'http://www.newpct.com/torrents/{:0>6}.torrent'.format(torrent_id)
 
     def parse_newpct1(self, url):
-        log.verbose('Newpct1 URL: %s' % url)
+        log.verbose('Newpct1 URL: %s', url)
         url = url.replace('newpct1.com/', 'newpct1.com/descarga-torrent/')
         page = requests.get(url)
         try:
--- a/flexget/plugins/urlrewrite/newpct.py
+++ b/flexget/plugins/urlrewrite/newpct.py
@@ -24,7 +24,7 @@ class UrlRewriteNewPCT(object):
     def url_rewritable(self, task, entry):
         url = entry['url']
         rewritable_regex = '^http:\/\/(www.)?newpct1?.com\/.*'
-        return re.match(rewritable_regex, url) and not url.startswith('http://www.newpct.com/descargar/')
+        return re.match(rewritable_regex, url) and not url.startswith('http://www.newpct1.com/download/') and not url.startswith('http://www.newpct.com/torrents/')
 
     # urlrewriter API
     def url_rewrite(self, task, entry):
@@ -32,6 +32,13 @@ class UrlRewriteNewPCT(object):
 
     @plugin.internet(log)
     def parse_download_page(self, url):
+        if 'newpct1' in url:
+            return self.parse_newpct1(url)
+        else:
+            return self.parse_newpct(url)
+
+    def parse_newpct(self, url):
+        log.verbose('Newpct URL: %s' % url)
         page = requests.get(url)
         try:
             soup = get_soup(page.text)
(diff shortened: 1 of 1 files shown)


flexget/plugins/urlrewrite/newpct.py
```diff
@@ -42,8 +49,22 @@ def parse_download_page(self, url):
         if len(torrent_ids) == 0:
             raise UrlRewritingError('Unable to locate torrent ID from url %s' % url)
         torrent_id = torrent_id_prog.search(torrent_ids[0]).group(1)
-        return 'http://www.newpct.com/descargar/torrent/%s/dummy.html' % torrent_id
+        return 'http://www.newpct.com/torrents/%s.torrent' % ('{:0>6}'.format(torrent_id))
```
cvium: Couldn't this be rewritten to `'http://www.newpct.com/torrents/{:0>6}.torrent'.format(torrent_id)`?

flexget/plugins/urlrewrite/newpct.py
```diff
@@ -24,14 +24,21 @@ class UrlRewriteNewPCT(object):
...
 
     @plugin.internet(log)
     def parse_download_page(self, url):
+        if 'newpct1' in url:
+            return self.parse_newpct1(url)
+        else:
+            return self.parse_newpct(url)
+
+    def parse_newpct(self, url):
+        log.verbose('Newpct URL: %s' % url)
```
cvium: `log` is capable of string formatting, so stuff like this should be written as `log.verbose('Newpct URL: %s', url)`

flexget/plugins/urlrewrite/newpct.py
```diff
@@ -24,14 +24,21 @@ class UrlRewriteNewPCT(object):
     def url_rewritable(self, task, entry):
         url = entry['url']
         rewritable_regex = '^http:\/\/(www.)?newpct1?.com\/.*'
-        return re.match(rewritable_regex, url) and not url.startswith('http://www.newpct.com/descargar/')
+        return re.match(rewritable_regex, url) and not url.startswith('http://www.newpct1.com/download/') and not url.startswith('http://www.newpct.com/torrents/')
```
cvium: Lines may not be longer than 120 characters.

flexget/plugins/urlrewrite/newpct.py
Lukeid: Yes, It could. I have just tested it and it works.

flexget/plugins/urlrewrite/newpct.py
Lukeid: Ok, this one and also the same one in parse_newpct1 method.

flexget/plugins/urlrewrite/newpct.py
```diff
@@ -24,14 +24,21 @@ class UrlRewriteNewPCT(object):
...
+        return re.match(rewritable_regex, url) and not url.endswith('.torrent')
 
     # urlrewriter API
     def url_rewrite(self, task, entry):
         entry['url'] = self.parse_download_page(entry['url'])
 
     @plugin.internet(log)
     def parse_download_page(self, url):
+        if 'newpct1' in url:
+            return self.parse_newpct1(url)
```
liiight: Looks like a lot of code duplication between `parse_newpct1` and `parse_newpct`. Are both needed?

flexget/plugins/urlrewrite/newpct.py
```diff
@@ -42,8 +49,22 @@ def parse_download_page(self, url):
...
-        return 'http://www.newpct.com/descargar/torrent/%s/dummy.html' % torrent_id
+        return 'http://www.newpct.com/torrents/{:0>6}.torrent'.format(torrent_id)
 
+    def parse_newpct1(self, url):
+        log.verbose('Newpct1 URL: %s', url)
+        url = url.replace('newpct1.com/', 'newpct1.com/descarga-torrent/')
+        page = requests.get(url)
+        try:
+            soup = get_soup(page.text)
+        except Exception as e:
```
liiight: Catching a broad Exception like that is very bad practice

flexget/plugins/urlrewrite/newpct.py
Lukeid: I know, both do basically the same. But regex are different, soup.findAll is also little different also just newpct1 need url replacement.
So I though I would be more clear keep them separated.

flexget/plugins/urlrewrite/newpct.py
Lukeid: I this case I just kept it as it was in the original newpct.py.

liiight: That's no excuse :-)

flexget/plugins/urlrewrite/newpct.py
liiight: i don't think it's best to have to different code paths like that

Lukeid: So, what shall we do?

liiight: write better code?

Lukeid: I'm not a Python expert (I just read some high level points to be able to modify current code) so I'm not sure what to do exactly to reach your code quality expectations. (But I can try if you give some tips :))
All I know by now is that it works... and original newpct rewriter doesn't. So it may help others as it is now. ;-)

liiight: try to make it into one function, see what conditions have to be made and decide on the code flow.

Lukeid:
Not really... I've checked all urlrewriters to see if any of them caugth more specific exceptions, but they do it the same way I did. (Well, I didn't do it I just kept it as it was... ;-))

cvium:
Most of the urlrewriters are old and newer ones just copy the old ones. While this is great for getting stuff out fast it is a pitfall of technical debt (yay, buzzwords).

Catching broad exceptions is symptomatic of a bad/lazy programmer.

Lukeid:
I understand and I really would like to do it better but right now I'm not able to do it other way.

Besides maybe would be a better idea that someone with more expertise working with BeautifulSoup caught all kind of exception inside the get_soup method so all urlrewriters get benefit of it.

flexget/plugins/urlrewrite/newpct.py
```diff
@@ -24,26 +24,42 @@ class UrlRewriteNewPCT(object):
...
 
     @plugin.internet(log)
     def parse_download_page(self, url):
+        if 'newpct1' in url:
+            log.verbose('Newpct1 URL: %s', url)
+            url = url.replace('newpct1.com/', 'newpct1.com/descarga-torrent/')
+        else:
+            log.verbose('Newpct URL: %s', url)
+
         page = requests.get(url)
```
cvium: This should probably be wrapped in a try-except.

flexget/plugins/urlrewrite/newpct.py
```diff
@@ -24,26 +24,42 @@ class UrlRewriteNewPCT(object):
...
         try:
             soup = get_soup(page.text)
         except Exception as e:
             raise UrlRewritingError(e)
```
cvium: It's probably fine in this instance since it is reraised as something else. I don't like it, but the only other exception I can find is `from html5lib.html5parser import ParseError`.

flexget/plugins/urlrewrite/newpct.py
Lukeid: It should never fail because first I check the string "newpct1" so it will always contain "newpct1.com" but if you want I can change it to "if 'newpct1.com' in url:"

cvium: I'm talking about `requests.get(url)`. It can raise an exception if it fails.

Lukeid: Ok, I added a try to that sentence.
Is it better now? Or would be better to unify the two try/except in one block with the get_soup?

PR #1152, opened 2016-05-08
New subtitle_list plugin, replaces subtitle_queue

cvium:
Torrent support removed (maintaining it is not very nice). Otherwise it should pretty much be the same.


flexget/plugins/list/subtitle_list.py
```diff
@@ -0,0 +1,261 @@
...
+    def get_list(config):
+        return SubtitleList(config)
+
+    def on_task_input(self, task, config):
+        subtitle_list = SubtitleList(config)
+        if config['check_subtitles']:
+            for file in subtitle_list:
+                if not os.path.isfile(file['location']):
+                    log.error('File %s does not exist. Removing from list.', file['location'])
+                    subtitle_list -= [file]
```
liiight: you can do `subtitle_list.remove(file)`, or `subtitle_list.discard(file)`, depending if you want to raise a KeyError on fail. 
Should you consider if task is run on test mode before manipulating list?

flexget/plugins/list/subtitle_list.py
```diff
@@ -0,0 +1,261 @@
...
+class SubtitleList(MutableSet):
+    schema = {
+        'type': 'object',
+        'properties': {
+            'list': {'type': 'string'},
+            'languages': {'type': 'array', 'items': {'type': 'string'}, 'minItems': 1},
+            'check_subtitles': {'type': 'boolean', 'default': True},
+            'remove_after': {'type': 'string', 'format': 'interval'}
+        },
+        'required': ['list'],
```
liiight: if `list` is the only required property, wouldn't it be more friendly to have the schema use `oneOf` string or object, so you could call the list with `subtitle_list: list_name`.
just a suggestion

flexget/plugins/list/subtitle_list.py
```diff
@@ -0,0 +1,261 @@
...
+            log.debug('deleting file %s', db_file)
+            session.delete(db_file)
+
+    def __contains__(self, entry):
+        return self._find_entry(entry) is not None
+
+    @with_session
+    def _find_entry(self, entry, session=None):
+        """Finds `SubtitleListFile` corresponding to this entry, if it exists."""
+        res = self._db_list(session).files.filter(SubtitleListFile.location == entry.get('location', '')).first()
```
liiight: does this work? shouldn't you need to join `SubtitleListFile` to `self._db_list` query?

paranoidi:
Do not have real release names in test or in filenames. Use bogus names. Rebase and get rid of them. They should not exist even in a history.

cvium:
@paranoidi I just did the same thing as the tvmaze tests (which are currently in master).

flexget/plugins/list/subtitle_list.py
gazpachoking: I think it's fine `SubtitleListList.files` is a dynamic relationship, which will be a query on `SubtitleListFile`s that are children of the `SubtitleListList`

PR #1155, opened 2016-05-09
FTP list interface plugin

liiight:
This plugin will replace the current (completely broken) `ftp_list` plugin. It utilizes the [list interface](http://flexget.com/wiki/Plugins/List).

Waiting for more feedback before finishing this since @paranoidi does not like using list interface for this usage.

To Do:

- [ ] Symlink support
- [ ] Regex support
- [ ] Make FTPS work
- [ ] Tests
- [ ] Direction?


paranoidi:
Discussion needs to happen before going down this road.

References: http://pastie.org/private/q8ihah3tgdrqs8s7ei7zoa

liiight:
Closed due to a lack of direction

PR #1156, opened 2016-05-09
New subtitle_list plugin, replaces subtitle_queue

cvium:
Now with bogus file names.

TODO:
- [x] @liiight's comments from previous PR


liiight:
This is ready, right?

PR #1158, opened 2016-05-09
Faulty variable refactor in sceneaccess plugin

qvazzler:
I cannot for the life of me get rid of the initial merge commit for clean-ness, please forgive me. :-)

Plugin is currently broken until this is fixed.


flexget/plugins/search/sceneaccess.py
```diff
@@ -211,11 +211,11 @@ def processCategories(self, config):
 
             for category in categories:
                 try:
-                    cat_id = CATEGORIES[scope][category]
+                    id = CATEGORIES[scope][category]
                 except KeyError:            # User provided category id directly
-                    cat_id = category
+                    id = category
```
cvium: `id` is a builtin Python function. Please don't overwrite it.

liiight:
What does this change even supposed to do?

qvazzler:
@liiight @cvium  I think there's a misconception that id and cat_id is the same variable where it is not. Actually cat_id is the array, and id is just an int. The code, at its current state, throws exception when trying to use .append to cat_id which is assigned to be an int.

I had no idea that id was a built in variable though, maybe a different variable name would be better.

PR #1161, opened 2016-05-10
FTP input plugin (replace existing one)

liiight:
A new FTP/S input plugin based on the `ftputil` lib.
Replaces the current `ftp_list` plugin.

To Do:

- [x] Verify over FTP/S 
- ~~Tests~~


crawln45:
@liiight https://github.com/Flexget/Flexget/blob/develop/flexget/plugins/input/ftp_list.py#L7 shouldn't that be inside the try block?

liiight:
Yeah, i thought I pushed that in develop...

PR #1169, opened 2016-05-14
Filter plugin to detect valid archives

ianstalk:
Add a filter plugin that accepts valid Zip/RAR archives, and rejects non-archive files. I also had to do a bit of refactoring to support these changes.


cvium:
The archive filter should be in its own file in `flexget/plugins/filter`. It doesn't have anything to do with decompress other than it also uses `rarfile`.

ianstalk:
Makes sense. Do you have a good suggestion for doing that without a whole
lot of code repetition?

flexget/plugins/output/decompress.py
```diff
@@ -18,6 +18,124 @@
...
+            else:
+                rarfile.UNRAR_TOOL = unrar_tool
+                log.debug('Set RarFile.unrar_tool to: %s', unrar_tool)
+
+def open_archive(entry):
+    """
+    Returns the appropriate archive object
+    """
+
+    archive_path = entry['location']
```
cvium: This will crash if the entry doesn't have `location` set.

cvium:
You can put the is_archive etc. in a utility file in `flexget/utils/` and then import them.

ianstalk:
Any feedback on my latest commit?

flexget/plugins/filter/archives.py
```diff
@@ -0,0 +1,73 @@
...
+        if isinstance(config, bool) and not config:
+            return
+
+        config = self.prepare_config(config)
+        archive.rarfile_set_tool_path(config)
+
+        for entry in task.entries:
+            archive_path = entry.get('location')
+
+            if archive.is_archive(archive_path):
```
cvium: What happens if `archive_path = None`?

flexget/plugins/output/decompress.py
```diff
@@ -4,70 +4,53 @@
...
+    try:
+        archive_path = entry['location']
+        arch = archive.open_archive(archive_path)
+    except KeyError:
+        log.error('Entry does not appear to represent a local file.')
+    except archive.BadArchive as error:
+        fail_entry_with_error(entry, 'Bad archive: %s (%s)' % (archive_path, error))
+    except archive.NeedFirstVolume:
+        log.error('Not the first volume: %s', archive_path)
+    except Exception as error:
```
cvium: Is catching such a broad exception really necessary?

flexget/plugins/output/decompress.py
```diff
@@ -80,65 +63,14 @@ def is_dir(info):
...
-                entry.reject()
 
+def makepath(path):
+    if not os.path.exists(path):
+        log.debug('Creating path: %s', path)
+        os.makedirs(path)
 
 class Decompress(object):
-    """
+    r"""
```
cvium: What's the point of this?

flexget/utils/archive.py
```diff
@@ -0,0 +1,186 @@
...
+    """
+
+    archive = None
+    ret = False
+
+    try:
+        archive = open_archive(path)
+        if archive:
+            archive.close()
+            ret = True
```
cvium: Might as well just use `return True`. You're using 3 lines to do what 2 lines could.

flexget/utils/archive.py
```diff
@@ -0,0 +1,186 @@
...
+        self.archive.close()
+
+    def delete(self):
+        """Delete the volumes that make up this archive"""
+        volumes = self.volumes()
+        self.close()
+
+        for volume in volumes:
+            log.debug('Deleting volume: %s', volume)
+            os.remove(volume)
```
cvium: Shouldn't you catch an OSError here?

liiight:
Tests would be nice too

flexget/utils/archive.py
ianstalk: What should I do instead, add another exception to raise in its stead? Obviously I still need to let the calling code know that something went wrong.

ianstalk:
Hopefully that latest update addresses your concerns.

liiight:
Looks good to me :thumbsup:

flexget/plugins/filter/archives.py
```diff
@@ -61,7 +61,7 @@ def on_task_filter(self, task, config):
         archive.rarfile_set_tool_path(config)
 
         for entry in task.entries:
-            archive_path = entry.get('location')
+            archive_path = entry.get('location') or ''
```
cvium: You can specify a default value in dict.get ie. `entry.get('location', '')`.

flexget/plugins/filter/archives.py
```diff
@@ -0,0 +1,73 @@
...
+        Task handler for archives
+        """
+        if isinstance(config, bool) and not config:
+            return
+
+        config = self.prepare_config(config)
+        archive.rarfile_set_tool_path(config)
+
+        for entry in task.entries:
+            archive_path = entry.get('location') or ''
```
cvium: Same here.

PR #1183, opened 2016-05-15
add queue position to transmission

cvium:
Untested


Addicted5:
I tested it & it Worked as intended.
Files are being added to queue top if i set "queue_position: 0"

PR #1190, opened 2016-05-18
IRC bot plugin (automatically download IRC announcements)

cvium:
### Current features
 - IRCConnectionManager continually polls the threads to check for aliveness. Restarts them if necessary.
 - IRCConnection continually sends out PING messages to the server(s) to make sure the connection doesn't die.
 - Supports all the trackers in autodl-trackers and automatically downloading them.
 - irc_bot is my own small IRC client implementation based on Python's asyncore. Once Python 2.7 support is dropped it should use Python's asyncio instead. For compatibility it's not possible right now though.

TODO
- [x] ~~Tests?~~ Tests have been scrapped as it is too complex to do reliably. Would have to spin up a small IRC server instance or connect to Freenode, then have another user/bot connect and send announcements. The plugin itself has been tested rigorously in a real world setting.


requirements.txt
```diff
@@ -30,4 +30,5 @@ flask-login>=0.3.2
 flask-cors>=2.1.2
 pyparsing>=2.0.3
 Safe>=0.4
-future>=0.15.2
\ No newline at end of file
+future>=0.15.2
+irc
```
liiight: I dunno if this should be a requirement, feels more like it should be optional

cvium: I feel the same way about flask.

liiight: what does that have to do with anything?

PR #1193, opened 2016-05-18
Fix #1178

### LINKED ISSUE #1178: BUG: t411_api invalid get requests with terms
When searching using terms the GET request sent to the server is sending invalid characters when chaining those terms. The result is those terms are not being filtered by t411 search server

For example, in this pcap file you can see the request and the response 
[t411_api_flexget.zip](https://github.com/Flexget/Flexget/files/264770/t411_api_flexget.zip)

The real request should be: 
/torrents/search/NCIS?cid=433&term%5B46%5D%5B%5D%27=949&limit=100&offset=0&term%5B45%5D%5B%5D%27=980&term%5B51%5D%5B%5D%27=1210&term%5B51%5D%5B%5D%27=1212&term%5B51%5D%5B%5D%27=1213&term%5B51%5D%5B%5D%27=1214

and it should give out 3 results only

{"query":"NCIS","offset":"0","limit":"100","total":"3","torrents":[{"id":"5505701","name":"NCIS.S13E12.FRENCH.HDTV.x264-FRiES","category":"433","rewritename":"ncis-s13e12-french-hdtv-x264-fries","seeders":"136","leechers":"1","comments":"1","isVerified":"1","added":"2016-05-05 23:02:43","size":"253477370","times_completed":"422","owner":"8039783","categoryname":"S\u00e9rie TV","categoryimage":"video-tv-series","username":"superbus49","privacy":"normal"},{"id":"5505743","name":"NCIS.S13E12.FRENCH.720p.HDTV.x264-FRiES","category":"433","rewritename":"ncis-s13e12-french-720p-hdtv-x264-fries","seeders":"169","leechers":"1","comments":"2","isVerified":"1","added":"2016-05-06 01:00:35","size":"923681846","times_completed":"459","owner":"8039783","categoryname":"S\u00e9rie TV","categoryimage":"video-tv-series","username":"superbus49","privacy":"normal"},{"id":"5505863","name":"NCIS.S13E12.FRENCH.HDTV.XviD-ZT","category":"433","rewritename":"ncis-s13e12-french-hdtv-xvid-zt","seeders":"219","leechers":"0","comments":"1","isVerified":"1","added":"2016-05-06 09:36:11","size":"326451286","times_completed":"668","owner":"99613071","categoryname":"S\u00e9rie TV","categoryimage":"video-tv-series","username":"Seuph","privacy":"normal"}]}

This is my config file
http://pastie.org/10809799

kevmegforest:


cvium:
http://python-future.org/compatible_idioms.html#unicode-text-string-literals

liiight:
Heh, beat me to it

kevmegforest:
It's fixed according to python-future.
Thanks a lot for the info.
I was searching for it on stack overflow and the only simple thing i found was the six package.

liiight:
I think this is the one you need http://python-future.org/compatible_idioms.html#unicode
 `from builtins import str`

flexget/plugins/api_t411.py
```diff
@@ -311,7 +311,7 @@ def search(self, query):
             url_params['offset'] = query['page_index']
         if query.get('terms') is not None:
             for (term_type_id, term_id) in query['terms']:
-                term_type_key_param = 'term[' + str(term_type_id) + '][]'
+                term_type_key_param = u"term[%s][]" % str(term_type_id)
```
gazpachoking: `'term[%s][]' % term_type_id` should do the same thing, and looks a bit nicer.

gazpachoking:
> I think this is the one you need http://python-future.org/compatible_idioms.html#unicode
> from builtins import str

`str` already refers to future's str via the catchall import. https://github.com/kevmegforest/Flexget/blob/b982776793abde2fe3225a043829d757cfaa8814/flexget/plugins/api_t411.py#L2

We can avoid that entirely with my code reply though.

flexget/plugins/api_t411.py
kevmegforest: if you check the issue you would see that it absolutely needs to be a unicode string. (or everything else needs to be binary string.

As for the str(). I'm not a python expert. But the referenced site said that it needs the str().

@liiight at the top of the file theres `from builtins import * `. So I don't think it's necessary

kevmegforest:
@gazpachoking I just tried and it keeps the unicode all the way and there's no error too.
I've pushed your changes

gazpachoking:
> @gazpachoking I just tried and it keeps the unicode all the way and there's no error too.
> I've pushed your changes

Cool. The explicit `u` prefix is actually unnecessary too, by virtue of us having unicode literals enabled. https://github.com/kevmegforest/Flexget/blob/34cd5bd0244b875210096301f2a795bcd8f592f1/flexget/plugins/api_t411.py#L1

kevmegforest:
@gazpachoking But why didn't it work before with the concatenation with + ?

gazpachoking:
Cool. Just trusting you that this fixes something, but the code looks pretty now. Thanks! :smile:

gazpachoking:
> But why didn't it work before with the concatenation with + ?

I'm not too sure... that is confusing. Maybe because the thing became a `newstr` from future rather than a native `unicode` string on python 2. Some libraries choke when they get a unicodelike string rather than an explicit builtin python one.

PR #1194, opened 2016-05-19
Imdb search bugfixes

drwyrm:
1) Since we now allow all titles (video games, tv movies, shorts, etc..) we may get lots of pollution on the search results page, allow more than 10 results to be processed
2) We should continue processing after skipping a result, fix it


flexget/utils/imdb.py
```diff
@@ -62,7 +62,7 @@ def __init__(self):
         self.min_diff = 0.01
         self.debug = False
 
-        self.max_results = 10
+        self.max_results = 50
```
liiight: i dunno about making it 50, it'll make the lookup way longer and heavier. maybe a lower default?
bonus points for makings this configurable :stuck_out_tongue:

drwyrm: The test i added is just number 13, not sure if any other movies in my collection get ranked lower in the search results page, lower it to 20?

About making it configurable... Well.. same problem as the geolocation nuke... I don't know how to add a config option for utils/imdb.py, it's not a plugin. If you give me instructions on how it should be done, sure, i will add it and also make the geolocation ip configurable while at it.

liiight: you can add the option to `imdb_lookup`. that can use [partial](https://docs.python.org/2/library/functools.html#functools.partial) when it calls `imdb.py` operations.

PR #1199, opened 2016-05-20
Pogcal

pilluli:
Since a couple of days access to pogdesign is failing due to changes in the web layout. This small push request seems to fix it in my machine.

### DIFF
--- a/flexget/plugins/input/pogcal.py
+++ b/flexget/plugins/input/pogcal.py
@@ -31,29 +31,38 @@ class InputPogDesign(object):
         session = requests.Session()
         data = {'username': config['username'], 'password': config['password'], 'sub_login': 'Account Login'}
         try:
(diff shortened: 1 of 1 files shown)


lepinkainen:
This seems to break stuff subtly:
`
DC's Legends of Tomorrow S01 720p HDTV x264-DIMENSION looks like series \n \n \n DC's Legends of Tomorrow\n \n \n \n but cannot find a(n) ep style identifier.
`

Note the extraneous spaces and newlines in the series name.

cvium:
Adding `.replace('\n', '').strip()` should fix that,

lepinkainen:
Yep, not a big issue.

The weird thing is that FlexGet internals don't do that already, as there is exactly zero need for a show name to have leading or following spaces and newlines.

PR #1206, opened 2016-05-24
Webui testing

Tommatheussen:
### Motivation for changes:

Testing the webui will prevent the missing of small bugs after changing some parts.

### Detailed changes:

- [ ] Check Provider Injection (possible rewrite of the router to take advantage of bardjs's routing mocking ability)

- [ ] Create tests for all plugins and components

- [ ] Create coverage report, try to integrate with Jenkins and Codacy?

- [ ] Create build process so we can run tests/visualise results in a browser? (modify example config from [John Papa's Hottowel,](https://github.com/johnpapa/generator-hottowel) which has this function, for our needs and structure)

- [ ] Maybe add JSHinting to the process as well, clean up some code while I'm at it.

- [ ] Create mock data files, which can be used in the tests and updated when needed so we don't have to look through all the test files.

### Addressed issues:

No related issues, but will hopefully result in less issues in the future

Currently only 2 simple tests present, both passing. Will update with the amount of tests present as I go along the way.

Currently tests can only be run using the `karma start` command in the flexget\ui folder.


stevezau:
hey @Tommatheussen hows this going??

Tommatheussen:
It's almost done, I had to rewrite most of the stuff to be testable ;) Right now only the log and execute plugin are left to do, I hope they'll be ready to merge in next week

stevezau:
ok cool, looking fw to taking a proper look :D

PR #1226, opened 2016-06-06
[wip] Series list

liiight:
### Motivation for changes:
A manged list for series, enabling using `series` plugin without the need to explicitly edit config file or use 3rd party service.
This will also allows using multiple lists for multiple purposes.

Note: This does not affect series tracking in any way, this just represents an input to which `configure_series` plugin can be used with (among others things).
### Detailed changes:

- New managed list type, `series_list`
- Manageable via API and CLI
- Had to add a small change to JSON decoder to support trying to decode None values.
- Added a global supported series IDs list to be shared between `configure_series` and `series_list` (and any other plugin that this can relate too, need to look into other `series` related plugins).

```
tasks:
  series_with_list:
    configure_series:
      from:
        series_list: main_shows
    rss: http://url.com/links
    download: /path/to/downloads
```
Tests included.

#### To Do:
- [ ] CLI
- [ ] API and tests

### DIFF
--- a/flexget/plugins/api/series_list.py
+++ b/flexget/plugins/api/series_list.py
@@ -45,6 +45,8 @@ class objects_container(object):
 
     return_lists = {'type': 'array', 'items': list_object}
 
+
+
     return_series = {
         'type': 'object',
         'properties': {
--- a/flexget/plugins/api/series_list.py
+++ b/flexget/plugins/api/series_list.py
@@ -45,9 +45,79 @@ class objects_container(object):
 
     return_lists = {'type': 'array', 'items': list_object}
 
+    input_series_list_id_object = {
+        'type': 'array',
+        'items': {
+            'type': 'object',
+            'minProperties': 1,
+            'additionalProperties': True
+        }
+    }
 
+    return_series_list_id_object = copy.deepcopy(input_series_list_id_object)
+    return_series_list_id_object.update(
+        {'properties': {
+            'id': {'type': 'integer'},
+            'added_on': {'type': 'string'},
+            'series_id': {'type': 'integer'}
+        }})
 
-    return_series = {
+    input_series_object = {
+        'type': 'object',
+        'properties': {
+            'title': {'type': 'string'},
+            'path': {'type': 'string'},
+            'alternate_name': {'type': 'array', 'items': {'type': 'string', 'format': 'regex'}},
+            'name_regexp': {'type': 'array', 'items': {'type': 'string', 'format': 'regex'}},
+            'ep_regexp': {'type': 'array', 'items': {'type': 'string', 'format': 'regex'}},
+            'date_regexp': {'type': 'array', 'items': {'type': 'string', 'format': 'regex'}},
+            'sequence_regexp': {'type': 'array', 'items': {'type': 'string', 'format': 'regex'}},
+            'id_regexp': {'type': 'array', 'items': {'type': 'string', 'format': 'regex'}},
+            'date_yearfirst': {'type': 'boolean'},
+            'date_dayfirst': {'type': 'boolean'},
+            'quality': {'type': 'string', 'format': 'quality_requirements'},
+            'qualities': {'type': 'array', 'items': {'type': 'string', 'format': 'quality_requirements'}},
+            'timeframe': {'type': 'string', 'format': 'interval'},
+            'upgrade': {'type': 'boolean'},
+            'target': {'type': 'string', 'format': 'quality_requirements'},
+            # Specials
+            'specials': {'type': 'boolean'},
+            # Propers (can be boolean, or an interval string)
+            'propers': {'type': ['boolean', 'string'], 'format': 'interval'},
+            # Identified by
+            'identified_by': {
+                'type': 'string', 'enum': ['ep', 'date', 'sequence', 'id', 'auto']
+            },
+            # Strict naming
+            'exact': {'type': 'boolean'},
+            # Begin takes an ep, sequence or date identifier
+            'begin': {
+                'oneOf': [
+                    {'name': 'ep identifier', 'type': 'string', 'pattern': r'(?i)^S\d{2,4}E\d{2,3}$',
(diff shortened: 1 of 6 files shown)


flexget/plugins/api/series_list.py
```diff
@@ -0,0 +1,31 @@
...
+import copy
+import logging
+from math import ceil
+
+from flask import jsonify
+from flask import request
+from sqlalchemy.orm.exc import NoResultFound
+from flexget.plugins.filter.series import FilterSeriesBase
+from flexget.api import api, APIResource
+from flexget.plugins.list import movie_list as ml
```
paranoidi: should not use movie_list

flexget/plugins/list/series_list.py
```diff
@@ -0,0 +1,283 @@
...
+
+    id = Column(Integer, primary_key=True)
+    added = Column(DateTime, default=datetime.now)
+    title = Column(Unicode)
+
+    # Internal series attributes
+    set = Column(Unicode)
+    path = Column(Unicode)
+    _alternate_name = Column('alternate_name', Unicode)
+    alternate_name = json_synonym('_alternate_name')
```
paranoidi: you're using json with relational database waaaay too much. There's no reason to use JSON here at all. Use proper schema.

flexget/plugins/list/series_list.py
```diff
@@ -0,0 +1,283 @@
...
+import logging
+from collections import MutableSet
+from datetime import datetime
+
+from builtins import *  # pylint: disable=unused-import, redefined-builtin
+from sqlalchemy import Column, Unicode, Integer, ForeignKey, Boolean, DateTime, String, func
+from sqlalchemy.orm import relationship
+from sqlalchemy.sql.elements import and_
+
+from flexget.plugins.filter.series import FilterSeriesBase
```
paranoidi: Hard dependencies between plugins are discouraged, try to use get_plugin_by_name (preferable where possible) or surround this with try...except

liiight: Will do, but what's the potential risk here? Can't we assume that our internal plugin is present?

flexget/plugins/list/series_list.py
liiight: The idea was to avoid defining numerous tables while still easily maintaining list structure. Any ideas how to achieve this otherwise?

flexget/plugins/list/series_list.py
paranoidi: all of them are kind of internal, the main idea is to make coupling harder in order to avoid creating spaghetti mess :)

flexget/plugins/list/series_list.py
```diff
@@ -0,0 +1,438 @@
...
+
+log = logging.getLogger('series_list')
+Base = versioned_base('series_list', 0)
+
+SUPPORTED_IDS = FilterSeriesBase().supported_ids
+SETTINGS_SCHEMA = FilterSeriesBase().settings_schema
+SERIES_ATTRIBUTES = SETTINGS_SCHEMA['properties']
+
+
+class SeriesListList(Base):
```
paranoidi: Why not just call this SeriesList ?

flexget/plugins/list/series_list.py
```diff
@@ -0,0 +1,438 @@
...
+    __tablename__ = 'series_list_id_regexps'
+
+    id = Column(Integer, primary_key=True)
+    regexp = Column(Unicode)
+    added = Column(DateTime, default=datetime.now)
+    series_id = Column(Integer, ForeignKey(SeriesListSeries.id))
+
+
+class SeriesListQuality(Base):
+    __tablename__ = 'series_list_id_quality'
```
paranoidi: why the "id" in name?

flexget/plugins/list/series_list.py
```diff
@@ -0,0 +1,438 @@
...
+
+    @staticmethod
+    def get_list(config):
+        return SeriesList(config)
+
+    def on_task_input(self, task, config):
+        return list(SeriesList(config))
+
+
+@event('plugin.register')
```
paranoidi: this is usually the last thing in the module

flexget/plugins/list/series_list.py
liiight: i retained the naming convention we used by other plugins. `PluginName` `TableType`

flexget/plugins/list/series_list.py
liiight: mistake

flexget/plugins/list/series_list.py
liiight: After this are API helper methods. I agree this isn't the prettiest format, just retained the structure we set with other plugins.

flexget/plugins/filter/series.py
```diff
@@ -949,6 +949,11 @@ def settings_schema(self):
             'additionalProperties': False
         }
 
+    @property
+    def supported_ids(self):
```
paranoidi: This sucks, series filter plugin should not have this.

liiight: where would you put it? it's used by both `configure_series` and now `series_list`. I thought filter_base was a natural place to put it

flexget/plugins/list/series_list.py
```diff
@@ -0,0 +1,438 @@
...
+        }
+        for attribute in SETTINGS_SCHEMA['properties']:
+            # `set` is not a real series attribute and is just a way to pass IDs in task.
+            if attribute == 'set':
+                continue
+            series_dict[attribute] = self.format_converter(attribute)
+        return series_dict
+
+
+class SeriesListID(Base):
```
paranoidi: I would call this SeriesListSeriesExternalID because that's what this is .. or at least describe a bit in a docstring

liiight: you got it. i plan to to some much better docs to

liiight:
Reopening from flexget/series_list

PR #1228, opened 2016-06-07
Series list

liiight:
### Motivation for changes:
A manged list for series, enabling using `series` plugin without the need to explicitly edit config file or use 3rd party service.
This will also allows using multiple lists for multiple purposes.

Note: This does not affect series tracking in any way, this just represents an input to which `configure_series` plugin can be used with (among others things).
### Detailed changes:

- New managed list type, `series_list`
- Manageable via API and CLI
- Had to add a small change to JSON decoder to support trying to decode None values.
- Added a global supported series IDs list to be shared between `configure_series` and `series_list` (and any other plugin that this can relate too, need to look into other `series` related plugins).

```
tasks:
  series_with_list:
    configure_series:
      from:
        series_list: main_shows
    rss: http://url.com/links
    download: /path/to/downloads
```
Tests included.

#### To Do:
- [x] CLI
- [x] API and tests
- [x] Create a method in lookup providers that will register valid series identifiers instead of using hard coded list
- [x] Change API input payload to directly inherit from `series` schema
- [x] Readd `set` attribute as it was wrong not to use it in the first place

### DIFF
--- a/flexget/plugins/api/series_list.py
+++ b/flexget/plugins/api/series_list.py
@@ -41,7 +41,12 @@ class objects_container(object):
     del list_input['properties']['id']
     del list_input['properties']['added_on']
 
-    return_lists = {'type': 'array', 'items': list_object}
+    return_lists = {
+        'type': 'object',
+        'properties': {
+            'series_lists': {'type': 'array', 'items': list_object}
+        }
+    }
 
     input_series_list_id_object = {
         'type': 'array',
@@ -156,7 +161,7 @@ class SeriesListAPI(APIResource):
         name = args.get('name')
         series_lists = [series_list.to_dict() for series_list in
                         sl.SeriesListDBContainer.get_series_lists(name=name, session=session)]
-        return jsonify(series_lists)
+        return jsonify({'series_lists': series_lists})
 
     @api.validate(new_list_schema)
     @api.response(201, model=list_object_schema)
--- a/flexget/plugins/api/series_list.py
+++ b/flexget/plugins/api/series_list.py
@@ -45,6 +45,8 @@ class objects_container(object):
 
     return_lists = {'type': 'array', 'items': list_object}
 
+
+
     return_series = {
         'type': 'object',
         'properties': {
--- a/flexget/plugins/api/series_list.py
+++ b/flexget/plugins/api/series_list.py
@@ -62,59 +62,12 @@ class objects_container(object):
             'added_on': {'type': 'string'},
             'series_id': {'type': 'integer'}
         }})
+    input_series_object = FilterSeriesBase().settings_schema
+    input_series_object['properties'].update({'title': {'type': 'string'}})
+    input_series_object['required'] = ['title']
+    input_series_object['additionalProperties'] = True
 
-    input_series_object = {
-        'type': 'object',
-        'properties': {
-            'title': {'type': 'string'},
-            'path': {'type': 'string'},
-            'alternate_name': {'type': 'array', 'items': {'type': 'string', 'format': 'regex'}},
-            'name_regexp': {'type': 'array', 'items': {'type': 'string', 'format': 'regex'}},
-            'ep_regexp': {'type': 'array', 'items': {'type': 'string', 'format': 'regex'}},
-            'date_regexp': {'type': 'array', 'items': {'type': 'string', 'format': 'regex'}},
-            'sequence_regexp': {'type': 'array', 'items': {'type': 'string', 'format': 'regex'}},
-            'id_regexp': {'type': 'array', 'items': {'type': 'string', 'format': 'regex'}},
-            'date_yearfirst': {'type': 'boolean'},
-            'date_dayfirst': {'type': 'boolean'},
-            'quality': {'type': 'string', 'format': 'quality_requirements'},
-            'qualities': {'type': 'array', 'items': {'type': 'string', 'format': 'quality_requirements'}},
-            'timeframe': {'type': 'string', 'format': 'interval'},
-            'upgrade': {'type': 'boolean'},
-            'target': {'type': 'string', 'format': 'quality_requirements'},
-            # Specials
-            'specials': {'type': 'boolean'},
-            # Propers (can be boolean, or an interval string)
-            'propers': {'type': ['boolean', 'string'], 'format': 'interval'},
-            # Identified by
-            'identified_by': {
-                'type': 'string', 'enum': ['ep', 'date', 'sequence', 'id', 'auto']
-            },
-            # Strict naming
-            'exact': {'type': 'boolean'},
-            # Begin takes an ep, sequence or date identifier
-            'begin': {
-                'oneOf': [
-                    {'name': 'ep identifier', 'type': 'string', 'pattern': r'(?i)^S\d{2,4}E\d{2,3}$',
-                     'error_pattern': 'episode identifiers should be in the form `SxxEyy`'},
-                    {'name': 'date identifier', 'type': 'string', 'pattern': r'^\d{4}-\d{2}-\d{2}$',
-                     'error_pattern': 'date identifiers must be in the form `YYYY-MM-DD`'},
-                    {'name': 'sequence identifier', 'type': 'integer', 'minimum': 0}
-                ]
-            },
-            'from_group': {'type': 'array', 'items': {'type': 'string'}},
-            'parse_only': {'type': 'boolean'},
-            'special_ids': {'type': 'array', 'items': {'type': 'string'}},
-            'prefer_specials': {'type': 'boolean'},
-            'assume_special': {'type': 'boolean'},
-            'tracking': {'type': ['boolean', 'string'], 'enum': [True, False, 'backfill']}
-        },
-        'required': ['title'],
-        'additionalProperties': True
-    }
-
-    edit_series_object = copy.deepcopy(input_series_object)
-    del edit_series_object['properties']['title']
-    del edit_series_object['required']
+    edit_series_object = FilterSeriesBase().settings_schema
 
     return_series_object = copy.deepcopy(input_series_object)
     return_series_object['properties']['series_list_identifiers'] = {'type': 'array',
--- a/flexget/plugins/api/series_list.py
+++ b/flexget/plugins/api/series_list.py
@@ -111,7 +111,7 @@ class SeriesListAPI(APIResource):
         args = series_list_parser.parse_args()
         name = args.get('name')
         series_lists = [series_list.to_dict() for series_list in
-                        sl.SeriesListDBContainer.get_series_lists(name=name, session=session)]
+                        sl.SeriesListDB.get_series_lists(name=name, session=session)]
         return jsonify({'series_lists': series_lists})
 
     @api.validate(new_list_schema)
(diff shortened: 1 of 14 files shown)


flexget/plugins/api/series_list.py
```diff
@@ -0,0 +1,341 @@
...
+
+
+series_parser = api.parser()
+series_parser.add_argument('sort_by', choices=('title', 'added'), default='title',
+                           help='Sort by attribute')
+series_parser.add_argument('order', choices=('desc', 'asc'), default='desc', help='Sorting order')
+series_parser.add_argument('page', type=int, default=1, help='Page number')
+series_parser.add_argument('page_size', type=int, default=10, help='Number of series per page')
+
+series_identifiers_doc = "Use series identifier using the following format:\n[{'ID_NAME: 'ID_VALUE'}]." \
```
paranoidi: This is executed when module is imported, it will make call to supported_ids .. which will get list of registered plugins, but plugins are not necessary yet loaded. Importing module should not have any side effects / dependencies.

flexget/plugins/filter/series.py
```diff
@@ -949,6 +949,11 @@ def settings_schema(self):
             'additionalProperties': False
         }
 
+    @property
```
paranoidi: remove as it is now unnecessary

flexget/plugins/plugin_configure_series.py
```diff
@@ -17,6 +17,14 @@
 Base = db_schema.versioned_base('import_series', 0)
 
 
+def supported_ids():
```
paranoidi: might be good in FilterSeries base so we don't need to copy paste this too much ..

flexget/plugins/plugin_configure_series.py
```diff
@@ -17,6 +17,14 @@
...
+    # Return a list of supported series identifier as registered via their plugins
+    ids = []
+    for plugin_ in plugin.get_plugins(group='series_metainfo'):
```
paranoidi: _ prepended variables suck, avoid if you can .. in this case single letter variable "p" is fine or something else

flexget/plugins/api/series_list.py
```diff
@@ -0,0 +1,341 @@
...
+from flask import request
+from sqlalchemy.orm.exc import NoResultFound
+from flexget import plugin
+from flexget.plugins.filter.series import FilterSeriesBase
+from flexget.api import api, APIResource
+from flexget.plugins.list import series_list as sl
+
+
+SETTINGS_SCHEMA = FilterSeriesBase().settings_schema
+SERIES_ATTRIBUTES = SETTINGS_SCHEMA['properties']
```
paranoidi: This does not appear to be used, can you get rid of whole module plugins.filter.series  dependency here?

flexget/plugins/list/series_list.py
```diff
@@ -0,0 +1,490 @@
...
+from flexget.event import event
+
+log = logging.getLogger('series_list')
+Base = versioned_base('series_list', 0)
+
+SETTINGS_SCHEMA = FilterSeriesBase().settings_schema
+SERIES_ATTRIBUTES = SETTINGS_SCHEMA['properties']
+
+
+def supported_ids():
```
paranoidi: if this is moved to filterseriesbase and that dependency can not be removed nicely this could be used from there instead

flexget/plugins/list/series_list.py
```diff
@@ -0,0 +1,490 @@
...
+        for attribute in SETTINGS_SCHEMA['properties']:
+            # `set` is not a real series attribute and is just a way to pass IDs in task.
+            if attribute == 'set':
+                continue
+            series_dict[attribute] = self.format_converter(attribute)
+        return series_dict
+
+
+class SeriesListSeriesExternalID(Base):
+    __tablename__ = 'series_list_series_external_ids'
```
paranoidi: you lost empty line separator by accident from here :)

flexget/plugins/list/series_list.py
```diff
@@ -0,0 +1,490 @@
...
+        db_series.id_regexp = [SeriesListIDRegexp(regexp=regexp) for regexp in entry.get('id_regexp')]
+    if entry.get('date_dayfirst'):
+        db_series.date_dayfirst = entry.get('date_dayfirst')
+    if entry.get('date_dayfirst'):
+        db_series.date_dayfirst = entry.get('date_dayfirst')
+    if entry.get('quality'):
+        db_series.quality = SeriesListQuality(quality=entry.get('quality'))
+    if entry.get('qualities'):
+        db_series.qualities = [SeriesListQualities(quality=quality) for quality in entry.get('qualities')]
+    if entry.get('timeframe'):
```
paranoidi: you can reduce this bit a lot by making a list/map and then loop over it with setattr

    fields = [timeframe, target, upgrade, ...]
    for field in fields:
      setattr(db_series, field, getattr(entry, field)

flexget/plugins/list/series_list.py
```diff
@@ -0,0 +1,490 @@
...
+        return list(SeriesList(config))
+
+
+@event('plugin.register')
+def register_plugin():
+    plugin.register(PluginSeriesList, 'series_list', api_ver=2, groups=['list'])
+
+
+@with_session
+def get_series_lists(name=None, session=None):
```
paranoidi: Might be good idea to have a class for database access and have all of these in there as static methods ..

flexget/plugins/list/series_list.py
liiight: The problem with that is that not all fields are handled equally. This is what I did when I used json synonym, but I have some relationships that should be handled differently.

flexget/plugins/api/series_list.py
liiight: yeah, it doesn't work as you expected. any suggestion how i can output this to api doc?

flexget/plugins/list/series_list.py
paranoidi: Yeah, but looks like most of them are just moving value from object to another so those could be done with that approach. Rest can be still done separately.

flexget/plugins/cli/series_list.py
```diff
@@ -0,0 +1,344 @@
...
+from flexget.utils import qualities
+from flexget.config_schema import parse_interval
+from flexget.plugins.list.series_list import SeriesListList, SeriesListDBContainer as slDb, SeriesList, get_db_series
+from flexget.plugins.filter.series import FilterSeriesBase
+
+SETTINGS_SCHEMA = FilterSeriesBase().settings_schema
+SERIES_ATTRIBUTES = SETTINGS_SCHEMA['properties']
+
+
+def supported_ids():
```
paranoidi: you don't need this duplication, you can access this via FilterSeriesBase.supported_ids .. correct?

flexget/plugins/cli/series_list.py
```diff
@@ -0,0 +1,344 @@
...
+
+from sqlalchemy.orm.exc import NoResultFound
+
+from flexget import options, plugin
+from flexget.event import event
+from flexget.logger import console
+from flexget.manager import Session
+from flexget.utils import qualities
+from flexget.config_schema import parse_interval
+from flexget.plugins.list.series_list import SeriesListList, SeriesListDBContainer as slDb, SeriesList, get_db_series
```
paranoidi: PEP8 violation with "slDb"

flexget/plugins/plugin_configure_series.py
```diff
@@ -16,6 +16,8 @@
 log = logging.getLogger('configure_series')
 Base = db_schema.versioned_base('import_series', 0)
 
+supported_ids = FilterSeriesBase().supported_ids()
```
paranoidi: this causes method to be called when module is imported and supported_ids will be set whatever happens to be loaded at that time, remove this from module level

flexget/plugins/list/series_list.py
paranoidi: todo: remove

flexget/plugins/list/series_list.py
```diff
@@ -0,0 +1,480 @@
...
+        db_series.quality = SeriesListQuality(quality=data.get('quality'))
+    if data.get('qualities'):
+        db_series.qualities = [SeriesListQualities(quality=quality) for quality in data.get('qualities')]
+    if data.get('from_group'):
+        db_series.from_group = [SeriesListFromGroup(group_name=name) for name in data.get('from_group')]
+    if data.get('special_ids'):
+        db_series.special_ids = [SeriesListSpecialID(id_name=name) for name in data.get('special_ids')]
+
+    # Get list of supported identifiers
+    for id_name in supported_ids():
```
paranoidi: use from FilterSeriesBase

flexget/plugins/list/series_list.py
```diff
@@ -0,0 +1,480 @@
...
+            log.debug('deleting series %s', db_series)
+            session.delete(db_series)
+
+    def __contains__(self, entry):
+        return self.find_entry(entry) is not None
+
+    @with_session
+    def find_entry(self, entry, session=None):
+        """Finds `SeriesListSeries` corresponding to this entry, if it exists."""
+        for id_name in supported_ids():
```
paranoidi: use from FilterSeriesBase

flexget/plugins/plugin_configure_series.py
```diff
@@ -78,8 +79,9 @@ def on_task_start(self, task, config):
 
             for entry in result:
                 s = series.setdefault(entry['title'], {})
-                if entry.get('tvdb_id'):
-                    s['set'] = {'tvdb_id': entry['tvdb_id']}
+                for supported_id in supported_ids:
```
paranoidi: use from FilterSeriesBase

flexget/plugins/api/series_list.py
```diff
@@ -0,0 +1,336 @@
...
+from math import ceil
+
+from flask import jsonify
+from flask import request
+from sqlalchemy.orm.exc import NoResultFound
+from flexget.plugins.filter.series import FilterSeriesBase
+from flexget.api import api, APIResource
+from flexget.plugins.list import series_list as sl
+
+supported_ids = FilterSeriesBase().supported_ids()
```
paranoidi: remove

flexget/plugins/cli/series_list.py
liiight: actually since i import `SeriesListDBContainer` I get a pep8 violation if i DON'T use camel case. at least according to pycharm that is...

paranoidi: weird ... I would drop the container word from class name.

flexget/plugins/list/series_list.py
```diff
@@ -0,0 +1,471 @@
...
+class SeriesListSpecialID(Base):
+    __tablename__ = 'series_list_special_ids'
+
+    id = Column(Integer, primary_key=True)
+    id_name = Column(Unicode)
+    added = Column(DateTime, default=datetime.now)
+    series_id = Column(Integer, ForeignKey(SeriesListSeries.id))
+
+
+def get_db_series(data, db_series=None):
```
paranoidi: shouldn't this be in SeriesListDBContainer?

liiight:
linked to https://github.com/Flexget/Flexget/projects/3#card-60653

liiight:
this will be redone while refactoring series.

PR #1234, opened 2016-06-14
Fix scheduler issues with APScheduler 3.1+

agronholm:
### Motivation for changes:
Compatibility with APScheduler 3.2 and beyond
### Detailed changes:
The scheduler plugin was modified to start the scheduler in paused mode, then manipulate the jobs and only then start the job processing. The way FlexGet was doing it before only worked because of a programming oversight which was corrected in 3.1. The 3.1 version however lacked the means to provide comparable functionality. This shortcoming was just addressed in the 3.2 release.
- 

### Addressed issues:

- Fixes # .

```
paste_config_here
```
========== 918 passed, 7 skipped, 1 xfailed, 3 xpassed, 1 pytest-warnings in 220.10 seconds ===========
paste output here
```


Job stores should not be accessed before they're started. The fact this was working in previous versions was an oversight. That said, APScheduler 3.1 lacked the ability to manipulate the jobs before the scheduler has been started, but 3.2 adds that feature.

### DIFF
--- a/flexget/plugins/daemon/scheduler.py
+++ b/flexget/plugins/daemon/scheduler.py
@@ -137,7 +137,7 @@ def setup_scheduler(manager):
     if not timezone:
         # The default sqlalchemy jobstore does not work when there isn't a name for the local timezone.
         # Just fall back to utc in this case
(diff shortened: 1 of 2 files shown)


gazpachoking:
Awesome, thanks!

PR #1248, opened 2016-06-23
bug fix for crash when letters are in SQLAlchemy version

M1chael:
### Motivation for changes:

Fix for error message "Failed to check SQLAlchemy version, you may need to upgrade it" for SQLAlchemy beta version (1.1.0b1).
### Detailed changes:

All letters are cutted out in version checker.

### DIFF
--- a/flexget/manager.py
+++ b/flexget/manager.py
@@ -18,6 +18,7 @@ from contextlib import contextmanager
 from datetime import datetime, timedelta
 
 import io
+import string
 
 import sqlalchemy
 import yaml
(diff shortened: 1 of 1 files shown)


liiight:
Have you considered using https://docs.python.org/2/distutils/apiref.html#module-distutils.version

flexget/manager.py
```diff
@@ -647,7 +648,7 @@ def validate_config(self, config=None):
     def init_sqlalchemy(self):
         """Initialize SQLAlchemy"""
         try:
-            if [int(part) for part in sqlalchemy.__version__.split('.')] < [0, 7, 0]:
+            if [int(part) for part in sqlalchemy.__version__.translate(None, string.letters).split('.')] < [0, 7, 0]:
```
liiight: Please use https://docs.python.org/2/distutils/apiref.html#module-distutils.version instead

PR #1281, opened 2016-07-18
Fix online tests

cvium:
### Motivation for changes:
It's annoying to look at failing Jenkins jobs
### Detailed changes:

- Some data changed, so tests have been updated
- Some tests have been skipped because they can't be run without the cassettes


stevezau:
Manually merged due to conflicts.

PR #1283, opened 2016-07-19
list_clear - new list plugin

cvium:
### Motivation for changes:
Wanted to be able to clear an entry_list before populating it eg. make entry_list mimick an RSS feed. It takes as input a `phase` param that specifies at which phase it should clear the list.
### Detailed changes:

- Added plugin
- Added three tests

```
list_clear:
  what:
    - entry_list: whatever
  phase: start
```

### DIFF
--- a/flexget/plugins/modify/list_clear.py
+++ b/flexget/plugins/modify/list_clear.py
@@ -26,6 +26,7 @@ class ListClear(object):
         'required': ['what']
     }
 
+    @plugin.priority(255)
     def on_task_start(self, task, config):
         for item in config['what']:
             for plugin_name, plugin_config in item.items():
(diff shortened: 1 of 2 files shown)


flexget/plugins/modify/list_clear.py
```diff
@@ -0,0 +1,58 @@
...
+        for item in config['what']:
+            for plugin_name, plugin_config in item.items():
+                try:
+                    thelist = plugin.get_plugin_by_name(plugin_name).instance.get_list(plugin_config)
+                except AttributeError:
+                    raise PluginError('Plugin %s does not support list interface' % plugin_name)
+                if thelist.immutable:
+                    raise plugin.PluginError(thelist.immutable)
+                if config['phase'] == task.current_phase:
+                    if task.manager.options.test:
```
liiight: other list plugins check for online attribute here:
`if task.manager.options.test and thelist.online:`

cvium: Yes and I don't fully agree with it. I would expect `--test` to not clear my entry_list.

liiight: since test mode creates a temp db, anything that is a part of that temp db only should persist in that scope, since it can affect other operations. Al plugins that rely on our db still work, so why not `list_clear`?

PR #1288, opened 2016-07-20
Next branch

liiight:
### Motivation for changes:
Next breaking changes for flexget v2.2.0
### Detailed changes:

- Moved `irc_bot` into a dependency. Install with `pip install irc_bot`
- Removed deprecated `thetvdb` plugins
- Removed `kat` plugin
- Removed all `uoccin` plugins, these can be found from flexget/extras respository
- Changed `along` in copy/move/delete plugins. See wiki for details.


flexget/plugins/daemon/irc.py
```diff
@@ -222,7 +228,7 @@ def __init__(self, config, config_name):
                          'invite_message': config.get('invite_message'),
                          'nickserv_password': config.get('nickserv_password'),
                          'use_ssl': config.get('use_ssl')}
-        irc_bot.IRCBot.__init__(self, ircbot_config)
+        IRCBot.__init__(self, ircbot_config)
```
paranoidi: This is not correct way to call super

flexget/plugins/daemon/irc.py
```diff
@@ -667,6 +673,13 @@ def irc_update_config(manager):
         stop_irc(manager)
         return
 
+    if irc_bot is None:
+        log.error('ImportError: irc_bot module not found. Shutting down daemon.')
+        stop_irc(manager)
+        from flexget.manager import manager
```
paranoidi: - Imports should be at the top of the module unless some very good reason
- Manager shutdown seems rather drastic, I don't think any plugin does that. Simply disable the plugin instead?

flexget/plugins/daemon/irc.py
cvium: That is the correct way to call super on old-style classes.

flexget/plugins/daemon/irc.py
cvium: It should shutdown imo. If it reaches that if, that means the user has specified irc but failed to install the dep. Since it's a daemon plugin it makes more sense to shut down the daemon.

PR #1290, opened 2016-07-20
[WIP] Tmdb list

Ram-Z:
### Motivation for changes:

Add a managed list for tmdb similar to trakt_list
### Detailed changes:
- [x] Add cli command to authenticate with themoviedb.org
- [ ] Add tmdb_list.py

``` yaml
#not sure yet... suggestion welcome
```

### DIFF
--- a/flexget/plugins/api_tmdb.py
+++ b/flexget/plugins/api_tmdb.py
@@ -2,6 +2,7 @@ from __future__ import unicode_literals, division, absolute_import
 from builtins import *  # pylint: disable=unused-import, redefined-builtin
 
 import logging
+
 from datetime import datetime, timedelta
 
 from sqlalchemy import Table, Column, Integer, Float, Unicode, Boolean, DateTime, func, or_
@@ -10,13 +11,16 @@ from sqlalchemy.schema import ForeignKey
 from sqlalchemy.orm import relation
 
 from flexget import db_schema, plugin
+from flexget import options
 from flexget.event import event
+from flexget.logger import console
+from flexget.manager import Session
 from flexget.plugin import get_plugin_by_name
 from flexget.utils import requests
 from flexget.utils.database import text_date_synonym, year_property, with_session
 
 log = logging.getLogger('api_tmdb')
-Base = db_schema.versioned_base('api_tmdb', 1)
+Base = db_schema.versioned_base('api_tmdb', 2)
 
 # This is a FlexGet API key
 API_KEY = 'bdfc018dbdb7c243dc7cb1454ff74b95'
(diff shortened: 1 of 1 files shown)


flexget/plugins/api_tmdb.py
```diff
@@ -29,14 +33,20 @@ def get_tmdb_config():
     """Gets and caches tmdb config on first call."""
     global _tmdb_config
     if not _tmdb_config:
-        _tmdb_config = tmdb_request('configuration')
+        _tmdb_config = tmdb_request('configuration').json()
     return _tmdb_config
 
 
+def tmdb_post(endpoint, **params):
```
liiight: i'd change `tmdb_request` to take `method` param and the do `return requests.request(full_url, method=method, params=params)` to avoid duplicate code paths.

flexget/plugins/api_tmdb.py
```diff
@@ -269,6 +279,132 @@ def lookup(title=None, year=None, tmdb_id=None, imdb_id=None, smart_match=None,
...
+    """Authenticates a session with TMDB, and grabs any IDs needed for getting/modifying list."""
+    try:
+        r = tmdb_request('authentication/token/new')
+        url = r.headers['Authentication-Callback']
+        body = r.json()
+        request_token = body['request_token']
+        expires_at = body['expires_at']
+
+        console('Please visit {0} and authorize Flexget. Your token expires {1}.'.format(url, expires_at))
+        input('Press `Enter` when the token has been authorized...')
```
liiight: This isn't ideal, a loop would be better, like in trakt:
https://github.com/Flexget/Flexget/blob/develop/flexget/plugins/api_trakt.py#L83

flexget/plugins/api_tmdb.py
```diff
@@ -269,6 +279,132 @@ def lookup(title=None, year=None, tmdb_id=None, imdb_id=None, smart_match=None,
...
+@event('options.register')
+def register_parser_arguments():
+    acc_text = 'local identifier which should be used in your config to refer these credentials'
+    # Register subcommand
+    parser = options.register_command('tmdb', do_cli, help='view and manage tmdb authentication.')
+    # Set up our subparsers
+    subparsers = parser.add_subparsers(title='actions', metavar='<action>', dest='action')
+    auth_parser = subparsers.add_parser('auth', help='authorize Flexget to access your themoviedb.org account')
+
+    auth_parser.add_argument('account', metavar='<account>', help=acc_text)
```
liiight: You could create another parser, add argument to it and use it as a parent parser instead of duplicating code to use the same arg for different subparsers. 
Example: https://github.com/Flexget/Flexget/blob/develop/flexget/plugins/cli/movie_list.py#L175
Also, if that argument is required, add the `required=True` param to its creating, instead of later checking it in code.
Lastly, using a metavar with the exact name as the argument is redundant.

Ram-Z: > Also, if that argument is required, add the required=True param to its creating, instead of later checking it in code.

It's not required for the `show` parser. So if I switch to using parent parser I can't use that.

I could add a new command `list` which does what `show` currently does when no `account` is given.

cvium:
Any progress on this?

Ram-Z:
No sorry.

I had very little time lately and will not have much more for the following month. I'm hoping to pick it up again in October.

cvium:
No worries.

PR #1302, opened 2016-07-29
Coerce title to unicode type in TorrentLeech search plugin

### LINKED ISSUE #1301: Search plugins storing BeautifulSoup objects in Entry cause excessive recursion in Entry::take_snapshot
### Expected behaviour:

Flexget doesn't crash.

### Actual behaviour:

Flexget crashes due to excessive recursion depth while taking the snapshot of an entry produced by the TorrentLeech search plugin.

### Steps to reproduce:

Use the TorrentLeech search plugin in a task.

#### Config (snippet):

```
tasks:
  tv-discover:
    interval: 12 hours
    discover:
      what:
        - next_series_episodes: yes
      from:
        - torrentleech:
            rss_key: "{{ secrets.torrentleech.rsskey }}"
            username: "{{ secrets.torrentleech.username }}"
            password: "{{ secrets.torrentleech.password }}"
            category: Episodes HD
    transmission: yes
```

### Additional information:

- Flexget Version: 2.2.1
- Python Version: 2.7.12
- Installation method: pip
- OS and version: Fedora 24 x86_64
- Link to crash log: https://gist.github.com/danielkza/d6e8c98b31037de731c5b5ac76733b39

### Debugging information

I'm not familiar with the Flexget codebase, but some quick debugging shows the problem is caused by the TL plugin inserting some text directly from the HTML tree produced by BeautifulSoup. That text is actually a `bs4.element.NavigableString` that contains references to the whole tree, which Flexget attempts to serialize when saving a snapshot of the entry.

The plugin should convert any such usages to `unicode` before storing it in the entry. It is possible other search plugins have the same issue, so a generic solution might be desirable.

danielkza:
Pulling it directly from the BeautifulSoup model will cause the whole
document tree to be stored in the Entry. It can cause excessive
recursion errors during serialization, and is not necessary at all.

Fixes #1301.

### DIFF
--- a/flexget/plugins/urlrewrite/torrentleech.py
+++ b/flexget/plugins/urlrewrite/torrentleech.py
@@ -139,7 +139,8 @@ class UrlRewriteTorrentleech(object):
                 log.debug('link phase: %s' % link.contents[0])
                 entry = Entry()
                 # extracts the contents of the <a>titlename/<a> tag
-                entry['title'] = link.contents[0]
+                # coerce to str to throw away the reference to the document tree
+                entry['title'] = str(link.contents[0])
 
                 # find download link
                 torrent_url = tr.find("a", attrs={'href': re.compile('/download/\d+/.*')}).get('href')


flexget/plugins/urlrewrite/torrentleech.py
```diff
@@ -139,7 +139,7 @@ def search(self, task, entry, config=None):
                 log.debug('link phase: %s' % link.contents[0])
                 entry = Entry()
                 # extracts the contents of the <a>titlename/<a> tag
-                entry['title'] = link.contents[0]
+                entry['title'] = unicode(link.contents[0])
```
cvium: That will not work in Py3 afaik. http://python-future.org/compatible_idioms.html#unicode

danielkza: Right. Since the `future.builtins.*` is already imported, it should be a matter of just using `str()` instead, correct?

paranoidi:
use:

    link.contents[0].text

danielkza:
Looking at the BS4 source code, it doesn't seem like `.text` provides any guarantee that it will return a plain string. `NavigableString` is *itself* an instance of `unicode` (`str` in Py3) mixed with the element trait, and the BS4 documentation [recommends](https://www.crummy.com/software/BeautifulSoup/bs4/doc/#navigablestring) calling `unicode()`.

paranoidi:
Hmm, I see ... then we have a few other places where the same fix needs to be done.

gazpachoking:
Hmm, I think we used to cast all strings to unicode when setting on entry and lost a bit of that on the py3 upgrade. Perhaps we should add something here to solve this for all cases?
https://github.com/Flexget/Flexget/blob/b92f47a265db04e8e4ec8dd2bff7a54b1a6f1581/flexget/entry.py#L197-L197
Maybe
```python
elif isinstance(value, str) and type(value) not in [str, text_type]:
            # Cast string subclasses (e.g. NavigableString) back to regular strings
            value = str(value)
```
`text_type` is from future.utils

liiight:
I'm with you @gazpachoking, that seems like a more correct approach

cvium:
Any progress on this?

danielkza:
I don't know enough about flexget and how DB interactions are handled to
know the best place for a global fix. I can try to work it with some help
on where to look :)

Em seg, 8 de ago de 2016 10:55, Claus Vium <notifications@github.com>
escreveu:

cvium:
Well, @gazpachoking proposed a fix. Try it.

danielkza:
I somehow missed his comment, my bad.

Em seg, 8 de ago de 2016 11:04, Claus Vium <notifications@github.com>
escreveu:

danielkza:
Closed in favor of generic solution in #1337.

PR #1304, opened 2016-07-31
Miscellaneous Subliminal plugin improvements

danielkza:
### Motivation/Detailed changes:

- Use the effective list of providers to validate configuration when possible. Avoids needing to manually update the list to follow upstream.
- Make it possible to use providers that require login by adding the `provider_configs` option.
- Add `hearing_impaired` option.
- Make debugging easier by silencing useless spam, and un-silencing subliminal itself.

#### Add `provider_configs` option
```yaml
subliminal:
  provider_configs:
    legendastv:
      username: myuser
      password: mypassword
```
#### Add `hearing_impaired` option
```yaml
subliminal:
  hearing_impaired: yes
```

### DIFF
--- a/flexget/plugins/output/subtitles_subliminal.py
+++ b/flexget/plugins/output/subtitles_subliminal.py
@@ -12,14 +12,17 @@ from flexget.event import event
 
 log = logging.getLogger('subtitles')
 
-PROVIDERS = [
-    'opensubtitles',
-    'thesubdb',
-    'podnapisi',
-    'addic7ed',
-    'tvsubtitles'
-]
-
+try:
+    from subliminal.extensions import provider_manager
+    PROVIDERS = provider_manager.names()
+except ImportError as e:
+    PROVIDERS = [
+        'opensubtitles',
+        'thesubdb',
+        'podnapisi',
+        'addic7ed',
+        'tvsubtitles'
+    ]
 
 class PluginSubliminal(object):
     """
--- a/flexget/plugins/output/subtitles_subliminal.py
+++ b/flexget/plugins/output/subtitles_subliminal.py
@@ -49,6 +49,7 @@ class PluginSubliminal(object):
           providers: addic7ed, opensubtitles
           single: no
           directory: /disk/subtitles
+          hearing_impaired: yes
           authentication:
             addic7ed:
               username: myuser
@@ -64,6 +65,7 @@ class PluginSubliminal(object):
             'providers': {'type': 'array', 'items': {'type': 'string', 'enum': PROVIDERS}},
             'single': {'type': 'boolean', 'default': True},
             'directory': {'type:': 'string'},
+            'hearing_impaired': {'type': 'boolean', 'default': False},
             'authentication': {'type': 'object', 'properties': AUTHENTICATION_SCHEMA},
         },
         'required': ['languages'],
@@ -94,6 +96,7 @@ class PluginSubliminal(object):
                 providers: List of providers from where to download subtitles.
                 single: Download subtitles in single mode (no language code added to subtitle filename).
                 directory: Path to directory where to save the subtitles, default is next to the video.
+                hearing_impaired: Prefer subtitles for the hearing impaired when available
                 authentication: >
                   Dictionary of configuration options for different providers.
                   Keys correspond to provider names, and values are dictionaries, usually specifying `username` and `password`.
@@ -132,6 +135,7 @@ class PluginSubliminal(object):
         # we ignore the configuration and add the language code to the
         # potentially downloaded files
         single_mode = config.get('single', '') and len(languages | alternative_languages) <= 1
+        hearing_impaired = config.get('hearing_impaired', False)
 
         with subliminal.core.ProviderPool(providers=providers_list, provider_configs=provider_configs) as provider_pool:
             for entry in task.accepted:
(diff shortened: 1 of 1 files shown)


danielkza:
Fixed Codacy issue.

liiight:
Do all providers need/have relevant config options? 
Also, do you think it'll be possible to define a provider directly under root level without needing the `providers_config` property?

danielkza:
> Do all providers need/have relevant config options? 

Currently only those that need authentication.

> Also, do you think it'll be possible to define a provider directly under root level without needing the providers_config property? 

I commented about that in the commit: using only the `providers` key would make it impossible to set options for some plugins without disabling all the others. An alternative would be to allow a boolean instead of a dictionary for plugins without configuration. It would still not cover the current case of omitting `providers` altogether and fetching from *all* the providers available.

danielkza:
@cvium I just edited my comment to explain why not.

liiight:
You could set type to be object or string, then check in code :
`'type':['object', 'string']`.
If that's not possible I guess leaving it as is is ok. 
I think naming the option something like `authentication` would be better but that's just a suggestion.

danielkza:
@liiight I don't think there is any specification that the configs will always be for authentication - it just happens to be the current situation. Using only the `providers` key does seem cleaner, but the `all` workaround is quite ugly. Is there any other plugin with a similar situation that we can use as a model?

cvium:
```
'providers': {'type': 'array', 'items': {
    'oneOf': [
        {'type': 'object',
        'additionalProperties': {
            'type': 'object',
            'properties': {
                'username': {'type': 'string'},
                'password': {'type': 'string'},
            }
        }},
        {'type': 'string'}
    ]
}},
```
The above will let you do this:
```YAML
    subliminal:
      providers:
        - legendastv:
            username: myuser
            password: mypassword
        - opensubtitles
```

danielkza:
@cvium That looks really nice. But what about the option of enabling all  available providers?

cvium:
I think I prefer having `provider_configs`

danielkza:
What about this:

```yaml
providers:
  all: yes
  configs: 
    legendastv:
      username: myuser
      password: mypassword
```

For the case of keeping all providers and configuring some. @cvium's last suggestion would also work.

paranoidi:
Nay for `provider_configs` from me, looks like nothing we have currently.

paranoidi:
Perhaps call the key `authentication` or `login` and it makes a lot more sense. I don't think there will be anything else in there?

cvium:
Any progress on this?

danielkza:
I've renamed the configuration key from `provider_configs` to `authentication` and rebased to the latest of the `develop`  branch.

flexget/plugins/output/subtitles_subliminal.py
```diff
@@ -12,14 +12,19 @@
...
-    'thesubdb',
-    'podnapisi',
-    'addic7ed',
-    'tvsubtitles'
-]
-
+try:
+    from subliminal.extensions import provider_manager
+    PROVIDERS = provider_manager.names()
+except ImportError as e:
```
cvium: Since you don't use the exception it's better not to include `as e`.

danielkza: It was a leftover from debugging, nice catch. Thanks!

flexget/plugins/output/subtitles_subliminal.py
```diff
@@ -84,6 +96,10 @@ def on_task_output(self, task, config):
                 providers: List of providers from where to download subtitles.
                 single: Download subtitles in single mode (no language code added to subtitle filename).
                 directory: Path to directory where to save the subtitles, default is next to the video.
+                hearing_impaired: Prefer subtitles for the hearing impaired when available
+                authentication: >
+                  Dictionary of configuration options for different providers.
+                  Keys correspond to provider names, and values are dictionaries, usually specifying `username` and `password`.
```
cvium: We have a max 120 char line length.

flexget/plugins/output/subtitles_subliminal.py
```diff
@@ -111,19 +134,27 @@ def on_task_output(self, task, config):
...
         single_mode = config.get('single', '') and len(languages | alternative_languages) <= 1
-        for entry in task.accepted:
-            if 'location' not in entry:
-                log.warning('Cannot act on entries that do not represent a local file.')
-            elif not os.path.exists(entry['location']):
-                entry.fail('file not found: %s' % entry['location'])
-            elif '$RECYCLE.BIN' not in entry['location']:  # ignore deleted files in Windows shares
+        hearing_impaired = config.get('hearing_impaired', False)
+
+        with subliminal.core.ProviderPool(providers=providers_list, provider_configs=provider_configs) as provider_pool:
```
cvium: Too lazy to check. Is this line <= 120 chars?

flexget/plugins/output/subtitles_subliminal.py
```diff
@@ -146,23 +177,30 @@ def on_task_output(self, task, config):
...
+                                                                          min_score=msc,
+                                                                          hearing_impaired=hearing_impaired)
+                        if subtitles:
+                            downloaded_subtitles[video].extend(subtitles)
                             log.info('Subtitles found for %s', entry['location'])
                         else:
                             # only try to download for alternatives that aren't alread downloaded
-                            subtitle = subliminal.download_best_subtitles({video}, alternative_languages,
-                                                                          providers=providers_list, min_score=msc)
+                            subtitles = provider_pool.download_best_subtitles(all_subtitles, video, alternative_languages,
```
cvium: This line looks a little long as well.

cvium:
Other than some PEP8 issues it looks fine.

danielkza:
Would it be preferable to fix the issues in an extra commit or just amend the previous ones?

liiight:
Another commit is fine

flexget/plugins/output/subtitles_subliminal.py
danielkza: 120 exactly.

flexget/plugins/output/subtitles_subliminal.py
danielkza: Fixed.

flexget/plugins/output/subtitles_subliminal.py
danielkza: Fixed.

danielkza:
Done.

danielkza:
I don't know what's wrong with Codacy, it shows no issues with the commit but fails anyway.

cvium:
We had miscommunication so our codacy account got deleted along with its config. We have yet to reconfigure the new one.

danielkza:
I update the Wiki page with the new additions. Thanks a lot!

PR #1310, opened 2016-08-04
Next version breaking changes

liiight:
### Motivation for changes:
*"New is always better."* - Barney Stinson

### Detailed changes:

- Removed `movie_queue` and all related plugins.
- Removed `imdb_required` plugin, switch to `imdb_lookup` and `require_field`
- Removed `subtitle_queue` plugin
- Removed `rutracker`  plugin
- Removed `torrentz` plugin
- Removed `newzleech` plugin
- Removed `publichd` plugin
- Removed `bt-chat` plugin
- Removed `btjunkie` plugin
- Removed `isohunt` plugin
- Removed `redskunk` plugin
- Removed `stmusic` plugin
- Pretty tables in terminal, https://github.com/Flexget/Flexget/pull/1311
- CLI command `flexget trakt show` has been renamed to `flexget trakt list`
- discover `release_estimations` now defaults to `strict` meaning anything with no release date or one that lies in the future will not be included in the search
- added blu-ray.com movie estimator (takes priority over the old one)
- discover will now keep searching for new episodes until it fails to find any (maximum 100 runs). Can be set to the old behaviour with `max_reruns` plugin.


flexget/options.py
```diff
@@ -473,3 +471,91 @@ def parse_args(self, *args, **kwargs):
...
+            'single': SingleTable,
+            'double': DoubleTable,
+            'github': GithubFlavoredMarkdownTable
+        }
+        if keys:
+            return list(table_types)
+        return table_types
+
+    @staticmethod
+    def colorize(text, color_tag, porcelain=False):
```
paranoidi: This is used outside from table context, in lines that have nothing to do with table. So this should probably be a plain function. Location is a bit icky though, coloring terminal output you need to import options.

Our console method is not much better being in logger.py, maybe both should be in "terminal.py" or so ...

flexget/plugins/cli/series.py
```diff
@@ -39,64 +64,58 @@ def display_summary(options):
...
-            console(formatting % ('Name', 'Latest', 'Age', 'Downloaded'))
-            console('-' * 79)
+            kwargs['sort_by'] = 'last_download_date'
 
-        for series in query.order_by(Series.name).yield_per(10):
+        query = get_series_summary(**kwargs)
+        header = ['Name', 'Latest', 'Age', 'Downloaded', 'Identified By', 'Latest status']
+        for index, value in enumerate(header):
+            if value.lower() == options.sort_by:
+                header[index] = color(value, SORT_COLUMN_COLOR)
```
paranoidi: Where is color defined?

flexget/options.py
liiight: yeah, i agree. ill see about splitting it

PR #1313, opened 2016-08-06
Populate series info for Plex episodes

pkoenig10:
### Motivation for changes:

Sometimes I download episodes into my Plex server through channels other than Flexget.  I would like to mark these episodes as seen/downloaded in Flexget.  Currently the series plugin cannot be used with Plex episode entries because the necessary field are not set, although the information is available.  These changes set the relevant fields so Flexget can get metainfo about Plex episodes.

### Detailed changes:

Added the fields `series_name`, `series_season`, `series_episode`, `series_id_type`, and `series_id` for Plex episode entires.  The necessary info was already used in Flexget for setting other fields, so my changes are entirely setting these fields.

### DIFF
--- a/flexget/plugins/input/plex.py
+++ b/flexget/plugins/input/plex.py
@@ -235,14 +235,21 @@ class InputPlex(object):
             if viewgroup == "episode":
                 e['plex_thumb'] = "http://%s:%d%s%s" % (
                     config['server'], config['port'], node.getAttribute('thumb'), urlappend)
+                e['series_name'] = title
                 season = int(node.getAttribute('parentIndex'))
                 if node.getAttribute('parentIndex') == node.getAttribute('year'):
                     season = node.getAttribute('originallyAvailableAt')
                     filenamemap = "%s_%s%s_%s_%s_%s.%s"
                     episode = ""
+                    e['series_id_type'] = 'date'
+                    e['series_date'] = season
                 elif node.getAttribute('index'):
                     episode = int(node.getAttribute('index'))
                     filenamemap = "%s_%02dx%02d_%s_%s_%s.%s"
+                    e['series_season'] = season
+                    e['series_episode'] = episode
+                    e['series_id_type'] = 'ep'
+                    e['series_id'] = 'S%02dE%02d' % (season, episode)
                 else:
                     log.debug("Could not get episode number for '%s' (Hint, ratingKey: %s)"
                               % (title, node.getAttribute('ratingKey')))


liiight:
Will this work with episodes by date?

pkoenig10:
Not entirely sure what you are asking.  The variables `season` and `episode` populated by the Plex plugin are always integers.  This information will be sufficient to fetch metainfo, which will contain the airdate, should that information be desired.

pkoenig10:
Updated the commit to handle episodes by date.

liiight:
Also, will this support other media types? Plex has music, movies etc libraries...

pkoenig10:
These changes are exclusive to entries determined to be Plex episodes.

As for movies, the only fields used by the movies metainfo plugins are `title`, `imdb_id`, and `tmdb_id`.  Plex already provides the title.  In most cases the title is sufficient to obtain the correct metainfo (although there is certainly the possibility of error). Plex can provide the `imdb_id` by parsing a url that looks like 'com.plexapp.agents.imdb://tt0338751?lang=en`.  However that seems out of the scope of this PR, since the title alone works in the majority of cases.  However I would be more than willing to add that as well if you think it would be a desired feature.

For other media (music, photos, and home videos), there aren't any metainfo plugins for these types of content, so there isn't any work to be done in this PR.

liiight:
I got no issues with relying on metainfo retrieved via Plex or any other input source for that matter. The rare situation that the metadata is incorrect in the source does not match up to the advantage of adding metadata without needing additional lookup. 

I was wondering if you added a check that knows if it should populate movie or series fields based on content type or library type or whatever

pkoenig10:
Okay, I'll add the metainfo id parsing to this PR.

That check already existed in this plugin (see there variable `viewgroup`). That info is determined in the XML returned by Plex. I'm using that to determine which fields to populate.

pkoenig10:
Currently, Flexget makes a single request to Plex and receives an XML response containing of list with limited info about the requested items in Plex.  This limited info does not include the fields necessary to populate the `imdb_id` or `tvdb_id`.  In order to populate these fields, Flexget would have to make an additional request for each entry listed in the original response.  While this is certainly possible, it is outside the scope of this commit.

As I mentioned earlier, the info returned by Plex is almost always sufficient to retrieve movie metainfo using other plugins.  This PR was motivated by the desire the ensure that the Plex plugin provides enough info for other metainfo plugins to retrieve info about series episodes.

liiight:
Fair enough, thanks for this!

PR #1317, opened 2016-08-08
Moved trakt cli stuff into its own file

cvium:
### Motivation for changes:
api_trakt is messy
### Detailed changes:

- Moved Trakt CLI into cli/trakt.py
- Moved trakt auth to its own versioned base to prevent nuking auth data when nuking cache

### DIFF
--- a/flexget/plugins/cli/trakt.py
+++ b/flexget/plugins/cli/trakt.py
@@ -1,0 +1,85 @@
+from __future__ import unicode_literals, division, absolute_import
+
+
+from flexget import options, plugin
+from flexget.event import event
+from flexget.logger import console
+from flexget.manager import Session
+from flexget.plugins.api_trakt import get_access_token, TraktUserAuth, delete_account, PIN_URL
+
+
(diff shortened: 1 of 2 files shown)


cvium:
Pushed to Next branch. Closing PR.

flexget/plugins/cli/trakt.py
```diff
@@ -0,0 +1,85 @@
...
+
+from flexget import options, plugin
+from flexget.event import event
+from flexget.logger import console
+from flexget.manager import Session
+from flexget.plugins.api_trakt import get_access_token, TraktUserAuth, delete_account, PIN_URL
+
+
+def do_cli(manager, options):
+    if options.action == 'auth':
```
paranoidi: Each of these if blocks should be own function, then you can call correct function from a dict like

    action_map = {'auth': action_auth, ...}
    action_map[options.action]()

cvium: Sure I can do that. I just moved the old function into a new file.

PR #1323, opened 2016-08-11
Added two new fields in Entry

queltosh:
### Motivation for changes:
I needed this two fields in order to be able to mark as seen and vote the episodes in tvshowtime and letterboxd websites automatically

plex_userrating: Rating given by the user. From 1 to 10
plex_key: Internal Plex identificator for the Media

### DIFF
--- a/flexget/plugins/input/plex.py
+++ b/flexget/plugins/input/plex.py
@@ -259,6 +259,8 @@ class InputPlex(object):
 
             e['plex_duration'] = node.getAttribute('duration')
             e['plex_summary'] = node.getAttribute('summary')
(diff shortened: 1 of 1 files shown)


liiight:
thanks. any chance you update the plugin wiki?

queltosh:
I'll try.

queltosh:
I think it is done. I have just updated the Entry page, the plugin page does not say anything about fileds.

liiight:
yeah, we should probably change that i guess. i think it makes more sense for the individual input plugins to declare their attributes than have it all under one big entry. thanks!

PR #1331, opened 2016-08-12
Limetorrents Search Plugin

OmegaZero:
### Motivation for changes:
Create a way to search limetorrents

```
Either
- limetorrents: yes
or
- limetorrents:
      order_by: (date|seeds)
      category: (all|anime|applications|games|movies|tv|other)

category defaults to all if not specified.  You may specify order_by and not category or vice versa.  You may only select one category.
```

### DIFF
--- a/flexget/plugins/sites/limetorrents.py
+++ b/flexget/plugins/sites/limetorrents.py
@@ -1,0 +1,113 @@
+from __future__ import unicode_literals, division, absolute_import
+from future.moves.urllib.parse import quote
+from builtins import *  # pylint: disable=unused-import, redefined-builtin
+
+import logging
+
+from flexget import plugin
+from flexget.entry import Entry
+from flexget.event import event
+from flexget.utils.requests import RequestException
+from flexget.utils.soup import get_soup
+from flexget.utils.search import torrent_availability
+from flexget.plugins.plugin_urlrewriting import UrlRewritingError
+
+log = logging.getLogger('limetorrents')
+
+class limetorrents(object):
+    """
+        limetorrents search plugin.
+    """
+
+    schema = {
+        'oneOf': [
+            {'type': 'boolean'},
+            {
+                'type': 'object',
+                'properties': {
+                    'category': {'type': 'string', 'enum': ['all', 'anime', 'applications', 'games', 'movies', 'music', 'tv', 'other'], 'default': 'all'},
+                    'order_by': {'type': 'string', 'enum': ['date', 'seeds'], 'default': 'date'}
+                },
+                'additionalProperties': False
+            }
+        ]
+    }
+
+    base_url = 'https://www.limetorrents.cc/'
+    errors = False
+
+
+    @plugin.internet(log)
+    def search(self, task, entry, config):
+        """
+            Search for entries on limetorents
+        """
+
+        if not isinstance(config, dict):
+            config = {'category': config}
+
+        order_by = ''
+        if isinstance(config.get('order_by'), str):
+            if config['order_by'] != 'date':
+                order_by = '{0}/1'.format(config['order_by'])
+
+        category = 'all'
+        if isinstance(config.get('category'), str):
+            category = '{0}'.format(config['category'])
+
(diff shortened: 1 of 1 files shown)


flexget/plugins/sites/limetorrents.py
```diff
@@ -0,0 +1,113 @@
...
+                    title = title[:title.rfind('-torrent')].replace('-', ' ')
+                    title = title[1:]
+
+                    data = row.findAll('td', attrs={'class': 'tdnormal'})
+                    size = str(data[1].text).replace(',','')
+
+                    seeds = int(row.find('td', attrs={'class': 'tdseed'}).text.replace(',',''))
+                    leeches = int(row.find('td', attrs={'class': 'tdleech'}).text.replace(',',''))
+
+                    if size.split()[1] == 'GB':
```
paranoidi: I think this bit is already in few places. We need it as generic utility ...

flexget/plugins/sites/limetorrents.py
```diff
@@ -0,0 +1,113 @@
...
+from flexget.entry import Entry
+from flexget.event import event
+from flexget.utils.requests import RequestException
+from flexget.utils.soup import get_soup
+from flexget.utils.search import torrent_availability
+from flexget.plugins.plugin_urlrewriting import UrlRewritingError
+
+log = logging.getLogger('limetorrents')
+
+class limetorrents(object):
```
cvium: I think pep8 dictates CamelCase for class names so it should be `Limetorrents`.

flexget/plugins/sites/limetorrents.py
```diff
@@ -0,0 +1,113 @@
...
+                },
+                'additionalProperties': False
+            }
+        ]
+    }
+
+    base_url = 'https://www.limetorrents.cc/'
+    errors = False
+
+
```
cvium: According to pep8 there should only be one newline between class properties/functions?

flexget/utils/tools.py
```diff
@@ -426,3 +426,24 @@ def get_latest_flexget_version_number():
 
 def get_current_flexget_version():
     return flexget.__version__
+
+def data_size_to_bytes(size):
+    """
+    Expands a data size into its bytes as an integer i.e. '1 MB' becomes 1048576
```
cvium: This is not entirely true assuming sane abbreviations.

1 MB is 1000000 bytes. 1 MiB is 1048576. I agree with @gazpachoking that the function should take an extra param that sets it to either base 2 or base 10.

If you need help getting these commits into its own branch/PR you know where to find us :)

flexget/utils/tools.py
```diff
@@ -426,3 +426,24 @@ def get_latest_flexget_version_number():
...
+
+    Args:
```
paranoidi: We use sphinx format for docstrings

flexget/plugins/sites/limetorrents.py
```diff
@@ -0,0 +1,112 @@
...
+        category = 'all'
+        if isinstance(config.get('category'), str):
+            category = '{0}'.format(config['category'])
+
+        entries = set()
+
+        for search_string in entry.get('search_strings', [entry['title']]):
+
+            query = 'search/{0}/{1}/{2}'.format(category,quote(search_string.encode('utf8')),order_by)
+            log.debug('Using search params: %s; category: %s; ordering by: %s',search_string, category, order_by or 'default')
```
cvium: This line looks quite long. Is it <= 120 chars?

flexget/plugins/sites/limetorrents.py
```diff
@@ -0,0 +1,114 @@
...
+import logging
+
+from flexget import plugin
+from flexget.entry import Entry
+from flexget.event import event
+from flexget.utils.requests import RequestException
+from flexget.utils.soup import get_soup
+from flexget.utils.search import torrent_availability
+
+log = logging.getLogger('Limetorrents')
```
cvium: Logger names are usually lowercase

PR #1336, opened 2016-08-14
API, lookup and estimator for Blu-ray.com

cvium:
### Motivation for changes:
Boredom
### Detailed changes:

- Blu-ray.com stuff

### TODO

- [x] Lookup
- [x] API
- [x] Estimator

### DIFF
--- a/flexget/plugins/estimators/est_released_movies_bluray.py
+++ b/flexget/plugins/estimators/est_released_movies_bluray.py
@@ -24,16 +24,18 @@ class EstimatesMoviesBluray(object):
 
         log.debug('Searching Blu-ray.com for release date of {} ({})'.format(movie_name, movie_year))
 
+        release_date = None
         try:
             with Session() as session:
                 lookup = get_plugin_by_name('api_bluray').instance.lookup
                 movie = lookup(title=movie_name, year=movie_year, session=session)
-                if movie and movie.release_date:
-                    log.debug('received release date: {0}'.format(movie.release_date))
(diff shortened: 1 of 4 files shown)


flexget/plugins/internal/api_bluray.py
```diff
@@ -0,0 +1,232 @@
...
+
+
+class BlurayMovie(Base):
+    __tablename__ = 'bluray_movies'
+
+    id = Column(Integer, primary_key=True, autoincrement=False, nullable=False)
+    name = Column(Unicode)
+    url = Column(Unicode)
+    _release_date = Column('release_date', DateTime)
+    release_date = text_date_synonym('_release_date')
```
liiight: Not familiar with this synonym, why not just use datetime?

flexget/plugins/estimators/est_released_movies_bluray.py
```diff
@@ -0,0 +1,41 @@
...
+
+class EstimatesMoviesBluray(object):
+
+    @plugin.priority(2)
+    def estimate(self, entry):
+        if 'movie_name' not in entry:
+            return
+
+        movie_name = entry['movie_name']
+        movie_year = entry.get('movie_year')
```
liiight: You could use the split title year util

cvium: To what end? The movie parser sets `movie_name` and `movie_year` fields.

PR #1338, opened 2016-08-15
Fix wrong umask calculation in download test

danielkza:
Calculating the umask by subtraction only works by coincidence in some cases. The correct way is to use bitwise operations to unset any bits in the umask from the expected permissions. I found this since I run with `umask 077` and the test always failed on my system.


cvium:
Is it only the test or also in the download plugin itself?

danielkza:
Very likely only the test, since it passes after the fix without changing the implementation at all.

cvium:
Alright, thanks

PR #1341, opened 2016-08-15
parse_filesize function and associated code cleanup

OmegaZero:
### Motivation for changes:
Many search plugins use the same math for parsing the file sizes.  Cherrypicking this into a function will make the code healthier.

### Detailed changes:

- Added parse_filesize to flexget/utils/tools.py and changed search plugins to use it.
- Added a flag `si` to disambiguate MB etc.


flexget/plugins/sites/site_1337x.py
```diff
@@ -63,12 +66,12 @@ def url_rewrite(self, task, entry):
 
         soup = get_soup(page.content)
 
-        magnetURL = str(soup.find('a',id='magnetdl').get('href')).lower()
-        torrentURL = str(soup.find('a',id='torrentdl').get('href')).lower()
+        magneturl = str(soup.find('a', id='magnetdl').get('href')).lower()
```
cvium: to be semantically pep8 compliant it should probably be `magnet_url` and `torrent_url`

flexget/plugins/sites/site_1337x.py
```diff
@@ -11,9 +11,11 @@
 from flexget.utils.requests import RequestException
 from flexget.utils.soup import get_soup
 from flexget.utils.search import torrent_availability
+from flexget.utils.tools import data_size_to_bytes
 from flexget.plugins.internal.urlrewriting import UrlRewritingError
 
-log = logging.getLogger('1337x')
+log = logging.getLogger('site_1337x')
```
cvium: plugin name is 1337x but logger is site_1337x?

flexget/utils/tools.py
```diff
@@ -426,3 +426,25 @@ def get_latest_flexget_version_number():
...
+    """
+    Expands a data size into its bytes as an integer i.e. '1 MB' becomes 1000000.
+    If base is specified then assume bibytes so i.e. '1 MB' becomes 1048576
+    """
+    if (base2):
+        split = 1024 ** 2
+    else:
+        split = 1
+
+    if size.split()[1] == 'GB' or size.split()[1 == "GiB"]:
```
cvium: `1 == "GiB"` will always be false :)

cvium: I might prefer `size.split()[1] in ['GB', 'GiB']` but dunno

flexget/utils/tools.py
```diff
@@ -426,3 +426,25 @@ def get_latest_flexget_version_number():
...
+    """
+    Expands a data size into its bytes as an integer i.e. '1 MB' becomes 1000000.
+    If base2 is specified then assume bibytes so i.e. '1 MB' becomes 1048576
+    """
+    if (base2):
+        split = 1024 ** 2
+    else:
+        split = 1000 ** 2
+
+    if size.split()[1] == 'GB' or size.split()[1] == "GiB":
```
gazpachoking: I'd probably do something like `unit, amount = size.split()` once, instead of referring to size.split()[0] and [1] each time.

flexget/utils/tools.py
```diff
@@ -426,3 +426,25 @@ def get_latest_flexget_version_number():
...
+        size = int(float(size.split()[0].replace(',', '')) * 1000 ** 3 / split)
```
gazpachoking: No reason to call `int(float(whatever))`, `int(whatever)` should be just the same.

gazpachoking:
I think MB vs MiB needs some work. Looks like it always divides by 1000 even in the case of MiB.

flexget/plugins/sites/piratebay.py
```diff
@@ -146,13 +147,9 @@ def search(self, task, entry, config=None):
...
                 size = re.search('Size ([\.\d]+)\xa0([GMK])iB', size)
-                if size:
-                    if size.group(2) == 'G':
-                        entry['content_size'] = int(float(size.group(1)) * 1000 ** 3 / 1024 ** 2)
-                    elif size.group(2) == 'M':
-                        entry['content_size'] = int(float(size.group(1)) * 1000 ** 2 / 1024 ** 2)
-                    else:
-                        entry['content_size'] = int(float(size.group(1)) * 1000 / 1024 ** 2)
+
+                entry['content_size'] = parse_filesize(size)
```
gazpachoking: I don't think `size` is a string here.

flexget/plugins/sites/horriblesubs.py
```diff
@@ -22,8 +22,13 @@ class HorribleSubs(object):
     @staticmethod
     def horrible_entries(requests, page_url):
         entries = []
-        soup = get_soup(requests.get(page_url).content)
-        
+
+        try:
+            soup = get_soup(requests.get(page_url).content)
+            log.debug('requesting: %s', page_url)
```
cvium: https://github.com/Flexget/Flexget/blob/next/flexget/utils/requests.py#L240 <-- this already logs the request url

flexget/plugins/sites/horriblesubs.py
```diff
@@ -22,8 +22,13 @@ class HorribleSubs(object):
...
+        except RequestException as e:
+            log.error('HorribleSubs request failed: %s', e)
```
cvium: should raise pluginerror instead

flexget/plugins/sites/horriblesubs.py
```diff
@@ -27,7 +27,7 @@ def horrible_entries(requests, page_url):
             soup = get_soup(requests.get(page_url).content)
             log.debug('requesting: %s', page_url)
         except RequestException as e:
-            log.error('HorribleSubs request failed: %s', e)
```
cvium: Should keep this logging since you're returning instead of raising an exception :P

gazpachoking:
I think our 'content_size' entry field is supposed to be in MiB, rather than bytes. Not sure what the best fix is, we can either:
- Return MiB from the parse function
- divide the result of the parse function by 1024 when setting content_size field
- add a parameter to the parse function so that return unit can be specified (might be overkill, as we'll probably always use the same unit)
- change the definition of content_size field to be bytes (I don't like this one, bytes are less useful unit and it would introduce backward compatibility issues)

We should also add a docstring to the new parse function which explains what it does, returns, and what the base2 param does.

gazpachoking:
Consensus on IRC is to go with returning MiB from the parse function. That's been our standard so far, so it should keep everything working the same.

flexget/utils/tools.py
```diff
@@ -428,20 +428,16 @@ def get_current_flexget_version():
...
     """
     prefix_order = {'': 0, 'k': 1, 'm': 2, 'g': 3, 't': 4, 'p': 5}
 
     unit, amount = text_size.lower().split()
     if not unit.endswith('b'):
         raise ValueError('%s does not look like a file size' % text_size)
-    unit.rstrip('b')
-    if unit.endswith('i'):
-        base2 = True
-        unit.rstrip('i')
```
gazpachoking: Why remove this? This will cause e.g. 'MiB' to be parsed as if it were 'MB' if base2 is False.

flexget/utils/tools.py
```diff
@@ -428,20 +428,16 @@ def get_current_flexget_version():
     return flexget.__version__
 
 
-def parse_filesize(text_size, base2=False):
+def parse_filesize(text_size, base2=True):
     """
-    Expands a data size into its bytes as an integer i.e. '1 MB' becomes 1000000.
-    If base2 is specified then assume bibytes so i.e. '1 MB' becomes 1048576
+    Expands a data size into its bibytes as an integer i.e. '1 MB' becomes 1048576.
```
gazpachoking: bibytes aren't a thing :wink: There are bytes (which is what the function currently returns) and then there are kibi-, mebi-, gibibytes (in the base2 system.) We are thinking the return should be changed to return mebibytes though, since that is what we use in the content_size field. To achieve that we need to divide the return by 1024*1024, and update the docstring.

flexget/utils/tools.py
```diff
@@ -428,20 +428,16 @@ def get_current_flexget_version():
...
+    If base2 is false then assume bytes so i.e. '1 MB' becomes 1000000
```
gazpachoking: Our docstring format goes something like this for parameters:
```rst
:param bool base2: If True, possibly ambiguous units like KB, MB, GB will be assumed to be base 2 units,
    rather than base 10.
```

flexget/plugins/sites/site1337x.py
```diff
@@ -11,10 +11,12 @@
 from flexget.utils.requests import RequestException
```
gazpachoking: Why the removal of the underscore in the filename?

flexget/plugins/sites/ptn.py
```diff
@@ -84,13 +85,9 @@ def create_entries(self, soup, passkey=None, imdb_id=None):
...
             entry['search_sort'] = torrent_availability(entry['torrent_seeds'], entry['torrent_leeches'])
-            size, unit = row.find(title='Torrent size').text.split(' ')
-            if unit == 'GB':
-                entry['content_size'] = int(float(size) * 1024)
-            elif unit == 'MB':
-                entry['content_size'] = int(float(size))
-            elif unit == 'KB':
-                entry['content_size'] = int(float(size) / 1024)
+
+            entry['content_size'] = parse_filesize(str(row.find(title='Torrent size').text))
```
gazpachoking: Looks like this one needs `si=False`. Also, don't think we need the str() here.

flexget/plugins/sites/cpasbien.py
```diff
@@ -113,15 +114,10 @@ def search(self, task, entry, config):
...
-                        unit = sizefull[-2:]
-                        if unit == 'GB':
-                            entry['content_size'] = int(float(size) * 1024)
-                        elif unit == 'MB':
-                            entry['content_size'] = int(float(size))
-                        elif unit == 'KB':
-                            entry['content_size'] = int(float(size) / 1024)
+                        size = result.find('div', attrs={'class': re.compile('poid')}).text
+
+                        entry['content_size'] = parse_filesize(size)
```
gazpachoking: This one looks like `si=False` too.

flexget/utils/tools.py
```diff
@@ -426,3 +426,30 @@ def get_latest_flexget_version_number():
...
 def get_current_flexget_version():
     return flexget.__version__
+
+
+def parse_filesize(text_size, si=True):
+    """
+    Parses a data size and returns its value in megabytes
+
+    :param string text_size: string containing the data size to parse i.e. "5 GB"
+    :param bool base2: If True, possibly ambiguous units like KB, MB, GB will be assumed to be base 10 units,
```
gazpachoking: Switch name to 'si' here too.

flexget/utils/tools.py
```diff
@@ -426,3 +426,30 @@ def get_latest_flexget_version_number():
...
+    rather than the default base 2. i.e. if si then 50 GB = 47684 else 50GB = 51200
+
+    :returns: an float with the data size in megabytes
```
gazpachoking: We are returning mebibytes

flexget/plugins/sites/site1337x.py
OmegaZero: paranoidi changed the class name to remove the underscore as it wasn't pep8.  I changed the name of the file to match the class name

gazpachoking: Ahh. The standard is actually different in pep8 for module names though, I think it's better with the underscore in. https://www.python.org/dev/peps/pep-0008/#package-and-module-names (and paranoidi is also the one who named it with the underscore :wink:)

PR #1342, opened 2016-08-16
Reorder Quality plugin

cvium:
### Motivation for changes:
Sometimes I disagree with quality priorities for some sites. With this plugin I can change the ordering.

### Detailed changes:

- Added plugin and one test to make sure it sets value at task start and restores original value at task exit/abort

```
reorder_quality:
  webrip:
    above: hdtv
```

### DIFF
--- a/flexget/plugins/modify/quality_priority.py
+++ b/flexget/plugins/modify/quality_priority.py
@@ -25,8 +25,8 @@ class QualityPriority(object):
         'additionalProperties': {
             'type': 'object',
             'properties': {
-                'above': {'type': 'string'},
-                'below': {'type': 'string'}
+                'above': {'type': 'string', 'format': 'quality_requirements'},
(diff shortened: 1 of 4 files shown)


liiight:
Thus requires knowing the value for each quality component, doesn't it? That's not very user friendly imo

cvium:
Same can be said for plugin_priority

liiight:
yep, i agree. That one sucks too. Doesn't mean we should perpetuate it.

cvium:
Going to reopen against develop later.

grr:
> Thus requires knowing the value for each quality component, doesn't it? That's not very user friendly imo

couldn't manipulate plugin be used to swap the qualities to accomplish the same thing? i.e. if hdtv is found, replace with webrip, and vice versa.

liiight:
This plugin doesn't manipulate entry quality, it enables overriding the hard coded quality component values to create custom comparisons.

PR #1357, opened 2016-08-20
Daemon will now auto-reload config file if it changes

cvium:
### Motivation for changes:
Why not
### Detailed changes:

- On initialization, a sha1 hash of the config file is saved in the manager.
- Before task execution the sha1 hash of the config file is compared to the cached one. Calls `load_config()` if they don't match and `self.is_daemon` is `True`. If new config is invalid, the hash is still saved so it doesn't try to reload it again, but the config changes will be rolled back to the last valid one (as has always been the case)

### DIFF
--- a/flexget/manager.py
+++ b/flexget/manager.py
@@ -245,6 +245,7 @@ class Manager(object):
         task_names = self.tasks
         # Only reload config if daemon
         if self.is_daemon and self.config_file_hash != self.hash_config():
+            log.info('Config change detected. Reloading.')
             self.load_config(output_to_console=False)
         # Handle --tasks
         if options.tasks:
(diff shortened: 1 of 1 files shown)


flexget/manager.py
```diff
@@ -243,6 +245,9 @@ def execute(self, options=None, output=None, loglevel=None, priority=1):
             options_namespace.__dict__.update(options)
             options = options_namespace
         task_names = self.tasks
+        # Only reload config if daemon
+        if self.is_daemon and self.config_file_hash != self.hash_config():
```
liiight: Maybe check if `self.config_file_hash` isn't None too to avoid reload on init?

paranoidi:
Log message on reload. How about if new config is invalid? It should say so and continue using old loaded one.

flexget/manager.py
cvium: Once the config is loaded during init, the config_file_hash will be set in manager.

PR #1364, opened 2016-08-23
[WIP] Webui add movie

Tommatheussen:
### Motivation for changes:
This PR allows users to add a movie to any existing movie_list from the web UI.

### Detailed changes:
Movies section of the UI has a new input field, allowing users to type a movie name (maybe later imdb/trakt/... ids?). This name will be searched for on the API using several different providers.

pastedgraphic-9

Add button only enabled after list is selected
pastedgraphic-8

Scrollable content
pastedgraphic-7

### Addressed issues:
No fix, just addition

Test are still left to do

### TODO:

- [ ] Tests

- [x] Fix buggy interface (as seen in second screenshot)

- [x] Layout stuff

- [x] Loading animation

- [x] Update list from already open panel instead of opening new one (creates stacked panels right now)

- [x] Get actual search endpoint and use it (currently mocked)

- [x] Use currently selected movie list (the selected tab) as default list to add a movie to


paranoidi:
I hope using dropdown is not needed when adding movie to list. By default it should go to default list without any additional selection.

Tommatheussen:
No indeed, I'm planning on using the currently selected movie list as default for the dropdowns.
That way ppl still have the change to add it to ad different list if wanted

liiight:
What is a default list? That exists only in the CLI scope. Should that be configured in UI side?

PR #1375, opened 2016-08-27
Fix trakt_list submit

cvium:
### Motivation for changes:
Adding episodes to a trakt_list required a lookup which is unnecessary.
### Detailed changes:

- Reverted aaf1a69384586dc5bd18e42462693ac4d37b2bcb and fixed the bugs in a saner way.

### DIFF
--- a/flexget/plugins/internal/api_trakt.py
+++ b/flexget/plugins/internal/api_trakt.py
@@ -228,13 +228,13 @@ def upgrade(ver, session):
     return ver
 
 
-def get_entry_ids(entry, episode=False):
+def get_entry_ids(entry):
     """Creates a trakt ids dict from id fields on an entry. Prefers already populated info over lazy lookups."""
     ids = {}
     for lazy in [False, True]:
         if entry.get('trakt_movie_id', eval_lazy=lazy):
             ids['trakt'] = entry['trakt_movie_id']
-        elif not episode and entry.get('trakt_show_id', eval_lazy=lazy):
+        elif entry.get('trakt_show_id', eval_lazy=lazy):
             ids['trakt'] = entry['trakt_show_id']
         elif entry.get('trakt_episode_id', eval_lazy=lazy):
             ids['trakt'] = entry['trakt_episode_id']
--- a/flexget/plugins/internal/api_trakt.py
+++ b/flexget/plugins/internal/api_trakt.py
@@ -228,13 +228,13 @@ def upgrade(ver, session):
     return ver
 
 
-def get_entry_ids(entry):
+def get_entry_ids(entry, episode=False):
     """Creates a trakt ids dict from id fields on an entry. Prefers already populated info over lazy lookups."""
     ids = {}
     for lazy in [False, True]:
         if entry.get('trakt_movie_id', eval_lazy=lazy):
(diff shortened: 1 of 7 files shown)


liiight:
Looks good. Maybe add a specific test for it?

gazpachoking:
What fields were not getting populated by trakt list that requires a separate lookup?

liiight:
Trakt list will probably populate correctly, the thing is when trying to add to collection from non-trakt sources, like feed, filesystem etc.
The internal lookup is just a convinience to remove the need to manually add `trakt_lookup` from the user. I think there's nothing wrong with 3rd party lists calling their respective lookup, this is not the same situation as was with `movie_queue`

cvium:
Episode ids are not populated from trakt list.

paranoidi:
You need to do this consistently in ALL lists, not just one.

gazpachoking:
> Episode ids are not populated from trakt list.

Is that because trakt doesn't return them when getting the list? If it does return them, lets make it so trakt_list populates them.

> The internal lookup is just a convinience to remove the need to manually add trakt_lookup from the user. I think there's nothing wrong with 3rd party lists calling their respective lookup, this is not the same situation as was with movie_queue

Isn't this the exact same situation as movie queue? I agree hard coding trakt lookup isn't as bad here as it would be for other plugins, since this is also a trakt plugin, but I'm still not wild about it. It would be perfectly valid to do an imdb_lookup rather than trakt lookup then add to a trakt list (in fact, that's what I do.)

cvium:
@gazpachoking It only performs a lookup if no other ids are present.

gazpachoking:
> @gazpachoking It only performs a lookup if no other ids are present.

Yeah, I get that. I just like the lookups being explicit by user in config since we don't have any global system to configure them when they are hard coded into other plugins. As I said though, if everyone else agrees that adding this is a good compromise for convenience, I can be persuaded.

gazpachoking:
> You need to do this consistently in ALL lists, not just one.

Not exactly sure what you mean? You mean entry_list should also have a trakt_lookup hard coded in it? (I don't actually think you mean that, just looking for clarification.)

cvium:
The problem for me is that shows and movies can be submitted to trakt without any lookups, but episodes can't. That causes confusion for some.

cvium:
> > You need to do this consistently in ALL lists, not just one.

>Not exactly sure what you mean? You mean entry_list should also have a trakt_lookup hard coded in it? (I don't actually think you mean that, just looking for clarification.)

I assume he means imdb_list should do an imdb lookup etc.

gazpachoking:
> The problem for me is that shows and movies can be submitted to trakt without any lookups, but episodes can't. That causes confusion for some.

Oh? Trakt supports submitting season and episode numbers to add episodes to lists, if we don't fall back to that in trakt_list I think that should be added rather than forcing a lookup.

cvium:
@gazpachoking Where do you see that?

gazpachoking:
I had a closer look and agree something needs to be fixed, but I don't think a lookup is the right fix. I think this is actually the change I don't agree with aaf1a69384586dc5bd. An entry should be detected and submitted as an episode just based on series_name, season and episode if no further ids are readily available.

PR #1382, opened 2016-08-30
Created cached endpoint and util

liiight:
### Motivation for changes:
Allows local caching of remote resources (mainly images), to make UI work quicker

### Detailed changes:

- Added `cached_resource()` to utils. Return a tuple of `file_path` and `mimetype`.
- Added `/cached/` endpoint that return local resource via `flask.send_file`. Original mime-type is passed 

`/cached/` has two params:
- `url` - URL to cache, required.
- `force` - Force fetching, as opposed to load if exist

#### To Do:

- [x] Change UI `series` and `movies` view to use this.
- [ ] Tests.

#### Notes:

- ~~the `cached_resources` dir can get very big, very fast, not sure how best to handle it~~. Added a `max_size` value, set to 250MB. Will start trimming least accessed files if value is met when adding,


flexget/utils/tools.py
```diff
@@ -455,3 +461,26 @@ def parse_filesize(text_size, si=True):
     amount = float(amount.replace(',', ''))
     base = 1000 if si else 1024
     return (amount * (base ** order)) / 1024 ** 2
+
+
+def cached_resource(url, force=False):
+    """
+    Caches a remote resource to local filesystem. Return a tuple of local file name and mime type, use primarily
+    for API/WebUI.
+    :param url: Resource URL
```
paranoidi: Empty line needed before `:param`

PR #1399, opened 2016-09-08
Add possibility to exclude trackers from clean_transmission

oxc:
### Motivation for changes:

I want to exclude torrents from a certain tracker to be removed from transmission

Currently, the clean_transmission plugin allows only to **include** torrents from a certain tracker, not **exclude** them. This does not fit the use case where you want to keep seeding torrents from your private trackers up to a certain ratio, but not other ones.
### Detailed changes:
- Add `tracker_mode` property which allows you to specify whether you want to _include_ torrents that have a tracker matching the `tracker` pattern (current behaviour) or  _exclude_ them.

```
      # clean all torrents but those from .se trackers
      clean_transmission: 
        finished_for: 2 hours
        tracker: '.se$'
        tracker_mode: 'exclude'
```

### DIFF
--- a/flexget/plugins/clients/transmission.py
+++ b/flexget/plugins/clients/transmission.py
@@ -642,6 +642,7 @@ class PluginTransmissionClean(TransmissionBase):
         advanced.accept('boolean', key='transmission_seed_limits')
         advanced.accept('boolean', key='delete_files')
         advanced.accept('regexp', key='tracker')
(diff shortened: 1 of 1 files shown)


liiight:
Like @gazpachoking said in gitter, I vote into making `clean_transmission` into a regular output plugin that can operate on entries with input such as `from_transmission`. Then it could benefit from all filter plugins (among other things)  and wouldn't require custom stuff like this.

liiight:
any progress on this?

oxc:
I gues the feature made it through in #1482. Oh well...

PR #1400, opened 2016-09-08
Explain why a torrent was not cleaned from transmission

oxc:
### Motivation for changes:

I want to know why a certain torrent was not cleaned from transmission

Currently the clean_transmission plugin collects all the information that might or might not be relevant to cleaning a certain torrent from the queue, and then decides whether it will remove the torrent or not. This prevents the code from logging why the torrent was in fact not removed, because the specific reason is burried in the whole if-block.
### Detailed changes:
- restructure code to follow the "early return" principle, and skips a torrent as soon as it's decided it does not meet the required criteria.
- this allows to add log statements why exactly a torrent has not been removed.
### Note:

This change depends on #1399


oxc:
Since the pull request this is based upon was not merged, I'm not rebasing this on the current state. Let me know if you're still interested.

PR #1408, opened 2016-09-16
add ssl support to webserver

b2un0:
### Motivation for changes:

- remote access to flexget

### Detailed changes:

- add ssl support to webserver.py (cherrypy)

```
web_server:
  bind: 0.0.0.0
  port: 3539
  ssl_certificate: '/etc/ssl/private/myCert.pem'
  ssl_private_key: '/etc/ssl/private/myKey.key'
api: yes
webui: yes
```

### DIFF
--- a/flexget/webserver.py
+++ b/flexget/webserver.py
@@ -235,11 +235,11 @@ class WebServer(threading.Thread):
             host = '127.0.0.1'
 
         if self.ssl_certificate and self.ssl_private_key:
-            proto = 'https'
+            protocol = 'https'
         else:
-            proto = 'http'
+            protocol = 'http'
 
-        log.info('Web interface available at %s://%s:%s' % (proto, host, self.port))
+        log.info('Web interface available at %s://%s:%s' % (protocol, host, self.port))
 
         # Start the CherryPy WSGI web server
         cherrypy.engine.start()
--- a/flexget/webserver.py
+++ b/flexget/webserver.py
@@ -36,6 +36,8 @@ web_config_schema = {
             'properties': {
                 'bind': {'type': 'string', 'format': 'ipv4', 'default': '0.0.0.0'},
                 'port': {'type': 'integer', 'default': 3539},
+                'ssl_certificate': {'type': 'string', 'format': 'string', 'default': ''},
+                'ssl_private_key': {'type': 'string', 'format': 'string', 'default': ''},
             },
             'additionalProperties': False
         }
@@ -154,6 +156,8 @@ def setup_server(manager, session=None):
     web_server = WebServer(
         bind=web_server_config['bind'],
         port=web_server_config['port'],
+        ssl_certificate=web_server_config['ssl_certificate'],
+        ssl_private_key=web_server_config['ssl_private_key'],
     )
 
     _default_app.secret_key = get_secret()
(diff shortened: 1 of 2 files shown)


b2un0:
image

flexget/webserver.py
```diff
@@ -36,6 +36,8 @@
             'properties': {
                 'bind': {'type': 'string', 'format': 'ipv4', 'default': '0.0.0.0'},
                 'port': {'type': 'integer', 'default': 3539},
+                'ssl_certificate': {'type': 'string', 'format': 'string', 'default': ''},
+                'ssl_private_key': {'type': 'string', 'format': 'string', 'default': ''},
```
paranoidi: I believe in schema you can tell that both of these are required or neither.

stevezau:
Ah! this has been on the todo list for a while.. 

I think we should support the generation of a self signed certificate?

flexget/webserver.py
b2un0: they are not required, if you don't set them, the webserver starts without ssl.

b2un0:
`I think we should support the generation of a self signed certificate?`
self signed certificate even works

Tommatheussen:
What I believe @stevezau meant is provide a way to the user that allows Flexget to create self-signed certificates

liiight:
@stevezau if that is what you meant, allow flexget to create the ssl, I don't think we need that. 

I believe @paranoidi meant json schema dependencies: https://spacetelescope.github.io/understanding-json-schema/reference/object.html#dependencies

Specifically bidirectional dependencies.

flexget/webserver.py
```diff
@@ -213,12 +219,22 @@ def _start_server(self):
...
+            })
+
         try:
             host = self.bind if self.bind != "0.0.0.0" else socket.gethostbyname(socket.gethostname())
         except socket.gaierror:
             host = '127.0.0.1'
 
-        log.info('Web interface available at http://%s:%s' % (host, self.port))
+        if self.ssl_certificate and self.ssl_private_key:
+            log.info('Web interface available at https://%s:%s' % (host, self.port))
```
cvium: Let the logger handle string formatting ie. `log.info('Web interface available at https://%s:%s' , host, self.port)`

liiight:
Please set the schema to reflect a bidirectional dependency:
```
'depenedencies': {
  'ssl_certificate': ['ssl_private_key'],
  'ssl_private_key': ['ssl_certificate']
}
```

b2un0:
dependencies was added

flexget/webserver.py
b2un0: done

flexget/webserver.py
```diff
@@ -232,9 +235,11 @@ def _start_server(self):
             host = '127.0.0.1'
 
         if self.ssl_certificate and self.ssl_private_key:
-            log.info('Web interface available at https://%s:%s' % (host, self.port))
+            proto = 'https'
```
cvium: Might as well call it `protocol`.

flexget/webserver.py
```diff
@@ -232,9 +235,11 @@ def _start_server(self):
...
         else:
-            log.info('Web interface available at http://%s:%s' % (host, self.port))
+            proto = 'http'
+
+        log.info('Web interface available at %s://%s:%s' % (proto, host, self.port))
```
cvium: Let the logger handle the string formatting here as well

liiight:
You forgot to change the logger formatting, that's the only thing left

b2un0:
i does not understand what you mean?!

liiight:
Change:
`log.info('Web interface available at %s://%s:%s' % (protocol, host, self.port))`
Into:
`log.info('Web interface available at %s://%s:%s' ,protocol, host, self.port)`
The logger module knows when it needs to interpolate the strings itself and it's considered a good practice.

b2un0:
ah, okay.

i just used the original line and add the protocol parameter;
https://github.com/Flexget/Flexget/blob/develop/flexget/webserver.py#L221

liiight:
yeah, it wasn't right before ;-) might as well do it right the second time.

b2un0:
finally done :)

liiight:
thanks for this

PR #1412, opened 2016-09-18
Release groups in internal parser

cvium:
### Motivation for changes:

Nice information to have.
### Detailed changes:

This is pretty much just a PoC. Some of the code is weird, some of it should be changed etc.
- Changed the series parser to work more like the movie parser in that it cuts the parsed bits. This means we'll end up with a string that contains the release group and very little else.

### DIFF
--- a/flexget/plugins/parsers/parser_common.py
+++ b/flexget/plugins/parsers/parser_common.py
@@ -15,9 +15,35 @@ log = logging.getLogger('parser')
 SERIES_ID_TYPES = ['ep', 'date', 'sequence', 'id']
 
 
+def remove_leading_trailing_junk(data):
+    """
+    Removes any leading and trailing junk characters such as comma, dot etc.
+
+    :param str data: data to be stripped of junk
+    :return str: stripped str
+    """
+    allowed_junk = ['[', ']']
+    # remove leading junk
+    for c in data:
+        if c.isalpha() or c.isdigit() or c in allowed_junk:
+            break
+        data = data[1:]
+    # remove trailing junk
+    for c in reversed(data):
+        if c.isalpha() or c.isdigit() or c in allowed_junk:
+            break
+        data = data[:-1]
+    return data
+
+
 def extract_group(data):
+    """
+
+    :param str data: String to extract release group from. Assumes everything has been removed (titles, qualities etc)
+    :return str: Returns the release group if found
+    """
     if not data.strip():
-        return None, None
+        return
     data_parts = re.split('[\s\.]', data)
     illegal_groups = ['mkv', 'avi', 'mp4']
     # start from the end
@@ -34,24 +60,22 @@ def extract_group(data):
         # remove leading and closing bracket
         if stripped_part[0] == '[' and stripped_part[-1] == ']':
             stripped_part = stripped_part[1:-1]
+
+        # Remove any brackets missing an opening/closing. Assumed to only have one.
         if '[' not in stripped_part and ']' in stripped_part:
             stripped_part = stripped_part.replace(']', '')
         if '[' in stripped_part and ']' not in stripped_part:
             stripped_part = stripped_part.replace('[', '')
 
-        # remove leading junk
-        for c in stripped_part:
-            if c.isalpha() or c.isdigit():
-                break
-            stripped_part = stripped_part[1:]
+        stripped_part = remove_leading_trailing_junk(stripped_part)
+
         if not stripped_part:
             continue
 
         match = re.search('^([A-Za-z0-9_\-\[\]]+)$', stripped_part)
         if match and len(match.group(1)) > 1:
             potential = match.group(1)
-            return potential, ''
-    return None, None
+            return potential
 
 
 def clean_value(name):
--- a/flexget/plugins/parsers/parser_common.py
+++ b/flexget/plugins/parsers/parser_common.py
@@ -15,6 +15,62 @@ log = logging.getLogger('parser')
 SERIES_ID_TYPES = ['ep', 'date', 'sequence', 'id']
 
 
+def extract_group(data):
+    if not data:
+        return None, None
+    data_parts = re.split('[\s\.]', data)
+    illegal_groups = ['mkv', 'avi', 'mp4']
+    # start from the end
+    for part in reversed(data_parts):
+        stripped_part = part
+        if not part or part in illegal_groups:
+            continue
+        # remove leading and closing bracket
+        if stripped_part[0] == '[' and stripped_part[-1] == ']':
+            stripped_part = stripped_part[1:-1]
+        elif '[' not in stripped_part and ']' in stripped_part:
+            stripped_part = stripped_part.replace(']', '')
+        elif '[' in stripped_part and ']' not in stripped_part:
+            stripped_part = stripped_part.replace('[', '')
+
+        # remove leading junk
+        for c in stripped_part:
+            if c.isalpha() or c.isdigit():
+                break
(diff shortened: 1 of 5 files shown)


flexget/utils/titles/series.py
```diff
@@ -408,31 +405,76 @@ def parse(self, data=None, field=None, quality=None):
...
+            self.quality = quality
+
+        # try to extract release group
+        self.group = extract_group(groups_data_stripped)
+
+        # allow group(s)
+        if self.allow_groups:
+            for group in self.allow_groups:
+                group = group.lower()
+                if self.group and group in self.group:
```
cvium: This line should probably be `if self.group and group == self.group` instead.

flexget/plugins/parsers/parser_common.py
```diff
@@ -15,6 +15,69 @@
...
+
+        if not stripped_part:
+            continue
+
+        match = re.search('^([A-Za-z0-9_\-\[\]]+)$', stripped_part)
+        if match and len(match.group(1)) > 1:
+            potential = match.group(1)
+            return potential
+
+
```
gazpachoking: I think this stuff should probably be in flexget.utils.titles.parser rather than here

flexget/tests/test_movieparser.py
```diff
@@ -21,37 +21,46 @@ def test_parsing(self, parse):
...
+        assert movie.group == 'flexget'
 
         movie = parse('TRON.Legacy.3D.2010.1080p.BluRay.Half.Over-Under.DTS.x264-FlexGet')
         assert movie.name == 'TRON Legacy', 'failed to parse %s' % movie.data
+        assert movie.group == 'flexget'
 
         movie = parse('[SomeThing]Up.2009.720p.x264-FlexGet')
         assert movie.name == 'Up', 'failed to parse %s (got %s)' % (movie.data, movie.name)
         assert movie.year == 2009, 'failed to parse year from %s' % movie.data
+        assert movie.group in ['flexget[something]', 'flexget']
```
gazpachoking: We aren't defining one correct result? Either is fine?

flexget/tests/test_seriesparser.py
```diff
@@ -44,23 +45,27 @@ def test_non_proper(self, parse):
...
         assert s.proper_count == 2
+        assert s.group is None
 
     def test_basic(self, parse, parse_invalid):
         """SeriesParser: basic parsing"""
         parse_invalid(name='Something Interesting', data='The.Something.Interesting.S01E02-FlexGet')
 
         s = parse(name='25', data='25.And.More.S01E02-FlexGet')
         assert s.valid, 'Fix the implementation, should be valid'
+        #assert s.group == 'flexget'
```
gazpachoking: What's the deal with these commented ones?

flexget/utils/titles/movie.py
```diff
@@ -76,10 +79,17 @@ def parse(self, data=None):
...
+        data = re.sub(r'^(\[(?:.*?)\])(.*)', r'\2', data)
 
-        for char in '[]()_,.':
+        for char in '()_,.':
             data = data.replace(char, ' ')
+            prefix = prefix.replace(char, ' ')
+        data = data.replace('[', ' ')
+        data = data.replace(']', ' ')
+
+        data += prefix
```
gazpachoking: Can we clean this up a bit, and update the comments to reflect what's happening? I think it's splitting a bracketed prefix, cleaning the 2 parts separately, then rejoining, no?

flexget/utils/titles/series.py
```diff
@@ -245,98 +244,87 @@ def parse(self, data=None, field=None, quality=None):
...
                 # Subtract 5 to leave room for fastsub propers before the normal release
                 self.proper_count -= 5
                 data_parts.remove(part)
+                groups_data_stripped = groups_data_stripped.replace(part, '')
             elif part in self.specials:
                 self.special = True
                 data_parts.remove(part)
+                groups_data_stripped = groups_data_stripped.replace(part, '')
 
         data_stripped = ' '.join(data_parts).strip()
```
gazpachoking: What is the difference between data_stripped and groups_data_stripped at this point?

flexget/utils/titles/series.py
```diff
@@ -245,98 +244,87 @@ def parse(self, data=None, field=None, quality=None):
...
                 if self.strict_name:
                     if date_match['match'].start() > 1:
-                        return
-                self.id = date_match['date']
-                self.id_groups = date_match['match'].groups()
-                self.id_type = 'date'
-                self.valid = True
-                if not (self.special and self.prefer_specials):
-                    return
+                        found_identifier = True
```
gazpachoking: It looks like self.valid was being used similarly to how you are using found_identifier. What is the function of tracking it separately? Also, why does it need to continue here now instead of just return?

flexget/tests/test_movieparser.py
cvium: I think it's guessit vs internal

flexget/tests/test_seriesparser.py
cvium: guessit failing

flexget/utils/titles/series.py
cvium: I think the problem was the way everything useful was stripped in `data_stripped` like dashes and such.

flexget/utils/titles/series.py
cvium: Not entirely sure about that line

flexget/utils/titles/movie.py
paranoidi: Preferably have methods/functions with a name and docstring

flexget/utils/titles/series.py
```diff
@@ -245,98 +244,87 @@ def parse(self, data=None, field=None, quality=None):
...
-                    self.episodes = 1
-                self.id_type = 'ep'
-                self.valid = True
-                if not (self.special and self.prefer_specials):
-                    return
+                        found_identifier = True
+                        matched_string = ep_match['match'].group(0)
+
+                if not found_identifier:
+
```
paranoidi: weird empty line

flexget/utils/titles/series.py
```diff
@@ -245,98 +244,87 @@ def parse(self, data=None, field=None, quality=None):
...
+                    if ep_match['end_episode'] and ep_match['end_episode'] > ep_match['episode'] + 2:
+                        # This is a pack of too many episodes, ignore it.
+                        log.debug('Series pack contains too many episodes (%d). Rejecting',
+                                  ep_match['end_episode'] - ep_match['episode'])
+                        found_identifier = True
+                        matched_string = ep_match['match'].group(0)
+                    if not found_identifier:
+
```
paranoidi: another weird empty line

flexget/utils/titles/series.py
```diff
@@ -408,31 +405,76 @@ def parse(self, data=None, field=None, quality=None):
...
+                msg += 'any series numbering.'
+            else:
+                msg += 'a(n) `%s` style identifier.' % self.identified_by
+            raise ParseWarning(self, msg)
+
+        # Remove episode id
+        if found_identifier and matched_string:
+            groups_data_stripped = re.sub(matched_string.replace(' ', '.*?'), '', groups_data_stripped)
+
+        # Find quality and clean from data
```
paranoidi: This method has grown beyond hope, consider extracting all the "# do stuff" segments into own methods

cvium:
I've been busy lately, so work on this will be slow, but I'll get it done eventually. Just FYI. (or someone else pick it up)

Issue #1409, opened 2016-09-16
Seriejunkies rewrite crashing


kasperj:
I commented line 168 to 171 and all works fine,
no time to have a deeper look right now,
maybe someone else has

```
2016-09-16 21:34 DEBUG    urlrewriter   serienjk_de_local Checking 2 entries
2016-09-16 21:34 DEBUG    urlrewriter   serienjk_de_local Url rewriting http://serienjunkies.org/gotham/gotham-staffel-2-blu-ray-sd720p1080p/
2016-09-16 21:34 DEBUG    utils.requests serienjk_de_local Fetching http://serienjunkies.org/gotham/gotham-staffel-2-blu-ray-sd720p1080p/
2016-09-16 21:34 DEBUG    serienjunkies serienjk_de_local Title seems to describe multiple episodes
2016-09-16 21:34 DEBUG    plugin        serienjk_de_local decorator caught ValueError. handled traceback:
Traceback (most recent call last):
  File "/usr/local/lib/python2.7/dist-packages/flexget/plugin.py", line 118, in wrapped_func
    return func(*args, **kwargs)
  File "/usr/local/lib/python2.7/dist-packages/flexget/plugins/sites/serienjunkies.py", line 116, in parse_downloads
    if not self.check_language(episode_lang):
  File "/usr/local/lib/python2.7/dist-packages/flexget/plugins/sites/serienjunkies.py", line 169, in check_language
    languages = languages[:languages.index("+")]
ValueError: substring not found
2016-09-16 21:34 WARNING  urlrewriter   serienjk_de_local URL rewriting serienjunkies failed: substring not found
```

cvium:
The string it's looking for has changed. It's not longer separate by a +. Also not very nice code. Would be better to just do `languages.split('+', 1)[0]`

PR #1416, opened 2016-09-22
fix serienjunkies language detection #1409

b2un0:
fix serienjunkies language detection #1409

### DIFF
--- a/flexget/plugins/sites/serienjunkies.py
+++ b/flexget/plugins/sites/serienjunkies.py
@@ -166,7 +166,7 @@ class UrlRewriteSerienjunkies(object):
     def check_language(self, languages):
         # Cut additional Subtitles
         try:
-            languages = languages[:languages.index("+")]
+            languages = languages.split('+', 1)[0]
         except IndexError:
             pass
 


flexget/plugins/sites/serienjunkies.py
```diff
@@ -166,7 +166,7 @@ def find_all_titles(self, search_title):
     def check_language(self, languages):
         # Cut additional Subtitles
         try:
-            languages = languages[:languages.index("+")]
+            languages = languages.split('+', 1)[0]
```
cvium: This doesn't really fix anything as the problem is that they stopped using `+` as a delimiter.

b2un0: i self opened some days ago a similar thread with exactly the same problem as in #1409 here
http://discuss.flexget.com/t/url-rewriting-serienjunkies-failed-substring-not-found/2690

the fix in #1409 from you solved my problem!

image

cvium: It only fixes the issue because it hides it. `languages.split('+', 1)[0]` will always return something. If there is no `+` in the string, then it simply returns the whole thing.

PR #1428, opened 2016-10-01
Use task.requests for using configured proxies

StefftheEmperor:
### Motivation for changes:

I needed to use a proxy but the serienjunkies urlrewrite plugin did not use the configured ones

Set Proxy in config.yml


flexget/plugins/sites/serienjunkies.py
```diff
@@ -82,8 +82,8 @@ def url_rewrite(self, task, entry):
         log.debug('Download URL: %s', download_urls)
 
     @plugin.internet(log)
-    def parse_downloads(self, series_url, search_title):
-        page = requests.get(series_url).content
+    def parse_downloads(self, task, series_url, search_title):
```
cvium: Why not just give it the requests session directly?

liiight:
Reopen if relevant

PR #1432, opened 2016-10-03
Update link to download-npo site

Carpetsmoker:
This link is better since I "own" it, and has a much higher guarantee to keep
working in the future (I may stop using BitBucket, it may shut down, etc.).


cvium:
Seems more like a Wiki thing.

PR #1437, opened 2016-10-10
nfo_lookup plugin: Get metadata from nfo file to aid IMDB search

darcamo:
### Motivation for changes:

New Plugin: nfo_lookup

When searching for metadata with the imdb_lookup plugin and the entries coming
from the filesystem I had many wrong values, as well as many movies with no
match. The problem is that my files are named in portuguese and, although IMDB
lookup still works, it is less reliable since all it has is the
filename. Furthermore, sometimes there are different movies with the same name.

These movies are in my **Kodi library** and I already have
an ['nfo' file](http://kodi.wiki/view/NFO_files) for each of them containing,
among other things, the IMDB id of the correct movie. Since I couldn't find a
way to use this with flexget I decided to get my hands dirty.

### Detailed changes:

- Add a new 'nfo_lookup' plugin (in the metadata folder).
- Add a test file for it (`test_nfo_lookup.py`) including a folder with test
  data.
  
  
The idea is that this plugin can be used to add metadata from the nfo file to
the entry when the entries come from the filesystem input. It will also set the
'imdb_id' field of the entry such that if the imdb_lookup plugin is also used
it will be able to find the correct movie no matter how the movie file is
named. Currently it only works for movies (the nfo file must have a `<movie>`
tag) but it should be easy to extended to series.

The plugin will do the following:

- Get the name of the file associated with the entry
- Change the file extension to '.nfo', read the content of the nfo file 
- Add the metadata in the nfo file as 'nfo_something' fields in the entry
- Set the 'imdb_id' field of the entry

**Note**: This is working for me and I have implemented unittests for it, but
it should be tested by someone else with nfo files generated by a different
program than the one I'm using (MediaElch).

### Addressed issues:

None in particular besides my own needs. 

```
nfo_lookup: yes
```

Note that the input must be the filesystem plugin. That is, the entry needs to
have a 'filename' and a 'location' fields.

Ex. of a full configuration

```
tasks:
  test_task:
    filesystem:
      path: some_path/
      mask: '*.mkv'
    nfo_lookup: yes
    imdb_lookup: yes
    <some_output_plugin>
```

```
paste output here
```


flexget/plugins/metainfo/nfo_lookup.py
```diff
@@ -0,0 +1,241 @@
...
+from builtins import *  # pylint: disable=unused-import, redefined-builtin
+
+import logging
+import os
+import xml.etree.ElementTree as ET
+
+from flexget import plugin
+from flexget.event import event
+from flexget.plugins.metainfo.imdb_lookup import ImdbLookup
+# from flexget.utils.imdb import make_url as make_imdb_url
```
liiight: Unused imports

flexget/plugins/metainfo/nfo_lookup.py
```diff
@@ -0,0 +1,241 @@
...
+
+    # Fields that appear only once in the nfo file
+    single_value_fields = ["title", "originaltitle", "sorttitle", "rating",
+                           "year", "votes", "plot", "runtime",
+                           "id", "filenameandpath", "trailer",
+                           # thumb has multiple values, but we will get the first one
+                           "thumb"]
+    # Fields that can appear multiple times in the nfo file
+    multiple_value_fields = [
+        # "thumb",
```
liiight: Probably need to be removed

flexget/plugins/metainfo/nfo_lookup.py
```diff
@@ -0,0 +1,241 @@
...
+        for entry in task.entries:
+            # If this entry was obtained from the filesystem plugin it should
+            # have a filename field. If it does not have one then there is
+            # nothing we can do in this plugin.
+            filename = entry.get('filename')
+            location = entry.get('location')
+
+            # If there is no 'filename' field there is also no nfo file
+            if filename is None or location is None:
+                log.warning("Entry '{0}' didn't come from the filesystem plugin".format(entry.get('title')))
```
liiight: I think you could probably just return here, since there is no need to proceed if entry didn't originate from filesystem plugin.

flexget/plugins/metainfo/nfo_lookup.py
```diff
@@ -0,0 +1,241 @@
...
+                    # continue
+
+            # Populate the fields from the information in the .nfo file
+            self.lookup(entry, nfo_filename)
+
+    def lookup(self, entry, nfo_filename):
+        # If there is already data from a previous parse then we don't need to
+        # do anything
+        if entry.get('nfo_id') is not None:
+            log.warning("Entry '{0}' was already parsed by nfo_lookup. "
```
liiight: Maybe warning isn't the Right level for this log?

flexget/plugins/metainfo/nfo_lookup.py
```diff
@@ -0,0 +1,241 @@
...
+            else:
+                # We will perform an IMDB search later the fields add here might be
+                # overwritten
+                self.add_imdb_fields_to_dict(fields)
+
+        # Let's search metadata in IMDB. If we found an nfo file containing an
+        # ID we have already added the 'imdb_id' field and this will make the
+        # IMDB lookup below find the correct movie no matter what the title and
+        # filename are.
+        imdb_lookup = ImdbLookup()
```
liiight: You shouldn't use IMDb or any other hard coded lookups. Adding the fields to the entry should suffice. If the user wants to use any others lookup plugin he should so explicitly

flexget/plugins/metainfo/nfo_lookup.py
```diff
@@ -0,0 +1,241 @@
...
+        nfo_full_filename = os.path.splitext(location)[0] + '.nfo'
+
+        if os.path.isfile(nfo_full_filename):
+            return nfo_full_filename
+        else:
+            return None
+
+    @property
+    def movie_identifier(self):
+        """Returns the plugin main identifier type"""
```
liiight: This property method is not relevant for this plugin, since this is not a lookup identifier

flexget/plugins/metainfo/nfo_lookup.py
```diff
@@ -0,0 +1,241 @@
...
+        return 'nfo_id'
+
+
+@event('plugin.register')
+def register_plugin():
+    plugin.register(NfoLookup, 'nfo_lookup', api_ver=2, groups=['movie_metainfo'])
```
liiight: This should not be in the `movie_metainfo` group,

flexget/plugins/metainfo/nfo_lookup.py
darcamo: Originally I was going to get other fields such as actors and directors as well from the nfo file. However, the nfo file only has the names and I would have to get the IMDB id of each of them to add to the fields `imdb_actors` and `imdb_directors`. 

For fields such as title and plot the nfo file might have different information, such as the translated title and plot. On the other hand, the names of actors and directors will be the same and thus I commented out this part to let the IMDB search fill the data.

If IMDB search is not performed anymore as part of this plugin then probably this should be kept and improved.

flexget/plugins/metainfo/nfo_lookup.py
darcamo: The reason I'm not returning here is that I'm currently using this plugin as a
replacement to imdb_lookup and thus if I returned here there would be no
lookup.

cvium: That's just bad design though

flexget/plugins/metainfo/nfo_lookup.py
darcamo: I'm actually new to flexget and looking at the source I saw some plugins defining this property. It could be something important that I didn't see in the documentation and that's why I have included. I will remove this then.

flexget/plugins/metainfo/nfo_lookup.py
darcamo: The nice thing about doing like this is that I can use something like the make_html plugin and get a nice html output, where some data not in the nfo file come from the IMDB search (which will find the correct movie due to setting imdb_id before the search) and the data in the nfo file overwrites what was obtained from the search. This way I can have all the metadata with only the things I care from the nfo file, such as the title (now in my language) and the plot.

I agree that it is better to not do a hard lookup and use this plugin and the
imdb_plugin together to get everything. However, the nfo_lookup plugin needs to
act before imdb_lookup in order to set the `imdb_id` field (otherwise
imdb_lookup will find the wrong movie, since all it has is the filename). But,
when I was calling imdb_lookup it was overwriting all the imdb fields and thus
when I set the `imdb_title`, `imdb_plot_outline` and other fields in the
nfo_lookup plugin it was overwritten by the subsequent imdb_lookup. That's why
I decided to not using the imdb_lookup plugin and instead call it inside nfo_lookup.

Now I'm thinking that my mistake was trying to set the corresponding
`imdb_something` fields so that make_html works nicely. Maybe a better approach
is to set only `imdb_id` to help the imdb_lookup plugin (and maybe other
similar IDs such as for trakt) and only add fields corresponding to
`nfo_something`. Then If I want to use make_html I could create a different
jinja template that will use the corresponding `nfo_something` field if
available. If it is not available then fallback to `imdb_something`.

Is this reasonable?

darcamo:
Overall I expected the code could have design problems, since I'm new to flexget and there are many details I still need to clarify.

For now I have removed the  movie_identifier property and the movie_metainfo group in the plugin registration.

Should I commit the "cassettes" created by the tests? I haven't used this before and I saw that there are many of them in the repository. On the other hand, If I refactor this to not call IMDB lookup anymore then I guess I would not have them anymore.

flexget/plugins/metainfo/nfo_lookup.py
darcamo: Maybe setting only `imdb_id` to help the imdb_lookup plugin and making sure that nfo_lookup executes before the imdb_lookup plugin would be a better option. 

Doing things as I did was a good first step for me to learn how things work and also use the make_html plugin, which works nicely with entries containing IMDB fields.

paranoidi:
@darcamo Some issues are to be expected, but don't sweat it. This is quite well made for first PR :)

flexget/plugins/metainfo/nfo_lookup.py
```diff
@@ -0,0 +1,234 @@
...
+
+    WARNING: This plugin will read a file with extension '.nfo' and the same
+    name as the entry filename as an XML file using xml.etree.ElementTree from
+    the standard python library. As such, it is vulnerable to XML
+    vulnerabilities described in the link below
+    https://docs.python.org/3/library/xml.html#xml-vulnerabilities
+
+    Use this only with nfo files you have created yourself.
+    """
+    # xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx
```
paranoidi: no separator comments in python code, empty lines where needed

flexget/plugins/metainfo/nfo_lookup.py
```diff
@@ -0,0 +1,234 @@
...
+
+            # If there is no 'filename' field there is also no nfo file
+            if filename is None or location is None:
+                log.warning("Entry '{0}' didn't come from the filesystem plugin".format(entry.get('title')))
+                nfo_filename = None
+            else:
+                nfo_filename = self.get_nfo_filename(entry)
+                # Stop if there is no nfo file
+                if nfo_filename is None:
+                    log.warning(
```
paranoidi: logged usage should not use string formatting
```
log.warn('aaa %s', foo)
```

flexget/plugins/metainfo/nfo_lookup.py
```diff
@@ -0,0 +1,234 @@
...
+            u'nfo_title': u'imdb_name',
+            u'nfo_originaltitle': u'imdb_original_name',
+            # u'nfo_votes' u'imdb_votes',
+            # u'nfo_year': u'imdb_year',
+            u'nfo_plot': u'imdb_plot_outline',
+            # u'nfo_rating': u'imdb_score',
+            # u'nfo_thumb': u'imdb_photo'
+        }
+
+        # multiple_field_mapping = {
```
paranoidi: remove leftovers

flexget/plugins/metainfo/nfo_lookup.py
liiight: yeah, if `imdb_id` is populated it will be used for lookup by `imdb_lookup`. to make it run before lookups look into the `@priority` decorator

darcamo:
I think I have implemented the suggested changes.
Except for "generic to match any format of log file". The nfo file used by Kodi
has a know set of XML tags.

I have separated the code concerning reading the nfo file into a `NfoReader`
class, while the `NfoLookup` class is in charge of updating the entries. I will
improve `NfoReader` in the future to also handle series.

Right now there a dictionary "_fields" in `NfoReader` specifying which XML tags
should be read from the nfo file. One possible improvement to make the plugin
more generic is to allow specifying more tags that should be read and add them
to this dictionary.

For something other then XML files a different `SomethingReader` class could be
implemented, if desirable (but in that case wouldn't a different be better?).

I have also committed the cassettes and now all the tests are passing (there is
one test failing for python 2.7 in travis, but it is in a different plugin).

flexget/plugins/metainfo/nfo_lookup.py
```diff
@@ -0,0 +1,256 @@
...
+            # If this entry was obtained from the filesystem plugin it should
+            # have a filename field. If it does not have one then there is
+            # nothing we can do in this plugin.
+            filename = entry.get('filename')
+            location = entry.get('location')
+
+            # If there is no 'filename' field there is also no nfo file
+            if filename is None or location is None:
+                log.warning("Entry %s didn't come from the filesystem plugin",
+                            entry.get('title'))
```
cvium: Unnecessary line break here. We "allow" up to 120 characters.

flexget/plugins/metainfo/nfo_lookup.py
```diff
@@ -0,0 +1,256 @@
...
+                return
+            else:
+                # This will be None if there is no nfo file
+                nfo_filename = self.get_nfo_filename(entry)
+                if nfo_filename is None:
+                    log.warning(
+                        "Entry %s has no corresponding %s file",
+                        entry.get('title'), self.nfo_file_extension)
```
cvium: Here as well

flexget/plugins/metainfo/nfo_lookup.py
```diff
@@ -0,0 +1,256 @@
...
+            # that at this point `nfo_filename` has the name of an existing
+            # '.nfo' file
+            self.lookup(entry, nfo_filename)
+
+    def lookup(self, entry, nfo_filename):
+        # If there is already data from a previous parse then we don't need to
+        # do anything
+        if entry.get('nfo_id') is not None:
+            log.warning("Entry %s was already parsed by nfo_lookup and it "
+                        "will be skipped. ", entry.get('title'))
```
cvium: Again

flexget/plugins/metainfo/nfo_lookup.py
```diff
@@ -0,0 +1,256 @@
...
+        str
+            The file name of the 'nfo' file, or None it there is no 'nfo' file.
+        """
+        location = entry.get('location')
+        nfo_full_filename = os.path.splitext(location)[0] + self.nfo_file_extension
+
+        if os.path.isfile(nfo_full_filename):
+            return nfo_full_filename
+        else:
+            return None
```
cvium: This is unnecessary. It will return None without the Else branch.

flexget/plugins/metainfo/nfo_lookup.py
```diff
@@ -0,0 +1,256 @@
...
+            tree = ET.parse(filename)
+            root = tree.getroot()
+        except ParseError:
+            raise BadXmlFile()
+
+        if os.path.exists(filename):
+            self._nfo_filename = filename
+            self._root = root
+        else:
+            raise BadXmlFile()
```
cvium: Seems a bit silly to use the same exception for ParseError and if the file doesn't exist. At the very least you should give it a proper error message.

flexget/plugins/metainfo/nfo_lookup.py
```diff
@@ -0,0 +1,256 @@
...
+    def _extract_single_field(self, name, getter_func):
+        """
+        Use this method to get fields from the root XML tree that only appear once,
+        such as 'title', 'year', etc.
+        """
+        f = self._root.find(name)
+        if f is not None:
+            return getter_func(f)
+        else:
+            return None
```
cvium: Again, not needed

flexget/plugins/metainfo/nfo_lookup.py
```diff
@@ -0,0 +1,256 @@
...
+        Use this method to get fields from the root XML tree that can appear more
+        than once, such as 'actor', 'genre', 'director', etc. The result will
+        be a list of values.
+        """
+        values = [getter_func(i) for i in self._root.findall(name)]
+
+        if len(values) > 0:
+            return values
+        else:
+            return None
```
cvium: .

flexget/plugins/metainfo/nfo_lookup.py
```diff
@@ -0,0 +1,256 @@
...
+
+        # TODO: Right now it only works for movies
+        if self._root.tag != 'movie':
+            return d
+
+        for name, values in self._fields.items():
+            multiple_bool = values[0]
+            getter_func = values[1]
+
+            nfo_field_name = u'nfo_{0}'.format(name)
```
cvium: I don't think the 'u' is necessary. Everything should already be unicode.

flexget/tests/test_nfo_lookup.py
```diff
@@ -0,0 +1,328 @@
+#!/usr/bin/env python
```
cvium: Why is this here?

flexget/tests/test_nfo_lookup.py
```diff
@@ -0,0 +1,328 @@
...
+    """.format(base)
+
+    def test_nfo_with_only_id(self, execute_task):
+        # run the task
+        task = execute_task('test_1')
+        for entry in task.entries:
+            # Get all 'nfo' keys in the entry. In this test the info file only
+            # has the 'id' field
+            nfo_keys = sorted([i for i in entry.keys() if i[:3] == 'nfo'])
+            assert nfo_keys == [u'nfo_id']
```
cvium: The 'u' should not be needed on any of the strings

flexget/tests/test_nfo_lookup.py
```diff
@@ -0,0 +1,328 @@
...
+            # Get all 'nfo' keys in the entry
+            nfo_keys = sorted([i for i in entry.keys() if i[:3] == 'nfo'])
+            assert nfo_keys == [u'nfo_id', u'nfo_plot', u'nfo_title']
+
+            assert entry['title'] == u'A Bela e a Fera'  # This will be == to filename
+
+            # Check that the 'nfo_id' field is set to the correct movie
+            assert entry['nfo_id'] == 'tt2316801'
+            assert entry['nfo_title'] == 'A Bela e a Fera'
+            assert entry['nfo_plot'] == u"Um romance inesperado floresce depois que a filha mais nova de um mercador em dificuldades se oferece para uma misteriosa besta com a qual seu pai ficou endividado."
```
cvium: This line is too long

flexget/tests/test_nfo_lookup.py
```diff
@@ -0,0 +1,328 @@
...
+
+            assert entry['nfo_actor'][4]['name'] == u"Myriam Charleins"
+            assert entry['nfo_actor'][4]['role'] == u"Astrid"
+
+            assert entry['nfo_country'] == ["France", "Germany"]
+            # Single director, but still a list
+            assert entry['nfo_director'] == ["Christophe Gans"]
+            assert entry['nfo_genre'] == ["Fantasia", "Romance"]
+            assert entry['nfo_originaltitle'] == u"La Belle et la Bête"
+            assert entry['nfo_plot'] == u"Um romance inesperado floresce depois que a filha mais nova de um mercador em dificuldades se oferece para uma misteriosa besta com a qual seu pai ficou endividado."
```
cvium: Too long

flexget/plugins/metainfo/nfo_lookup.py
```diff
@@ -0,0 +1,256 @@
...
+    Class in charge of parsing the '.nfo' file and getting a dictionary of
+    fields.
+
+    The '.nfo' file is an XML file. Some fields can only appear once, such as
+    'title', 'id', 'plot', etc., while other fields can appear multiple times
+    (with different values), such as 'thumb', 'genre', etc. These fields are
+    listed in the `_fields` attribute.
+    """
+    def __init__(self, filename):
+        from xml.etree.ElementTree import ParseError
```
cvium: Why is this import not in global scope?

flexget/plugins/metainfo/nfo_lookup.py
darcamo: This exception is only used here when, for any reason, we can't read information from the filename that was specified. Therefore, NfoLookup can capture it and return without changing the entry. It's better then raising a general `Exception`, but I didn't think too much about it.

darcamo:
I have implemented the new suggestion.

Now that I understand more about the code I see that I could not set `imdb_id`
at all and use the manipulate plugin to set `imdb_id` from the `nfo_id` field
to allow `imdb_lookup` to find the correct match. I tested and it works, but I
kept the code setting `imdb_id` because I don't see a drawback and I think it
is easier for users like this (manipulate plugin is an advanced plugin, after
all).

flexget/tests/test_nfo_lookup.py
```diff
@@ -297,7 +273,9 @@ def test_nfo_with_all_fields(self, execute_task):
             assert entry['nfo_director'] == ["Christophe Gans"]
             assert entry['nfo_genre'] == ["Fantasia", "Romance"]
             assert entry['nfo_originaltitle'] == u"La Belle et la Bête"
-            assert entry['nfo_plot'] == u"Um romance inesperado floresce depois que a filha mais nova de um mercador em dificuldades se oferece para uma misteriosa besta com a qual seu pai ficou endividado."
+            assert entry['nfo_plot'] == (u"Um romance inesperado floresce depois que a filha mais nova de um mercador "
```
cvium: Missed a couple of u's

cvium:
I don't see anything wrong with setting imdb_id if that's what you're parsing. Just make sure it's valid.

flexget/tests/test_nfo_lookup.py
darcamo: Searched u' and replaced with '.
I forgot u".

Done.

flexget/plugins/metainfo/nfo_lookup.py
```diff
@@ -0,0 +1,226 @@
...
+        except BadXmlFile:
+            log.warning("Invalid '.nfo' file for entry %s", entry.get('title'))
+            return
+
+        entry.update(fields)
+
+        # If the IMDB id was found in the nfo file, set the imdb_id field of the entry. This will help the imdb_lookup
+        # plugin to get the correct data if it is also used.
+        if 'nfo_id' in fields:
+            entry.update({'imdb_id': fields['nfo_id']})
```
cvium: Maybe you should use some regex or whatever to verify that it is a valid imdb_id. There is probably a helper function somewhere in the imdb plugins.

darcamo: There is `extract_id` in `utils.imdb.py`, but it matches "title IDs" (start with "tt") and "people IDs" (start with "nm"). Therefore, I added a `is_valid_imdb_title_id` method that only matches title IDs. The regex is simple: Just 'tt' followed by 7 digits.

flexget/plugins/metainfo/nfo_lookup.py
```diff
@@ -104,6 +108,13 @@ def get_nfo_filename(self, entry):
         if os.path.isfile(nfo_full_filename):
             return nfo_full_filename
 
+    @staticmethod
+    def is_valid_imdb_title_id(value=None):
```
cvium: I think it makes more sense to put it in utils/imdb

darcamo: I didn't want to touch code in other files, but that was were I first looked to see if such a function existed. I'll move the function to there and create a similar one for "person ID".

flexget/utils/imdb.py
```diff
@@ -37,22 +37,26 @@ def is_imdb_url(url):
...
     return re.match('tt[\d]{7}', value) is not None
 
 
-def is_valid_imdb_person_id(value=None):
+def is_valid_imdb_person_id(value):
     """
     Return True if `value` is a valid IMDB ID for a person.
     """
-    if not isinstance(value, basestring):
+    if value is None:
```
cvium: Why this check?

flexget/utils/imdb.py
```diff
@@ -37,6 +37,26 @@ def is_imdb_url(url):
     return re.match(r'https?://[^/]*imdb\.com/', url)
 
 
+def is_valid_imdb_title_id(value=None):
+    """
+    Return True if `value` is a valid IMDB ID for titles (movies, series, etc).
+    """
+    if not isinstance(value, basestring):
+        return False
```
darcamo: I'm now raising a `TypeError` in case `value` is not a string. However, I think it is better to not raise an exception and instead return `False` if `value` is `None`. The reason is that we can write something as

```
if is_valid_imdb_title_id(entry.get('nfo_id')):
    do_some_work_using_nfo_id_field()
```

In case the entry does not have the field then `entry.get` returns `None`,
`is_valid_imdb_title_id` returns `False` and we don't try to use the field
value. If `is_valid_imdb_title_id` raises an exception for `None` then everyone would have to test for that.

flexget/utils/imdb.py
darcamo: You are fast!
I was writing the message explaining the reason and after I finished I saw a new comment.

The reason is that we can write something as

```
if is_valid_imdb_title_id(entry.get('nfo_id')):
    do_some_work_using_nfo_id_field()
```

In case the entry does not have the field then `entry.get` returns `None`,
`is_valid_imdb_title_id` returns `False` and we don't try to use the field
value. If `is_valid_imdb_title_id` raises an exception for `None` then everyone
would have to test for that.

flexget/utils/imdb.py
cvium: Or you could just do `entry.get('nfo_id', '')` and avoid the None value altogether. It's not the responsibility of this function to hold the developer's hand.

darcamo: Makes sense. I'll make the change.

liiight:
thanks for this! you wanna update the wiki?

darcamo:
Sure.
I'll have a look on it.

2016-10-19 7:06 GMT-03:00 Or Carmi <notifications@github.com>:

PR #1438, opened 2016-10-10
add ssl chain support to webserver #1417

### LINKED ISSUE #1417: Web server SSL chain
### Expected behaviour:

Web Server SSL should work on all browsers, with certificates from Let's Encrypt
### Actual behaviour:

Some browser say the SSL certificate is invalid
### Steps to reproduce:
- Step 1: Configure Web server with SSL (Let's Encrypt with Android Chrome was the problem for me)
### Additional information:

I noticed this is because the intermediate certificates cannot be resolved correctly. Adding the chain to the config (from cherrypy) seems to fix this issue.
Cherrypy supports adding the chain in the config directly (http://docs.cherrypy.org/en/latest/deploy.html#ssl-support), it might be a good idea if we can add this as an optional param to the web_server config?
- Flexget Version: 2.3.31
- Python Version: 2.7.6
- Installation method: pip
- OS and version: Ubuntu 14.04

b2un0:
### Addressed issues:

- Fixes #1417


liiight:
Can you resync this please?

PR #1441, opened 2016-10-11
support size unit of "B". if sizes doesn't match log and ignore.

lostsubstance:
some results were so small they reported bytes and the regex didn't match. (resulting in an exception being thrown) This makes missed parses of size not trigger an exception, as well as supporting Bytes.

### DIFF
--- a/flexget/plugins/sites/piratebay.py
+++ b/flexget/plugins/sites/piratebay.py
@@ -151,7 +151,7 @@ class UrlRewritePirateBay(object):
                     if size:
                         entry['content_size'] = parse_filesize(size.group(1))
                     else:
-                        log.error('Malformed search result? Title: "%s", No size? %s' % (entry['title'], sizeText))
+                        log.error('Malformed search result? Title: "%s", No size? %s', (entry['title'], sizeText))
 
                 entries.add(entry)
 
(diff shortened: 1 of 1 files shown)


flexget/plugins/sites/piratebay.py
```diff
@@ -145,10 +145,13 @@ def search(self, task, entry, config=None):
...
-                size = re.search('Size (\d+(\.\d+)?\xa0(?:[PTGMK])iB)', size)
-
-                entry['content_size'] = parse_filesize(size.group(1))
+                sizeText = link.find_next(attrs={'class': 'detDesc'}).get_text()
+                if sizeText:
+                    size = re.search('Size (\d+(\.\d+)?\xa0(?:[PTGMK])?i?B)', sizeText)
+                    if size:
+                        entry['content_size'] = parse_filesize(size.group(1))
+                    else:
+                        log.error('Malformed search result? Title: "%s", No size? %s' % (entry['title'], sizeText))
```
cvium: Let the logger handle string formatting. Use a comma instead of %

flexget/plugins/sites/piratebay.py
```diff
@@ -151,7 +151,7 @@ def search(self, task, entry, config=None):
                     if size:
                         entry['content_size'] = parse_filesize(size.group(1))
                     else:
-                        log.error('Malformed search result? Title: "%s", No size? %s' % (entry['title'], sizeText))
+                        log.error('Malformed search result? Title: "%s", No size? %s', (entry['title'], sizeText))
```
cvium: That's not gonna work. You need to remove the parenthesis around the variables. This is gonna crash since you're passing a tuple.

cvium:
Thanks

PR #1446, opened 2016-10-13
Add extra fields in entries from HTML input to allow filtering later

darcamo:
### Motivation for changes:

Currently the entries returned by the HTML input plugin have basically the URL
field. Some more data could be useful to allow more flexible filtering in the
filter phase in flexget. I my particular case, I need to gather only links that
are in h2 headers, but at the filtering phase that information is lost. This
pull request adds the name of the parent and grandparent of the link, as well
as the list of classes for the link, the parent and the grandparent nodes.

With this, suppose the links you are interested in appear in `<h2>`
headers. You could then use the 'if' filter as

```
test-task:
  html: http://example.com
  if:
    - "html_parent == 'h2'": accept
  <some_output_plugin>
```
### Detailed changes:

Add the following fields to entries from the html input plugin:
- `html_classes`: Classes of the anchor tag containing the returned link
- `html_parent`: The element containing the anchor tag (a string)
- `html_parent_classes`: The classes of the parent (this is a list of strings)
- `html_grandparent`: The element containing the parent of the anchor tag (a string)
- `html_grandparent_classes`: The classes of the grandparent (this is a list of strings)

The fields with classes are only added if there is at least one class.

There is no change in configuration. On the other hand, if adding these fields
is considered an overhead then an "add_extra_fields" option could be added to
html plugin config.

### DIFF
--- a/flexget/plugins/input/html.py
+++ b/flexget/plugins/input/html.py
@@ -275,6 +275,25 @@ class InputHtml(object):
             entry['url'] = url
             entry['title'] = title
 
+            # If the element has no classes then None is returned
+            link_classes = link.get('class')
+            if link_classes:
+                entry['html_classes'] = link_classes
+
+            # Get the parent and grandparent HTML nodes of the link node. We
+            # will add extra fields to the entry with the tag and classes of
+            # these nodes to allow filtering the entry in subsequent phases.
+            html_parent = link.parent
+            entry['html_parent'] = html_parent.name
+            parent_classes = html_parent.get('class')
+            if parent_classes:
+                entry['html_parent_classes'] = parent_classes
+            html_grandparent = html_parent.parent
+            entry['html_grandparent'] = html_grandparent.name
+            grandparent_classes = html_grandparent.get('class')
+            if grandparent_classes:
+                entry['html_grandparent_classes'] = html_grandparent.get('class')
+
             if 'username' in config and 'password' in config:
                 entry['download_auth'] = (config['username'], config['password'])
 


flexget/plugins/input/html.py
```diff
@@ -275,6 +275,25 @@ def title_exists(title):
...
+
+            # Get the parent and grandparent HTML nodes of the link node. We
+            # will add extra fields to the entry with the tag and classes of
+            # these nodes to allow filtering the entry in subsequent phases.
+            html_parent = link.parent
+            entry['html_parent'] = html_parent.name
+            parent_classes = html_parent.get('class')
+            if parent_classes:
+                entry['html_parent_classes'] = parent_classes
+            html_grandparent = html_parent.parent
```
cvium: What happens if it doesn't have a parent or grandparent?

paranoidi:
If we add this information to entry .. what else "can be useful" and when does it stop? This seems like a never ending quest

darcamo:
These are the fields that I think would cover most of the common cases. In my case I only needed "html_parent" and I added the other fields biased by my own personal interpretation of when to stop. I'm not so sure that this is the best approach, but without some extra (structure) information available when the html input plugin is doing it work, filtering the desired entries later might be very difficult or impossible. Creating a pull request seemed like a good idea to discuss the idea and either go forward with it or change to a better approach.

The particular website that I'm trying to get links from has more data that I want to add to the entry, such as the imdb_id, and the actual magnet links are in a download page (I need to request the download page and only then get the magnet URL to put in the entry). In the end I created a whole input plugin, but now it is specific to one website.

One generic and flexible approach could be adding an option to html plugin to specify a python file that would be imported and defines a function with a know signature to do the data gathering. Something that would be called inside `InputHhtml.create_entries` with the `soup` as argument and that should return a dictionary with data to add to the entry (the returned dictionary should have at least the 'title' and 'url' keys). Does this make sense?

**Note:** If this is better discussed in the forums than here I can close the
pull request and create a post there.

paranoidi:
I think some sort of notation for getting extra info would be much more preferable than hard coded specific cases. Just from top of my head:

```
html:
  url: ....
  fields:
    html_parent: parent
    html_parent_classes: parent.classes
    html_grandparent: parent.parent
```

PR #1449, opened 2016-10-18
Prepare v2.3.47.dev

Amiganer:
### Motivation for changes:
csv input not working on python 3

### Detailed changes:
ERROR: needs byte string in PY2 and unicode in PY3
- 

### Addressed issues:
input_csv.py
- Fixes # .
from __future__ import unicode_literals, division, absolute_import
from builtins import *  # pylint: disable=unused-import, redefined-builtin
from future.utils import PY3

import logging
import csv

from requests import RequestException

from flexget import plugin
from flexget.utils.cached_input import cached

log = logging.getLogger('csv')

class InputCSV(object):
    """
        Adds support for CSV format. Configuration may seem a bit complex,
        but this has advantage of being universal solution regardless of CSV
        and internal entry fields.

        Configuration format:

        csv:
          url: <url>
          values:
            <field>: <number>

        Example DB-fansubs:

        csv:
          url: http://www.dattebayo.com/t/dump
          values:
            title: 3  # title is in 3th field
            url: 1    # download url is in 1st field

        Fields title and url are mandatory. First field is 1.
        List of other common (optional) fields can be found from wiki.
    """

    schema = {
        'type': 'object',
        'properties': {
            'url': {'type': 'string', 'format': 'url'},
            'values': {
                'type': 'object',
                'additionalProperties': {'type': 'integer'},
                'required': ['title', 'url']
            }
        },
        'required': ['url', 'values'],
        'additionalProperties': False
    }

    @cached('csv')
        entries = []
        try:
            r = task.requests.get(config['url'])
        except RequestException as e:
            raise plugin.PluginError('Error fetching `%s`: %s' % (config['url'], e))
        # CSV module needs byte strings, we'll convert back to unicode later
        if PY3:
            page = r.text.splitlines()
        else:
            page = r.text.encode('utf-8').splitlines()
        for row in csv.reader(page):
            if not row:
                continue
            entry = Entry()
            for name, index in list(config.get('values', {}).items()):
                try:
                    # Convert the value back to unicode
                    if PY3:
                        entry[name] = row[index - 1].strip()
                    else:
                        entry[name] = row[index - 1].decode('utf-8').strip()
                except IndexError:
                    raise plugin.PluginError('Field `%s` index is out of range' % name)

            entries.append(entry)
        return entries

@event('plugin.register')
def register_plugin():
    plugin.register(InputCSV, 'csv', api_ver=2)

```
paste_config_here
```
```
paste output here
```

### DIFF
--- a/flexget/_version.py
+++ b/flexget/_version.py
@@ -7,4 +7,4 @@ The version should always be set to the <next release version>.dev
 The jenkins release job will automatically strip the .dev for release,
 and update the version again for continued development.
 """
-__version__ = '2.3.46'
+__version__ = '2.3.47.dev'


Amiganer:
I add an test to PY3 in order to not use the byte conversion.

cvium:
Wrong branch. All PRs must be to Develop.

liiight:
dude, that is really not how you open a pr...

cvium:
I think https://help.github.com/articles/creating-a-pull-request-from-a-fork/ is more relevant

PR #1450, opened 2016-10-18
Getting CSV input on PY3 to work

Amiganer:
### Motivation for changes:
csv input was not working on PY3

### Detailed changes:
CSV reader in Python 2 is broken and does not support unicode strings. It therefore needs byte strings, but Python 3 needs unicode strings.

csv plugin has been changed to transform the string to the proper format based on Python version.

### Addressed issues:

http://discuss.flexget.com/t/input-csv-py-not-working-on-python3/2797


cvium:
Thanks

PR #1454, opened 2016-10-19
Support different TVDB languages lookup 

### LINKED ISSUE #1453: tvdb_lookup: support all/specified language
### Expected behaviour:
If using tvdb_lookup to look-up a TV show with a localized name, it should be able to find it. Either automagically, or by specifying the language to search for.

### Actual behaviour:
If using tvdb_lookup to look-up a TV show with its non-English name, it plainly fails.
It currently searches for English-name only.

### Steps to reproduce:
- Step 1: Search for 'Tegenlicht' (=Dutch) -> fails to find
- Step 2: Search for 'Backlight' -> that works, but we don't have that name...

### Additional information:

- Flexget Version: 2.40.
- See: https://github.com/Flexget/Flexget/pull/1334#issuecomment-254772015

liiight:
### Motivation for changes:
#1453 

### Detailed changes:

- Changed schema:

  ```python
  schema = {'oneOf': [
        {'type': 'boolean'},
        {'type': 'object',
         'properties': {
             'language': {'type': 'string', 'default': 'en'}
         }}
    ]}
  ```
- Sends `Accept-Language` header to relevant thetvdb API requests.

### Addressed issues:

- Fixes #1453

```yaml
tasks:
  test:
    mock:
      - {title: 'Tegenlicht.S010E01.HDTV.XViD-FlexGet'}
    series:
      - Tegenlicht
    thetvdb_lookup:
      language: nl
    mock_output: yes
```
Tests added

#### To Do:
- [x] Add API support.
- [x] Some more tests

### DIFF
--- a/flexget/api/plugins/tvdb_lookup.py
+++ b/flexget/api/plugins/tvdb_lookup.py
@@ -56,7 +56,7 @@ class ObjectsContainer(object):
             'overview': {'type': 'string'},
             'director': {'type': 'string'},
             'rating': {'type': 'number'},
-            'image': {'type': 'string'},
+            'image': {'type': ['string', 'null']},
             'first_aired': {'type': 'string'},
             'series_id': {'type': 'integer'}
         },
--- a/flexget/api/plugins/tvdb_lookup.py
+++ b/flexget/api/plugins/tvdb_lookup.py
@@ -87,7 +87,10 @@ tvdb_series_schema = api.schema('tvdb_series_schema', ObjectsContainer.tvdb_seri
 tvdb_episode_schema = api.schema('tvdb_episode_schema', ObjectsContainer.episode_object)
 search_results_schema = api.schema('tvdb_search_results_schema', ObjectsContainer.search_results_object)
 
-series_parser = api.parser()
+base_parser = api.parser()
+base_parser.add_argument('language', default='en', help='Language abbreviation string for different language support')
+
+series_parser = base_parser.copy()
 series_parser.add_argument('include_actors', type=inputs.boolean, help='Include actors in response')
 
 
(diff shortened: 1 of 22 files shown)


mfonville:
The `language: nl` can be a set of multiple languages that are passed in the accept-header?

liiight:
no, only a single language is permitted. so your input need to consist of the same language

liiight:
a default, not a fallback. if you look with a language header and there's no result, you'll get no data

mfonville:
ok, that will work for the npo_watchlist :-)
Everything there is in Dutch, but I can imagine that maybe with some future scrapers that could scrape television series from multiple-lingual countries (Belgium, Switzerland etc) sometimes the languages could be mixed. But I guess we will see how to tackle such a thing a that time ;-)

liiight:
cool, ill merge this and let me know any issues you encounter

mfonville:
It works perfectly for 'Tegenlicht' (Backlight), and it is able to find the episode information using the airdate with @jeroenl his recent enhancement.

But for some reason 'Zondag met Lubach' (Sunday with Lubach) it does not work. When I run in debug mode, I first see:
```
api_tvdb      Looking up tvdb information for Zondag met Lubach None
api_tvdb      Series <name=Zondag met Lubach,tvdb_id=None> not found in cache, looking up from tvdb.
thetvdb_lookup Error looking up tvdb episode information for Zondag met Lubach (VPWON_1261999): Unable to get search results for Zondag met Lubach: 404 Client Error: Not Found for url: https://api.thetvdb.com/search/series?name=Zondag+met+Lubach
```
and then a bit further in the log it does:
```
api_tvdb      Looking up tvdb information for Zondag met Lubach None
api_tvdb      Series <name=Zondag met Lubach,tvdb_id=None> not found in cache, looking up from tvdb.
utils.requests Fetching https://api.thetvdb.com/series/288799
api_tvdb      Looking up tvdb information for Zondag met Lubach 288799
api_tvdb      Series <name=Zondag met Lubach,tvdb_id=288799> information restored from cache.
```
so to it seems to first trying to search the episode information using the Dutch string, but not finding any results because it does not have either the English name or the tvdb_id yet.
And only as second step it starts to search and resolve using the Dutch language parameter, and finds the tvdb_id. After it has this tvdb_id the episode-searching goes ok.

We probably would have to enforce it somehow that it first searches for the tvdb_id using the localized name, before it tries to search for the specific episode (and/or allow that the episode searching also provides the language to search in).

Also a spin-off issue from this enhancement: I also use a trakt_lookup, but trakt only allows searching by the English name. Could the tvdb_lookup provide the English name that can be passed on to trakt?
Probably letting tvdb_lookup replace series_name with its English counterpart (and keep the localized/original around in another variable?) would resolve this issue, and the one above if we would be certain it runs as first.

Should I create new issues for this/these?

liiight:
The last issue, supplying english name in entry, is a good idea, please open an issue for that.
As for the first issue, I'm not sure I follow. 
When doing a lookup, we first search our local cache via the entry title. If not finding results we then look in tvdb. Only after finding results we cache the series with its original search title, so the next time it'll just retreive from cache.

liiight:
I looked in `trakt_lookup` and it should find a match based on provided `tvdb_id` if provided regardless of title. You only need to enforce `thetvdb_lookup` to happen first. Try using [`plugin_priority`](http://flexget.com/Plugins/plugin_priority) plugin.

mfonville:
ok, I will make a new issue for setting the English series_name;

For the priority/order, I do understand that with tvdb_id it should go ok, but sometimes it doesn't have that yet (like on 1st run).
Ofc it can be fixed by plugin_priority, but I think in essence we have still a small bug here: episode lookup does apparently not use the language-accept (or it should have found it). I am not sure if that is something that should be fixed in `tvdb_lookup` or whether it should be provided by `npo_watchlist`. But also a more robust mechanism would be if tvdb_lookup itself would internally structure to first search for series-details, before searching for the specific episode. That should avoid this race-condition and would make sure it always has `tvdb_id` if it exists.

liiight:
ok, please open a new issue and provide relevant config and logs

PR #1455, opened 2016-10-19
Writers in imdb lookup

Daniseijo:
### Motivation for changes:

I am trying to improve the current html.template with some extra info so the first step was to add the movie writers into the `imdb_lookup` plugin. I also used the movie writers to filter inside the imdb filter.

### Detailed changes:

- Added `imdb_writers` to the imdb_lookup plugin. It returns the main writers of the film.
- Added some code on the imdb parser to find the writers on the imdb page. They are always inside an span with `itemprop="creator"`
- Added `reject_writers` and `accept_writers` to the imdb filter file.

Once the `imdb_lookup` plugin is called, now the entry will have an `imdb_writers` label.

To use the new filters use the following commands:
```
# accept all movies by these writers
accept_writers:
  - nm0000318

# reject movies by these writers
reject_writers:
  - nm0093051
```


flexget/plugins/metainfo/imdb_lookup.py
```diff
@@ -52,6 +58,7 @@ class Movie(Base):
     genres = relation('Genre', secondary=genres_table, backref='movies')
     actors = relation('Actor', secondary=actors_table, backref='movies')
     directors = relation('Director', secondary=directors_table, backref='movies')
+    writers = relation('Writer', secondary=writers_table, backref='movies')
```
liiight: When you change the DB Schema, you need to bump the Base version and update the upgrade method.

flexget/utils/imdb.py
```diff
@@ -329,6 +330,15 @@ def parse(self, imdb_id):
                 director_name = None
             self.directors[director_id] = director_name
 
+        # get writer(s)
+        for writer in title_overview.select('[itemprop="creator"] > a'):
```
liiight: Can you add tests for this?

Daniseijo:
I think I messed it up with that last commit. I'm sorry but I am really new with all this github things. I added the tests and update de DB Schema version. If the final commit is wrong I can make a new pull request.

liiight:
all good, thanks for this. can you update the wiki where relevant please?

PR #1465, opened 2016-10-26
Improve internal series parser performance?

gazpachoking:
### Motivation for changes:

Currently, calls to the series parser from series plugin end up instantiating one parser per series per entry. Instantiating a new parser involves compiling regexes for that series again, which may be slow.
### Detailed changes:

Add an lru cache where internal parser keeps instantiated parsers for configured series. This means the parsers should be instantiated once per series in your config, rather than once per series per entry.
### Caveats and a Request

I haven't done any performance testing, this was just a gut reaction. If anyone would like to do some scientific performance testing, or just test this branch out with your (series heavy) config and post results it would be appreciated.

### DIFF
--- a/flexget/plugins/parsers/parser_internal.py
+++ b/flexget/plugins/parsers/parser_internal.py
@@ -1,6 +1,7 @@
 from __future__ import unicode_literals, division, absolute_import
 from builtins import *  # pylint: disable=unused-import, redefined-builtin
 
+import functools
 import logging
 import time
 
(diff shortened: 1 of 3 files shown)


cvium:
A while back I made a plugin that can test this rather easily. Task:
```
  mock_series:
    parsing:
      series: guessit
    mock_series:
      number: 100
      with_episode_id: yes
    configure_series:
      from:
        mock_series:
          number: 1500
      settings:
        identified_by: ep
```
Plugin: https://gist.github.com/cvium/eeb40652570dd8fc501b75cae29348f1

gazpachoking:
I did some very cursory testing without significant result. Thinking this isn't worth it.

PR #1475, opened 2016-10-30
Next 

liiight:
### Motivation for changes:

Next stuff
### Detailed changes:
- `move`/`copy` plugins:  
  - Changed `filename` option to `rename` so it'll be possible to override filename set by `filesystem` plugin
  - Entry render error will now raise `PluginError` instead of `PluginWarning` since that caused a situation where operation continued to unspecified location, if it was possible to fallback on entry values
- daemon changes:
  - `flexget daemon reload` has been renamed to `flexget daemon reload-config` to avoid confusion
  - `--config-autoreload` has been renamed to `--autoreload-config` to be more consistent with `reload-config`
  - `flexget daemon enable-autoreload` and `flexget daemon disable-autoreload` have been removed as their use was limited and ill-conceived.


flexget/manager.py
```diff
@@ -420,7 +420,7 @@ def daemon_command(self, options):
             self.ipc_server.start()
             self.task_queue.wait()
             fire_event('manager.daemon.completed', self)
-        elif options.action in ['stop', 'reload', 'status', 'enable-autoreload', 'disable-autoreload']:
+        elif options.action in ['stop', 'config-reload', 'status']:
```
liiight: shouldn't this be `reload-config` instead?

PR #1482, opened 2016-11-02
Preserve torrents based on presence of trackers

mannionpatrick:
### Motivation for changes:

Forum discussion about skipping cleanup for certain torrents based on the presence of a tracker in the torrent's tracker list

### Detailed changes:
Creation of a new config setting "preserve_tracker" and adding corresponding logic

### Addressed issues:

None

clean_transmission:
  host: localhost
  port: 9091
  username: myusername
  password: mypassword
  tracker: nyaa|animebytes
  preserve_tracker: alpharatio|what

### DIFF
--- a/flexget/plugins/clients/transmission.py
+++ b/flexget/plugins/clients/transmission.py
@@ -638,6 +638,7 @@ class PluginTransmissionClean(TransmissionBase):
         advanced.accept('boolean', key='transmission_seed_limits')
         advanced.accept('boolean', key='delete_files')
         advanced.accept('regexp', key='tracker')
+        advanced.accept('regexp', key='preserve_tracker')
         directories_re = advanced.accept('list', key='directories')
         directories_re.accept('regexp')
         return root
@@ -653,6 +654,7 @@ class PluginTransmissionClean(TransmissionBase):
         delete_files = bool(config['delete_files']) if 'delete_files' in config else False
         trans_checks = bool(config['transmission_seed_limits']) if 'transmission_seed_limits' in config else False
         tracker_re = re.compile(config['tracker'], re.IGNORECASE) if 'tracker' in config else None
+        preserve_tracker_re = re.compile(config['preserve_tracker'], re.IGNORECASE) if 'preserve_tracker' in config else None
         directories_re = config.get('directories')
 
         session = self.client.get_session()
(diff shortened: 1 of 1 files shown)


gazpachoking:
My first reaction is that I don't really like the current design of clean_transmission, and think we should move to a system where it can live in its own task with from_transmission input, and use normal filter plugins on that task to control what it cleans. The current model only begets this sort of option adding for internal filtering. That being said, I'm not sure if that should hold up this PR or not.

flexget/plugins/clients/transmission.py
```diff
@@ -682,7 +685,7 @@ def on_task_exit(self, task, config):
                                 is_minratio_reached or
                                 (is_torrent_seed_only and is_torrent_idlelimit_since_added_reached) or
                                 (not is_torrent_seed_only and is_torrent_idlelimit_since_finished_reached))
-                    and is_tracker_matching and is_directories_matching):
+                    and is_directories_matching (not is_preserve_tracker_matching and is_tracker_matching)):
```
gazpachoking: >  and is_directories_matching (not is_preserve_tracker_matching and is_tracker_matching)

is there meant to be an 'and' or an 'or' in between is_directories_matching and that parenthetical?

mannionpatrick: Whoops yes. Fixing now...

flexget/plugins/clients/transmission.py
```diff
@@ -682,7 +685,7 @@ def on_task_exit(self, task, config):
                                 is_minratio_reached or
                                 (is_torrent_seed_only and is_torrent_idlelimit_since_added_reached) or
                                 (not is_torrent_seed_only and is_torrent_idlelimit_since_finished_reached))
-                    and is_tracker_matching and is_directories_matching):
+                    and is_directories_matching and (not is_preserve_tracker_matching and is_tracker_matching)):
```
cvium: There's no need for that parenthesis.

cvium:
I don't think it makes sense to not allow these kinds of PRs. Small changes that improve QoL. Should Transmission plugins be refactored completely? Yes, but that's quite a daunting task for most, so who's to say when and if that will happen.

liiight:
Like we discussed in gitter, I'm @gazpachoking on this one. I think that any feature that is added to the current state of this plugin will just deter any work on modifying it.

marine1988:
I think thes feature is awesome i do download form alot of place but I only and to keep torrents from 2 or 3 trackers

PR #1494, opened 2016-11-10
Pending approval plugin

liiight:
### Motivation for changes:
A more generic approach to the `pending` idea, this will work with a single task, not list dependent.

### Detailed changes:

- Each entry in `task.entries` will be cached.
- If `db_entry.approved` is True entry will be accepted

Currently this operates inside a single task, but schema can take task name and try to inject it into any other task if needed. Not sure if that's useful or not, it's @gazpachoking idea ;-)

#### To Do:
- [x] CLI
- [x] API
- [x] Tests
- [x] DB Cleanup

### DIFF
--- a/flexget/plugins/filter/pending_approval.py
+++ b/flexget/plugins/filter/pending_approval.py
@@ -2,7 +2,7 @@ from __future__ import unicode_literals, division, absolute_import
 from builtins import *  # noqa pylint: disable=unused-import, redefined-builtin
 
 import logging
-from datetime import datetime
+from datetime import datetime, timedelta
 
 from flexget import db_schema, plugin
 from flexget.event import event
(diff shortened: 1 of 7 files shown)


flexget/plugins/filter/pending_approval.py
```diff
@@ -0,0 +1,95 @@
...
+    def on_task_input(self, task, config):
+        if not config:
+            return
+
+        approved_entries = []
+        with Session() as session:
+            # Let details plugin know that it is ok if this task doesn't produce any entries
+            task.no_entries_ok = True
+
+            for entry in task.entries:
```
gazpachoking: I'm thinking we should do this at the end of the filter phase, and only add entries that are accepted. At this point, we'd also reject any entries in the task that aren't approved in the db yet.

PR #1495, opened 2016-11-10
[WIP] Created daemon restart method

liiight:
### Motivation for changes:
An easy way to reload daemon.

### Detailed changes:

- Created `restart` method
- Renamed `manager,_shutdown` to `_cleanup`
- Made `wait` argument shared between `restart` and `stop` CLI actions
- Added `/server/restart/` API endpoint

Note: This currently only works when running `flexget_vanilla`, more work needed.

### DIFF
--- a/flexget/manager.py
+++ b/flexget/manager.py
@@ -324,7 +324,7 @@ class Manager(object):
         with self.acquire_lock():
             self.initialize()
             self.handle_cli()
-            self._shutdown()
+            self._cleanup()
 
     def handle_cli(self, options=None):
         """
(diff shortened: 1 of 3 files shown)


flexget/manager.py
```diff
@@ -916,7 +918,7 @@ def shutdown(self, finish_queue=True):
         fire_event('manager.shutdown_requested', self)
         self.task_queue.shutdown(finish_queue)
 
-    def _shutdown(self):
+    def _cleanup(self):
```
paranoidi: This rename is a bit inconvenient as it fires event shutdown ? Seems fine otherwise ...

liiight: this only works when running vanilla, making this universal seems a lot more tricky than I thought

gazpachoking:
I think using a native service manager is a better way to handle the daemon start/stop/restart. If there is some way we can make systemd/windows service/other thing easier to set up for users, (without adding too much maintenance burden on us,) I'd prefer that. Not sure if that should be better docs, or included helper scripts or what.

paranoidi:
I would vote to abandon this, being likely way too hard to do well.

PR #1498, opened 2016-11-13
Update trakt.tv api url

segelkma:
### Motivation for changes:
Old url was used during api version 2 development while the mein url had v1. v2 is finished development so it can be changed now. Old url may not be supported in future.
http://docs.trakt.apiary.io/#introduction/api-url

### Detailed changes:
- changed trakt.tv api url

### Addressed issues:
- none


cvium:
Sadly it's not that simple. You'll have to recreate all Trakt test cassettes as well

segelkma:
Oh my, I did not think for a second that this would require changes there. But having a look into the cassettes, it is pretty obvious why.
I'll have a look into it the next days.

segelkma:
Oops, I messed up No changes. Still new to git. Please close this pull request, I will make another one later.

PR #1499, opened 2016-11-14
Notifier interface and plugin

liiight:
### Motivation for changes:
Notification are currently just output plugins that operate only on `task.accepted`. This new interface and plugin allow to tweak that.

### Detailed changes:

- Maintain backwards compatibility, no upgrade actions required.
- Converted `pushover` plugin to use a new interface
- Added `render() ` to `Task` for consistent render calls. 
- Added a way to suppress missing phase plugin messages in log.
- Added traceback to `task.traceback` on crash.
- Created `notify` plugin that operates on these `notifiers` group plugins:
  - Can now select to iterate over any entry container, not just `accepted`.
  - Can set the `scope` to task, and send only 1 notification per task, regardless of entry containers
 - Operates on `exit` phase only.
- Created `notify_crash` plugin that operate `on_task_abort` and calls `notify` plugin with a hardcoded config to operate on task scope.

Schema:
```python
schema = {
        'type': 'object',
        'properties': {
            'to': {'type': 'array', 'items':
                {'allOf': [
                    {'$ref': '/schema/plugins?group=notifiers'},
                    {'maxProperties': 1,
                     'error_maxProperties': 'Plugin options within notify plugin must be indented '
                                            '2 more spaces than the first letter of the plugin name.',
                     'minProperties': 1}]}},
            'scope': {'type': 'string', 'enum': ['task', 'entries']},
            'what': one_or_more({'type': 'string', 'enum': ENTRY_CONTAINERS})
        },
        'required': ['to'],
        'additionalProperties': False
    }
```
Usage:
```
test_notify:
  mock:
    - {title: "foo", url: 'http://bla.com'}
    - {title: "bar2", url: 'http://bla2.com'}
  regexp:
    accept:
      - bar2
  notify:
    to:
      - pushover:
          userkey: userkey
          apikey: apikey
    what:
      - undecided
      - accepted
  scope: entries
  disable: [seen]
  manual: yes
```
Notify crash schema:
```python
    schema = {
        'type': 'object',
        'properties': {
            'to': {'type': 'array', 'items':
                {'allOf': [
                    {'$ref': '/schema/plugins?group=notifiers'},
                    {'maxProperties': 1,
                     'error_maxProperties': 'Plugin options within notify plugin must be indented '
                                            '2 more spaces than the first letter of the plugin name.',
                     'minProperties': 1}]}}

        },
        'required': ['to'],
        'additionalProperties': False
    }
```
Usage:
```yaml
notify_crash:
  to:
    - pushover:
        userkey: userkey
        apikey: apikey
        message: "Task has crashed {{task.abort_reason}} {{task.traceback}}"
```
`notify_crash` plugin could probably best be used in a global template.

#### To Do:
- [ ] Convert all notifier plugin to follow new interface:
  - [x] `notify_osd`
  - [x] `notify_sns`
  - [ ] `notify_xmpp`
  - [x] `notifymyandroid`
  - [ ] `prowl`
  - [x] `pushalot`
  - [x] `pushbullet`
  - [x] `rapidpush`
  - [ ] `send_email`
  - [x] `slack`
  - [ ] `sms_ru`
  - [ ] `telegram`
  - [x] `pushover`
-  [ ] Think of way to test this


flexget/plugins/output/notify.py
```diff
@@ -0,0 +1,98 @@
...
+from flexget import plugin
+from flexget.config_schema import one_or_more
+from flexget.event import event
+
+log = logging.getLogger('notify')
+
+HANDLED_PHASES = ['start', 'input', 'filter', 'output', 'exit']
+
+
+class Notify(object):
```
gazpachoking: Do we need a base class for these? Could we use the existing paradigm of registering notification plugins to a 'notification' group? Being in the notification group would mean you need to implement a 'send_notification' method, or somesuch, that gets passed the appropriate variables describing the desired notification.

flexget/plugins/output/notify.py
```diff
@@ -0,0 +1,98 @@
...
+
+    @staticmethod
+    def prepare_config(config):
+        if not isinstance(config['phase'], list):
+            config['phase'] = [config['phase']]
+        if not isinstance(config['what'], list):
+            config['what'] = [config['what']]
+        config.setdefault('what', ['accepted'])
+        config.setdefault('scope', 'entries')
+        config.setdefault('phase', ['output'])
```
gazpachoking: Can you think of any reason we'd need alternate phases? If not, I'd skip this as a config option and add it in later if we think of some use.

flexget/plugins/output/notify.py
liiight: sure, I can do that. why not base class though?

flexget/plugins/output/notify.py
liiight: I guess I was over zealous with phases, I'll remove

gazpachoking:
@liiight Yeah, I dig the general structure now. Should be very easy to implement new notifiers, as well as to send messages about new things.

liiight:
Closed and opened #1510 on public branch for easier testing

PR #1505, opened 2016-11-17
Allow user to specify extra keys wanted to be populated with from_deluge

Tommatheussen:
### Motivation for changes:
The from_deluge only loads a default set of keys, but Deluge can send a lot more.
Adding an optional keys property allows the user to specify which extra keys the plugin should load.

### Detailed changes:
- Added `keys` array property to `from_deluge`
- Can only be populated with a valid item from the `extra_settings_map` dict
- `from_deluge` will load the extra keys and perform conversions necessary (like seconds -> hours etc)

```
from_deluge:
  keys:
    - 'active_time'
    - 'remove_at_ratio'
    - ...
```
```
Nothing yet, can't test deluge at work, will provide tonight
```

### DIFF
--- a/flexget/plugins/clients/deluge.py
+++ b/flexget/plugins/clients/deluge.py
@@ -355,7 +355,10 @@ class InputDeluge(DelugePlugin):
                     else:
                         log.warning('Did not find torrent file at %s' % torrent_path)
                 for key, value in torrent_dict.items():
-                    flexget_key = self.settings_map[key]
+                    if key in self.settings_map:
+                        flexget_key = self.settings_map[key]
+                    else:
+                        flexget_key = self.extra_settings_map[key]
                     if isinstance(flexget_key, tuple):
                         flexget_key, format_func = flexget_key
                         value = format_func(value)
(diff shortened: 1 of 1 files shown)


flexget/plugins/clients/deluge.py
```diff
@@ -317,8 +367,12 @@ def on_get_torrents_status(torrents):
             client.disconnect()
 
         filter = config.get('filter', {})
+
+        extraKeys = config.get('keys', [])
```
paranoidi: pep8

flexget/plugins/clients/deluge.py
```diff
@@ -317,8 +367,12 @@ def on_get_torrents_status(torrents):
             client.disconnect()
 
         filter = config.get('filter', {})
+
+        extra_keys = config.get('keys', [])
+        combined_list = self.settings_map.keys() + extra_keys
```
cvium: I know deluge doesn't work in py3, but this particular line won't work in py3 either iirc. Need to convert `self.settings_map.keys()` to a list first.
Also, is there even a need for that variable, `extra_keys`? Couldn't it just be one line?

Tommatheussen: hmm, guess it's not really needed, not sure why I did it like that in the first place.

flexget/plugins/clients/deluge.py
```diff
@@ -262,6 +302,13 @@ def __init__(self):
                             }
                         },
                         'additionalProperties': False
+                    },
+                    'keys': {
+                        'type': 'array',
+                        'items': {
+                            'type': 'string',
+                            'enum': extra_settings_map.keys()
```
cvium: This line is failing the tests in py3 for the same reason. Need to convert it to a list because of how py3 changed dict.

liiight:
this is ready, right?

Tommatheussen:
Some fields might need additional formatting, but we could do that after it's merged as well

PR #1510, opened 2016-11-20
Notifier plugin and interface

liiight:
### Motivation for changes:
Notification are currently just output plugins that operate only on `task.accepted`. This new interface and plugin allow to tweak that.

### Detailed changes:

- Created `notify` plugin that operates on these `notifiers` group plugins:
  - Can now select to iterate over any entry container, not just `accepted`.
  - Can set the `scope` to task, and send only 1 notification per task, regardless of entry containers
 - Operates on `exit` phase only.
- Created `notify_crash` plugin that operate `on_task_abort` and calls `notify` plugin with a hardcoded config to operate on task scope.
- Added support to render file templates
- Added `render() ` to `Task` for consistent render calls. 
- Added a way to suppress missing phase plugin messages in log.
- Added traceback to `task.traceback` on crash.
- Created two new default file templates

Breaking changes:
- Renamed `notify_sns` to `sns`
- Renamed `notify_xmpp` to `xmpp`
- All notifier plugin must pass dict in schema, removed all other options (mainly string and bool).
- Changed almost all of the schema, added `file_template`

Schema:
```python
schema = {
        'type': 'object',
        'properties': {
            'to': {'type': 'array', 'items':
                {'allOf': [
                    {'$ref': '/schema/plugins?group=notifiers'},
                    {'maxProperties': 1,
                     'error_maxProperties': 'Plugin options within notify plugin must be indented '
                                            '2 more spaces than the first letter of the plugin name.',
                     'minProperties': 1}]}},
            'scope': {'type': 'string', 'enum': ['task', 'entries']},
            'what': one_or_more({'type': 'string', 'enum': ENTRY_CONTAINERS})
        },
        'required': ['to'],
        'additionalProperties': False
    }
```
Usage:
```
test_notify:
  mock:
    - {title: "foo", url: 'http://bla.com'}
    - {title: "bar2", url: 'http://bla2.com'}
  regexp:
    accept:
      - bar2
  notify:
    to:
      - pushover:
          userkey: userkey
          apikey: apikey
    what:
      - undecided
      - accepted
  scope: entries
  disable: [seen]
  manual: yes
```
Notify crash schema:
```python
    schema = {
        'type': 'object',
        'properties': {
            'to': {'type': 'array', 'items':
                {'allOf': [
                    {'$ref': '/schema/plugins?group=notifiers'},
                    {'maxProperties': 1,
                     'error_maxProperties': 'Plugin options within notify plugin must be indented '
                                            '2 more spaces than the first letter of the plugin name.',
                     'minProperties': 1}]}}

        },
        'required': ['to'],
        'additionalProperties': False
    }
```
Usage:
```yaml
notify_crash:
  to:
    - pushover:
        userkey: userkey
        apikey: apikey
        message: "Task has crashed {{task.abort_reason}} {{task.traceback}}"
```
`notify_crash` plugin could probably best be used in a global template.

#### To Do:
- [x] Convert all notifier plugin to follow new interface:
  - [x] `notify_osd`
  - [x] `notify_sns`
  - [x] `notify_xmpp`
  - [x] `notifymyandroid`
  - [x] `prowl`
  - [x] `pushalot`
  - [x] `pushbullet`
  - [x] `rapidpush`
  - [x] `email`
  - [x] `slack`
  - [x] `sms_ru`
  - [x] `telegram`
  - [x] `pushover`
- [ ] Figure out how to correctly merge using task templates with different scopes
- [x] Make attributes generic across all notifiers.
- [x] Enable using global attributes in addition to specific notifier ones
-  [x] Think of way to test this


paranoidi:
* Notify plugin should be the way to do multi recipient notifications, deprecate such features from some of the notification plugins.

* Ability to specify template in "notify" plugin so it does not need to be done in each notify plugin?

* Option "what" is almost never anything else than "accepted". Should not be mandatory and default to that?

* Task render shoud work with template filenames somehow?

Idea was that we should have nice build in templates in "templates/plugin_name/xxxx.html"

(ehh, looks like there are now completely different kind of templates there ... swagger-ui.html, api_response.html. It was NOT ment for UI templates.)

flexget/utils/template.py
```diff
@@ -229,4 +230,17 @@ def render_from_task(template, task):
     :param task: Task to render the template from.
     :return: The rendered template text.
     """
-    return render(template, {'task': task, 'now': datetime.now()})
+    variables = {'task': task, 'now': datetime.now(), 'task_name': task.name}
+    result = render(template, variables)
+    # Only try string replacement if jinja didn't do anything
```
paranoidi: We could start dropping support for string replacement format ... add warning for it?

liiight:
* `Notify plugin should probably be THE way to do multi recipient notifications, deprecate such features from some of the notification plugins?` **By multi recipient do you mean several apikeys per config? Because we do not offer a way to do that via the `notify` plugin. This kinda relates to the second point I guess**.
* `Ability to specify template in "notify" plugin so it does not need to be done in each notify plugin?` **That was my initial thinking, that each plugin config will constitute ONLY of its connection data, and it will not have anything related to the message data itself. Is that what you mean?**
* `Option "what" is almost never anything else than "accepted". Should not be mandatory and default to that?` **It's not mandatory and `accepted` is already the default** :smile: 
* `Task render should work with template filenames somehow?` **Yeah, I'm not sure how those work...**

paranoidi:
1) Yeah, like

```
  notify:
    to:
      - pushover:
          userkey: userkey
          apikey: apikey
      - pushover:
          userkey: another_userkey
          apikey: another_apikey
```

I believe some notifiers accept multiple recipients already via config, but this would make it unnecessary feature. Although it may be preferable in some cases.

2) Probably same idea, although I don't see example template use in description.

3) Excellent

4) Templates work a bit poorly ATM. Look at "output/html.py", which is still best implementation I believe. This needs to be unified somehow. So each plugin could have a default template in a filesystem, and a set of other choices. Build in ones should be possible to use simply with "name" rather than full path.

liiight:
1. This already works, `to` takes a list of plugins, they do not have to be unique. So both ways can be supported at the same time.
2 + 4. OK, i see how templates are used. Should be no problem adding a `template` attribute and rendering it. The thing that bugs me about having global attribute names which will remove all data handling from the notifer plugins themselves, is what if a user wanted to send several notification per `notify` plugin usage but with different data:
```yaml
notify:
  to:
    - pushover: <opts>
         message: "foo"
    - telegram: <opts>
        message: "bar"
```
but maybe that's not a common scenario...

flexget/utils/template.py
liiight: hmm, i didn't really pay attention to that. how is that even being used currently?

flexget/plugins/notifiers/notify.py
```diff
@@ -0,0 +1,161 @@
...
+        scope = config.pop('scope')
+        what = config.pop('what')
+        notifiers = config.pop('to')
+
+        # Creating iterator based on config scope
+        iterate_on = [getattr(task, container) for container in what] if scope == 'entries' else [[task]]
+
+        for notifier in notifiers:
+            for notifier_name, notifier_config in notifier.items():
+                notifier = plugin.get_plugin_by_name(notifier_name).instance
```
paranoidi: you might want to extract this into a method, or split this otherwise as it's long side

PR #1527, opened 2016-12-01
Seen `learn` event

liiight:
## Motivation for changes:
Pending plugin works is a unique way, and it needs to be able to reject entries that weren't accepted in the past. This is achieved by calling `seen` plugin using the new `learn` event

### Detailed changes:

- Created the `learn` event
- Minor change to `seen` plugin to use the same code path
- Firing event via `pending_approval` on `PendingEntry` creation in DB

### Problem: 
When the `learn` event is called, the `local` var is set to `None` by default. But if the user uses `seen: local` in his task, then the seen entry won't match and it'll get re-created on new execution. Need to figure out a way to couple `local` seen value to `learn` event call...

### DIFF
--- a/flexget/plugins/filter/seen.py
+++ b/flexget/plugins/filter/seen.py
@@ -170,6 +170,29 @@ def forget(value):
     return count, field_count
 
 
+@event('learn')
+def learn(task, entry, fields=None, reason=None, local=False):
+    """Marks entry as seen"""
+    # no explicit fields given, use default
+    if not fields:
+        fields = ['title', 'url', 'original_url']
+    se = SeenEntry(entry['title'], str(task.name), reason, local)
+    remembered = []
+    for field in fields:
+        if field not in entry:
+            continue
+        # removes duplicate values (eg. url, original_url are usually same)
+        if entry[field] in remembered:
+            continue
+        remembered.append(entry[field])
+        sf = SeenField(str(field), str(entry[field]))
+        se.fields.append(sf)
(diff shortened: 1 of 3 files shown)


flexget/plugins/filter/seen.py
```diff
@@ -170,6 +170,29 @@ def forget(value):
     return count, field_count
 
 
+@event('learn')
+def learn(task, entry, fields=None, reason=None, local=False):
+    """Marks entry as seen"""
+    # no explicit fields given, use default
+    if not fields:
+        fields = ['title', 'url', 'original_url']
+    se = SeenEntry(entry['title'], str(task.name), reason, local)
```
cvium: se? Please no.

liiight: Yeah, i agree. it was already there, but ill change it

flexget/plugins/filter/seen.py
```diff
@@ -170,6 +170,29 @@ def forget(value):
...
+    remembered = []
+    for field in fields:
+        if field not in entry:
+            continue
+        # removes duplicate values (eg. url, original_url are usually same)
+        if entry[field] in remembered:
+            continue
+        remembered.append(entry[field])
+        sf = SeenField(str(field), str(entry[field]))
+        se.fields.append(sf)
```
cvium: Read this line in a vacuum and then try to understand it.

liiight:
pending will iterate over accepted entries, making this redundant (not that it's working anyway)

PR #1529, opened 2016-12-01
Fetch more of the series data exposed by NPO.nl

mfonville:
### Motivation for changes:
NPO.nl exposes more information about the series than currently offered by the 'searching for broadcasts' URL that we use (e.g. http://www.npo.nl/als-de-dijken-breken/VPWON_1261083/search?category=broadcasts )

If you visit and fetch the official series URL (e.g. http://www.npo.nl/als-de-dijken-breken/VPWON_1261083 ) this extra information is exposed via a small snippet:
```
<div class="schema-org-props hidden-item-props"><div itemscope="itemscope" itemtype="http://schema.org/TVSeries"><a itemprop="url" href="/als-de-dijken-breken/VPWON_1261083"><span itemprop="name">Als de dijken breken</span></a><span itemprop="description">Serie over een hedendaagse watersnoodramp in Nederland en delen van Vlaanderen.</span><span itemprop="inLanguage">nl</span></div></div>
```

### Detailed changes:

- I tried to add a function to fetch more information about the series itself
- I would like to expose this meta-data so that it can be re-used by other plugins or the user

I do need some further guidance how to integrate this further, and make sure the fields are exposed in a proper manner.

### DIFF
--- a/flexget/plugins/input/npo_watchlist.py
+++ b/flexget/plugins/input/npo_watchlist.py
@@ -167,24 +167,38 @@ class NPOWatchlist(object):
 
         return entries
 
-    def _get_series_episodes(self, task, config, series_name, series_url):
+    def _get_series_info(self, task, config, series_name, series_url):
+        log.info('Retrieving series info for %s', series_name)
+        response = task.requests.get(series_url)
+        page = get_soup(response.content)
+        tvseries = page.find('div', itemscope='itemscope', itemtype='http://schema.org/TVSeries')
+        series_info = Entry()  # create a stub to store the common values for all episodes of this series
+        series_info['npo_url'] = tvseries.find('a', itemprop='url')['href']
+        series_info['npo_name'] = tvseries.find('span', itemprop='name').contents[0]
+        series_info['npo_description'] = tvseries.find('span', itemprop='description').contents[0]
+        series_info['npo_language'] = tvseries.find('span', itemprop='inLanguage').contents[0]
+
+        return series_info
+
+
+    def _get_series_episodes(self, task, config, series_name, series_url, series_info):
         log.info('Retrieving new episodes for %s', series_name)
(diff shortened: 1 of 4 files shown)


liiight:
just add the values to the entries you return: `entry['npo_language'] = npo_language` for example

mfonville:
Ok thanks, for that hint.

I incorporated that change in my updated commit, and now return several `npo_*` values;

How can I now make sure that e.g. the value of `npo_language` gets passed onto `tvdb_lookup` for the language-dependent lookup?

Also, should I add any tests for these changes? And is the coding-style ok, or should I do some things in a different way?

liiight:
add a `language` attribute to the entry, in addition to `npo_language`. i'll slightly modify `thetvdb_lookup` so it'll be able to accept that from the entry.
About tests, the answer is always yes :smile:

mfonville:
ok thanks, I now added the `language' attribute to to it.

I also started a test-class, but having some trouble, because there was not on yet for npo_watchlist.

Would you be able to help me out a bit?
Because normally npo_watchlist logs in to a npo.nl account, but in this situation that does not apply, so I just want a mock that it can compare to [Als de de dijken breken](http://www.npo.nl/als-de-dijken-breken/VPWON_1261083) series information. But I have no clue how to achieve this :-/

liiight:
if you could set up a real account for flexget and use that it would be ideal

mfonville:
@liiight I set-up the account:
```
email: chetrutrok@throwam.com
pass: Fl3xg3t
```

I added *Als de dijken breken* to its favourites (so that are the series indexed by `npo_watchlist`)

mfonville:
Added some tests to it, thank you @liiight for your help with that! It should be ok to merge at this point

liiight:
np, thank you for contributing. update wiki if relevant

PR #1530, opened 2016-12-02
rewrite of newznab plugin:

Andy2244:
### Motivation for changes:
Better metadata lookup support and adding all of the usefully features the newznap api offers. Parsing/storing of advanced attributes that can be directly used by the Age/If and Sort_by plugins.

### Detailed changes:
Added:
- official and unofficial newznab categories via `category`
- ability to build custom search query strings from existing fields via` custom_query`
- automatic and manual release age support via `max_age`
- support for ill behaving indexers via `force_quotes`
- meta data lookup if configured
- tvdb support for meta id based searches
- additional stat support via 'newznab' namespace fields
- support for nzbhydra extra search stats

Changed:
- `website `renamed to `api_server_url`
- `apikey `renamed to `api_key`

Entry changes:
- added the following 'newznab' namespace fields if found:
```
    'newznab_age'                   - age in days of this release
    'newznab_pubdate'               - date the indexer added the nzb to its database (aka age)
    'newznab_guid'                  - unique guid of this release
    'newznab_grabs'                 - number of grabs
    'newznab_size'                  - size in bytes of the release
    'newznab_files'                 - number of files this release (archive) has
    'newznab_usenetdate'            - date the release was posted on usenet
    'newznab_password'              - if the release uses a password ('0' no, '1' rar pass, '2' contains inner archive)
    'newznab_usenet_group'          - the usenet group this was posted on
    'newznab_tvairdate'             - the airdate reported by the indexer
    'newznab_season'                - the season reported by the indexer
    'newznab_episode'               - the episode reported by the indexer
    'newznab_hydra_indexer_name'    - the name set in nzbhydra
    'newznab_hydra_indexer_host'    - the host url used by nzbhydra
    'newznab_hydra_indexer_score'   - the priority score set by nzbhydra config for this indexer
```

```
Newznab search plugin
    Provide the 'api_server_url' + 'api_key' and one or more search categories via 'category'
    Most common categories are: 'tv', 'movie', 'tv/hd', 'movie/hd'
    Valid meta names for 'use_metadata': 'tvdb', 'trakt', 'tvmaze', imdb', 'tmdb'
    TIP: Use nzbhydra to perform searches on multiple indexers with one config https://github.com/theotherp/nzbhydra

    Config example:
    # search by name: search in the 'tv' category, using existing names for the type ('title'...)
    newznab:
        api_server_url: https://api.nzbindexer.com
        api_key: my_apikey
        category: tv

    # search by name + age: search in the 'tv' category, using existing names for the type ('title'...)
    # maxage: auto will try to calculate the maximum possible age for the release and use it to confine the search
    newznab:
        api_server_url: https://api.nzbindexer.com
        api_key: my_apikey
        category: tv
        max_age: auto

    # meta id based: try using existing metaid's (tvdb/ragetv/imdb) to search only in the tv/hd,uhd categories
    newznab:
        api_server_url: https://api.nzbindexer.com
        api_key: my_apikey
        category:
            - tv/hd
            - tv/uhd
        use_metadata: yes

    # meta id based: try using metaid's (tvdb/ragetv/imdb) and if needed do a trakt/imdb lookup,
                     only in the movie/uhd category for releases that are not older than 1 week
    newznab:
        api_server_url: https://api.nzbindexer.com
        api_key: my_apikey
        category: movie/uhd
        use_metadata:
            - trakt
            - imdb
        max_age: 1 week

    # custom search string: builds the search string and searches in the tv/hd and a custom 5999 category
    newznab:
        api_server_url: https://api.nzbindexer.com
        api_key: my_apikey
        custom_query: "{{trakt_series_name}} {{series_id}}"
        category:
            - tv/hd
            - 5999

    # combined (meta id + custom search string): try using metaid's (tvdb/ragetv/imdb) and if needed does a tvdb lookup,
                                                also builds/uses the custom search string.
    newznab:
        api_server_url: https://api.nzbindexer.com
        api_key: my_apikey
        custom_query: "{{trakt_series_name}} {{series_id}}"
        category: tv
        use_metadata: tvdb

    # forces quotes around the search query, some ill behaving indexers wont correctly handle multiterm queries like: 'title s01e01'
    # If you notice you get results of titles that are not in the search query or wrong season/episodes, try this.
    # NOTE: will most likely break well behaving indexers, so use in a separate task/entry or use nzbhydra
    newznab:
        api_server_url: https://api.nzbindexer.com
        category: tv
        force_quotes: yes
```


flexget/plugins/sites/newznab.py
```diff
@@ -1,124 +1,561 @@
...
 
 from flexget import plugin
 from flexget.entry import Entry
 from flexget.event import event
-import feedparser
+from flexget.config_schema import one_or_more
+from flexget.utils.tools import parse_timedelta
+from flexget.utils.template import render_from_entry, RenderError
+from flexget.utils.search import normalize_unicode
+from flexget.plugins.parsers.parser_common import normalize_name, remove_dirt
```
paranoidi: Ni direct imports for other plugins

flexget/plugins/sites/newznab.py
```diff
@@ -1,124 +1,561 @@
...
+        do_show_lookup = False
+        do_movie_lookup = False
+        if not isinstance(config['search_type'], basestring) and config['search_type'].get('custom_query'):
+            if entry.get('series_name') or entry.get('tvdb_id'):
+                do_show_lookup = True
+            else:
+                do_movie_lookup = True
+
+        # update entry metadata based on task config
+        if config['search_type'] == 'tv' or do_show_lookup:
```
paranoidi: No probing task.config, or enabling lookups for user.

flexget/plugins/sites/newznab.py
```diff
@@ -1,124 +1,561 @@
...
+        return number
 
-        if not len(rss.entries):
-            log.info('No results returned')
+    def string_to_bool(self, value):
+        return value.lower() in ('yes', 'true', 't', '1', 'y')
 
-        for rss_entry in rss.entries:
+    def dump_entry(self, entry):
+        log.verbose('#####################################################################################')
```
paranoidi: No separator lines

flexget/plugins/sites/newznab.py
cvium: That's not a plugin

flexget/plugins/sites/newznab.py
cvium: I'm assuming it's only meant for debugging while he's developing.

Andy2244: Yes only for myself, so i can easily spot the start/end of entrys and check if they have all the new fields with good values.

flexget/plugins/sites/newznab.py
Andy2244: `flexget.plugins.parsers.parser_common`
Its the same import as in series.py, movie_list, metainfo_movie.

flexget/plugins/sites/newznab.py
Andy2244: I don't understand this one, can you explain what is wrong here?

flexget/plugins/sites/newznab.py
```diff
@@ -1,124 +1,561 @@
...
+                url_param = '&tvdbid=%s' % entry.get('tvdb_id')
+            elif entry.get('tvrage_id'):
+                url_param = '&rid=%s' % entry.get('tvrage_id')
+            # lets not use those for now, rarely supported!
+            # elif entry.get('tvmaze_series_id'):
+            #    url_param = '&tvmazeid=%s' % entry.get('tvmaze_series_id')
+            # elif entry.get('trakt_id'):
+            #    url_param = '&traktid=%s' % entry.get('trakt_id')
+            if not url_param:
+                log.error('Could not get valid meta id for series < %s >' % (entry.get('series_name', entry['title'])))
```
Andy2244: Sure, will change the syntax to match the rest, just my own programmer habit of wanting to distinguish between static and dynamic parts of log values.

flexget/plugins/sites/newznab.py
paranoidi: @cvium It's under plugins so it sure looks like one. If it's not a plugin it should not be under plugins package .. :)

@Andy2244 You should not rely on guessit or any other lookup plugins. Search plugins should not issue any metadata or other side effects. It could use them though if they are available. Can you explain how series / movies metadata is being used by newznab?

Andy2244: It can use most of the available meta id's, i have only enabled tvdb/imdb/ragetvid for now, since the other require more complex capabilities checks for the used indexer. If it can't find a id it tries the series_name/movie_name and title fields. The parser plugin is mainly used to fix missing episode numbers on tv lookups, if it is not used with the next_episode plugin for example.

Andy2244: Is the main objection on how to handle metadata lookup, that a search plugin should not update existing entries with meta data? If so i can refactor it so that it will not change the entries and instead just use a copy for lookup? I guess i also should than remove the task probing and do my own config setting for lookup?

paranoidi: The next_series_episodes plugin already sets "series_name" field, as well bunch of others. No need to do any lookups for that.

For IDs I would say use them if they are available ...

paranoidi: Oh right, it happens within discover plugin, not during task execution. This is ongoing design challenge, it's not right to make quick solutions with one search plugin though.

Andy2244:
Oki if someone has the time to take a look plz, i'm happy so far with it.

liiight:
This is a big one... Any chance you adding tests?

Andy2244:
Is there a template for this kind of search plugin? How do i simulate the results that come from the indexers?

liiight:
It returns xml and/or json files right? You can just save those as files in the tests dir, then mock them and return it

Andy2244:
I can take a look at the other tests and see if i can adapt those.

flexget/plugins/sites/newznab.py
```diff
@@ -1,125 +1,719 @@
...
+
+NAMESPACE_NAME = 'newznab'
+NAMESPACE_URL = 'http://www.newznab.com/DTD/2010/feeds/attributes/'
+NAMESPACE_TAGNAME = 'attr'
+NAMESPACE_PREFIX = 'newznab_'
+ENCLOSURE_TYPE = 'application/x-nzb'
+
+
+# utils
+def list_insert_unique(in_list, idx, element, caseinsensitive=False):
```
paranoidi: Check python set

It's not case insensitive though by nature, but it would be a LOT more elegant if you could utilize that instead as a basis.

flexget/plugins/sites/newznab.py
```diff
@@ -1,125 +1,719 @@
...
+        if element.isspace() or element in in_list:
+            return
+        elif caseinsensitive is True and element.lower() in in_list:
+            return
+    elif element in in_list:
+        return
+    in_list.insert(idx, element)
+
+
+def list_append_unique(in_list, element, caseinsensitive=False):
```
paranoidi: Check python set

flexget/plugins/sites/newznab.py
```diff
@@ -1,125 +1,719 @@
...
+    list_insert_unique(in_list, len(in_list), element, caseinsensitive=caseinsensitive)
+
+
+def list_combine_unique(list_dst, list_source, caseinsensitive=False):
```
paranoidi: Check python set

flexget/plugins/sites/newznab.py
```diff
@@ -1,125 +1,719 @@
...
+    if not isinstance(list_dst, list) or not isinstance(list_source, list):
+        log.error('Cant combine lists, src/dst is not a listtype: %s %s' % (list_dst, list_source))
+    for item in list_source:
+        list_append_unique(list_dst, item, caseinsensitive=caseinsensitive)
+
+
+def safe_get(in_object, keynames, types=basestring, default=None, caseinsensitive=True):
```
paranoidi: This is not pythonic, it needs to go away one way or another :)

flexget/plugins/sites/newznab.py
Andy2244: I did and set will not keep the list order, this combines two lists, but keeps the order and also deals with case-sensitivity.

flexget/plugins/sites/newznab.py
```diff
@@ -1,125 +1,719 @@
...
-        imdb_id = arg_entry['imdb_id'].replace('tt', '')
-        url = config['url'] + '&imdbid=' + imdb_id
-        return self.fill_entries_for_url(url, task)
+        # use custom query in all cases (not 100% well defined behaviour in metaid cases)
+        list_insert_unique(query_list, 0, custom_query_string, caseinsensitive=True)
+        # do search without list
+        if query_list is None or len(query_list) == 0:
+            return self.fill_entries_for_url(url, task)
+        else:
+            results_set = set()
```
paranoidi: Ah, I see you are familiar with set :)

flexget/plugins/sites/newznab.py
Andy2244: Set will not keep the order, i specifically need the order to-be intact here.

paranoidi:
This requires some cleanups IMO

flexget/plugins/sites/newznab.py
```diff
@@ -1,125 +1,719 @@
...
-        if 'url' not in config:
-            if 'apikey' in config and 'website' in config:
-                params = {
-                    't': config['category'],
-                    'apikey': config['apikey'],
-                    'extended': 1
-                }
-                config['url'] = config['website'] + '/api?' + urlencode(params)
+    def has_supported_meta_id(self, entry):
+        if self.is_tv_entry(entry) is True:
```
Andy2244: Yeah missed to fix this, at the time i did not get how python handles 'is' vs ==.

flexget/plugins/sites/newznab.py
Andy2244: What exactly?
I'm asking a untrusted object to get me a value of a specific type i want? So doing no checks is preferred over my paranoid checks?

This just mimics entry[] and list.get() methods, but i do a extra attribute, type check, empty string check, None check and allow caseinsensitive by default?

So this is preferred:
```python
if 'name' in entry and entry['name'] is not None and isinstance(value, str) and not entry['name'].isspace():
  result = do_stuff(entry['name'])
```
This still misses 'Name' over 'name'.

over this ?
```python
result = safe_get(entry, 'name')
```

flexget/plugins/sites/newznab.py
```diff
@@ -1,125 +1,719 @@
...
+}
+
+# NOTE: all lowercase only
+NAMESPACE_ATTRIBUTE_MAP = {
+    'grabs': int,
+    'size': int,
+    'files': int,
+    'usenetdate': datetime,
+    'password': bool,
+    'guid': basestring,
```
gazpachoking: We should pretend that this is python 3, and using 'str' instead of basestring (our `from builtins import *` makes this a real python 3 `str` (text) even on python 2.) If there is an actual possibility that the thing will be either text (str) or bytes (bytes), we should handle those cases explicitly.

flexget/plugins/sites/newznab.py
gazpachoking: Well, for one, I'm not exactly sure what it's trying to do, or why. From the looks of the way this function is being used, it could be replaced with `if all(field in entry for field in ['tvdb_id', 'other', fields]):` We use duck typing extensively in flexget (and python in general,) if some plugin is filling these fields with data types that other plugins aren't expecting, we sortof want everything to fail and fix the issue rather than everything silently degrading.

flexget/plugins/sites/newznab.py
gazpachoking: Didn't you write `is_tv_entry` yourself? That method should be consistent with its return value, and you won't have to worry about unexpectedly typed results.

Andy2244: I mean is this really worth discussing?
It helps me avoid logic errors, that's all. Its a simple reminder that i do a Boolean check and not check vs invalid values like None, int = 0, Empty-string. 
```python
if do_stuff_a(entry):
if do_stuff_b(entry) is True:
```
I can directly tell that do_stuff_b is a Boolean return, while do_stuff_a might also want to check for None, int = 0, empty-string cases.

Andy2244:
I really try to adapt, but its really 'strange' to me how anyone can trust outside sources, without doing some basic checks. There are a lot of json inputs, where fields are copied into entry, without ever checking for valid input values, let alone types. So emptystrings or space only strings, negative sizes, dates with timezone information are all mingled with valid `expected` values.

So if a space-only string is stored, most code now tries to-do some regex or other operation on it, simply because its not Nonetype. Now all kind of weird stuff can happen, without even raising some warning/error. 

I tried to at least ensure i get a valid string to operate on, yet i feel like i get scolded for doing some basic checks that should be mandatory in my opinion. I really can't understand why its even possible to store a string or any other type than int in the `content_size` field. There should be a direct exception/error if this happens. The same should be true for all other fields where a certain type is expected. In the same sense why is it allowed to store Nonetype or emptystring or space-strings in Entry? Are there cases where a "   " is a valid entry value? Why not check this and raise a value/type exception, so we can safely work with the fields, without extra error checks. In my opinion if a field exists in Entry, i should be certain what type it is and that it has a valid value.

paranoidi:
> I tried to at least ensure i get a valid string to operate on, yet i feel like i get scolded for doing some basic checks that should be mandatory in my opinion. 

I hope you're not feeling scolded, we are quite friendly small community of developers :)
You're doing pretty well for a guy coming from vastly different paradigm and obviously experienced in many areas.

> I really can't understand why its even possible to store a string or any other type than int in the content_size field. 

Dictionaries do not enforce type information and we have not built any limitations to that, consider that users can set any field with any value they like with manipulate plugin for example. I have never really seen a problem because of types happening related to entry fields, I'm sure that has occasionally happened though.

LBYL vs. EAFP might be worthwhile to read if you're not familiar with it. Some quick links from google that seemed good.

https://jeffknupp.com/blog/2013/02/06/write-cleaner-python-use-exceptions/
http://docs.quantifiedcode.com/python-code-patterns/readability/asking_for_permission_instead_of_forgiveness_when_working_with_files.html

If we wanted to constraint entry fields we sure could do it, jsonschema and explicitly registering entry fields from plugins would be a possibility. Not worth the hazzle probably .. but I have considered that :)

So, we have implicit contract that certain fields are certain types. Plugins never should violate it so type should always remain same. Cases where field value can be either string or None should be eliminated. For missing data leaving field out from entry is preferred.

> Yet i somehow can't just let go of everything i learned and accepted as "good" design, code practice. So just bare with my questions and give me some time to adapt.

FlexGet is not perfect design by any means, it's almost 10 years old now and has grown quite a bit during that time. As well as our skills with python ..

Andy2244:
> So, we have implicit contract that certain fields are certain types. Plugins never should violate it so type should always remain same. Cases where field value can be either string or None should be eliminated. For missing data leaving field out from entry is preferred.

This works because you are already very familiar with the field usages and general setup. Here are just some questions/problems i encountered and had problems finding quick answers.

- Is 'imdb_id' a int or string field, does it always have the 'tt' prefix?
- Are the other `_id` fields also strings with some prefix?
- Why is `content_size` in MB, is it a plugin error or general rule, are all size fields in MB?
- Why can't i find a defined way to distinguish between a series or movie entry? Is checking 'series_name' vs ' movie_name' always valid?
- Is it correct that imdb_lookup and tmdb_lookup only works for movies, since both databases also store series.
- Is the `search_strings` entry a default way of using alternate 'title' names or just from next_episode plugin
- What exactly is stored in 'alts'.
- Are date fields always datetime types and are they timezone corrected and if so is UTZ conversion used?

flexget/plugins/sites/newznab.py
cvium: If your function can return multiple value types, you're doing something wrong and you end up with spaghetti where you have to do explicit checks that have different (side) effects.

cvium:
> Is imdb_id a int or string field, does it always have the 'tt' prefix?
imdb_id should always be a string with a 'tt' prefix. If it's not the case in some plugins, then those need to be changed imo.

> Are the other _id fields also strings with some prefix?
No. Most of them are ints. imdb_id is a special case because it contains a 'tt'-prefix. 

> Why is content_size in MB, is it a plugin error or general rule, should all size fields be stored in MB?
content_size is always MiB. It's just another implicit contract.

> Why can't i find a defined way to distinguish between a series or movie entry? Is checking 'series_name' vs ' movie_name' always valid?
Yes. There are cases where a movie is parsed a series and vice versa, but that's how every other plugin checks the entry type. We have to assume the parser works (and fix it if it doesn't).

> Is it correct that imdb_lookup and tmdb_lookup only works for movies, since both databases also store series.
Yes. imdb_lookup should probably be generalized though.

> Is the search_strings entry a default way of using alternate 'title' names or just from next_episode plugin
search_strings is a list of strings that is used in search plugins. It's to avoid having to create 'search strings' explicitly in every search plugin. 

> What exactly is stored in 'alts'.
If you mean in a series context, it's alternate names for a series.

> Are date fields always datetime types and are they timezone corrected and if so is UTZ conversion used?
Always DateTime or Date. Movie release fields are Date, everything else is DateTime.

> Are the score and votes fields stored as strings or int/float?
Should be ints.

flexget/plugins/sites/newznab.py
Andy2244: My function does return multiple value types like most other functions, that use the NoneType as return. In this case it simply signals that the entry is neither a series or "not a" series aka movie. So it will return Nonetype.
As noted this is simply more verbose and reminds me that i need to take special care with If (statement) checks on integer fields, where 0 is a valid value, like `episode_number`.

paranoidi:
>>   Why is content_size in MB, is it a plugin error or general rule, should all size fields be stored in MB?
> content_size is always MiB. It's just another implicit contract.

In retrospective I should just have used bytes, but such is life.

>>Are the score and votes fields stored as strings or int/float?
> Should be ints.

Depends on site, some use floats like imdb (6.2) and some ints (1-10, or 1-5) etc. This is implicit contract as well, we know imdb uses floats.

>> Are date fields always datetime types and are they timezone corrected and if so is UTZ conversion used?
> Always DateTime or Date. Movie release fields are Date, everything else is DateTime.

DateTime usually contains timezone information too ..

We are not building operating system or banking software so we generally go for best miles per effort when doing things :)

Perfect is the enemy of good etc.

Andy2244:
@paranoidi , thx for the links i was reading something similar the other day. I have a basic understanding of how EAFP is supposed to work and what the pro/cons are, its simply that i strongly disagree with this paradigm, which is a result of my past work and experience. On the other hand i'm also used to adapt to coding styles that i disagree with personally, but working as a team requires.

I will just stop wondering and refactor all my stuff to EAFP and keep on adding actual flexget features :)

paranoidi:
I find that as time goes on and with more hands on experience you learn to accept or even appreciate different approaches. It's not like either of these ideas are wrong or inelegant. Just different.

(Where as I would say JS, Perl, PHP are inelegant languages compared to many other choices, some may disagree ... but how can this be quantified objectively ...).

flexget/plugins/sites/newznab.py
```diff
@@ -1,125 +1,655 @@
...
+        elif meta_url and self._is_tv_entry(entry) is False:
+            url += '&t=movie' + meta_url
+        else:
+            url += '&t=search'
+            # use existing search_strings
+            query_list = entry.get('search_strings', default=[entry['title']])
+            list_append_unique(query_list, entry['title'], caseinsensitive=True)  # we trust external list?
+
+        # use custom query in all cases (not 100% well defined behaviour in metaid cases)
+        list_insert_unique(query_list, 0, custom_query_string, caseinsensitive=True)
```
paranoidi: I think you can axe the one use method simply with inline:

```
if custom_query_string.lower() not in [v.lower() for v in query_list]:
  query_list.insert(0, custom_query_string)
```

Also, in python we don't usually add type information into variables (some_list, stuff_dict, etc), this is especially true if they are introduced within the scope where type is clearly visible. For method/func params the docstring is the place for this information (and type hinting in the future when we can be only py3 compatible)

Andy2244: Its no so much that i want to encode the type, but its easier to use the typename instead of coming up with a alternative name. So in this case i just wanted to distinguish between the query text input, the parsed version, the url_fragment and the "collection" object. So sure i could have named query_list/string also query_text/collection. If i could load python syntax into my brain, as easy as i can switch in the IDE than this would be no problem :)

flexget/plugins/sites/newznab.py
```diff
@@ -1,125 +1,608 @@
...
+                    config['category'] = [config['category']]
+                categories = config['category']
+                # Convert named categories to its respective categories id number
+                categories = [c if isinstance(c, int) else CATEGORIES[c] for c in categories]
+                if len(categories) > 0:
+                    config['category_string'] = ','.join(str(c) for c in categories)
+        return config
+
+    def _is_tv_entry(self, entry):
+        if any(entry.get(field) for field in ['series_name', 'tvdb_id', 'tvrage_id', 'tvmaze_series_id', 'trakt_id']):
```
paranoidi: There is a way to do this without hard coding this list here.

````
[p.instance.movie_identifier for p in plugin.get_plugins(group='series_metainfo')]
````
(you can search this for usages)

Might not include series_name though, but at least gets mostly everything dynamically. You need to apply this other places too, we don't want to manually maintain these lists across numerous plugins.

Comment (Andy2244) [inline flexget/plugins/sites/newznab.py]: Should i add this property to all the missing series metainfos plugins as well, so it is consistent with the 'movie_metainfo'?

**Comment by Andy2244** [comment]
Is a missing unitetest all that i have to-do on this one or is there anything else?

File: flexget/plugins/sites/newznab.py [hunk 14]
```diff
@@ -1,125 +1,615 @@
...
-                params = {
-                    't': config['category'],
-                    'apikey': config['apikey'],
-                    'extended': 1
-                }
-                config['url'] = config['website'] + '/api?' + urlencode(params)
 
+    def prepare_config(self, config):
+        if 'use_metadata' in config:
+            if 'plugins_list' not in config:
```
cvium: What is `plugins_list`? I don't see it in the schema and why are the two ifs on different lines?

flexget/plugins/sites/newznab.py
```diff
@@ -1,125 +1,615 @@
...
-        if 'imdb_id' not in arg_entry:
-            return entries
-
-        imdb_id = arg_entry['imdb_id'].replace('tt', '')
-        url = config['url'] + '&imdbid=' + imdb_id
-        return self.fill_entries_for_url(url, task)
+    def _build_metaid_url_fragment(self, entry):
+        url_param = None
+        # use first valid meta id
+        if self._is_tv_entry(entry) is True:
```
cvium: This is not pythonic. `if self._is_tv_entry(entry)` is better and if your function returns something that isn't `True` or `False`, then it needs to be rewritten.

flexget/plugins/sites/newznab.py
```diff
@@ -1,125 +1,615 @@
...
+            if entry.get('tvdb_id'):
+                url_param = '&tvdbid=%s' % entry['tvdb_id']
+            elif entry.get('tvrage_id'):
+                url_param = '&rid=%s' % entry['tvrage_id']
+                # lets not use those for now, rarely supported!
+                # elif self.safe_get(entry, 'tvmaze_series_id', [basestring, int]):
+                #    url_param = '&tvmazeid=%s' % entry['tvmaze_series_id']
+                # elif self.safe_get(entry, 'trakt_id', [basestring, int]):
+                #    url_param = '&traktid=%s' % entry['trakt_id']
+        elif self._is_tv_entry(entry) is False:
```
cvium: See above.

flexget/plugins/sites/newznab.py
```diff
@@ -1,125 +1,615 @@
...
-        if not len(rss.entries):
-            log.info('No results returned')
+        search_entry = Entry(entry)
+        for search_string in search_list:
+            search_string = search_string.strip()
+            if not search_string:
+                continue
+            search_entry['title'] = search_string
+            # update entry metadata
+            if show_lookup is True:
```
cvium: All these `if X is True` needs to go.

flexget/plugins/sites/newznab.py
```diff
@@ -1,125 +1,615 @@
...
+                search_entry['series_name'] = search_string
+                for plugin_name in config['plugins_list']:
+                    if plugin_name == 'trakt_lookup':
```
cvium: Why are all of these hardcoded? You should determine whether the entry is a series or a movie and simply use `plugin_name` to get the plugin instance.

flexget/plugins/sites/newznab.py
```diff
@@ -1,125 +1,615 @@
...
+                        plugin.get_plugin_by_name('imdb_lookup').instance.lookup(search_entry)
+                        log.verbose('Doing `%s` for movie: %s', plugin_name, search_string)
+            if self._has_meta_id(search_entry) is not None:
+                search_entry['title'] = entry['title']  # keep org. title
+                return search_entry
+        return entry
 
-        for rss_entry in rss.entries:
+    def _build_base_url(self, entry, config):
+        # TODO @Andy: do 't=caps' check (many indexers wont correctly handle this, nzbhydra uses 'brute force' since 0.2.169)
```
cvium: All of these TODOs are useless clutter. Either "do them" or delete. No one is going to handle a TODO in someone else's code.

flexget/plugins/sites/newznab.py
```diff
@@ -1,125 +1,615 @@
...
+                new_entry['newznab_age'] = datetime.now() - new_entry['newznab_pubdate']
 
-            for key in list(rss_entry.keys()):
-                new_entry[key] = rss_entry[key]
-            new_entry['url'] = new_entry['link']
-            if rss_entry.enclosures:
-                size = int(rss_entry.enclosures[0]['length'])  # B
-                new_entry['content_size'] = old_div(size, 2 ** 20)  # MB
             entries.append(new_entry)
+            #_debug_dump_entry(new_entry)  # debug
```
cvium: remove

flexget/plugins/sites/newznab.py
```diff
@@ -1,125 +1,615 @@
...
+    def _build_metaid_url_fragment(self, entry):
+        url_param = None
+        # use first valid meta id
+        if self._is_tv_entry(entry) is True:
+            if entry.get('tvdb_id'):
+                url_param = '&tvdbid=%s' % entry['tvdb_id']
+            elif entry.get('tvrage_id'):
+                url_param = '&rid=%s' % entry['tvrage_id']
+                # lets not use those for now, rarely supported!
+                # elif self.safe_get(entry, 'tvmaze_series_id', [basestring, int]):
```
cvium: remove

flexget/plugins/sites/newznab.py
```diff
@@ -1,125 +1,615 @@
...
+
+
+def list_combine_unique(list_dst, list_source, caseinsensitive=False):
+    if not isinstance(list_dst, list) or not isinstance(list_source, list):
+        log.error('Cant combine lists, src/dst is not a listtype: %s, %s', list_dst, list_source)
+    for item in list_source:
+        list_append_unique(list_dst, item, caseinsensitive=caseinsensitive)
+
+
+def _debug_dump_entry(entry):
```
cvium: remove

flexget/plugins/sites/newznab.py
Andy2244: There are over 100+ "TODO" in the existing code, so i assumed those are used the same way in python/Flexget than on any other project i worked on? Is this just about the "@Andy" or "TODO" comments in general? So i'm not allowed to add TODO's in flexget?

flexget/plugins/sites/newznab.py
Andy2244: Whats the problem with private debug functions?
So i'm not allowed to submit debug functions for my modules and have to separately maintain those locally? This means i need to remove and copy&paste/merge them in again after each commit/update?

flexget/plugins/sites/newznab.py
cvium: You are :) At a glance it just seems like there are quite many. And usually TODOs are never followed up, so I don't particularly like them.

flexget/plugins/sites/newznab.py
```diff
@@ -1,125 +1,615 @@
...
+}
+
+NAMESPACE_NAME = 'newznab'
+NAMESPACE_URL = 'http://www.newznab.com/DTD/2010/feeds/attributes/'
+NAMESPACE_TAGNAME = 'attr'
+ENCLOSURE_TYPE = 'application/x-nzb'
+
+
+# list utils
+def list_insert_unique(in_list, idx, element, caseinsensitive=False):
```
Andy2244: Since this is the third time this function is deemed "problematic" and i'm tired of arguing i will just remove it and do the logic manually.

flexget/plugins/sites/newznab.py
Andy2244: I understand that, but that's because they are annoying "TODO" or we would actually "do" them in the first place. I still think its better than nothing and is just a quick reminder of "ideas". So i don't like keeping a separated textfile or something more complex around per module. I'm well aware the most TODO will never done, yet i also read some TODO in other modules that actually helped understand what was still missing.

flexget/plugins/sites/newznab.py
Andy2244: Again? 
This function can return None, which simply means its neither a movie or tv entry. I can rename the function to get_EntryType() and use a enum/string to identify its type, i simply thought this would be overkill.

cvium: If you wish to support entries that are neither movies nor series, then I'd rather you made it more explicit and had separate boolean functions for each. You have to consider that other people will be reading your code eventually and if a boolean function has a third special return value, then it becomes more difficult to read.

flexget/plugins/sites/newznab.py
cvium: It's fine then.

flexget/plugins/sites/newznab.py
cvium: It adds unnecessary clutter imo. No end-user is going to use it.

Andy2244: There is no _DEBUG define system in python and its not for the end-user its for me the developer. Its already hard enough to debug flexget in the first place. 
So this is what my workflow should look like for each commit?
- remove debug code for checkin
- do checkin
- copy&paste back debug code

Is there no way how python supports debug code?

crawln45: pycharm shows you every variable set according to breakpoints. Very easy to debug.

Andy2244: What are you talking about? I want to output formatted debug information, this has nothing to-do with breakpoints.

flexget/plugins/sites/newznab.py
Andy2244: I can shorten the ones that are identical, but "thetvdb_lookup" uses a extra parameter and trakt_lookup has "lazy_movie_lookup" instead of just "lookup".

cvium: We should probably fix that.

Btw, is there a reason that lookups need to be done inside this plugin? Usually, if a plugin needs lookup metadata, it's the user's responsibility to enable them.

Andy2244: Thats a limitation on how flexget handles metadata for input+search plugins and the next_series_ plugin.

flexget/plugins/sites/newznab.py
Andy2244: Its added in prepare_config().
I find it easier to read and maintain instead of combining main-branch and sub-branch checks here.

flexget/plugins/sites/newznab.py
Andy2244: Oki this turns out to-be more complicated doing manual than keeping the functions.
The problem is that the user can "strip" the year out of input entries, yet the most safe way to-do meta lookups is by using the name + year first and than try without it if nothing was found.
So i want to have at least those 2 entries first in a list and than add whatever was provided by the extra input "search_strings". Yet i also want to prevent hammering external api's with the same lookups, that why i use the "caseinsensitive=True" option.

flexget/plugins/sites/newznab.py
```diff
@@ -333,19 +332,19 @@ def update_metadata(self, entry, config):
         if 'plugins_list' not in config:
             return entry
 
-        show_lookup = self._is_tv_entry(entry)
+        show_lookup = self._get_entry_type(entry)
```
cvium: Should probably rename this variable.

flexget/plugins/sites/newznab.py
Andy2244: I had to-do like 60 changes to this module and you expect me to remove this line on each commit and than copy it back afterwards or rewrite it each time?
Is the point really to make my life as developer harder, just so the code has 1 more less debug line?

flexget/plugins/sites/newznab.py
```diff
@@ -1,125 +1,599 @@
...
+        if element.isspace() or element in in_list:
+            return
+        elif caseinsensitive is True and element.lower() in in_list:
+            return
+    elif element in in_list:
+        return
+    in_list.insert(idx, element)
+
+
+def list_append_unique(in_list, element, caseinsensitive=False):
```
cvium: Seems like `caseinsensitive` is always `True`.

Andy2244: What? I don't understand, the default case is defined as False?

cvium: Yes, but you never use the default from what I can see. Maybe the function should really just be case insensitive then.

flexget/plugins/sites/newznab.py
cvium: Your personal debugging code is fine for now, but once the PR is ready to be merged, it must be removed. Once it's upstream, there should be no need for that code.

flexget/plugins/sites/newznab.py
Andy2244: The language default for appending string objects to a list/set is case-sensitive, so i think its better to keep this expectation. This way its visible if a call wants to override this behavior.

flexget/plugins/sites/newznab.py
Andy2244: This makes no sense to me, sorry. 
So after a PR was merged and i want to add a new feature or fix a bug, i have to write the debug code again? So by this logic i also have to remove the "_add_debug_info" from my sort_by_weight plugin and rewrite it every-time i have to debug a problem?

PR #1539, opened 2016-12-05
Notifiers tweaks

liiight:
### Motivation for changes:
Hindsight is 20/20
### Detailed changes:

- Refactored `notify` so that base will be reused, and deprecated it.
- Created `notify_entries` plugin that sets the scope to `entries`.
- Created `notify_task` plugin that sets the scope to `task`.
- Created `notify_abort` plugin .
- Removed`notify_crash` .
- Added `templates` CLI that lists all file templates with their content.
- Removed sting replacement code from `render()`
- Refactored file templates to be used by scope instead of by file name (modified `make_html` and `make_rss` to use it, tested)
- Moved API templates to api dir


flexget/plugins/notifiers/notify.py
```diff
@@ -36,28 +36,34 @@
...
+                                                ' plugin name.',
+                         'minProperties': 1}]}}
         },
         'required': ['to'],
         'additionalProperties': True
     }
 
+
+class Notify(NotifyBase):
+    schema = NotifyBase.schema
```
gazpachoking: Should we put a deprecated message on this one?

flexget/plugins/notifiers/notify.py
```diff
@@ -156,6 +162,53 @@ def on_task_start(self, task, config):
...
+        notify_config = {'to': config['to'],
+                         'scope': 'task',
+                         'title': title,
+                         'message': message}
+        log.debug('sending abort notification')
+        plugin.get_plugin_by_name('notify').instance.send_notification(task, notify_config)
+
+
+class NotifyCrash(NotifyBase):
+    def on_task_abort(self, task, config):
```
gazpachoking: I was thinking notify_crash would be a root level plugin, such that it would notify you about crashes in the scheduler or other contexts outside of tasks as well. Right now this is just the same as notify_abort, no?

gazpachoking:
Now I'm starting to wonder about notifications from handled errors that cause log messages. I also started wondering if we want the ability to consolidate those errors so they don't get spammy. I think there should be some sort of way to notify about errors outside of task contexts, I'm just not exactly sure what sort of errors to include, and how to handle them best. Would also have to think about overlap with task level notify plugins if we have that.

flexget/plugins/notifiers/notify.py
liiight: you can use it in a global template. I think it feels better than a root level, since it has no merit or functionality on its own.

liiight:
Yeah, i though about hooking events to logger calls (specifically on `critical`, `error` and `warning`). It might still be a valid idea, but I rather get this PR out of the way first before adding new functionality.

gazpachoking:
> you can use it in a global template. I think it feels better than a root level, since it has no merit or functionality on its own.

I'm talking about reporting errors outside of the task context. For example: scheduler, webui, api, irc. I think maybe a plugin to send log messages to notifier would be good, but I agree, that can wait. Should we just delete notify_crash for now then?

liiight:
notify_crash is not similar to notify_abort. it checks for existence of task.traceback which gets populated only on crashes

gazpachoking:
Hmm, task.traceback? Where is that coming from? I think we may be better off just sticking with the abort one now. I think for crash notification, we really wouldn't want to confine that to individual tasks would we?

liiight:
dunno, feels fine to me using it in global but I'm cool with root level too. I'll drop it

gazpachoking:
Seems good to me!

liiight:
I'll update wiki asap

PR #1544, opened 2016-12-10
New notifier refactor

gazpachoking:
__NOTE:__ ~Everything here is still a draft, mostly untested, and meant for discussion still.~ Getting mostly up to snuff at this point, though still open for other changes.

### Motivation for changes:
Further separate the different types of notification generation plugins for simplification, and easier customization of each message type. Simplify the framework to not need advanced knowledge about the context of a message.

### Detailed changes:
- The notification framework only sends one message at a time now, and doesn't have knowledge of the 'context' of the message. (also makes a future log notifier easier, which isn't task or entry context)
- plugins in the 'notifier' group now get params as `(title, message, config)` with all templates already rendered. title and message will be the message as constructed by one of the notification message generator plugins (notify_abort, notify_task, etc.), and config is the config for the transport plugin (email, prowl, etc.) as specified and validated by its schema (any text fields in the config will have jinja templates rendered against the appropriate context before being passed to transport plugin though)
- to send a message, you send `(title, message, outputs, [template_renderer])` where outputs is a list of configured notifier output plugins. template_renderer should be a function that renders jinja strings or templates against the appropriate context for the message. It will be applied to title, message, and any string keys in the output plugin configs.
- [x] Remove title and message params from each notifier plugin?
- A function can be passed in which will be used to render all text fields in the notification config for convenience (so each notification generator doesn't have to do it themselves.)
- The common base class was removed from the notification generators, and split into multiple separate files which now handle the different contexts of the notifications themselves.
- Move notifications to a new `notify` phase. Add a way to suppress task aborts from this phase. (Is the way I did this okay? Is there something nicer?)
- notify_task now only notifies when there are accepted or failed entries. Removed and tweaked templates so that we don't generate blank emails anymore.
- Notifications are sent on --test runs now too.
- [x] Update the notification output plugins to take the title and message (add url as a default message field too?) separate than the output plugin config. (all of them now done, question about url as a default field remains open though)

### Breaking changes
I've changed some names that I thought made sense, but we could backtrack if they don't seem worth it to break backwards compatibility in places.
- [x] ~notify_entries -> notify_entry ? I feel this has more parity with the singular notify_task, plus it sends a notification per entry like notify_task does per task~
- [x] combined notify_entries and notify_task, and notify_abort back into notify plugin, with 'task',  'entries', and 'abort' subkeys.
- [x] the template folder 'entries' to 'entry', as that's the rendering context
- [x] file_template -> template I changed the option in notify_task and notify_entry to just 'template' to specify one of the templates from disk.
- [x] 'to' -> 'via' any of the plugins that send messages take the notifier configs under the 'via' key now.
- [x] Backwards compatibility with the notifiers allowed directly in task is gone.
- [x] The notifier transport plugins no longer allow title, message, or template (or file_template) in their schemas.
- [x] 'notify' plugin was used as a framework for any other plugins that wanted to send messages before. Put that functionality into 'notify_framework' plugin now.
- [x] sns is now a regular output plugin again. (I did move it to the new notify phase so that errors don't fail the task, but I'm not sure if that's good or not.)
- [x] notify_osd was renamed toast, and (beta) Windows support was added.

### Things to finish before merging
- [x] Make the sns notifier a regular output plugin again.
- [x] Combine notify_task, _entry, _abort, again? I'm still not really certain about that one, I sorta like the split.
- [x] Notify_task and _entries are combined, should _abort go in there too?
- [x] The common framework used to send notifications is now named `notification_framework` so that the task plugin can be named `notify`. Not great, but it's only used from other plugins, not user facing.
- [ ] Consider the case where there are no other output plugins in a task but a notifier. In this instance it may be desirable to let failures in the notification cause entry failures/task aborts, as there won't be any side effects.
- ~Consider whether the notification framework should provide 'url' as a default field. There are several notifiers which all support 'url', but it wouldn't be 100%. 'subject' is already included though, and that isn't used 100% either, so url might be just as appropriate.~ Leaning to 'no' on this.


flexget/plugins/notifiers/email.py
```diff
@@ -104,56 +101,47 @@ class EmailNotifier(object):
         'additionalProperties': False,
     }
 
-    def notify(self, to, message, title, smtp_host, smtp_port, smtp_username=None, smtp_password=None, smtp_tls=None,
-               smtp_ssl=None, html=None, **kwargs):
+    def notify(self, title, message, config):
```
liiight: This is how I initially did it, by passing config dict. @paranoidi advised me to pass it a literal params for clarity's sake. I can't say I disagree

flexget/plugins/notifiers/notify_entry.py
```diff
@@ -0,0 +1,74 @@
...
+                'default': '{% if series_name is defined %}'
+                           '{{ tvdb_series_name|d(series_name) }} '
+                           '{{series_id}} {{tvdb_ep_name|d('')}}'
+                           '{% elif imdb_name is defined %}'
+                           '{{imdb_name}} {{imdb_year}}'
+                           '{% elif title is defined %}'
+                           '{{ title }}'
+                           '{% endif %}'
+            },
+            'template': {'type': 'string'},
```
liiight: I used `file_templates` to differentiate this from `template` plugin

flexget/plugins/notifiers/notify_entry.py
```diff
@@ -0,0 +1,74 @@
...
+    def on_task_notify(self, task, config):
+        send_notification = plugin.get_plugin_by_name('notify').instance.send_notification
+        config = self.prepare_config(config)
+        entries = list(itertools.chain(getattr(task, what) for what in config['what']))
+        if not entries:
+            log.debug('No entries to notify about.')
+            return
+        # If a file template is defined, it overrides message
+        if config.get('template'):
+            body = get_template(config['template'], 'entry')
```
liiight: This could raise `ValueError`

flexget/plugins/notifiers/email.py
gazpachoking: But... it wasn't passing in literal params, it was passing in a config dict, with other stuff added to it, which was being keyword argument expanded. This way we are passing in our actual literal params, which are the ones generated by the notification system, and the config, as specified and validated by the schema of the plugin, like we do for our other plugin methods.

flexget/plugins/notifiers/notify_entry.py
gazpachoking: Good point. Need to catch that.

gazpachoking:
I was also wondering if 'via' might be more appropriate for the list of notifier outputs, e.g.:

```yaml
notify_task:
  via:
  - email:
      to: blah@blah.com
```
Eliminates the double nested 'to' option in the email case. I think it's a bit more accurate description as well, i.e. the notification is being sent 'via' one of the transport plugins, and 'to' a recipient defined in that transport plugin config. (BTW, maybe need some more consistent terminology to discuss these plugins. What do we call the message generators, (notify_task, notify_abort, etc.,) vs the transport ones, (email, pushbullet, etc.?) similar to the problem of 'list plugins' (list_match, list_add, etc.) vs 'list_plugins' (imdb_list, trakt_list, etc.))

flexget/plugins/notifiers/notify.py
```diff
@@ -2,208 +2,67 @@
...
-        scope = config.pop('scope')
-        what = config.pop('what')
-        notifiers = config.pop('to')
-
-        # Creating iterator based on config scope
-        iterate_on = [getattr(task, container) for container in what] if scope == 'entries' else [[task]]
-
+                title = template_renderer(title)
+            except RenderError as e:
+                log.error('Error rendering notification title: %s', e)
```
liiight: I liked having a fallback dict, but it's no biggie I guess

flexget/plugins/notifiers/notify_entry.py
liiight: notification template maybe? seems long and too verbose

liiight:
I like using `via` but I'm not sure if this warrants ANOTHER upgrade actions in spite all the noise this change already made...
About the semantics, I suggest we lock down the manged list semantics first :stuck_out_tongue:

gazpachoking:
> About the semantics, I suggest we lock down the manged list semantics first

I think the terminology should be the same. We have a pattern of defining an api the plugins in a certain group should implement. We then have plugins in that group, and plugins that use plugins in that group. This applies both to the list plugins, and the notification plugins. I just want to figure out what the hell to call things when discussing. 😛

flexget/plugins/notifiers/notify.py
gazpachoking: I sorta feel like having the unrendered string will be a better indication of error when debugging why their config isn't working rather than getting some string that isn't to do with the title/message they defined.

alneven:
for me the email should be very simple, it should be HTML and be sent at every new DOWNLOAD, like this:

[http://i.imgur.com/l5E0Pwh.png](url)

alneven:
and for me one thing is not clear, what is the different between 

> notify_task

 and 

> notify_entry

?

Thanks for the hard work to make this happen!

liiight:
@alneven `notify_task` sends notification per task execution, `notify_entry` (or `notify_entires` as it currently exists) will send a notification per entry in the task, for each accepted entry by default.,

alneven:
In my case I would need `notify_entry` as I have one task to download TV shows, and for each downloaded episode I would need one email as notification. With the old HTML email template, pimped with the "title" as release_name.

liiight:
This really isn't the place for this, pick this up in the forum.

flexget/plugins/notifiers/email.py
paranoidi: @liiight What you had was different, it was a single dict called "data" and contained message and configuration in same dict. This is what I proposed as first ideal solution but there were some challenges with that at the time. This is fine.

flexget/plugins/notifiers/notify.py
```diff
@@ -1,212 +1,131 @@
...
-        config.setdefault('what', ['accepted'])
-        if not isinstance(config['what'], list):
-            config['what'] = [config['what']]
-        config.setdefault('title', DEFAULT_DICTS[config['scope']]['title'])
-        config.setdefault('message', DEFAULT_DICTS[config['scope']]['message'])
-        config.setdefault('url', DEFAULT_DICTS[config['scope']]['url'])
+def render_config(config, template_renderer, _path=''):
+    """
+    Recurse through config data structures attempting to render any string fields against a given context.
+    :param config: Any simple data structure as retrieved from the FlexGet config.
```
paranoidi: Empty line before :param

flexget/plugin.py
```diff
@@ -177,14 +178,19 @@ def decorator(target):
 _new_phase_queue = {}
 
 
-def register_task_phase(name, before=None, after=None):
-    """Adds a new task phase to the available phases."""
+def register_task_phase(name, before=None, after=None, suppress_abort=False):
+    """
+    Adds a new task phase to the available phases.
+    :param suppress_abort: If True, errors during this phase will be suppressed, and not affect task result.
```
paranoidi: Empty line before :param

liiight:
@gazpachoking i'd say add `url` as a default and keep the separation of `notify_task` and `notify_entry`.

flexget/plugins/notifiers/notify_entry.py
```diff
@@ -65,7 +64,7 @@ def on_task_notify(self, task, config):
             try:
                 message = get_template(config['template'], scope='entry')
             except ValueError:
-                raise plugin.PluginError('Cannot locate template on disk: %s', config['template'])
+                raise plugin.PluginError('Cannot locate template on disk: %s' % config['template'])
```
gazpachoking: I thought we settled on doing it the lazy way?

gazpachoking:
> i'd say add url as a default and keep the separation of notify_task and notify_entry.

I'm leaning the same way on keeping notify_task and _entry split.

I'm starting to lean the other way on the url thing though, here's some of my questions:
- Should url have a default value if user hasn't configured it?
  - What would it be for notify_task (or notify_entry for that matter?)
- What should plugins that don't take the url field do? Edit the message to include it? Ignore it?
- Might be a bit confusing that configuring that field does nothing depending on your notifier transport.
(I actually have similar concerns about the notifiers that don't accept 'subject', but I'm not really as concerned about that.)

gazpachoking:
Just converted sns plugin back out of the notifier system. I don't think it fit, as it ignored the 'message' parameter that the notification system passed in and made its own (it seems to output structured json, not just a message.) I moved it to the on_task_notify phase, which means the task won't fail (and entries won't be tried to add to sns after fail) if the sns posting doesn't work. I'm a bit hesitant on the whole thing because I have absolutely no idea what the sns plugin is being used for, and whether these change will affect said users.

liiight:
I probably shouldn't have refactored sns. I'd say just revert to to how it was before 2.8.

liiight:
about url, I agree that it doesn't make sense to when using task scope, that's why I had it default to empty string. But I do think it's a nice default to have in entry scope,  maybe default to imdb or tvdb url. 
someday we'll have generic lookup support we could use that instead

gazpachoking:
Okay, I finished the combination of notify_task and notify_entries back under just notify plugin. Right now I renamed the framework plugin (that other plugins use to send notifications,) to `notify_`, which isn't so nice. Open for suggestions on what to do with that.

gazpachoking:
Also, should notify_abort be combined back to notify plugin as well?

gazpachoking:
@paranoidi @liiight Any comments on the last 3 todo items?

liiight:
I'd say making `notify_abort` work like the other two makes sense, no need to have two schema paradigms.
About having no output plugins in the task, I don't know how common of an issue that is, and if it's worth supporting as a real life scenario. maybe handle that problem if and when it arises?

andocromn:
@liiight I wanted to comment on the "no other output plugins" consideration since I was the one to raise this concern.  Although I agree its probably not common, the email plugin used to be just a normal output so there was nothing wrong with this configuration.  I raised the concern to address the changes in the behavior resulting from introduction of the notify phase and discuss how to best maintain the previous functionality while still implementing the improvements.  As a result of the changes associated with the notify_phase, a task that only has a notification output will no longer retry sending the notification if the notification fails.  I think that if the notification is the only output, and the notification fails, it should consider that entry to be failed 

As an alterantive, I'm not sure if this was considered, but the best way of handling may be to create a notification queue.  This way, regardless of the output, any failed notifications could be retried on the next run of the task and would be sent to the user without the program re-downloading

liiight:
@gazpachoking notifyosd is a crappy name but it's not original name https://wiki.ubuntu.com/NotifyOSD

gazpachoking:
> @gazpachoking notifyosd is a crappy name but it's not original name https://wiki.ubuntu.com/NotifyOSD

Oh hmm. It works with this too, right? https://wiki.archlinux.org/index.php/Desktop_notifications 'notify osd' seems to be just one implementation for unity.

liiight:
I have no idea...

liiight:
Sweet, windows osd. Works on 7 and up I suppose?

gazpachoking:
> Sweet, windows osd. Works on 7 and up I suppose

Works on 10 at least, (though not optimal integration with the new action center,) not sure how far back it goes, I assume it works fine on 7.

gazpachoking:
What should we call the osd notifier now that it's a bit more generic? Would 'popup' or 'desktop' or something be more appropriate?

liiight:
I think `on_screen_display` is generic enough. Also, I think it's schema should also be a Boolean, so you could just do `on_screen_display: yes`

liiight:
Hmm,got this crash while testing it:
```
4 VERBOSE  task          ser             ACCEPTED: `series.s01e01.720p.hdtv-flexget` by accept_all plugin                           2017-01-06 01:14 VERBOSE  details       ser             Summary - Accepted: 1 (Rejected: 0 Undecided: 0 Failed: 0)                                 2017-01-06 01:14 WARNING  task          ser             Task doesn't have any output plugins, you should add (at least) one!                       2017-01-06 01:14 CRITICAL task          ser             BUG: Unhandled error in plugin notify: Missing parentheses in call to 'print' (__init__.py, line 39)                                        Traceback (most recent call last):
  File "/data/data/com.termux/files/home/Flexget/flexget/task.py", line 490, in __run_plugin          return method(*args, **kwargs)
  File "/data/data/com.termux/files/home/Flexget/flexget/event.py", line 23, in __call__              return self.func(*args, **kwargs)              File "/data/data/com.termux/files/home/Flexget/flexget/plugins/notifiers/notify.py", line 129, in on_task_notify                                     send_notification(config['task']['title'], template, config['task']['via'], template_renderer=task.render)                                       File "/data/data/com.termux/files/home/Flexget/flexget/plugins/notifiers/notification_framework.py", line 120, in send_notification                  notifier.notify(title, message, rendered_config)  # TODO: Update notifiers for new api          File "/data/data/com.termux/files/home/Flexget/flexget/plugins/notifiers/on_screen_display.py", line 28, in linux_notify                             from gi.repository import Notify               File "/data/data/com.termux/files/usr/lib/python3.5/site-packages/gi/__init__.py", line 39          print url                                                ^                                    SyntaxError: Missing parentheses in call to 'print'                      
```

liiight:
I'm not so sure anymore about the `via` keyword now that we have the `task` and `entries` levels:
```yaml
notify:                                                                                            task:                                                                                    via:                                                                                     - on_screen_display: {}
``` 
This feels too much to me...

gazpachoking:
> Hmm,got this crash while testing it:

Very tough to read that traceback. Looks like maybe the gobject introspection doesn't support python 3? I'm thinking of reverting the changes to osd for now and dealing with that in a separate PR. Maybe include the rename, and allow for bool schema, that way we could upgrade it outside of next branch.

gazpachoking:
> I'm not so sure anymore about the via keyword now that we have the task and entries levels:

Dunno what there is to be done about that. We could move the via to directly under notify, but that would mean you couldn't configure e.g. aborts to notify via a different method than entries.

PR #1549, opened 2016-12-11
anidb_feed input plugin

Andy2244:
### Motivation for changes:
Anidb file feed support, for global or private file notification feeds. The namespace fields are quite unique so the normal rss handler wont do the job. Entries also needed special care to build valid title names.

### Detailed changes:
new anidb_feed plugin
creates entries that can be used as input

```
        Creates an entry for each movie or series in the AniDB file notification feeds.
        Your personalized notification feed url is shown under [Account/Settings/Notifications/link to your personal atom feed]
        See: https://wiki.anidb.net/w/Notifications
        The general main anidb file feed can also be used url: http://anidb.net/feeds/files.atom
        The feeds update every 15 minutes, which is the cache time we also use in Flexget.

        Example:
            anidb_feed:
                url: http://anidb.net/perl-bin/animedbfeed.pl?id=xxxxx................
                valid_only: yes

            valid_only: yes
            # If enabled only 'valid' entries are returned, that have a size, release group, resolution and source field
            # This happens after the file was added to anidb and has passed some initial tests/parsing

            priority: (all, medium-high, high)
            # defaults to 'all', which are the anidb notification priorities

            include: ('special', 'op-ending', 'trailer-promo')
            # Also report none regular feed file types, by default any none regular file is skipped.
            anidb_feed:
                url: http://anidb.net/perl-bin/animedbfeed.pl?id=xxxxx................
                include: special

        Adds those extra Entry fields if found:
        'content_size'
        'rss_pubdate'
        'anidb_fid'             # anidb file_id
        'anidb_file_version'    # fileversion aka v1-5
        'anidb_file_crc32'      # the anidb validated hash for this file relese
        'anidb_feed_added'
        'anidb_feed_source'
        'anidb_feed_resolution'
        'anidb_feed_crc_status'
        'anidb_feed_language'
        'anidb_feed_group'
        'anidb_feed_grouptag'
        'anidb_feed_subtitle_language'
        'anidb_feed_priority'
        'anidb_feed_quality'
        'anidb_feed_size'
```


liiight:
any chance you'd consider making this a manged list plugin? if not, no biggie

Andy2244:
Not yet, i'm working on a anidb api meta lookup plugin next and will see how this works out. We might need a very special anime managed list plugin anyway, because of the mind mending way episodes/seasons are named and to get valid "next" episodes, some magic needs to happen. The meta lookup is needed to get from the fid to the aid, so matching is exact.

PS: How can i add external python libs support to flexget? I would like to use https://github.com/winterbird-code/adbb or https://github.com/lad1337/adba

liiight:
put import in try except block. look into some plugins for reference

paranoidi:
Name needs some tweaking, we don't have anything "_feed" named. Maybe just "anidb" ? :3

flexget/plugins/input/anidb_feed.py
```diff
@@ -0,0 +1,436 @@
...
+    def is_valid_entry(entry, ns_prefix_name=''):
+        for key in VALIDATION_LIST:
+            key_name = ns_prefix_name + key
+            if not entry.get(key_name):  # int(0) is invalid
+                return False
+            elif entry[key_name] in ['Raw/Unknown', 'N/A', 'Unchecked']:
+                return False
+        return True
+
+    def parse_from_xml(self, xml_entries, config):
```
paranoidi: You should try to split this, too long method. Some say correct size is about 10 lines :)

flexget/plugins/input/anidb_feed.py
```diff
@@ -0,0 +1,436 @@
...
+from flexget import plugin
+from flexget.event import event
+from flexget.utils.cached_input import cached
+from flexget.entry import Entry
+from flexget.utils.requests import Session, TimedLimiter
+
+
+log = logging.getLogger('anidb_feed')
+
+requests = Session()
```
paranoidi: Module level initialization should be avoided.

flexget/plugins/input/anidb_feed.py
```diff
@@ -0,0 +1,436 @@
...
+
+# def dump_entry(entry):
+#     log.verbose('#####################################################################################')
+#     for key in entry:
+#         log.verbose('Entry: [%s] = %s' % (key, entry[key]))
+#     log.verbose('#####################################################################################')
+
+
+class AnidbFeed(object):
+    """"Creates an entry for each movie or series in the AniDB notification feed.
```
paranoidi: Too long docstring lines

Andy2244:
There is already a 'anidb_list' plugin and i saw 'twitterfeed'. I means it creates entries based on a file feed, so i could only think of 'anidb_filefeed' or 'anidb_feed' ? I did not want to use the '_list' prefix, since it only works on the feed provided, not some selectable list status like watchlist, wishlist....

If you have naming suggestion feel free to comment.

liiight:
Is there a reason for both this and anidb list to exist simultaneously?

Andy2244:
I think so, one should ideally handle all your anidb lists entries, while the other is able to handle anidb file feeds. 
I guess you could also merge the two, are there other plugins that handle rss feeds and more complex list scraping in the same plugin?

paranoidi:
Search for DependencyError from codebase to see how to deal with non included 3rd party libraries. Remember that we aim to be py2.7, 3.3+ compatible.

As for naming twitterfeed should also be called just twitter. Many of the inputs we have are feed types but we don't name them that way still.

I hope you are not discouraged by my comments, keep up the good work :)

Andy2244:
@paranoidi its fine, i wish i could use c++, c#, java since i have a really hard time adapting to python, especially the automatic type conversion drives me crazy.
So far most of the comments on my stuff was helpfully and i found ways to improve the quality or basic design. So of course i value the inputs from the Flexget veterans.

flexget/plugins/input/anidb_feed.py
```diff
@@ -0,0 +1,479 @@
...
+        try:
+            number = float(valuestring)
+        except ValueError:
+            log.trace('Invalid number field: %s' % valuestring)
+    except Exception as ex:
+        log.trace('Invalid number field: %s error: %s' % (valuestring, ex))
+    return number
+
+
+def _debug_dump_entry(entry):
```
paranoidi: we have --dump which you might find more useful :)

remove this

flexget/plugins/input/anidb_feed.py
```diff
@@ -0,0 +1,479 @@
...
+            return []
+
+        if not len(parsed_xml.entries) > 0:
+            log.info('No entries returned from xml.')
+            return []
+
+        entries = self.parse_from_xml(parsed_xml.entries, config)
+        if len(entries) == 0:
+            log.verbose('No entries parsed from xml.')
+        # else:
```
paranoidi: leftovers

flexget/plugins/input/anidb_feed.py
```diff
@@ -0,0 +1,479 @@
...
+                                if dt_node.childNodes[0].nodeValue is not None and dd_nodes[idx].childNodes[0].nodeValue is not None:
+                                    dt_node.tagName = NAMESPACE_NAME + ':%s' % dt_node.childNodes[0].nodeValue.replace(' ', '_')
+                                    dt_node.setAttribute('name', dt_node.childNodes[0].nodeValue)
+                                    dt_node.setAttribute('value', dd_nodes[idx].childNodes[0].nodeValue)
+                                    dt_node.removeChild(dt_node.childNodes[0])
+                                    node_ns.removeChild(dd_nodes[idx])
+                                    entry.appendChild(dt_node)
+        except Exception as ex:
+            logging.debug('Unable to rename nodes in XML: %s' % ex)
+            return None
```
paranoidi: Way too wide exception catch that completely hides crashes. You want to log traceback if doing such wide catch.
You are returning empty list or None in many error cases. This is bad design, for error cases one should rely exceptions, not return values.

Later when some code tries to do something with the return value it will crash, or alternatively you will need always to check if you have proper value or not, which is usually more work than exceptions / worse.

Check if python has some good builtin exception to be used, or if this is fatal raise PluginError

flexget/plugins/input/anidb_feed.py
```diff
@@ -0,0 +1,479 @@
...
+    def get_value_via_regex(self, key, entry, re_string, group_nr=1):
+        result = None
+        if entry.get(key):
+            value_string = entry[key]
+            if isinstance(value_string, basestring) and not value_string.isspace():
+                match = re.search(re_string, value_string)
+                if match and match.group(group_nr) and not match.group(group_nr).isspace():
+                    result = match.group(group_nr)
+            else:
+                log.error('Expecting string for re lookup got: %s, tyep: %s' % (value_string, type(value_string)))
```
paranoidi: Don't use string formatting with logger

use params

```
log.error('Expecting string for re lookup got: %s, tyep: %s', value_string, type(value_string)))
```

flexget/plugins/input/anidb_feed.py
```diff
@@ -0,0 +1,479 @@
...
+                title_new += ' - [%s]' % group_tag  # TODO: do we always add group?
+                new_entry[NAMESPACE_PREFIX + 'grouptag'] = group_tag
+            if new_entry.get('quality') and hasattr(new_entry['quality'], 'resolution'):
+                if new_entry['quality'].resolution.name != 'unknown':
+                    title_new += ' [%s]' % new_entry['quality'].resolution
+
+            new_entry['title'] = title_new
+            new_entry[NAMESPACE_PREFIX_MAIN + 'name'] = anidb_name
+            entries.append(new_entry)
+            #_debug_dump_entry(new_entry)
```
paranoidi: leftovers

flexget/plugins/input/anidb_feed.py
```diff
@@ -0,0 +1,479 @@
...
+        return entries
+
+    def fill_entries_for_url(self, url, task, config):
+        log.verbose('Fetching %s' % url)
+
+        try:
+            r = task.requests.get(url, headers=HEADER, timeout=20)
+        except Exception as ex:
+            log.error("Failed fetching url: %s error: %s" % (url, ex))
+            return []
```
paranoidi: See my longer comment on error handling

flexget/plugins/input/anidb_feed.py
```diff
@@ -0,0 +1,479 @@
...
+                            'Error code: %s detail: %s' % (xml_feed['error']['code'], xml_feed['error']['description']))
+                    else:
+                        log.error('Error code: %s' % xml_feed['error']['code'])
+        except Exception as ex:
+            log.error('Unable to parse the XML from url: %s error: %s' % (url, ex))
+            return []
+
+        if not len(parsed_xml.entries) > 0:
+            log.info('No entries returned from xml.')
+            return []
```
paranoidi: See my longer comment on error handling

flexget/plugins/input/anidb_feed.py
```diff
@@ -0,0 +1,479 @@
...
+            if 'error' in xml_feed:
+                if 'code' in xml_feed['error']:
+                    if 'description' in xml_feed['error']:
+                        log.error(
+                            'Error code: %s detail: %s' % (xml_feed['error']['code'], xml_feed['error']['description']))
+                    else:
+                        log.error('Error code: %s' % xml_feed['error']['code'])
+        except Exception as ex:
+            log.error('Unable to parse the XML from url: %s error: %s' % (url, ex))
+            return []
```
paranoidi: See my longer comment on error handling

flexget/plugins/input/anidb_feed.py
```diff
@@ -0,0 +1,479 @@
...
+                log.warning('Unsupported attribute type: %s via name: %s' % (in_type, tagname))
+
+    def fill_namespace_attributes(self, xml_entry, entry):
+        for key in NAMESPACE_ATTRIBUTE_MAP:
+            self.set_ns_attribute(key, xml_entry, entry, NAMESPACE_NAME, NAMESPACE_PREFIX, NAMESPACE_ATTRIBUTE_MAP[key])
+
+
+@event('plugin.register')
+def register_plugin():
+    plugin.register(AnidbFeed, 'anidb_feed', api_ver=2, groups=['list'])
```
paranoidi: This does not implement list interface so it should not belong to that group.

flexget/plugins/input/anidb_feed.py
Andy2244: > Way too wide exception catch that completely hides crashes.
Yes i still have problems understanding how to handle errors in flexget/python, can you give me an example on how to handle this case?
I use the empty/None returns on errors that i won't handle, so should i just write the code as if no error might happen and put a try on the topmost caller? Should i expect the caller of my plugin to have a try/catch? I mean in this specific case i can either throw an exception and flexget just stops/crashes or i can just do a error log and use the designed alternate code-path?

I don't fully understand the issue with none/empty objects, if a caller ask for a object and the function cant get a valid object, i'm supposed to throw a exception? If that's the case, sure i can change those returns to raise specific exceptions instead?

PS: Unfortunately my c++ background is not helping either, since the projects i was involved in the past not even had exception handling enabled. So i'm used to try always return a valid object, even if its a empty object and as fallback return a none object that the caller has to handles. In the past i noticed that no one really handles specific exceptions anyway and just lets the application crash/stop, which is also bad design. I really liked code contracts in c#, so exceptions handling is not my strong side.

flexget/plugins/input/anidb_feed.py
Andy2244: I was assuming this is needed to register the plugin as some sort of input plugin? Seems i looked at the wrong examples, since all i checked also had this 'list' type, without using the list interface. Guess those are all old plugins than.

flexget/plugins/input/anidb_feed.py
Andy2244: > Way too wide exception catch that completely hides crashes.

That's actually the point in this case, this purely optional fixup code, so if anything is not as expected we just skip and use the unfixed version of the xml.

> You are returning empty list or None in many error cases. This is bad design, for error cases one should rely exceptions, not return values.

Thats because i cant see/find a way to enforce a contract with the caller, to handle my exceptions. What i see is a consensus that code checks vs None or work with empty lists.

> Check if python has some good builtin exception to be used, or if this is fatal raise PluginError

That seems reasonable, but my problem is that outside of a top level error, why should i raise a exception for a unimportant field/operation of a single entry, if i could instead just skip this field/entry and continue working on the rest of the entries?

flexget/plugins/input/anidb_feed.py
Andy2244: How do you handle debug code in general? Its somewhat unpractical to remove and add debug code, after each commit. I'm used to use _DEBUG defines, so whats the workflow in python?

flexget/plugins/input/anidb_feed.py
Andy2244: Can you explain why and whats the difference plz? I was assuming both lines are similar/same?

flexget/plugins/input/anidb_feed.py
Andy2244: Not really, `dump()` uses too short left anchor, which results in hard readable line entries and has no ability to output the actual field type, which i was interested in this case.

flexget/plugins/input/anidb_feed.py
paranoidi: Debug code can be left in if really frequently used, it's quite easy to add such trivial debug bits back in if needed later. Using log.debug is fine too.

flexget/plugins/input/anidb_feed.py
paranoidi: Empty list can be ok return value sometimes ... but for failing to retrieve the thing we need? Usually we abort the task then.

flexget/plugins/input/anidb_feed.py
paranoidi: Even those plugins don't inherit / implement any explicit interface. Duck typing being used here heavily. We are in the talks of changing this parameter name to "interfaces" (from groups).

flexget/plugins/input/anidb_feed.py
paranoidi: That is proper way to use logging, it is possible to uniquely identify this error line. If you use string substition they all are different (think of hash of the first param). We don't use these capabilities though ATM, but it's better to start using it correctly nonetheless :)

flexget/plugins/input/anidb_feed.py
```diff
@@ -0,0 +1,479 @@
...
+        elif tagname in xml_entry:
+            if isinstance(xml_entry[tagname], dict) and 'value' in xml_entry[tagname]:
+                value = xml_entry[tagname]['value']
+            elif isinstance(xml_entry[tagname], basestring):
+                value = xml_entry[tagname]
+
+        if out_prefix_name is not None and not out_prefix_name.isspace():
+            tagname = out_prefix_name + name
+
+        if value is not None:
```
paranoidi: Readability would benefit of reverting the if check to "if value is None" -> log, return so you don't need to indent whole block. Extracting into own method also worthwhile to consider..

paranoidi:
Seems mostly OK with some proposed changes

flexget/plugins/input/anidb_feed.py
gazpachoking: The 'list' interface (nee 'group') should mean the plugin implements the `get_list` method. Plugins that implement the `on_task_X' methods don't need any group at the moment, but once the 'group'->'interface' [change](#1555) goes live they should declare the 'task' interface.

flexget/plugins/input/anidb_feed.py
gazpachoking: The params way also avoids some extra work in certain cases. When a message won't be displayed at the current logging level, the system never even does the string replacement. If you don't use parameters, the string replacement work has to be done no matter if the log message is going to be displayed or not. This is normally not a big issue, but could be bad inside some heavy loops.

Andy2244: oki doki, will replace my log output.

Andy2244:
implemented suggested changes

flexget/plugins/input/anidb_feed.py
```diff
@@ -0,0 +1,485 @@
...
+
+from flexget.utils.qualities import Quality
+from flexget.utils.tools import str_to_boolean
+
+from flexget import plugin
+from flexget.event import event
+from flexget.utils.cached_input import cached
+from flexget.entry import Entry
+from flexget.utils.requests import TimedLimiter
+from requests import RequestException
```
cvium: This one is a third-party import though.

flexget/plugins/input/anidb_feed.py
```diff
@@ -0,0 +1,485 @@
...
+from xml.dom import minidom
+
+import feedparser
+from time import mktime
+from dateutil.parser import parse as dateutil_parse
+from dateutil.tz import tz
+from datetime import datetime
+
+from flexget.config_schema import one_or_more
+
```
Andy2244: oki, sorry i still rely on pyCharms auto import feature

flexget/plugins/input/anidb_feed.py
```diff
@@ -0,0 +1,485 @@
...
+}
+
+ANIDB_EXTRA_TYPES = [
+    'special',
+    'op-ending',
+    'trailer-promo'
+]
+
+
+def convert_to_naive_utc(valuestring):
```
cvium: See comment on `convert_to_number`

flexget/plugins/input/anidb_feed.py
cvium: I stopped using it a few months ago because it started doing weird stuff like this.

flexget/plugins/input/anidb_feed.py
```diff
@@ -0,0 +1,485 @@
...
+        ep_nr = self.get_value_via_regex(key, entry, r'^.*? - ([0-9]{1,3})(v([1-5]))? -')
+        file_version = self.get_value_via_regex(key, entry, r'^.*? - T|S|C|ED|OP|[0-9]{1,3}[a-f]?v([1-5]) -')
+        is_special = self.get_value_via_regex(key, entry, r'^.*? - (S[0-9]{1,3})(v([1-5]))? -')
+        is_op_ed = self.get_value_via_regex(key, entry, r'^.*? - ((C|OP|ED)[0-9]{1,2}[a-f]?) -')
+        is_trailer = self.get_value_via_regex(key, entry, r'^.*? - (T[0-9]{1,2}) -')
+        if ep_nr:
+            try:
+                # anidb always uses absolute ep numbering, even movies get "- 1 -" !!
+                ep_nr = int(ep_nr)
+            except Exception as ex:
```
cvium: `int()` should only raise `ValueError` afaik

flexget/plugins/input/anidb_feed.py
```diff
@@ -0,0 +1,485 @@
...
+                if self.is_valid_entry(new_entry, NAMESPACE_PREFIX) is False:
+                    continue
+            # now manually fix known fields
+            # 344.18 MB (360.895.922)
+            size_string = self.get_value_via_regex(NAMESPACE_PREFIX + 'size', new_entry, r'\((([0-9]{1,3}\.|[0-9]{1,3}){1,5})\)')
+            if size_string:
+                bytes_string = size_string.replace('.', '')
+                try:
+                    size_bytes = int(bytes_string)
+                    size_mb = int(size_bytes / 1024 / 1024)  # MB
```
cvium: We have a utility function for something like this. Can it be used here? https://github.com/Flexget/Flexget/blob/develop/flexget/utils/tools.py#L435

flexget/plugins/input/anidb_feed.py
```diff
@@ -0,0 +1,485 @@
...
+                log.error('Skipping entry, invalid anidb name parsed from: %s', new_entry['title'])
+                continue
+
+            # fix and add to new_entry()
+            quality_string = ''
+            if resolution_string:
+                quality_string += resolution_string
+            if source_string:
+                quality_string += ' ' + source_string
+            if not quality_string.isspace():
```
cvium: Won't this be true if neither `resolution_string` nor `source_string` contain a value?

flexget/plugins/input/anidb_feed.py
Andy2244: Kinda, but i want to store the org 'safe' size in bytes anyway. Using the byte string at the end of the input string seemed safer to me, than relying on proper MB/GB/KB tokens.
Is there a version that only parses valid byte like numbers?

flexget/plugins/input/anidb_feed.py
Andy2244: yup good catch, thx

Andy2244:
I'm confused why does it fail the check for 2.7/3.4, but not 3.3/3.5 ? It seems to crash in some trakt code, what i'm supposed to now?

cvium:
Just ignore it.

flexget/plugins/input/anidb_feed.py
```diff
@@ -0,0 +1,453 @@
...
+                number = str_to_number(value)
+                if number is not None and number > 0:
+                    boolvalue = True
+                else:
+                    boolvalue = False
+            entry[tagname] = boolvalue
+
+    def _fill_namespace_attributes(self, xml_entry, entry):
+        for key in NAMESPACE_ATTRIBUTE_MAP:
+            self._set_ns_attribute(key, xml_entry, entry, NAMESPACE_NAME, NAMESPACE_PREFIX, NAMESPACE_ATTRIBUTE_MAP[key])
```
paranoidi: Mutable methods are harder to read, it would be probably much more readable if this could be done like

`entry[key] = self._parse_xml_entry(xml_entry, ....)`

Throw exception if value can not be parsed/set

flexget/plugins/input/anidb_feed.py
```diff
@@ -0,0 +1,453 @@
...
+                else:
+                    log.error('Error code: %s', feed['error']['code'])
+            raise plugin.PluginError('Parsed XML is a error return for url: %s' % url)
+
+        if len(parsed_xml.entries) == 0:
+            log.verbose('No entries returned from xml.')
+            return []
+
+        entries = self.parse_from_xml(parsed_xml.entries, config)
+        if len(entries) == 0:
```
paranoidi: `if not entries:`

len(iter) == 0 is antipattern

flexget/utils/tools.py
```diff
@@ -505,3 +507,32 @@ def parse_episode_identifier(ep_id):
...
+        except ValueError:
+            parsed_date = parsed_date.replace(tzinfo=None)
+    except ValueError as ex:
+        log.warning('Invalid datetime field format: %s error: %s', string, ex)
+    except Exception as ex:
+        log.debug('Unexpected datetime field: %s error: %s', string, ex)
+    return parsed_date
+
+
+def str_to_number(string):
```
paranoidi: string shadows builtin string, call variable something else (eg, value)

flexget/plugins/input/anidb_feed.py
```diff
@@ -0,0 +1,453 @@
...
+                else:
+                    boolvalue = False
+            entry[tagname] = boolvalue
+
+    def _fill_namespace_attributes(self, xml_entry, entry):
+        for key in NAMESPACE_ATTRIBUTE_MAP:
+            self._set_ns_attribute(key, xml_entry, entry, NAMESPACE_NAME, NAMESPACE_PREFIX, NAMESPACE_ATTRIBUTE_MAP[key])
+
+
+@event('plugin.register')
```
paranoidi: registration is usually at the bottom of the module

Andy2244: Um, it is at the bottom, are we looking at the same code?

flexget/plugins/input/anidb_feed.py
```diff
@@ -0,0 +1,453 @@
...
+            priority = 0
+            if config['priority'] == 'medium-high':
+                priority = 1
+            elif config['priority'] == 'high':
+                priority = 2
+            priority_fragment = '&pri=%s' % priority
+        return self.fill_entries_for_url(url + priority_fragment, task, config)
+
+    @staticmethod
+    def is_valid_entry(entry, ns_prefix_name=''):
```
cvium: Is there a particular reason for defaulting to the empty string? It seems like you're always providing one. I also hate abbreviations. I can reasonably guess that `ns` is namespace, but I shouldn't have to.

flexget/plugins/input/anidb_feed.py
```diff
@@ -0,0 +1,453 @@
...
+
+HEADER = {'User-agent': 'Mozilla/5.0'}  # anidb needs valid header or we get '403 Client Error'
+LIMITER = TimedLimiter('anidb.net', '2 seconds')  # default anidb api limit
+
+
+NAMESPACE_NAME = 'xhtml'
+NAMESPACE_URL = 'http://www.w3.org/1999/xhtml'
+NAMESPACE_TAGNAME = 'dl'
+NAMESPACE_PREFIX_MAIN = 'anidb_'
+NAMESPACE_PREFIX = 'anidb_feed_'
```
cvium: I'm not a fan of this naming convention. Namespace is very C++/C#-esque. At a glance, it seems like you're using prefix and prefix_main as prefixes for entry fields. I'd rather the variables were called something like `field_prefix`.

flexget/plugins/input/anidb_feed.py
Andy2244: I just got inspired by `update_using_map` and added value conversions.
`def update_using_map(self, field_map, source_item, ignore_none=False):`

So what exactly should i do? So instead of using a easy to refactor/maintain input_map i should change this to individual calls like this?
`entry[key] = self._parse_xml_entry(xml_entry, ....)`

I want to use a input_map, similar to `update_using_map`, i just do a automatic type conversion so i don't need to individually handle this.

flexget/plugins/input/anidb_feed.py
Andy2244: I guess so i can directly do this, without checking `ns_prefix_name`
`key_name = ns_prefix_name + key`
Its funny the parameter actually was named `namespace_prefix_name`, but i felt it was too long if used via `is_valid_entry(entry, namespace_prefix_name="mynamepsace")`

flexget/plugins/input/anidb_feed.py
Andy2244: Not sure what you mean by "C++/C#-esque", it actually just reflects what was stored in the logic xml namespace. So if i get `anidb` values and store it via `anidb_xy` than to me `anidb` is the namespace name used and adding `_` and putting it in front makes it to a prefix.
So to me in `anidb_feed_`, anidb is the root/main namespace name and `feed` is the namespace of the subcategory.
I think its just semantic, since to me flexget already uses namespaces all over the place, but if this is confusing then sure i will change the naming.

cvium: It's not something I feel strongly about. Maybe @paranoidi has some input.

flexget/plugins/input/anidb_feed.py
Andy2244: Let me try if i somehow can use `update_using_map`, but i have no clue how to use the syntax with a external util call to extract/convert the values? Can i just use a external util instead of a lambda?

```python
field_map = {
  'size': lambda node: _convert_to(_get_from_xml(node, 'size_node', type=int))
}
```

flexget/plugins/input/anidb_feed.py
Andy2244: Should not exist anymore if i refactor all to use `update_using_map`.

flexget/plugins/input/anidb_feed.py
paranoidi: ah, diff looks weird at github then

flexget/plugins/input/anidb_feed.py
paranoidi: great, also, update_using_map is a method of entry so it's fine if it changes itself :)

Andy2244:
Thats what i could come up with and will probably get scolded for :)

Problem is the xml node may not exist, so i need the exception check in update_from_map.

flexget/entry.py
```diff
@@ -254,7 +254,7 @@ def take_snapshot(self, name):
                 log.warning('Snapshot `%s` is being overwritten for `%s`' % (name, self['title']))
             self.snapshots[name] = snapshot
 
-    def update_using_map(self, field_map, source_item, ignore_none=False):
+    def update_using_map(self, field_map, source_item, ignore_none=False, ignored_values=None):
```
cvium: The naming should be in line with ignore_none ie. `ignore_values`

Andy2244:
Oki i did a full refactor to embrace the field_map system and the only code that remains is special fixup code, i could not fit into this logic. This should allow to easily tweak the fieldmap if changes occur.

Andy2244:
Wasn't sure how to implement disabling file logging, while still getting log output. I often work via a wan network drive and file logging kills test speed. Its also nice to disable file logging on systems like openwrt, to reduce nvram wear or safe ramdisk space.

flexget/utils/tools.py
```diff
@@ -14,58 +14,171 @@
...
+        try:
+            parsed_date = parsed_date.astimezone(tz.tzutc()).replace(tzinfo=None)
+        except ValueError:
+            parsed_date = parsed_date.replace(tzinfo=None)
+    except ValueError as ex:
+        log.warning('Invalid datetime field format: %s error: %s', value, ex)
+    return parsed_date
+
+
+def value_to_naive_utc(value):
```
cvium: Why do you have two functions that ultimately do the same thing?

Andy2244: The `str_to` versions can only deal with string inputs, while the `value_to` versions can deal with different input types. I needed this since sometimes xml fields are already in some specific date/number format, instead of a string.

cvium: Doesn't really explain why it's two separate functions.

Andy2244: oki fixed

flexget/manager.py
```diff
@@ -160,12 +160,15 @@ def __init__(self, args):
             logger.start(level=self.options.loglevel.upper(), to_file=False)
             raise
         else:
-            log_file = os.path.expanduser(self.options.logfile)
-            # If an absolute path is not specified, use the config directory.
-            if not os.path.isabs(log_file):
-                log_file = os.path.join(self.config_base, log_file)
+            if self.options.logfile.lower() == 'none':  # disable logfile output
```
paranoidi: This needs to go into a separate PR, unrelated to anidb

paranoidi:
@Andy2244 You're so close, don't give up yet!

Andy2244:
Kinda lost track, so what is preventing the merge?

flexget/utils/tools.py
```diff
@@ -160,12 +276,12 @@ def merge_dict_from_to(d1, d2):
                     merge_dict_from_to(d1[k], d2[k])
                 elif isinstance(v, list):
                     d2[k].extend(copy.deepcopy(v))
-                elif isinstance(v, (basestring, bool, int, float, type(None))):
```
cvium: Why was basestring removed?

Andy2244: Because i was told that basestring should not be used anymore and i spotted this here.

cvium: Who said that and in what context? Afaik, basestring still needs to be used.

Andy2244: You noted this in my other plugin, so this is only inside plugins and not tools?

> `'guid': basestring,`
We should pretend that this is python 3, and using 'str' instead of basestring (our from builtins import * makes this a real python 3 str (text) even on python 2.) If there is an actual possibility that the thing will be either text (str) or bytes (bytes), we should handle those cases explicitly.

cvium: I guess @gazpachoking should clarify then

Andy2244: So leave it as str or change all my new tools helpers to basestring?

Andy2244:
Btw can some explain the cassette system quickly? I cant find any info in the dev docs.
It seems if i run a local unittest that a file is created that mirrors the online request and i have to add this file to the project?

Andy2244:
What are the outstanding issues on this one?

PR #1551, opened 2016-12-12
Convert if plugin to use jinja expressions instead of python eval.

gazpachoking:
### Motivation for changes:
Python eval is not safe for user input. Jinja expressions are much more appropriate.

### Detailed changes:
- `if` plugin statements are now evaluated as jinja [expressions](http://jinja.pocoo.org/docs/dev/templates/#expressions)
- As jinja expressions have mostly the same syntax as the limited form of python we intended with the if plugin, most current configs should still work.
- We can now also use jinja [filters](http://jinja.pocoo.org/docs/dev/templates/#filters) and [tests](http://jinja.pocoo.org/docs/dev/templates/#tests) in if statements.
- implemented the expression evaluation via `flexget.utils.evaluate_expression` which allows for lazy fields to stay lazy for evaluation. (which the fully default jinja method wouldn't do by itself)

Our tests seem to just pass, not sure if there are cases where config changes could be needed.

### DIFF
--- a/flexget/plugins/filter/if_condition.py
+++ b/flexget/plugins/filter/if_condition.py
@@ -1,29 +1,20 @@
 from __future__ import unicode_literals, division, absolute_import
 from builtins import *  # noqa pylint: disable=unused-import, redefined-builtin
 from future.moves import builtins
-from past.builtins import basestring
 
 import logging
-import re
 import datetime
 from copy import copy
+
+from jinja2 import UndefinedError
 
 from flexget import plugin
 from flexget.event import event
 from flexget.task import Task
 from flexget.entry import Entry
+from flexget.utils.template import evaluate_expression
 
 log = logging.getLogger('if')
-
-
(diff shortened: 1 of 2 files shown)


gazpachoking:
I'd sortof like to change the if plugin config to bring the expressions to the value side of a dict, so that we can validate it with a schema. We could add a custom schema 'format' for these expressions to catch syntax errors during validation. Not exactly sure what the nicest way to do that is, and if it's worth the extra validation ability though.

liiight:
I think the implicit None is fine, `if` plugin was always programmatic in nature, this fits in well. 
schema format is a good idea but since a lot of fields in flexget use jinja not sure if it's relevant only to this plugin.

paranoidi:
Awesome! I always told you eval is evil! :)

gazpachoking:
I think we'll just go with the engine change for now. Config format can be revisited later.

PR #1554, opened 2016-12-12
[updated] fill in all the trakt fields

Andy2244:
### Motivation for changes: 
I noticed that without a extra trakt_lookup the next_trakt_episodes plugin did not fill in all the available trakt fields for the api call.

### Detailed changes:
fields now also filled:
`trakt_series_name, trakt_series_year, tvrage_id, imdb_id, tmdb_id, trakt_slug`


liiight:
thanks!

PR #1572, opened 2016-12-17
Remove Trakt images

cvium:
### Motivation for changes:
Trakt no longer provides images.
### Detailed changes:

- Removed images and updated cassettes.


liiight:
Looks good to me

PR #1573, opened 2016-12-17
add missing trakt fields from trakt_list

Andy2244:
### Motivation for changes:
Some trakt fields where not set for trakt_list imports

### Detailed changes:
added: `trakt_movie_name, trakt_movie_year, series_year, trakt_series_name, trakt_series_year`

I also noticed that there are multiple ways of handling the year and it looks somewhat inconsistent between plugins.
`year`
`series_year`
`movie_year`


flexget/plugins/input/next_trakt_episodes.py
```diff
@@ -75,6 +75,7 @@ def on_task_input(self, task, config):
                     trakt_id = item['show']['ids']['trakt']
                     listed_series[trakt_id] = {
                         'series_name': '%s (%s)' % (item['show']['title'], item['show']['year']),
+                        'series_year': item['show']['year'],
```
gazpachoking: I'm not sure we should set this. The `series_` fields are meant to come from the series plugin with no lookups required.

gazpachoking:
Looks like tests need a bit of updating as well.

flexget/plugins/input/next_trakt_episodes.py
Andy2244: I mean `movie_year` is also handled this way, so why not `series_` why not fill in all valid information we can get at a particular stage?

gazpachoking: Well, we have more of a built in concept of year for the movie stuff, and the offline parser fills it. With series, due to historical reasons, we have not had an explicit year involved, it's always just been lumped in in the series title.

gazpachoking: > why should it not also fill in the series_year, so those two fields are consistent.

Because it seems to imply that the series plugin has some sort of support for series_year, when it doesn't. I think it would be confusing to see that field populated for your series when you enable the trakt_lookup plugin, but not have it be there otherwise. When you enable the trakt_lookup plugin, you can explicitly access the year via trakt_series_year.

cvium: It's got nothing to do with 'trusted source'. The field `series_year` is not used anywhere, so there's no point starting now. `series` plugin needs a bigger overhaul and that is beyond the scope of this PR.

I can see that it is set here https://github.com/Flexget/Flexget/blob/1b2b13c205a65c39c6f3b0755fc23464db9d64f5/flexget/plugins/list/imdb_list.py#L234, but it shouldn't be.

Andy2244: It has nothing to-do where it is used in code, the user can use the fields via set/if plugins, custom.render(). 
So just to give a example, you want to set the title for a task to 'series_name' + year, assuming you normally omit this from you inputs. Now instead of using 'series_year' the same way you could use 'movie_year', you need specific information what metadata_lookup was used and successfully did the lookup and than separately check the namespace fields like `imdb_year, trakt_series_year, tvmaze_series_year`. My argument is why do i care what meta lookup populated this field as long as its valid, if the user has multiple meta lookup defined, why should he care which of those had a successful lookup and populated the year field, so it can be used?

cvium: I get your intentions, but as we've said, it's for historical reasons. Doesn't mean you don't have a valid point, it's just not something that belongs in this PR.

Andy2244:
Are there still problems with this PR?

cvium:
Well, you haven't fixed the tests https://travis-ci.org/Flexget/Flexget/jobs/185735702

PR #1579, opened 2016-12-19
Upgrade of 'secrets' plugin to 'variables'

### LINKED ISSUE #1578: Secrets plugin breaks jinja default filter
When a secrets file is used the jinja default filter displays the default value even when the entry exists. Occurs even if the secrets file is empty

#### Config:
```
secrets: secrets.yml
tasks:
  test:
    disable:
      - seen
    filesystem:
      path: ~/flexget/
      mask: '*.yml'
    accept_all: yes
    set:
      test1: "{{ title }}"
      test2: "{{ title|default('bla') }}"
    exec:
      - echo "{{test1}}{{test2}}"
```
  
#### Log:
With secrets file:
```
-- Accepted: ---------------------------
title            : config
url              : file:///home/user/flexget/config.yml
...
test1            : config
test2            : bla
```
Without secrets file (expected output):
```
-- Accepted: ---------------------------
title            : config
url              : file:///home/user/flexget/config.yml
...
test1            : config
test2            : config
```

### Additional information:

- Flexget Version: 2.8.7
- Python Version: 2.7.12+
- Installation method: pip in virtualenv
- OS and version: Ubuntu 16.10

gazpachoking:
### Motivation for changes:
There were some conflicts with secrets plugin when jinja was being used in the config (#1578). The name 'secrets' also seemed too limited for what was possible with the plugin.

### Detailed changes:
- Renames the 'secrets' plugin to 'variables'
- Changes jinja tag to {? ?} for replacement to avoid conflicts with jinja in the config
- Removes the need for the 'secrets' prefix on every variable

### Addressed issues:

- Fixes #1578

- Root level 'secrets' in config will need to be renamed to 'variables'.
- Any use of secrets in the config will need to have the surrounding tags replaced with {? ?} and remove the reference to 'secrets'. e.g.: ```{{ secrets.username }}``` becomes ```{? username ?}```


gazpachoking:
Started the wiki page upgrade at https://flexget.com/Plugins/variables We'll need to point secrets page there once this goes live.

PR #1583, opened 2016-12-21
input plugin for kitsu.io

NikkyAI:
### Motivation for changes:

making Flexget able to accept anime series based on their status in the kitsu.io library (former hummingbird.me)

### Detailed changes:

- updatted the old hummingbiird plugins [here](https://github.com/matthewdias/hummingbird_list) to the new api
- Entry fields
  - always:
    - title
    - url `= ''`
  - when available:
    - kitsu_title_en
    - kitsu_title_en_jp
    - kitsu_title_ja_jp
  - if `latest` is set:
    - series_episode
    - series_id_type `= 'sequential'`

```yaml
Syntax:
    kitsu:
      username: <value>
      lists:
        - <current|planned|completed|on_hold|dropped>
        - <current|planned|completed|on_hold|dropped>
      latest: <yes|no>
      limit_to: <airing|finished>

Example:
    configure_series:
      settings:
        identified_by: sequence
      from:
        kitsu:
          username: NikkyAi
          lists:
            - current
            - planned
          limit_to: airing
          latest: no
```

### possible additional features
- authetication (oath)
  - shows series marked as private
  - allows to update the series progress
- kitsu.io drama (but atm it seems like the drama category is a placeholder)
- kitsu.io manga (would people use flexget for that?)

### DIFF
--- a/flexget/plugins/input/kitsu.py
+++ b/flexget/plugins/input/kitsu.py
@@ -42,9 +42,9 @@ class KitsuAnime(object):
         try:
             user_response = task.requests.get('https://kitsu.io/api/edge/users', params=user_payload)
         except RequestException as e:
-            status = getattr(getattr(e, 'response', None), 'status_code', None)
-            error_message = 'Error finding User url: {url} status: {status}'.format(
-                url=e.request.url, status=status)
+            error_message = 'Error finding User url: {url}'.format(url=e.request.url)
+            if hasattr(e, 'response'):
+                error_message += ' status: {status}'.format(status=e.response.status_code)
             log.debug(error_message, exc_info=True)
             raise plugin.PluginError(error_message)
         user = user_response.json()
@@ -56,9 +56,9 @@ class KitsuAnime(object):
         try:
             response = task.requests.get(next_url, params=payload)
         except RequestException as e:
-            status = getattr(getattr(e, 'response', None), 'status_code', None)
-            error_message = 'Error getting list from {url} status: {status}'.format(
-                url=e.request.url, status=status)
+            error_message = 'Error getting list from {url}'.format(url=e.request.url)
+            if hasattr(e, 'response'):
+                error_message += ' status: {status}'.format(status=e.response.status_code)
             log.debug(error_message, exc_info=True)
             log.info(error_message, exc_info=True)
             raise plugin.PluginError(error_message)
@@ -99,9 +99,9 @@ class KitsuAnime(object):
                 try:
                     response = task.requests.get(next_url)
                 except RequestException as e:
-                    status = getattr(getattr(e, 'response', None), 'status_code', None)
-                    error_message = 'Error getting list from next page url: {url} status: {status}'.format(
-                        url=e.request.url, status=status)
+                    error_message = 'Error getting list from next page url: {url}'.format(url=e.request.url)
+                    if hasattr(e, 'response'):
+                        error_message += ' status: {status}'.format(status=e.response.status_code)
                     log.debug(error_message, exc_info=True)
                     raise plugin.PluginError(error_message)
             else:
--- a/flexget/plugins/input/kitsu.py
+++ b/flexget/plugins/input/kitsu.py
@@ -1,0 +1,107 @@
+from __future__ import unicode_literals, division, absolute_import
+import logging
+
+from flexget import plugin
+from flexget.entry import Entry
+from flexget.event import event
+from flexget.utils.cached_input import cached
+from requests import RequestException
+
+log = logging.getLogger('kitsu')
+
+
+class KitsuAnime(object):
+    """Creates an entry for each item in your kitsu.io list.
+    Syntax:
+    kitsu:
+      username: <value>
+      lists:
+      	- <current|planned|completed|on_hold|dropped>
+      	- <current|planned|completed|on_hold|dropped>
+      latest: <yes|no>
+      finishedonly: <yes|no>
+      currentonly: <yes|no>
+    """
+
+    schema = {
+        'type': 'object',
+        'properties': {
+            'username': {'type': 'string'},
+            'lists': {'type': 'array',
+                      'items': {'type': 'string', 'enum': ['current', 'planned', 'completed', 'on_hold', 'dropped']}},
+            'latest': {'type': 'boolean', 'default': False},
+            'currentonly': {'type': 'boolean', 'default': False},
+            'finishedonly': {'type': 'boolean', 'default': False}
+        },
+        'required': ['username'],
+        'additionalProperties': False,
+    }
+
+    @cached('kitsu', persist='2 hours')
+    def on_task_input(self, task, config):
+        entries = []
+        try:
+            user_payload = {'filter[name]': config['username']}
+            user_response = task.requests.get('https://kitsu.io/api/edge/users', params=user_payload)
+            user_response.raise_for_status()
+            user = user_response.json()
+            if len(user['data']) < 1:
+                raise plugin.PluginError('no such username found on kitsu.io')
+            userId = user['data'][0]['id']
+        except RequestException:
+            raise plugin.PluginError('Error getting User ID from kitsu.io')
+
+        next_url = 'https://kitsu.io/api/edge/users/{userId}/library-entries'.format(**locals())
+        payload = {'filter[status]': ','.join(config['lists']), 'filter[media_type]': 'Anime', 'include': 'media',
+                   'page[limit]': 20}
+
+        try:
+            response = task.requests.get(next_url, params=payload)
+            response.raise_for_status()
+        except RequestException:
+            raise plugin.PluginError('Error getting list from kitsu.io')
+
+        while response:
+            json_data = response.json()
+
+            for item, anime in zip(json_data['data'], json_data['included']):
+                if item['relationships']['media']['data']['id'] != anime['id']:
+                    raise ValueError
+                if config.get('finishedonly') and anime['attributes']['endDate'] == None:
+                    continue
+                if config.get('currentonly') and anime['attributes']['endDate'] is not None:
+                    continue
+
+                entry = Entry()
+                entry['title'] = anime['attributes']['canonicalTitle']
+                if anime['attributes']['titles']['en']:
+                    entry['kitsu_title_en'] = anime['attributes']['titles']['en']
+                if anime['attributes']['titles']['en_jp']:
+                    entry['kitsu_title_en_jp'] = anime['attributes']['titles']['en_jp']
+                if anime['attributes']['titles']['ja_jp']:
+                    entry['kitsu_title_ja_jp'] = anime['attributes']['titles']['ja_jp']
+                entry['kitsu_url'] = anime['links']['self']
+                entry['url'] = ''
+                if entry.isvalid():
+                    if config.get('latest'):
+                        entry['series_episode'] = item['progress']
+                        entry['series_id_type'] = 'sequence'
+                        entry['title'] += ' ' + str(entry['progress'])
+                    entries.append(entry)
+
+            next_url = json_data['links'].get('next')
+            if next_url:
+                try:
+                    response = task.requests.get(next_url)
+                    response.raise_for_status()
+                except RequestException:
+                    raise plugin.PluginError('Error getting next url from kitsu.io')
+            else:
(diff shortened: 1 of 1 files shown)


flexget/plugins/input/kitsu.py
```diff
@@ -0,0 +1,107 @@
...
+                        entry['title'] += ' ' + str(entry['progress'])
+                    entries.append(entry)
+
+            next_url = json_data['links'].get('next')
+            if next_url:
+                try:
+                    response = task.requests.get(next_url)
+                    response.raise_for_status()
+                except RequestException:
+                    raise plugin.PluginError('Error getting next url from kitsu.io')
```
gazpachoking: We should display a bit more info about the error here.

flexget/plugins/input/kitsu.py
```diff
@@ -0,0 +1,107 @@
...
+                entry = Entry()
+                entry['title'] = anime['attributes']['canonicalTitle']
+                if anime['attributes']['titles']['en']:
+                    entry['kitsu_title_en'] = anime['attributes']['titles']['en']
+                if anime['attributes']['titles']['en_jp']:
+                    entry['kitsu_title_en_jp'] = anime['attributes']['titles']['en_jp']
+                if anime['attributes']['titles']['ja_jp']:
+                    entry['kitsu_title_ja_jp'] = anime['attributes']['titles']['ja_jp']
+                entry['kitsu_url'] = anime['links']['self']
+                entry['url'] = ''
```
gazpachoking: I think it's better if the kitsu url goes in here instead of being blank.

flexget/plugins/input/kitsu.py
```diff
@@ -0,0 +1,107 @@
...
+            user_response = task.requests.get('https://kitsu.io/api/edge/users', params=user_payload)
+            user_response.raise_for_status()
+            user = user_response.json()
+            if len(user['data']) < 1:
+                raise plugin.PluginError('no such username found on kitsu.io')
+            userId = user['data'][0]['id']
+        except RequestException:
+            raise plugin.PluginError('Error getting User ID from kitsu.io')
+
+        next_url = 'https://kitsu.io/api/edge/users/{userId}/library-entries'.format(**locals())
```
gazpachoking: I'd prefer not to use `locals()` even though it's fairly explicit what's happening in this case.

flexget/plugins/input/kitsu.py
```diff
@@ -0,0 +1,107 @@
...
+    def on_task_input(self, task, config):
+        entries = []
+        try:
+            user_payload = {'filter[name]': config['username']}
+            user_response = task.requests.get('https://kitsu.io/api/edge/users', params=user_payload)
+            user_response.raise_for_status()
+            user = user_response.json()
+            if len(user['data']) < 1:
+                raise plugin.PluginError('no such username found on kitsu.io')
+            userId = user['data'][0]['id']
```
gazpachoking: We don't like to do camelCase variable names in python. `user_id` is better

flexget/plugins/input/kitsu.py
```diff
@@ -0,0 +1,107 @@
...
+      latest: <yes|no>
+      finishedonly: <yes|no>
+      currentonly: <yes|no>
+    """
+
+    schema = {
+        'type': 'object',
+        'properties': {
+            'username': {'type': 'string'},
+            'lists': {'type': 'array',
```
liiight: i'd consider using out `one_or_more` tool here

flexget/plugins/input/kitsu.py
```diff
@@ -0,0 +1,107 @@
...
+                      'items': {'type': 'string', 'enum': ['current', 'planned', 'completed', 'on_hold', 'dropped']}},
+            'latest': {'type': 'boolean', 'default': False},
+            'currentonly': {'type': 'boolean', 'default': False},
```
liiight: please add underscore for readaiblity

flexget/plugins/input/kitsu.py
```diff
@@ -0,0 +1,107 @@
...
+            'finishedonly': {'type': 'boolean', 'default': False}
```
liiight: underscore

flexget/plugins/input/kitsu.py
```diff
@@ -0,0 +1,107 @@
...
+        },
+        'required': ['username'],
+        'additionalProperties': False,
+    }
+
+    @cached('kitsu', persist='2 hours')
+    def on_task_input(self, task, config):
+        entries = []
+        try:
```
liiight: try to keep the except blocks as minimal as possible

flexget/plugins/input/kitsu.py
```diff
@@ -0,0 +1,107 @@
...
+            user_payload = {'filter[name]': config['username']}
+            user_response = task.requests.get('https://kitsu.io/api/edge/users', params=user_payload)
+            user_response.raise_for_status()
```
liiight: I'm not sure this is a good idea, as it will probably not be caught and just cause a crash.
Also, `raise_for_status()` raises for every non 200 status, some of which (like 404 for example) may be handled better by reraising a PluginError

flexget/plugins/input/kitsu.py
```diff
@@ -0,0 +1,107 @@
...
+            response.raise_for_status()
+        except RequestException:
+            raise plugin.PluginError('Error getting list from kitsu.io')
+
+        while response:
+            json_data = response.json()
+
+            for item, anime in zip(json_data['data'], json_data['included']):
+                if item['relationships']['media']['data']['id'] != anime['id']:
+                    raise ValueError
```
liiight: please add some data to this ValueError

flexget/plugins/input/kitsu.py
```diff
@@ -0,0 +1,115 @@
...
+    @cached('kitsu', persist='2 hours')
+    def on_task_input(self, task, config):
+        entries = []
+        user_payload = {'filter[name]': config['username']}
+        try:
+            user_response = task.requests.get('https://kitsu.io/api/edge/users', params=user_payload)
+            user_response.raise_for_status()
+        except RequestException as e:
+            error_message = 'Error finding User url: {url} status: {status}'.format(
+                url=e.request.url, status=e.response.status_code)
```
liiight: This is a but note that e doesn't necessarily mean that response attribute. Best to check with `hasattr`

flexget/plugins/input/kitsu.py
```diff
@@ -0,0 +1,115 @@
...
+        payload = {'filter[status]': ','.join(config['lists']), 'filter[media_type]': 'Anime', 'include': 'media',
+                   'page[limit]': 20}
+        try:
+            response = task.requests.get(next_url, params=payload)
+        except RequestException as e:
+            error_message = 'Error getting list from {url}'.format(url=e.request.url)
+            if hasattr(e, 'response'):
+                error_message += ' status: {status}'.format(status=e.response.status_code)
+            log.debug(error_message, exc_info=True)
+            log.info(error_message, exc_info=True)
```
liiight: There's no reason to log exactly the same thing using different log levels

NikkyAI: no idea why i did that.. o.o

flexget/plugins/input/kitsu.py
```diff
@@ -0,0 +1,114 @@
+from __future__ import unicode_literals, division, absolute_import
+import logging
+
+from flexget import plugin
+from flexget.config_schema import one_or_more
+from flexget.entry import Entry
+from flexget.event import event
+from flexget.utils.cached_input import cached
+from requests import RequestException
```
cvium: Should fix your imports.

0. __future__ related stuff
1. builtin Python modules
2. third party libs
3. Flexget

flexget/plugins/input/kitsu.py
```diff
@@ -0,0 +1,114 @@
...
+                try:
+                    response = task.requests.get(next_url)
+                except RequestException as e:
+                    error_message = 'Error getting list from next page url: {url}'.format(url=e.request.url)
+                    if hasattr(e, 'response'):
+                        error_message += ' status: {status}'.format(status=e.response.status_code)
+                    log.debug(error_message, exc_info=True)
+                    raise plugin.PluginError(error_message)
+            else:
+                response = None
```
cvium: What exactly is the purpose of this?

flexget/plugins/input/kitsu.py
```diff
@@ -0,0 +1,114 @@
...
+                    if limit_to == 'finished' and anime['attributes']['endDate'] is None:
+                        continue
+
+                entry = Entry()
+                entry['title'] = anime['attributes']['canonicalTitle']
+                if anime['attributes']['titles']['en']:
+                    entry['kitsu_title_en'] = anime['attributes']['titles']['en']
+                if anime['attributes']['titles']['en_jp']:
+                    entry['kitsu_title_en_jp'] = anime['attributes']['titles']['en_jp']
+                if anime['attributes']['titles']['ja_jp']:
```
cvium: Do these json fields always exist? It will crash if they don't.

flexget/plugins/input/kitsu.py
```diff
@@ -0,0 +1,114 @@
...
+      latest: <yes|no>
+    """
+
+    schema = {
+        'type': 'object',
+        'properties': {
+            'username': {'type': 'string'},
+            'lists': one_or_more({'type': 'string', 'enum': ['current', 'planned', 'completed', 'on_hold', 'dropped']}),
+            'latest': {'type': 'boolean', 'default': False},
+            'limit_to': {'type': 'string', 'enum': ['airing', 'finished']}
```
cvium: Would it not make more sense to call it something like `status` instead of `limit_to`?

flexget/plugins/input/kitsu.py
```diff
@@ -0,0 +1,114 @@
...
+        try:
+            user_response = task.requests.get('https://kitsu.io/api/edge/users', params=user_payload)
+        except RequestException as e:
+            error_message = 'Error finding User url: {url}'.format(url=e.request.url)
+            if hasattr(e, 'response'):
+                error_message += ' status: {status}'.format(status=e.response.status_code)
+            log.debug(error_message, exc_info=True)
+            raise plugin.PluginError(error_message)
+        user = user_response.json()
+        if len(user['data']) < 1:
```
cvium: More Pythonic to write `if not len(user['data'])`

flexget/plugins/input/kitsu.py
NikkyAI: you are correct, that was a stupid way to solve it

flexget/plugins/input/kitsu.py
NikkyAI: they always exist but are eventally null or empty strings

NikkyAI: if those will not be used and are uncertain then i can also just remove thise.. the canonicalName is always going to be set as far as i can tell

cvium: It's fine to keep it, but you could/should check for existence.

NikkyAI:
@gazpachoking reminder ping to look over this (i think your change request stop it from being merged)
i will create a second PR once the stable api is rolled out on namespace /v17/

gazpachoking:
Sorry about that @NikkyAI. Looks pretty good to me. Want to update the wiki?

NikkyAI:
what kind of info needs to be in the wiki entry ?

gazpachoking:
Add a link here https://flexget.com/Plugins/#3rd-party-sites-input to a new page with some info on what the plugin does and config options. You can take a peek at some of the other plugin pages for a bit of guidance.

NikkyAI:
only info that might need to be added apart from the config is that limit_to: araing|finished relies on the fact that there is a finished date on some entries.. this breaks on movies and some shows.. so people might want to just leave out that config option and not limit anything
latest will only add the latest episode to be downloaded.. 
that is both functionality that was easier achieved with the old api and i tried to emulate it as good as possible
i still hope they add that back in
if its not in when the stable namespace gets deployed i will remove those options though

NikkyAI:
i added a page for my plugin, based on anidb_list and others as well as a entry in the 3rd party input list
i grouped it with the other anime related input plugins

gazpachoking:
Awesome, thanks for the contribution!

Issue #1591, opened 2016-12-29
Bakabt URL change


xrsxj:
Bakabt url has change from http://www.bakabt.com to https://bakabt.me, the url rewriter plugin needs to be updated to match

PR #1598, opened 2016-12-31
[fix] bakaBT and nyaa TLDs changed

### LINKED ISSUE #1594: nyaa changed TLD
hi peeps. it seems they switched TLD from .eu to .se

i changed my local flexget/plugins/sites/nyaa.py, removed the pyc & reloaded the daemon. its pulling stuff. but i aint got the skills to send a pull request, so i thought i'd do the next best thing and say something

if you don't want to do anything, i guess thats fine too. the old is redirecting to the new

zosky:
### Motivation for changes:
nothing to exciting here. i learned how to do pull requests so i thought i'd deal with an issue i created yesterday & one very similar 

### Detailed changes:
- bakabt.com > bakabt.me
- nyaa.eu > nyaa.se

### Addressed issues:
- Fixes #1594 #1591

### DIFF
--- a/flexget/plugins/sites/bakabt.py
+++ b/flexget/plugins/sites/bakabt.py
@@ -17,9 +17,9 @@ class UrlRewriteBakaBT(object):
     # urlrewriter API
     def url_rewritable(self, task, entry):
         url = entry['url']
-        if url.startswith('http://www.bakabt.com/download/'):
+        if url.startswith('http://www.bakabt.me/download/'):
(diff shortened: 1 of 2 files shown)


cvium:
I feel this should be separate PRs as the sites are not really related.

zosky:
both sites are for anime, that was my logic for lumping them together
but if you want 2 PRs, i guess that makes sense too

PR #1603, opened 2017-01-03
Fix torrentleech URLs after changes on the site

yawor:
### Motivation for changes:

TL changed login page/URL and now requires HTTPS for all links. Also the "www." subdomain is now required (requesting addresses without www causes redirects). This change caused the plugin not to be able to find any entries.

### Detailed changes:

- Corrected login page URL and login form data.
- Corrected all addresses to start with ```https://www.```.


liiight:
Thanks!

PR #1610, opened 2017-01-05
add missing movie_metainfo, series_metainfo property's

Andy2244:
### Motivation for changes:
Some plugins did not support the movie_metainfo, series_metainfo group tag/property.

### Detailed changes:
added movie_metainfo to rottentomatoes_lookup
added series_metainfo to thetvdb_lookup, tvmaze_lookup

### DIFF
--- a/flexget/plugins/metainfo/rottentomatoes_lookup.py
+++ b/flexget/plugins/metainfo/rottentomatoes_lookup.py
@@ -116,7 +116,12 @@ class PluginRottenTomatoesLookup(object):
         for entry in task.entries:
             entry.register_lazy_func(self.lazy_loader, self.field_map)
 
+    @property
+    def movie_identifier(self):
+        """Returns the plugin main identifier type"""
+        return 'rt_id'
+
 
 @event('plugin.register')
 def register_plugin():
-    plugin.register(PluginRottenTomatoesLookup, 'rottentomatoes_lookup', api_ver=2)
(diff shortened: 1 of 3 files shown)


Andy2244:
Regarding this logic, i noticed a small logic problem. If a metainfo plugin supports multiple id's and no actual plugin for those id's exist than its impossible to retrieve this field name.

Example:
`series_identifier_fields = [p.instance.series_identifier for p in plugin.get_plugins(group='series_metainfo')]`
This will never get the `tvrage_id`, so this field still needs to-be hard-coded.

So maybe there should be a extra property or additional list return of all supported id fields. So that we can always get a complete list of all supported id fields for each category per plugin.

Maybe like this?
```
    @property
    def series_identifier(self):
        """Returns the plugin main identifier type"""
        return ('tvmaze_id', 'tvrage_id', 'tvdb_id')
or
        return ('tvmaze_id', ['tvrage_id', 'tvdb_id'])
or
        return ['tvmaze_id', 'tvrage_id', 'tvdb_id'])  # index 0 is main id?
```
or
```
    @property
    def series_identifier_supported(self):
        """Returns all supported series identifier fields"""
        return ['tvmaze_id', 'tvrage_id', 'tvdb_id']
```

Andy2244:
tvmaze has support for: tvmaze_id, tvdb_id, tvrage_id
tvdb has tvdb_id and fills (imdb_id, zap2it_id)
tmdb has tmdb_id, imdb (the actual api also supports tvdb, tvrage for series)

In my case i wanted to get all possible series id fields that can be used to identify a series and as noted i still have to hardcode `tvrage_id`.

PR #1611, opened 2017-01-05
Make sure series_name_plain is really plain

mfonville:
Needs to remove also some special character processing to produce filesystem-safe strings

### Motivation for changes:
Current series_name_plain produced outputs that are unprocessable on e.g. Windows filesystems or Samba shares; for passing a better value to `download-npo` a further processed plain string is more practical

### Detailed changes:

- Rewrote the `_strip_accents` function into a `_convert_plain` function

### DIFF
--- a/flexget/plugins/input/npo_watchlist.py
+++ b/flexget/plugins/input/npo_watchlist.py
@@ -61,9 +61,9 @@ class NPOWatchlist(object):
 
     csrf_token = None
 
-    def _strip_accents(self, s):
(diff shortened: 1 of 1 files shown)


cvium:
Looks like you forgot to update it in a few places.

mfonville:
Ah, whoops, I thought I had them all. I will update the commit/PR

mfonville:
@cvium it should now be ok :)

PR #1616, opened 2017-01-08
NPO parameter has changed for only showing full episodes

mfonville:
### Motivation for changes:
NPO API has changed

### Detailed changes:

- What used to be e.g. http://www.npo.nl/als-de-dijken-breken/VPWON_1261083/search?category=broadcasts is now http://www.npo.nl/als-de-dijken-breken/VPWON_1261083/search?media_type=broadcast

### Addressed issues:

- Without this fix, also other video objects, like small snippets of episodes are fetched

### DIFF
--- a/flexget/plugins/input/npo_watchlist.py
+++ b/flexget/plugins/input/npo_watchlist.py
@@ -181,7 +181,7 @@ class NPOWatchlist(object):
 
     def _get_series_episodes(self, task, config, series_name, series_url, series_info):
         log.info('Retrieving new episodes for %s', series_name)
-        response = task.requests.get(series_url + '/search?category=broadcasts')
+        response = task.requests.get(series_url + '/search?media_type=broadcast')  # only shows full episodes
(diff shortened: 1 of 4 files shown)


liiight:
You forgot to redo cassettes

mfonville:
I lack at the moment (access) to a system to re-run the cassettes on :-/
If somebody else could run them, I'd really appreciate it. Otherwise it will have to wait awhile I fear

mfonville:
Another I noticed now, is that the testcase also needs an update.
Because it assumed the Trailer was the first season of the series, while in reality it wasn't (I just hadn't discovered yet that the parameter had changed; I only found that out just now, while toying a bit with the HTML and URLs)

mfonville:
I now at least updated the test-case; so if somebody could just re-run the test an update the cassette from that run, it should be good to go

mfonville:
Ok, I got hold of a machine for a moment where I could run the test and update the cassettes

PR #1618, opened 2017-01-09
Support searching for titles which contain quotes

v17al:
Single quotes weren't matching anything; just replace with whitespace

### DIFF
--- a/flexget/plugins/sites/piratebay.py
+++ b/flexget/plugins/sites/piratebay.py
@@ -126,8 +126,10 @@ class UrlRewritePirateBay(object):
         entries = set()
         for search_string in entry.get('search_strings', [entry['title']]):
             query = normalize_unicode(search_string)
-            # TPB search doesn't like dashes
-            query = query.replace('-', ' ')
+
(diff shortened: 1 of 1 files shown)


cvium:
Why whitespace?

v17al:
Whitespace will match the results which have a quote in them, whereas stripping the quote matches results which don't have a quote:

https://thepiratebay.org/search/Miss%20Peregrines%20Home%20for%20Peculiar%20Children/0/99/0
https://thepiratebay.org/search/Miss%20Peregrine%20s%20Home%20for%20Peculiar%20Children/0/99/0

mfonville:
I checked and can confirm, i.e. the movie *O'horten*:
No results for exact name: https://thepiratebay.org/search/o'horten/0/99/0
No results if you drop the *`*: https://thepiratebay.org/search/ohorten/0/99/0
But there are results if you replace it with a space (or some other kind of characters like *`* or with a dot *.* or a hypen *-*): https://thepiratebay.org/search/o%20horten/0/99/0

PR #1620, opened 2017-01-11
relplace Safe by zxcvbn for password strength testing

arcresu:
### Motivation for changes:
Currently in the webserver password strength is checked using the Python library [Safe](https://github.com/lepture/safe/). This library has not seen very much attention from its author in a while and it is quite simplistic in its approach.

Instead I propose to use the Python implementation of Dropbox's zxcvbn library [dwolfhub/zxcvbn-python](https://github.com/dwolfhub/zxcvbn-python). This is a more sophisticated password strength checker which is actively maintained. It is pure Python and does not introduce extra dependencies. Note that there are older forks of the code but this is the current officially endorsed one.

### Detailed changes:
Replace the dependency on Safe with one on zxcvbn-python.

It can provide a more detailed password analysis but at the most basic level it returns a score from the following list:

```
  0 # too guessable: risky password. (guesses < 10^3)
  1 # very guessable: protection from throttled online attacks. (guesses < 10^6)
  2 # somewhat guessable: protection from unthrottled online attacks. (guesses < 10^8)
  3 # safely unguessable: moderate protection from offline slow-hash scenario. (guesses < 10^10)
  4 # very unguessable: strong protection from offline slow-hash scenario. (guesses >= 10^10)
```

I require that the password have a score of at least 3 in order to be considered secure.

### DIFF
--- a/flexget/webserver.py
+++ b/flexget/webserver.py
@@ -8,7 +8,7 @@ import socket
 import threading
 
 import cherrypy
-import safe
+import zxcvbn
 from flask import Flask, abort, redirect
 from flask_login import UserMixin
 from sqlalchemy import Column, Integer, Unicode
@@ -218,8 +218,8 @@ def get_user(username='flexget', session=None):
 
 @with_session
 def change_password(username='flexget', password='', session=None):
-    check = safe.check(password)
-    if check.strength not in ['medium', 'strong']:
+    check = zxcvbn.zxcvbn(password, user_inputs=[username])
+    if check['score'] < 3:
         raise WeakPassword('Password {0} is not strong enough'.format(password))
 
     user = get_user(username=username, session=session)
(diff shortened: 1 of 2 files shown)


liiight:
From what I can see, safe has no 3rd party requirements too. Other than that, I'm not sure this change has merit just for change sake. Safe is maybe simple but it does its job.

arcresu:
It's not change for the sake of it - it's about reducing the dependency on unmaintained code. It seems to be an easy change and zxcvbn is better than safe in a couple of respects.

liiight:
ok, did you test this with the CLI command? `flexget web passwd`?

arcresu:
I tested with the CLI command, yes. It works as intended and seems to have a reasonable threshold for password complexity.

liiight:
Thanks!

Issue #748, opened 2015-12-26
Revisit entire qualities specifications


blastcodem:
Could be cool to have 4k support :)

liiight:
any chance you do some research and find exactly all of the syntax for this resolution in the real world? so I can add it to the regex?

blastcodem:
As far as I know there is:

2160p
4k blu-ray
4k
UHD
UHDTV
3840x2160
4K UHDTV

liiight:
those are refresh rates, right? I don't believe we even have that attribute as part of the quality today

blastcodem:
Yes 4k will have both 30fps and 60fps

Kodi have already a great deal on decoding filenames.
There is a good subject about 3D and regex on kodi doc: here
http://kodi.wiki/view/3D

and here for videos
http://kodi.wiki/view/Media_flags#Video_flags

For general video tags here is some, taken from kodi source code.
https://github.com/xbmc/xbmc/blob/3aa56113345cf9c03a87924189c6c45a64c060f7/xbmc/settings/AdvancedSettings.cpp

ac3|dts|custom|dc|remastered|divx|divx5|dsr|dsrip|dutch|dvd|dvd5|dvd9|dvdrip|dvdscr|dvdscreener|screener|dvdivx|cam|fragment|fs|hdtv|hdrip|hdtvrip|internal|limited|multisubs|ntsc|ogg|ogm|pal|pdtv|proper|repack|rerip|retail|r3|r5|bd5|se|svcd|swedish|german|read.nfo|nfofix|unrated|extended|ws|telesync|ts|telecine|tc|brrip|bdrip|480p|480i|576p|576i|720p|720i|1080p|1080i|3d|hrhd|hrhdtv|hddvd|bluray|x264|h264|xvid|xvidvd|
  m_stereoscopicregex_3d = "[-. _]3d[-. _]";
  m_stereoscopicregex_sbs = "[-. _]h?sbs[-. _]";
  m_stereoscopicregex_tab = "[-. _]h?tab[-. _]";

  
Missing future h265 codec:
h265
H.265
HEVC

Else than 4k there will be HDR also.. No idea how it will be named, but I guess it will be something like this:
HDR
10bits

They are talking about 8k already, but wouldn't have to bother about that yet (hopefully)

For VR I have no idea... Not into that yet.

liiight:
This is great stuff, thanks! 
Btw, we already support 10bits

liiight:
[...]

PR #1625, opened 2017-01-12
Support for 2160p

cvium:
### Motivation for changes:
Bored

### Related
 - #748

### Detailed changes:

- Added 2160p support. 3840x2160 or 2160p or 4K are all valid. KNOWN LIMITATION: guessit cannot parse 'Some title 4K' as 2160p, but there's no reason the internal parser shouldn't be allowed to (I think).


Mirgolth:
Best motivation for change award ;)

PR #1628, opened 2017-01-14
Expansion of quality resolution matching

mfonville:
- Added 8K UHDTV
- Added extra matching options for other HD name formats that are defined in their respective standards, and sometimes used in naming schemes

### Motivation for changes:
Saw this respective commit, and expressed comment here: https://github.com/Flexget/Flexget/commit/bdc19ba3d59835ef924a39aacfee681fe06aa22b#commitcomment-20457492


liiight:
Please update the tests with the new qualities and new regexp matches.

cvium:
`SyntaxError('invalid syntax', ('/home/travis/build/Flexget/Flexget/flexget/utils/qualities.py', 131, 20, "    QualityComponent('resolution', 100, '4320p', '((7680x)?4320(p|uhd(tv)?)?x?(50)?)|8k')  # technically also named an ultra hd format\n"))`

I also don't fully agree with these changes as I don't really feel there's a need for them.

mfonville:
well, I made the PR mostly to at least discuss improving it.
Not *all* changes would have to be added, but e.g. if you type in some well known search engines `fullhd` or `uhd` you get plenty of results that would not have matched against the current filters.

So I do think that at least expanding the match-set with some of these keyword would be an improvement.

It is up to you to decide which ones you would want. And I will update the PR then accordingly, including the test.

cvium:
When searching for fullhd all I find is porn. 

I'm not too keen on adding more regex; it's a dark hole you may find yourself in when it goes wrong. If you can show with vigorous testing that it has no adverse effects, then I won't object much, but there should also be something to gain. Scene rules are pretty strict and it's pretty much those rules we try to adhere to and I can't remember a scene release containing the word `fullhd`.

mfonville:
@cvium ok, I am not very acquinted with the *scene rules* per se, just with the phrasing used in specs and applications, that use these terms.

I will propose than a very limited regex update, instead of a big one.

flexget/utils/qualities.py
```diff
@@ -126,8 +126,9 @@ def __deepcopy__(self, memo=None):
     QualityComponent('resolution', 50, '720i'),
     QualityComponent('resolution', 60, '720p', '(1280x)?720(p|hd)?x?(50)?'),
     QualityComponent('resolution', 70, '1080i'),
-    QualityComponent('resolution', 80, '1080p', '(1920x)?1080p?x?(50)?'),
-    QualityComponent('resolution', 90, '2160p', '((3840x)?2160p?x?(50)?)|4k')
+    QualityComponent('resolution', 80, '1080p', '(1920x)?1080p?x?(50)?|f(ull ?)?hd'),
+    QualityComponent('resolution', 90, '2160p', '((3840x)?2160p?x?(50)?)|4k|u(ltra ?)?hd')
```
cvium: You're missing a comma.

flexget/tests/test_qualities.py
```diff
@@ -30,7 +30,14 @@ def parser(self, request):
 
     @pytest.mark.parametrize("test_quality", [
         ('Test.File 1080p.web', '1080p webdl'),
+        ('Test.File.fullhd.web', '1080p webdl'),
+        ('Test.File.fhd.web', '1080p webdl'),
         ('Test.File.2160p.web', '2160p webdl'),
+        ('Test.File.ultrahd.web', '2160p webdl'),
+        ('Test.File.uhd.web', '2160p webdl'),
+        ('Test.File.4k.web', '2160p webdl'),
```
cvium: This test will fail for guessit parser

flexget/tests/test_qualities.py
```diff
@@ -64,7 +71,10 @@ def parser(self, request):
         ('Test.File.720P', '720p'),
         ('Test.File.1920x1080', '1080p'),
         ('Test.File.3840x2160', '2160p'),
+        ('Test.File.7680x4320', '2160p'),
```
cvium: This is wrong

flexget/tests/test_qualities.py
```diff
@@ -30,7 +30,14 @@ def parser(self, request):
...
     @pytest.mark.parametrize("test_quality", [
         ('Test.File 1080p.web', '1080p webdl'),
+        ('Test.File.fullhd.web', '1080p webdl'),
+        ('Test.File.fhd.web', '1080p webdl'),
         ('Test.File.2160p.web', '2160p webdl'),
+        ('Test.File.ultrahd.web', '2160p webdl'),
+        ('Test.File.uhd.web', '2160p webdl'),
+        ('Test.File.4k.web', '2160p webdl'),
+        ('Test.File.4320p.web', '4320p webdl'),
+        ('Test.File.8k.web', '4320p webdl'),
```
cvium: This too will fail

cvium:
Guessit has no support for 4320p, 8k, f(ull)hd nor u(ltra)hd. I think it's better to postpone this until there's actual 8k content available.

mfonville:
Yes, I see guessit lacks the support, what a pity.

Let's keep the PR indeed open until the guessit library will add 8k support;
Some 8K broadcasts are already done in Japan, last year the first feature film was shot in 8K, and at CES2017 there are now many 8K monitors announced. So slowly the first (non-consumer) material will slowly start to appear.

Another thing I noticed: in the regex there is a match for 50fps but not for 60fps, is that on purpose? Because in guessit the 60fps is included.

cvium:
I'm not sure about the fps. Don't think I've ever seen 1080p50. 50 fps usually means it's interlaced.

Anyway, that's a discussion better fit for the forums or Gitter. Closing this.

mfonville:
Just giving a heads-up that GuessIt added UHD detecting filters;
i.e. see https://github.com/guessit-io/guessit/commit/ac054b4bbfd534f2f84952b1abf4daea56415580 
https://github.com/guessit-io/guessit/commit/5c2a79719d9cee94d9140763777efcc8913f7031

PR #1630, opened 2017-01-15
Add url rewrite support for bt.hliang.com

jonathanou:
bt.hliang.com provides asian tv show torrents.  However, it does not support rss feeds.  To auto download shows, use http plugin to download the homepage of bt.hliang.com, and this url rewrite plugin will do the following:

1.  Go through each entry url and follow the link to open the details page.
2.  In the details page, search for the actual torrent download url (any link that contains "down.php?[unique id]"
3. Return the url if found, or raise UrlRewriteError.

### DIFF
--- a/flexget/plugins/sites/hliang.py
+++ b/flexget/plugins/sites/hliang.py
@@ -33,7 +33,7 @@ class UrlRewriteHliang(object):
             soup = get_soup(page.text)
         except Exception as e:
             raise UrlRewritingError(e)
-        down_link = soup.find('a', attrs={'href': re.compile("down.php?.*")})
+        down_link = soup.find('a', attrs={'href': re.compile("down\.php\?.*")})
         if not down_link:
             raise UrlRewritingError('Unable to locate download link from url "%s"' % url)
         return 'http://bt.hliang.com/' + down_link.get('href')
--- a/flexget/plugins/sites/hliang.py
+++ b/flexget/plugins/sites/hliang.py
@@ -30,7 +30,7 @@ class UrlRewriteHliang(object):
         try:
             page = requests.get(url, headers=txheaders)
         except requests.exceptions.RequestException as e:
-            log.exception('Cannot open "%s"' % url, e)
+            log.error('Cannot open "%s" : %s', url, e)
             raise UrlRewritingError(e)
 
         try:
--- a/flexget/plugins/sites/hliang.py
+++ b/flexget/plugins/sites/hliang.py
@@ -30,13 +30,14 @@ class UrlRewriteHliang(object):
         try:
             page = requests.get(url, headers=txheaders)
         except requests.exceptions.RequestException as e:
-            log.error('Cannot open "%s" : %s', url, e)
-            raise UrlRewritingError(e)
+            msg = 'Cannot open "%s" : %s'% (url, e.message)
+            log.error(msg)
+            raise UrlRewritingError(msg)
 
         try:
             soup = get_soup(page.text)
         except Exception as e:
-            raise UrlRewritingError(e)
+            raise UrlRewritingError(e.message)
         
         down_link = soup.find('a', attrs={'href': re.compile("down\.php\?.*")})
         if not down_link:
--- a/flexget/plugins/sites/hliang.py
+++ b/flexget/plugins/sites/hliang.py
@@ -30,14 +30,14 @@ class UrlRewriteHliang(object):
         try:
             page = requests.get(url, headers=txheaders)
         except requests.exceptions.RequestException as e:
-            msg = 'Cannot open "%s" : %s'% (url, e.message)
+            msg = 'Cannot open "%s" : %s'% (url, str(e))
             log.error(msg)
             raise UrlRewritingError(msg)
 
         try:
             soup = get_soup(page.text)
         except Exception as e:
-            raise UrlRewritingError(e.message)
+            raise UrlRewritingError(str(e))
         
         down_link = soup.find('a', attrs={'href': re.compile("down\.php\?.*")})
         if not down_link:
--- a/flexget/plugins/sites/hliang.py
+++ b/flexget/plugins/sites/hliang.py
@@ -29,9 +29,15 @@ class UrlRewriteHliang(object):
         txheaders = {'User-agent': 'Mozilla/4.0 (compatible; MSIE 5.5; Windows NT)'}
         try:
             page = requests.get(url, headers=txheaders)
+        except requests.exceptions.RequestException as e:
+            log.exception('Cannot open "%s"' % url, e)
+            raise UrlRewritingError(e)
+
+        try:
             soup = get_soup(page.text)
         except Exception as e:
             raise UrlRewritingError(e)
+        
         down_link = soup.find('a', attrs={'href': re.compile("down\.php\?.*")})
         if not down_link:
             raise UrlRewritingError('Unable to locate download link from url "%s"' % url)
--- a/flexget/plugins/sites/hliang.py
+++ b/flexget/plugins/sites/hliang.py
@@ -1,0 +1,44 @@
+from __future__ import unicode_literals, division, absolute_import
+import urllib2
+import logging
+from flexget import plugin
+from flexget.event import event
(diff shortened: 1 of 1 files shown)


flexget/plugins/sites/hliang.py
```diff
@@ -0,0 +1,44 @@
...
+        return False
+
+    # urlrewriter API
+    def url_rewrite(self, task, entry):
+        entry['url'] = self.parse_download_page(entry['url'], task.requests)
+
+    @plugin.internet(log)
+    def parse_download_page(self, url, requests):
+        txheaders = {'User-agent': 'Mozilla/4.0 (compatible; MSIE 5.5; Windows NT)'}
+        page = requests.get(url, headers=txheaders)
```
liiight: This needs to be in a try except block

flexget/plugins/sites/hliang.py
```diff
@@ -0,0 +1,44 @@
...
+        except Exception as e:
+            raise UrlRewritingError(e)
+        down_link = soup.find('a', attrs={'href': re.compile("down\.php\?.*")})
+        if not down_link:
+            raise UrlRewritingError('Unable to locate download link from url "%s"' % url)
+        return 'http://bt.hliang.com/' + down_link.get('href')
+
+@event('plugin.register')
+def register_plugin():
+    plugin.register(UrlRewriteHliang, 'hliang', groups=['urlrewriter'], api_ver=2)
```
liiight: `groups` was changed `interfaces`

flexget/plugins/sites/hliang.py
```diff
@@ -0,0 +1,43 @@
...
+
+    # urlrewriter API
+    def url_rewrite(self, task, entry):
+        entry['url'] = self.parse_download_page(entry['url'], task.requests)
+
+    @plugin.internet(log)
+    def parse_download_page(self, url, requests):
+        txheaders = {'User-agent': 'Mozilla/4.0 (compatible; MSIE 5.5; Windows NT)'}
+        try:
+            page = requests.get(url, headers=txheaders)
```
liiight: Please split this to two try except blocks and try avoiding catching Exception if it can be done

jonathanou: Can you explain why we need 2 try-except blocks instead of one?  Also, why is catching Exception not desirable?  I'm new to the flexget code, is there some top level catch-all that will handle uncaught exceptions?  I think I've seen logs where flexget crashed because Exceptions weren't caught.

liiight: Yeah, there is a catch all but this is irrelevant to my point. 
You have two separate operations, request and soup, and each one can fail independently. Having one big block with all poss exceptions is just poor code which is hard to debug and maintain

jonathanou: I apologize for the questions, I'm not super experienced in python programming.  I appreciate any insight you can give me.

So just to clarify a bit:

1.  Do you have an example somewhere that demonstrates what you're looking for here?  I based my code from Flexget/flexget/plugins/sites/deadfrog.py, which doesn't even put the request call in the try block (a bug, I guess).  I took a brief look at other py files in the sites folder and they are all very similar.

2.  In my experience with C++/C# programming, when a client calls an interface, it needs to know what exception types to expect if the operation fails.  UrlRewritingError seems to be this agreed exception type.  

Therefore my thought was it is not appropriate to leak exceptions solely due to implementation details out to the caller.  I could one day change my implementation and throw a completely different set of exceptions.  Catching all exception types, wrapping it in UrlRewritingError, and rethrowing allows the caller to only ever expect one type of exception.  It seems like the right thing to do.

3.  I don't quite understand the logic behind using different try blocks for different function calls.  Yes each one can fail independently, but if all I'm going to do is wrap the exception and throw UrlRewritingError, why is it better for debugging and maintainability for separate try-blocks?  We can set breakpoints to figure out which line threw exceptions.  

And is having separate try blocks really that much maintainable than a single try with all your implementation in it?  I'd almost argue it seems like the code is more readable with a single try-block.

liiight: 1. Just that a plugin code is what it is, doesn't mean it's OK. We try to constantly improve the code quality of stuff, especially with new PRs where it's easier (since the plugin dev is still around).
2. I didn't mean that you shouldn't raise a UrlRewritingError, I meant that if possible, avoiding catching all Exception, since that can swallow up stuff you maybe would like to treat differently. Again, this is more of a global guide line than a rule, and it certainly can be that this fits here.
3. By having as minimal code as possible in each `try`, you make sure the outcome is expected. This is especially true since you are catching `Exception` in the second part, which could be more relevant to the soup part than request part. So if ANY request error that isn't expected we will still raise that error. 

I hope this comes off in the manner which I mean it to, which is just trying to improve the global health of the plugin and not deter you from this in any way.

jonathanou: Sure, I certainly can understand wanting to maintain a high level of code quality.  I just wanted to make sure I understood everything.

I've updated the PR after reading what Exception type is thrown by the requests library.  For the soup library, however, it did not seem to guarantee a specific type of Exception.  So I've left it still catching Exception as before.

flexget/plugins/sites/hliang.py
```diff
@@ -29,9 +29,15 @@ def parse_download_page(self, url, requests):
         txheaders = {'User-agent': 'Mozilla/4.0 (compatible; MSIE 5.5; Windows NT)'}
         try:
             page = requests.get(url, headers=txheaders)
+        except requests.exceptions.RequestException as e:
+            log.exception('Cannot open "%s"' % url, e)
```
liiight: this should be `log.error`

jonathanou: I'll change it, but can you explain why we should prefer log.error over log.exception?  I read the [documentation](https://docs.python.org/2/library/logging.html#logging.Logger.exception) that said log.exception is the same as log.error and should be called from the exception handler.

From [stackoverflow](https://stackoverflow.com/questions/5191830/best-way-to-log-a-python-exception#5191885) it seems that log.exception will automatically show the traceback and log.error won't, is this something we want in the log?

flexget/plugins/sites/hliang.py
```diff
@@ -0,0 +1,49 @@
...
+        try:
+            page = requests.get(url, headers=txheaders)
+        except requests.exceptions.RequestException as e:
+            log.error('Cannot open "%s" : %s', url, e)
+            raise UrlRewritingError(e)
+
+        try:
+            soup = get_soup(page.text)
+        except Exception as e:
+            raise UrlRewritingError(e)
```
liiight: Last thing, you need to use `str(e)` or `e.args[0]` since you need to pass a str

flexget/plugins/sites/hliang.py
```diff
@@ -0,0 +1,49 @@
...
+        entry['url'] = self.parse_download_page(entry['url'], task.requests)
+
+    @plugin.internet(log)
+    def parse_download_page(self, url, requests):
+        txheaders = {'User-agent': 'Mozilla/4.0 (compatible; MSIE 5.5; Windows NT)'}
+        try:
+            page = requests.get(url, headers=txheaders)
+        except requests.exceptions.RequestException as e:
+            log.error('Cannot open "%s" : %s', url, e)
+            raise UrlRewritingError(e)
```
liiight: Same

flexget/plugins/sites/hliang.py
jonathanou: I used e.message instead, I think that gives better information.

cvium: Exception.message has been deprecated. Don't use it.

jonathanou: So what should I use? `str(e)` like @liiight mentioned?

jonathanou: Ah I found the [answer](https://stackoverflow.com/questions/1272138/baseexception-message-deprecated-in-python-2-6), I'll use `str(e)`, thanks.

jonathanou:
@liiight , it's been a couple of days now, is there anything else you'd like me to change?  Or is this ready to be merged?

liiight:
Thanks for this!

PR #1632, opened 2017-01-16
Python 3.6 compatibility

liiight:
### Motivation for changes:
Add support for python 3.6
### Detailed changes:

- ~Added to `.travis.yml`~
- Added to CircleCI
- Added explicit commit after adding a new column due to change in [python API](https://docs.python.org/3/whatsnew/3.6.html#changes-in-the-python-api) in 3.6: `sqlite3 no longer implicitly commits an open transaction before DDL statements.`

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -122,6 +122,41 @@ jobs:
             . venv/bin/activate
             python-codacy-coverage -r coverage.xml
 
+  test-py36:
+    docker:
+      - image: flexget/cci-python:3.6
+    environment:
+      - VCR_RECORD_MODE=none
+    steps:
+      - checkout
+      - restore_cache:
+          keys:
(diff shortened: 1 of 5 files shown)


flexget/plugins/generic/db_vacuum.py
```diff
@@ -12,12 +15,18 @@
...
     if not last_vacuum or last_vacuum < datetime.now() - VACUUM_INTERVAL:
         log.info('Running VACUUM on database to improve performance and decrease db size.')
-        session.execute('VACUUM')
-        persistence['last_vacuum'] = datetime.now()
+        with Session() as session:
+            try:
+                session.execute('VACUUM')
+            except OperationalError as e:
+                # Does not work on python 3.6, github issue #1596
+                log.error('Could not execute VACUUM command: %s', e.args[0])
```
cvium: Why e.args?

liiight: force of habit, ill just pass to logger then

PR #1638, opened 2017-01-16
Remove notify phase

### LINKED ISSUE #1634: After the new notification system: "Task doesn't have any output plugins, you should add (at least) one!"
### Expected behaviour:
Task should run without any warning...

### Actual behaviour:
Notification is working. However on every run I have these:
```
WARNING  task          TaskName1     Task doesn't have any output plugins, you should add (at least) 
one!
WARNING  task          TaskNameN     Task doesn't have any output plugins, you should add (at least) 
one!
```
### Steps to reproduce:
- Step 1: Have the config above.
- Step 2: Start daemon...

#### Config:
```yaml
schedules:
  - tasks: '*'
    interval:
      minutes: 16

variables: secrets.yml

templates:
  dwn:
    notify:
      entries:
        title: '{{title}}'
        message: '  ^ {{ rss_pubdate|formatdate("%H:%M:%S %a, %d.") }} {{task}}'
        via:
          - pushbullet:
              api_key: '{? var.pushbullet.apikey ?}'
              url: '{{url}}'

  site_rss:
    rss:
      url: https://example.com
    headers:
      Cookie: "nick={? var.site.nick ?}; pass={? var.site.pass ?}"

tasks:
  site_auto:
    template:
      - dwn
      - site_rss
    regexp:
      reject_excluding:
        - Sls \(LNG
      reject:
        - xvid
    series:
      settings:
        sls:
          upgrade: yes
          timeframe: 3 days
          target: h264 1080p
          quality: h264 360p+

      sls:
        - Sls1
```
  
#### Log:
see above

### Additional information:

- Flexget Version: From 2.9.0...
- Python Version: 2.7.9
- Installation method: pip
- OS and version: Raspbian 8

gazpachoking:
### Motivation for changes:
Notify phase was added to enable plugins that don't end up propagating errors to task. We since combined all notify plugins into one master 'notify' plugin, which means there is already a common place to catch the errors without a special phase.

### Detailed changes:
- Remove notify phase. 'notify' plugin runs on output phase and catches child plugin errors itself

### Addressed issues:

- Fixes #1634

### DIFF
--- a/flexget/plugin.py
+++ b/flexget/plugin.py
@@ -156,7 +156,6 @@ DEFAULT_PRIORITY = 128
 # task phases, in order of their execution; note that this can be extended by
 # registering new phases at runtime
 task_phases = ['start', 'input', 'metainfo', 'filter', 'download', 'modify', 'output', 'learn', 'exit']
-suppress_abort_phases = set()
 
 # map phase names to method names
 phase_methods = {
@@ -176,7 +175,7 @@ _plugin_options = []
 _new_phase_queue = {}
 
 
-def register_task_phase(name, before=None, after=None, suppress_abort=False):
+def register_task_phase(name, before=None, after=None):
     """
     Adds a new task phase to the available phases.
 
@@ -188,8 +187,6 @@ def register_task_phase(name, before=None, after=None, suppress_abort=False):
         raise RegisterException('You must specify either a before or after phase.')
     if name in task_phases or name in _new_phase_queue:
         raise RegisterException('Phase %s already exists.' % name)
-    if suppress_abort:
-        suppress_abort_phases.add(name)
 
     def add_phase(phase_name, before, after):
         if before is not None and before not in task_phases:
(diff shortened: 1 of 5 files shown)


paranoidi:
Much improved, +1

liiight:
I think there's a potential issue. What if a`tasj` is defined under `entries`and task has several reruns? The user will get a notification per task run right?

gazpachoking:
> I think there's a potential issue. What if a task is defined under entries and task has several reruns? The user will get a notification per task run right?

Not quite sure what you mean by "a task is defined under entries", but yes, task notifications will be sent once per rerun. I don't think that behavior has changed though, and don't really see a way around it.

liiight:
Sorry I meant task configured under notify, you got the idea. 
Can we pass stuff under notify > task to run on exit phase?

gazpachoking:
> Can we pass stuff under notify > task to run on exit phase?

Sure, but that you'll only get notified about entries in the last rerun of the task, which I think is even less desirable than multiple notifications.

gazpachoking:
To do proper combined emails you sorta need a digesting task (at the moment.)

liiight:
Well I guess it's out of the scope of this, but aggregating entries in different reruns would be nice. I see no other reason not to merge this then.

PR #1654, opened 2017-01-25
added extra recognized group format

mosterdt:
### Motivation for changes:
This change adds a new format to the from_group filter in the series plugin.
The existing formats do not cover AnimeBytes' torrent feeds.
### Detailed changes:
At the moment, upload-groups are only recognized in a '-Group' or '[Group]' format.
This PR adds recognition for the '(Group)' format.
In addition, I added unit test cases that check the recognition of this new format.


liiight:
Looks good, thanks!

PR #1660, opened 2017-01-27
added filename->out option for aria2

mitalhp:
### Motivation for changes:

I am pulling down a podcast feed which uses random strings for filenames, but there was no way to pass in the "--out" parameter to aria to make it rename the file

### Detailed changes:

Used the same logic that was there for path for filename 

### Example config:
```
tasks:
  podcast-1:
    rss:
      url: https://podcast.feed/rss.xml
    regexp:
      accept:
        - .*
    aria2:
      path: '/downloads'
      filename: '{{title|pathscrub}}.mp4'
```

### DIFF
--- a/flexget/plugins/clients/aria2.py
+++ b/flexget/plugins/clients/aria2.py
@@ -33,6 +33,7 @@ class OutputAria2(object):
             'username': {'type': 'string', 'default': ''}, # NOTE: To be deprecated by aria2
             'password': {'type': 'string', 'default': ''},
             'path': {'type': 'string'},
+            'filename': {'type': 'string'},
             'options': {
                 'type': 'object',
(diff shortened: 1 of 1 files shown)


flexget/plugins/clients/aria2.py
```diff
@@ -106,6 +107,11 @@ def add_entry(self, aria2, entry, config):
         except RenderError as e:
             entry.fail('failed to render \'path\': %s' % e)
             return
+        try:
+            options['out'] = os.path.expanduser(entry.render(config['filename']))
```
cvium: This will throw a KeyError exception if 'filename' is not specified

mitalhp: ahh, didnt test that.  will update and resubmit.

PR #1663, opened 2017-01-30
FIX Unicode-objects must be encoded before hashing

ghyster:
### Motivation for changes:
plugin betaseries_list has a bug when using python 3 (BUG: Unhandled error in plugin configure_series: Unicode-objects must be encoded before hashing)

### Detailed changes:

- encode the password in utf-8 to avoid crash

### Addressed issues:

- Fixes # .

```
paste_config_here
```
```
paste output here
```
Tested on a local installation


gazpachoking:
Looks good to me, thanks!

PR #1664, opened 2017-01-30
Update npo_watchlist

mfonville:
* The NPO3 site has been updated to list fragments with the episode too, and list 'articles' also; new trick to only find episodes by looking at running time
* Add running time as npo_runtime (should also help with the date parsing)
* Update the test, added npo_runtime check; but also make sure to test for that things that should NOT be listed are asserted as None

### Motivation for changes:

NPO.nl site has been changed again


flexget/plugins/input/npo_watchlist.py
```diff
@@ -201,33 +201,36 @@ def _parse_episodes_npo3(self, task, config, series_name, page, series_info):
 
         entries = list()
         for listItem in page.findAll('div', class_='item'):
```
liiight: I know this wasn't introduced in this PR but please change `listItem` to `list_item`

flexget/plugins/input/npo_watchlist.py
```diff
@@ -267,6 +270,8 @@ def _parse_episodes_std(self, task, config, series_name, page, series_info):
             e['npo_name'] = series_info['npo_name']
             e['npo_description'] = series_info['npo_description']
             e['npo_language'] = series_info['npo_language']
+            if subtitle[1]:  # if runtime is available in the subtext
```
liiight: Will the previous split always work? If not, this may raise a KeyError

flexget/plugins/input/npo_watchlist.py
mfonville: alright, will do

flexget/plugins/input/npo_watchlist.py
mfonville: The previous split might indeed fail, because there is no easy way to check if 100% of the site confirms to this new subtext format. I will add an extra check

mfonville: Hmm, I checked the current way, it and this should actually work already correctly... and the split() always returns a list; (even if the splitted symbol is not there);
and the `if`-statement correctly handles the situation if the key exists in the list. It does not spawn any error with any input.

cvium:
```
Python 2.7.11 (v2.7.11:6d1b6a68f775, Dec  5 2015, 20:40:30) [MSC v.1500 64 bit (AMD64)] on win32
>>> l = 'nodots'.split('.')
>>> l
['nodots']
>>> l[1]
  File "<stdin>", line 1, in <module>
IndexError: list index out of range
>>>
```

mfonville: but it is not `l[1]`

it is `if l[1]:`

cvium: That will still fail. No matter what precedes `l[1]` it is a direct access to the 2nd element in the list.
```
>>> if l[1]:
...     pass
...
  File "<stdin>", line 1, in <module>
IndexError: list index out of range
>>>
```

liiight: It'll still throw that error. Do `if len(l) > 1`

mfonville: Ok, I will update it

PR #1667, opened 2017-02-01
Add Interfaces column to plugins output.

### LINKED ISSUE #1666: Adding new column to `plugins` output
I'm trying to understand the concepts of 'phase' and 'interface' of a plugin. The output of the CLI command `flexget plugins` has been helpful. But I think I noticed that the output is missing a column. It has 'Keyword', 'Phases' and 'Flags', but no 'Interfaces'. I found out that all plugins _do_ defined a list of interfaces for themselves. 

Shall I create a PR adding the column 'Interfaces' to the output?
I wanted to ask before I put the effort in.

And if someone can explain what a 'phase' and 'interface' is I'll write up a wiki page. ;)

haarts:
Issue #1666

### Motivation for changes:
The output of the command `$ flexget plugins` included 'Keyword', 'Phases' and 'Flags', the logical extension was to include 'Interfaces' as well thus completely describing the plugins.

### Detailed changes:

- Adds a header field.
- Adds a simple compaction of the interfaces list in the plugin interface field of the output.
- Added some white space between the logical constructions of the plugin fields.

### Addressed issues:

- Fixes #1666

### DIFF
--- a/flexget/plugins/cli/plugins.py
+++ b/flexget/plugins/cli/plugins.py
@@ -15,11 +15,12 @@ log = logging.getLogger('plugins')
 def plugins_summary(manager, options):
     if options.table_type == 'porcelain':
         disable_all_colors()
-    header = ['Keyword', 'Phases', 'Flags']
+    header = ['Keyword', 'Interfaces', 'Phases', 'Flags']
     table_data = [header]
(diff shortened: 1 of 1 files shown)


liiight:
Thanks for this!

PR #1677, opened 2017-02-09
[WIP] Subtitle list refactor

cvium:
### Motivation for changes:
It's a stupid mess because I rushed it.
### Detailed changes:

- Removed `remove_after` config option. Same functionality can be done with `age` plugin and a separate task, which I think is much cleaner.


paranoidi:
Kind of agree with you that it would be more idiomatic to have a separate task where items are purged, then again purging subtitle queue is probably wanted almost always and the option is handy ...

PR #1688, opened 2017-02-14
flask-restplus upgrade to 0.10.1

liiight:
### Motivation for changes:
Native JSON schema support, latest stuff...

Currently inflicted by https://github.com/noirbizarre/flask-restplus/issues/194 which causes Swagger page to crash, tests and UI work


tubedogg:
The Flask-RESTPlus issue has been closed.

>> The ``AttributeError`` exception is hiding the real error so each time it can be a different issue.
This won't be the case anymore and you will be able to either fix it if this is possible or report the real issue if necessary.

liiight:
I know, waiting on The 0.11 release

Issue #1692, opened 2017-02-17
rtorrent plugin crashed when adding torrent with special utf-8 characters


xiayyu:
### Expected behaviour:
Adding entry to rtorrent without crash
### Actual behaviour:
Adding entry to rtorrent but flexget crashed. 
As a consequence, all the accepted entry with this run won't write to the database marked seen, and they will be downloaded again and again. However all the accepted torrent will be sent to the rtorrent, but the crash reports are bothersome.
### Steps to reproduce:
When an entry with special utf-8 characters in name or path accepted and sent to rtorrent plugin.
Such as the example from the log, "Al final del túnel AKA At the End of the Tunnel 2016 1080p BluRay TrueHD 5.1 x264-PbK", "ú" is a special utf-8 characters.
The details of this torrent
```
 'info': {'files': [{'length': 14152687941,
                     'path': ['Al.final.del.t\xc3\xbanel.AKA.At.the.End.of.the.Tunnel.2016.1080p.BluRay.TrueHD.5.1.x264-PbK.mkv']},
                    {'length': 1281,
                     'path': ['Al.final.del.t\xc3\xbanel.AKA.At.the.End.of.the.Tunnel.2016.1080p.BluRay.TrueHD.5.1.x264-PbK.nfo']}],
          'name': 'Al.final.del.t\xc3\xbanel.AKA.At.the.End.of.the.Tunnel.2016.1080p.BluRay.TrueHD.5.1.x264-PbK',
          'piece length': 4194304,
          'pieces': '<3375 piece hashes>',
          'private': 1,
          'source': '[www.hdpter.net] HDPter'},
```
#### Config:
```
  hdp:
    rss: **********
    accept_all: yes
    rtorrent:
      uri: scgi://127.0.0.1:5000
      path: /mnt/V3T/_Movies/_HD
      custom1: HD

```
  
#### Log:
```
2017-02-17 10:37 VERBOSE  task          hdp             REJECTED: `Hacksaw Ridge 2016 1080p BluRay TrueHD Atmos 7.1 x264-PbK` by seen plugin because entry with url `http://www.hdpter.net/download.php?id=10438&passkey=ab0ee1ba02ec1f4cd892219cab827967` is already marked seen in the task hdp at 2017-02-12 22:03
[...]

PR #1694, opened 2017-02-17
issues #1692

xiayyu:
details discussed [here](https://github.com/Flexget/Flexget/issues/1692)


cvium:
That is not the correct solution to this problem. @gazpachoking knows more about this though.

xiayyu:
If insist keeping the `unicode_literals`, change the way to init `response_body`, use`response_body = response.read(1024)` instead of `response_body = ''`.  This is the most straight-forward way for me. maybe there are other standard modification.

PR #1699, opened 2017-02-18
Escape tvmaze URL query parameters

pkoenig10:
### Motivation for changes:

When searching for shows using the TVmaze API, Flexget does not escape any query parameters.  This leads to incorrect searches.  My log file showing the error is linked below.  You can see that Flexget attempted to search for "Law & Order: Special Victims Unit", but the "&" was not escaped.  If you go to the requested URL you will see that the received search term was just "Law", resulting in the plugin returning Newton's Law instead of Law & Order SVU.

### Detailed changes:

- Pass any query parameters to requests as a dictionary so they are properly escaped.

### Log output:

https://gist.github.com/anonymous/460c9231e44062b6ce0d9c354fbc6ea4


liiight:
Very nice, thanks for this!

PR #1707, opened 2017-02-24
Fix tvdb lookup in myepisodes plugin

oxc:
The myepisodes service used the 'api_tvdb' plugin, which seems to be no longer available on its own. This change updates it to use the thetvdb_lookup plugin.

### Motivation for changes:
Configurations using myepisodes plugin no longer work.

### Detailed changes:
Use the thetvdb_lookup plugin to update the entry. Since this does not throw a LookupError, restructure to use early returns to keep the flow minimal.

### DIFF
--- a/flexget/plugins/services/myepisodes.py
+++ b/flexget/plugins/services/myepisodes.py
@@ -204,20 +204,15 @@ class MyEpisodes(object):
 
         Return: myepisode id or None
         """
-        search_value = entry['series_name']
-
         # Get the series name from thetvdb to increase match chance on myepisodes
-        if entry.get('tvdb_series_name'):
-            search_value = entry['tvdb_series_name']
-        else:
-            try:
-                series = plugin.get_plugin_by_name('api_tvdb').instance.lookup_series(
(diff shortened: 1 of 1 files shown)


liiight:
What's the problem with `api_tvdb`?

oxc:
The job failed with 
> CRITICAL task          series_torrents Plugin `myepisodes` cannot be used because dependency `api_tvdb` is missing.

I have not found any other usage as in the myepisodes plugin, and figured it missed out on an update again (which has happened in the past).
No plugin seems to register under this name, and judging by a quick search, the internal api by that name seems to be only used by other tvdb plugins.
Also, doesn't it make more sense to use the lookup wrapper, than the "raw" api?

liiight:
Hmm looks like that it's indeed missing it's plugin register event hook, weird. The fix should be just to add it, I have no idea why it's gone. Not sure if it was removed accidentally, need to lookup plugin history. 
About the second question, `api_tvdb` is the lookup wrapper, `thetvdb_lookup` is a plugin that allows it to be used via the task interface. The correct approach is to call it directly, like it currently is.

oxc:
Ok, I'll close this one then.

PR #1710, opened 2017-02-26
Season packs support - Phase 1

liiight:
### Motivation for changes:
Native season pack support is long overdue. [Design](https://flexget.com/Drafts/Season-pack-support)
### Detailed changes:

- Added basic support for season pack type in parser
- Added `Season` table in series DB
- Added logical test when tracking series for checking if season pack has been downloaded
- Split `Release` table to `EpisodeRelease` and `SeasonRelease`
- Added 4 modes for `season_packs` config: boolean, integer, `always` and 'only'. These will affect episode threshold control:
    By default (or by passing `season_packs: yes` the default episode threshold will be set to 0,    meaning that if a series has 1 or more episode with a downloaded release, a season pack will be rejected.
    By passing a specific integer (`season_packs: 5` for example), that threshold can be raised.
Setting `season_packs: always` will always accept the season pack, regardless of how many episodes were already downloaded for the series.
    By setting `season_packs: only`, series plugin will accept only season packs, and reject all episodes.
    All modes still abide to checking if season is not completed. 
- Added lookup by season to `api_tvmaze`, `tvmaze_lookup` and `est_release_series_tvmaze`

```
series:
- foo:
    season_packs: yes
- baz:
    season_packs: 5
- boo:
    season_packs: always
- bla:
    season_packs: only
```
### Existing issues:

- Guessit does not support season pack ATM. Its parser will always return False on the `is_season_pack` call.

#### To Do:

- [x] Create a default threshold that'll check if any existing eps for the season have already been downloaded before accepting season pack.
~- [ ] Add support for partial season packs (will not mark season as complete)~ TBD, this does not fit well with the current DB relations and flow
- [x] DB Migration. Will do this when DB schema will lock.
- [x] More tests.
- [x] Edit `next_series_episodes` to take completed season into account.
- [x] Create `next_series_seasons`
- [x] Enable estimators to search for season
- [x] Verify no metadata plugins crash.
- [x] Fix series CLI


flexget/plugins/filter/series.py
```diff
@@ -302,6 +304,61 @@ def __repr__(self):
...
+    def __str__(self):
+        return '<Season(id=%s,identifier=%s,season=%s)>' % \
+               (self.id, self.identifier, self.season)
+
+    def __repr__(self):
+        return str(self).encode('ascii', 'replace')
+
+    def __eq__(self, other):
+        if not isinstance(other, Season):
+            return NotImplemented
```
cvium: raise?

liiight: oops

gazpachoking:
With regards to guessit, when it returns type 'episode' with a season, but not an episode number, is that a reliable indication of season pack?

Also, from IRC: Switch (at least by default) to not download season packs once individual episodes have already been downloaded from a given season. Possibly allow an option to define how many individual eps are allowed to be downloaded and still consider getting a full season pack.

liiight:
Dunno about guessit, even if that were to case, I plan to add support for partial season packs (`Foo.S01.PART1-Flexget`) and that would require explicit support.

Toilal:
It has support for this

liiight:
Oh, I guess its a matter of just updating the flexget plugin. You have an open pr for this right?

Toilal:
I'm not sure flexget reads the part property returned by guessit. And yes,
it should be updated, I opened a pull request some times ago but 1 or 2
tests was failing and i don't have time to investigate more.

flexget/plugins/filter/series.py
```diff
@@ -313,7 +373,11 @@ class Episode(Base):
 
     identified_by = Column(String)
     series_id = Column(Integer, ForeignKey('series.id'), nullable=False)
-    releases = relation('Release', backref='episode', cascade='all, delete, delete-orphan')
+    releases = relation('EpisodeRelease', backref='episode', cascade='all, delete, delete-orphan')
+
+    @property
+    def is_season(self):
+        return False
```
liiight: We're using it in a bunch of places actually, like https://github.com/Flexget/Flexget/blob/5b2988d31cdb980b546b490120bd532d81791663/flexget/plugins/filter/series.py#L1486

What's the benefit of setting a class variable as opposed to a property method?

liiight: nm, did it already, pushing soon

liiight:
I changed a lot of the comparison stuff. A lot of it was just dumb copy paste form Episode. It should make more sense now

tubedogg:
I installed this branch for testing. From my (admittedly limited) testing this is not working with ``all_series`` - nothing is being decided upon and there are no ``series_*`` fields in the entry dump. Is resolving this within the scope of Phase 1?

(test filename: Show.Name.S02.720p.HDTV.x264-GROUP.torrent)

liiight:
Thanks for testing. Please paste config and debug logs. Admittedly, I did not test this with `all_series` at all

tubedogg:
Here it is. I should've posted that in the first place, my bad.

Config:
```
templates:
  watchdir_filesystem_tv:
    filesystem:
      path: "/path/to/torrents/tv"
      mask: '*.torrent'
  watchdir_move:
    move:
      to: '/path/to/torrents/_processed/'
  watchdir_disable:
    disable:
      - seen_info_hash
      - seen

tasks:
  watchdir_tv_eps:
    manual: yes
    template:
      - deluge-connect_tv
      - notify_tv
      - watchdir_disable
      - watchdir_move
      - watchdir_filesystem_tv
    set:
      tracker: tv
      step: 1
      notify_title: 'TV - Watch to Deluge'
    no_entries_ok: yes
    all_series:
      tracking: no
      season_packs: yes
```

Debug log:
https://gist.github.com/tubedogg/1caeb1cdf3b4949ab4b9d1207a21d111

liiight:
Hmm, looks like it didn't parse that to be a season pack. 
Please set the value [here](https://github.com/Flexget/Flexget/blob/develop/flexget/utils/titles/series.py#L22) to `debug` and run again.

tubedogg:
The only thing that added to the log was this:

```
2017-04-02 01:48 DEBUG    seriesparser  watchdir_tv_eps No name for series `Show.Name.S02.720p.HDTV.x264-GROUP` supplied, guessing name.
2017-04-02 01:48 DEBUG    seriesparser  watchdir_tv_eps Could not determine a series name
```

I think [guess_name](https://github.com/Flexget/Flexget/blob/develop/flexget/utils/titles/series.py#L160) only tries to match it as an episode, not a season pack, and gives up if it's not an episode.

liiight:
Yeah, i think you're right. Good catch, I'll test and fix soon

liiight:
try now @tubedogg

tubedogg:
@liiight Awesome, works a treat now. 👍  Thanks for doing this - super-excited to finally have season pack support!

liiight:
sure thing. while I have your attention, it'd be great if you could also test `next_series_seasons`, preferably with real life sites. those scenarios are the hardest to antcipate.

tubedogg:
I modified the BTN search plugin to support season packs and gave it a go with``next_series_seasons``. Seems to work great. :)

(For the BTN plugin, do you want me to just tell you what I did and you can include it as part of this PR, or just hold onto it until this is released and submit a PR for BTN?)

liiight:
You can tell me, but it's outside the scope of this pr

tubedogg:
I just put it in a gist, two quick changes. https://gist.github.com/tubedogg/3ae75ab928b8a21fc9b5a51a7cf59ac6

PR #1716, opened 2017-03-04
Extend Slack notification to support attachments

aidan-:
### Motivation for changes:

- Extend existing Slack notification to support [message attachments](https://api.slack.com/docs/message-attachments)  

### Detailed changes:

- modify `flexget/plugins/notifiers/slack.py` and add the additional `attachments` schema.
- send the new attachment field in the notification payload

### Config usage:
Extended from the existing example on the wiki:
```
tasks:
  move-test:
    rss: <redacted>
    imdb_lookup: yes
    imdb:
      min_year: 2016
    notify:
      entries:
        message: >
          {{task}} - Download started:
          {%- if imdb_name is defined %}
          {%- if imdb_url is defined %}
          <{{imdb_url}}|{{imdb_name}} {{imdb_year}}>
          {%- else %}
          {{imdb_name}} {{imdb_year}}
          {%- endif %}
          - {{quality}}
          {%- else -%}
          {{title}}
          {%- endif %}
        via:
          - slack:
              web_hook_url: <redacted>
              attachments:
                - image_url: '{{imdb_photo}}'
                - title: '{{imdb_name}} ({{imdb_year}})'
                  title_link: '{{imdb_url}}'
                  fallback: '{{imdb_name}} {{imdb_year}}'
                  text: '{{imdb_plot_outline}}'
                  color: "#3AA3E3"
                  footer: Flexget download
                  footer_icon: 'https://flexget.com/attachments/LogoContest/PastedGraphic-3.png'
                  fields:
                    - title: Votes
                      value: '{{imdb_votes}}'
                      short: true
                    - title: Score
                      value: '{{imdb_score}}'
                      short: true
                    - title: Genres
                      value: '{{imdb_genres|join(", ")}}'
                      short: true
                    - title: Rating
                      value: '{{imdb_mpaa_rating}}'
                      short: true
                    - title: Actors
                      value: >
                        {% for key, value in imdb_actors.iteritems() %}
                        {{value}}{% if not loop.last %},{% endif %}
                        {% endfor %}
                      short: false
```
###  Output:
A picture of the above config's output in slack:
screen shot 2017-03-04 at 3 48 00 pm
#### To Do:

- Update the [notifications documentation](https://flexget.com/Plugins/Notifiers/slack) to explain how the attachment field works as well as include some example configuration.

### DIFF
--- a/flexget/plugins/notifiers/slack.py
+++ b/flexget/plugins/notifiers/slack.py
@@ -26,6 +26,7 @@ class SlackNotifier(object):
         [username: <string>] (override username)
         [icon_emoji: <string>] (override emoji icon)
         [icon_url: <string>] (override emoji icon)
+        [attachments: <array>[<object>]] (send a more complex 'message' as part of the notification)
 
     """
     schema = {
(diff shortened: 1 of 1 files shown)


flexget/plugins/notifiers/slack.py
```diff
@@ -35,7 +36,74 @@ class SlackNotifier(object):
             'channel': {'type': 'string'},
             'username': {'type': 'string', 'default': 'Flexget'},
             'icon_emoji': {'type': 'string'},
-            'icon_url': {'type': 'string', 'format': 'url'}
+            'icon_url': {'type': 'string', 'format': 'url'},
+            'attachments': {
```
liiight: Attachments schema should have some required fields according to slack docs if I remember correctly.

liiight: Also I'd consider some default values for some of the more crucial fields

aidan-: I just read over the documentation, it doesn't look like any of the fields are mandatory?  This is shown in my example config which is actually two attachments (first being the picture, second being the content).  

I'm also not sure what the benefit would be in setting defaults.  My example config relies heavily on `imdb_lookup` but I can imagine we can't assume people are using that plugin let alone want to use attachments instead of a basic message.

liiight:
https://api.slack.com/docs/message-attachments#attachment_structure
Looks like there are several required fields, anything that isn't listed as optional. 

About defaults, it's perfectly valid to use imdb lookup data as defaults, we do that with other templates as well. Just make sure to doc it properly.

liiight:
You wanna update this?

liiight:
Reopen if relevant

PR #1723, opened 2017-03-08
Call log.warning instead of raising PluginWarning

jonathanou:
### Motivation for changes:
I use the [delete plugin](https://www.flexget.com/Plugins/delete) to clean folders older than 30 days.  The delete plugin has a feature to clean the source folder (`clean_source: [folder size]`) once the folder is under a certain size.  When using this feature, I often run into this scenario:

1. Folder contains 2 files, A.txt and B.txt, both 0.5 MB large.
2. The delete plugin specifies `clean_source: 1`, so it will clean the source folder once the folder size is less than 1 MB.
3. The delete plugin accepts A.txt, B.txt, as well as other files in other folders as its list of entries to delete.  
4. The delete plugin removes A.txt, the folder size is now 0.5 MB.
5. Immediately after removing A.txt, the plugin checks if the folder size is less than 1 MB.  It is, so it removes the entire folder (including B.txt).
6. The delete plugin moves on to remove B.txt, but it does not exist anymore.
7. `PluginWarning` is raised claiming it can't find B.txt and stops the task.
8. Subsequent files that were scheduled to be deleted are not processed.

### Detailed changes:

- Instead of raising `PluginWarning` in `move.py` when the code detects something is wrong, we call `log.warning` to still log the message, but will allow us to move on and process the next file.


liiight:
It sounds like the correct approach would be to break the loop instead of just logging a warning

jonathanou:
But breaking the loop stops other files from being processed.  If I'm deleting a bunch of files and one of them don't exist on disk anymore, why should it stop the other files from being deleted?

liiight:
Maybe I misunderstood the scenario. You describe a situation where during iteration over files in a folder, that folder gets deleted, and trying to operate on subsequent files in the loop cannot work, right? 
If so, when deleting the folder the loop need to be handled correctly. 

If I didn't under this correctly I apologize.

jonathanou:
Yes, that's right.  Here's a more conrete example:

1. I have `clean_source: 1` specified (delete the folder if less than 1 MB).  
2. I'm trying to delete the following files `Folder_A/A.txt`, `Folder_A/B.txt`, and `Folder_B/C.txt`.
3. The loop deletes `Folder_A/A.txt`, then checks if `Folder_A` is less than 1 MB.  Let's say it is, and so it deletes `Folder_A`.
4. The loop iterates to `Folder_A/B.txt`, but finds that it does not exist on disk anymore.  The current implementation will raise PluginWarning and the entire task aborts.
5. `Folder_B/C.txt` never gets processed.

I feel that the loop should carry on to the next iteration *regardless* if it can perform the operation on the current item or not.  Even if deleting the folder is handled by the loop gracefully, there could still be other reasons to cause the delete to fail, like permission denied.

I think the easiest way is to log a warning and continue, like what I have proposed.

liiight:
The easiest way and the best way are not always interchangeable. That being said, I'm not sure you're wrong. I rather wait for some more feedback on this.

jonathanou:
Was there anyone you wanted to include to discuss this?

liiight:
@cvium and @paranoidi maybe. I imagine they saw this already

gazpachoking:
This change seems good to me. We don't want errors handling one entry to stop all subsequent entries from processing. I'm a bit confused by all the extra commits in this PR, but I suppose they don't matter as we can just squash it down when merging.

jonathanou:
The extra commits in the PR was for a separate PR that was already merged back in January.  I reused the same fork to add this change.  

I'm quite new to Git PRs so I'm probably not doing it right.  Is it recommended to delete the fork once a PR is merged, and then later re-fork if you want to create another PR?  Or is there a way to reuse the fork but keep the commit history clean?

liiight:
Best practice is to maintain your fork in sync and open a branch per feature /pr

PR #1732, opened 2017-03-12
Set is_clean_all when intended.

autojuji:
### Motivation for changes:
When setting "clean_transmission: yes" nothing is cleaned before this patch.

### Detailed changes:
Code used to set a value to False and test for None.


gazpachoking:
The code for the options in clean_transmission is a bit crazy, but your logic seems sound. LGTM

PR #1737, opened 2017-03-15
[change] Extend series list help message. (#1720)

### LINKED ISSUE #1720: Add helpful message when `flexget series list` is empty
When no series are tracked/configured the command `flexget series list` returns an empty list.
That is not to say flexget doesn't know about *any* series, they might simply be unconfigured. The series can be viewed with the `flexget series list all` (or `flexget series list unconfigured`) command.

Currently the output of `flexget series list` is:
```shell
$ flexget series list

lqqqqqqwqqqqqqqqwqqqqqwqqqqqqqqqqqqwqqqqqqqqqqqqqqqk
x Name x Latest x Age x Downloaded x Identified By x
mqqqqqqvqqqqqqqqvqqqqqvqqqqqqqqqqqqvqqqqqqqqqqqqqqqj
Use `flexget series show NAME` to get detailed information
```

Like the helpful note on how to get more detailed information, a similar note could be added on the three available filters ('configured', 'unconfigured', 'all'). This message could be made conditional on the list being empty.

Additionally the output could lead with which filter the list is shown to begin with. Something like: 
```shell
$ flexget series list {,unconfigured,configured,all}
{Unconfigured,Configured,All} series are:
lqqqqqqwqqqqqqqqwqqqqqwqqqqqqqqqqqqwqqqqqqqqqqqqqqqk
x Name x Latest x Age x Downloaded x Identified By x
mqqqqqqvqqqqqqqqvqqqqqvqqqqqqqqqqqqvqqqqqqqqqqqqqqqj
Use `flexget series show NAME` to get detailed information
Use `flexget series FILTER` to use different filter
```

### Additional information:

- Flexget Version: 2.10.10
- Python Version: 3.5.2
- Installation method: pip3
- OS and version: Ubuntu 16.04

haarts:
### Motivation for changes:
See #1720 

### Detailed changes:

- Added 1 line of helpful text.

### Addressed issues:

- Fixes #1720.

#### To Do:

When 'implementing' this I noticed that most of the CLI plugins don't return anything when no additional CLI arguments are given. For example `flexget series` just returns as does `flexget archive`. `flexget regex` crashes (because it uses a map). That's rather confusing (for me at least).
I believe it would be helpful to print the help page with a leader like 'no option selected'. Would you agree?

### DIFF
--- a/flexget/plugins/cli/series.py
+++ b/flexget/plugins/cli/series.py
@@ -71,7 +71,7 @@ def display_summary(options):
         for index, value in enumerate(header):
             if value.lower() == options.sort_by:
                 header[index] = colorize(SORT_COLUMN_COLOR, value)
-        footer = 'Use `flexget series show NAME` to get detailed information'
+
         table_data = [header]
         for series in query:
             name_column = series.name
@@ -109,7 +109,10 @@ def display_summary(options):
         console('ERROR: %s' % str(e))
         return
     if not porcelain:
-        console(footer)
+        if not query.count():
+            console('Use `flexget series list all` to view all known series.')
+        else:
+            console('Use `flexget series show NAME` to get detailed information.')
 
 
 def begin(manager, options):


liiight:
That message is more confusing than helpful imo. Also, I'd do this when the default series list return nothing, like what happened to you. 

About what you said with cli, I agree with you and I imagine this could be done by forcing our subparser options to contain at least one option.

haarts:
Would it help if I'd write smt like: 'Use `flexget series all` to view all known series.'?
I'll make showing it conditional. 
If alright I'll make the other footer text conditional then too. It makes little sense to show it when displaying an empty list.

liiight:
Cool

flexget/plugins/cli/series.py
```diff
@@ -109,7 +109,10 @@ def display_summary(options):
         console('ERROR: %s' % str(e))
         return
     if not porcelain:
-        console(footer)
+        if query.count() == 0:
```
liiight: The python way is to do:
```python
if not query.count():
```
Also, I think this should be before the table construction, since displaying an empty table is weird IMO.
And maybe add some thing like: "No results found. Try using flexget series all to view all known series" to make it more clear.

haarts: Excellent!
I was a bit reluctant to change too much but I considered doing just that. Will do.

haarts:
@liiight Not sure if you saw the latest commit I pushed.

liiight:
I'm not sure why this over-complication is needed. just check if `not query.count`, send a message to console and return...

haarts:
I merely split the original function into two. The diff might look daunting but it's almost trivial. Reasoning behind doing so is the fact that the original function became/was quite long hampering (my) understanding. But you're the owner. If you prefer I'll revert it.

liiight:
I agree that it's quite long but i don't think this is the right time (or way) to address that.

haarts:
I'll amend. 
Just out of curiosity; what would be the way to address it. In general terms of course.

liiight:
Try to make it a DRY as possible, but since this is a UI plugin, and one that holds a significant amount of data and data types, some cluster is bound to exist.

haarts:
I'm unsure why CI failed. The log is blank?

PR #1738, opened 2017-03-17
Update NPO.nl schema.org URL

mfonville:
### Motivation for changes:
NPO.nl site has been updated; the schema.org URL has been updated

### Detailed changes:

- changed http to https in the schema.org URL
- updated the tests


mfonville:
fyi this patch is critical to make the plugin work again; otherwise it can't find and process the correct `div` element on the webpage and crashes

PR #1760, opened 2017-03-28
Change trakt URLs to TLS and plural collections.

haarts:
### Motivation for changes:
The URLs generate by flexget in various places still used http as
protocol. This was changed by Trakt to https some time ago. The commit
reflects that fact.

Several URLs have changed from using `serie` and `movie` in the path to
their plural form. Trakt.tv redirected these singular forms to the
canonical plural form.

### Detailed changes:

In essence this was a search and replace exercise.

### Addressed issues:

None. All the 'wrong' URLs are correctly redirected by Trakt.tv. They might stop doing that at some point.

### DIFF
--- a/flexget/plugins/list/trakt_list.py
+++ b/flexget/plugins/list/trakt_list.py
@@ -193,10 +193,10 @@ class TraktSet(MutableSet):
                     continue
                 entry = Entry()
                 if list_type == 'episode':
-                    entry['url'] = 'http://trakt.tv/shows/%s/seasons/%s/episodes/%s' % (
+                    entry['url'] = 'https://trakt.tv/shows/%s/seasons/%s/episodes/%s' % (
                         item['show']['ids']['slug'], item['episode']['season'], item['episode']['number'])
                 else:
-                    entry['url'] = 'http://trakt.tv/%s/%s' % (list_type, item[list_type]['ids'].get('slug'))
+                    entry['url'] = 'https://trakt.tv/%ss/%s' % (list_type, item[list_type]['ids'].get('slug'))
                 entry.update_using_map(field_maps[list_type], item)
                 # Override the title if strip_dates is on. TODO: a better way?
                 if self.config.get('strip_dates'):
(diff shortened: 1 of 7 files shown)


flexget/plugins/list/trakt_list.py
```diff
@@ -193,10 +193,10 @@ def items(self):
                     continue
                 entry = Entry()
                 if list_type == 'episode':
-                    entry['url'] = 'http://trakt.tv/shows/%s/seasons/%s/episodes/%s' % (
+                    entry['url'] = 'https://trakt.tv/shows/%s/seasons/%s/episodes/%s' % (
                         item['show']['ids']['slug'], item['episode']['season'], item['episode']['number'])
                 else:
-                    entry['url'] = 'http://trakt.tv/%s/%s' % (list_type, item[list_type]['ids'].get('slug'))
+                    entry['url'] = 'https://trakt.tv/%ss/%s' % (list_type, item[list_type]['ids'].get('slug'))
```
liiight: `%ss`?

cvium:
> Several URLs

What URLs? I don't think it makes sense to make partial changes right now if they're going to change things again soon.

flexget/plugins/list/trakt_list.py
haarts: A couple of lines up `list_type` is defined of which the trailing `s` is stripped, yielding 'serie'/'movie'/'episode'. That variable is use as the key to the Trakt API response dictionary. And it is _also_ used here to construct the URL. Which requires a trailing 's' again.

haarts:
@cvium I have no reason to believe the Trakt operators are going to change anything anymore. I'm sorry if I gave that impression. This commit encompasses a comprehensive change.

The URLs changed are:
- every 'http' became 'https'
- all paths '/movie/*' and '/show/*' became '/movies/*' and '/series/*'

flexget/plugins/list/trakt_list.py
mfonville: Isn't it more easy to then just drop the stripping of the `s` and change those other locations where it is compared?

haarts: @mfonville well... The variable `list_type` is used in two ways. One, and the most prevalent, is as the index or as the comparison to a dictionary index. The other, used only once, as part of this URL generation. 
I've opted to only make the URL generation slightly off and keep the majority the same.

haarts:
Is there anything I can do to move this PR forward?

liiight:
convince @cvium to merge it

haarts:
I'll give that a go then!
@cvium does my explanation address your question?

PR #1762, opened 2017-03-31
Update api_t411.py

poulpito:
Api changed from http to https only 
without scheme change, no json could be retrieve from api


poulpito:
looks like i missed something.  Build localy and it's working. i guess i need to modify test files also.
i'll have a look to understand how it's working 

Sry

PR #1770, opened 2017-04-04
Enable notify plugin to use entry data when sending notification

### LINKED ISSUE #1767: Allow using set plugin to override notification settings
### Expected behaviour:
Using the `set` plugin to override things like prowl's api_key, on a per-section basis. In the example config, the default should be set, then overridden for the 'wife' and 'daughter' sections. The 'family' and 'dad' should use the default. If there's a way to not have to have the entire notify -> entries -> via -> prowl chain for each override, that would be nice, but not essential.

### Actual behaviour:
It does not work, and whatever the api_key is set to first, "wins" for every notification.

#### Config:
```
tasks:
  tv-shows:
    notify:
      entries:
        via:
          - prowl:
              api_key: xxxxxxxxxxx
    series:
      settings:
        wife:
          set:
            notify:
              entries:
                via:
                  - prowl:
                      api_key: yyyyyy
        daughter:
          set:
            notify:
              entries:
                via:
                  - prowl:
                      api_key: zzzzzz
    wife:
      - "Murder Death Kill"
    daughter:
      - "Ponies and Kittens"
    family:
      - "Innuendo and Laugh Tracks"
    dad:
      - "Actual Good TV"
```

### Additional information:

- Flexget Version: 2.10.7
- Python Version: 2.7.13
- Installation method: pip
- OS and version: Debian 9.0/4.9.0-2-amd64

liiight:
### Motivation for changes:

Default behavior used to be so that entry data takes precedence as a rule of thumb. That ability was lost in a recent refactor, this PR brings it back.

### Detailed changes:

- Going over entry fields and trying to match `plugin_name_` as a key prefix
- This will be applied only when using `entries` scope notifications.
- Updated `pushover` plugin to use this.

### Usage:
```yaml
tasks:
  not_task:
    set:
      pushover_url: http://new.url.com
    mock:
    - title: bla
    accept_all: yes
    notify:
      entries:
        via:
        - pushover:
             api_key: token
             url: http://old.url.com
```

### Addressed issues:

- Fixes #1767

### DIFF
--- a/flexget/plugins/notifiers/email.py
+++ b/flexget/plugins/notifiers/email.py
@@ -14,6 +14,7 @@ from flexget import plugin
 from flexget.config_schema import one_or_more
 from flexget.event import event
 from flexget.plugin import PluginWarning
+from flexget.utils.tools import merge_by_prefix
 
 plugin_name = 'email'
 log = logging.getLogger(plugin_name)
@@ -137,7 +138,7 @@ class EmailNotifier(object):
         'additionalProperties': False,
     }
 
-    def notify(self, title, message, config):
+    def notify(self, title, message, config, entry=None):
         """
         Send an email notification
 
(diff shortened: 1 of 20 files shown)


liiight:
This won't work. a deeper solution needs to be implemented.

PR #1763, opened 2017-03-31
Update t411 api to https

poulpito:
T411 modify api to respond only to https (http fallback to https)
without modification the api cannot retrieve json

PR #1777, opened 2017-04-06
Update torrent411.py

ProphecyKt:
2 littles problems for the old t411 plugins
- one "li" to "ai" missing
- force download .torrent in https.

Thanks

### DIFF
--- a/flexget/plugins/sites/torrent411.py
+++ b/flexget/plugins/sites/torrent411.py
@@ -398,7 +398,7 @@ class UrlRewriteTorrent411(object):
     #   urlrewriter API
     def url_rewritable(self, task, entry):
         url = entry['url']
-        if re.match(r'^(https?://)?(www\.)?t411\.li/torrents/(?!download/)[-A-Za-z0-9+&@#/%|?=~_|!:,.;]+', url):
+        if re.match(r'^(https?://)?(www\.)?t411\.ai/torrents/(?!download/)[-A-Za-z0-9+&@#/%|?=~_|!:,.;]+', url):
             return True
         return False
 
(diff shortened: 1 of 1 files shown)


liiight:
Whats the difference between [this](https://github.com/Flexget/Flexget/blob/develop/flexget/plugins/sites/torrent411.py) and [this](https://github.com/Flexget/Flexget/blob/develop/flexget/plugins/sites/t411.py)?

ProphecyKt:
@liiight The first is just an url rewritter (the one I use) the second is a full plugin (with api) but it has some problem for me like the fact i can't use windows category because it appears twice.
I use the first for a long time even if it's deprecated

liiight:
Why not just fix the newer one then? Sustaining both seems silly

ProphecyKt:
Because for me the new one is not complete and it seems it's not updated since a while.

poulpito:
@ProphecyKt  is the api still working for you ? 
was using t411.py until now and it was working great 

can you gave me the documentation link about the other plugin use ? 
thks :)

poulpito:
i give a try the first one (with the https patch) and yes it's still a pretty good alternative to the newer one (especially like now when the api is down)   
@ProphecyKt   thanks for the alternative :)

liiight:
So this pr still applies? It'll fix that bug? This situation is very confusing, not sure how we managed to have two plugins for the same site

ProphecyKt:
The first one is not a plugins, it's a url rewritter. There is not official documentation but you can find one at https://wiki.t411.ai/doku.php/logiciels/flexget

poulpito:
@liiight   PR still applies  and fix https issue 
from my point of view it's good to have both as one use normal webpage page search and the other use api (confirmed not working for now on #1783 )

ProphecyKt:
@poulpito my pleasure but it's not my work (it's lildadou) I've just made a little pull.

karma007:
Hello, i'm on version 2.10.30 and still have aborting tasks with error : Expecting value: line 1 column 1 (char 0)

thanks

cvium:
See https://github.com/Flexget/Flexget/pull/1763#issuecomment-293243463

PR #1790, opened 2017-04-13
[add] deluge - Rename top-level torrent folder

andocromn:
### Motivation for changes:
Currently there is no way of doing this.  content_filename will rename the folder only if a main file can be identified.  This does not work well with Season Packs which support has recently been added for.

### Detailed changes:

- Adds the option container_directory to the Deluge plugin (the container_directory verbiage was adopted from a deluge_mod.py shared to me by tubedogg)
- container_directory is a jinja2 template used by client.core.rename_folder to rename the top level folder - which is effectively the torrent name in deluge 
- added 'name' to status_keys array to support this
- rename_folder function has been added after code for content_filename to prevent interference 

```
tasks:
  season-packs:
    rss: http://www.****.com/followed.rss
    deluge:
      label: seasonpacks
      addpaused: yes
      path: /media/tv/{{series_name}}/
      movedone: /media/tv/{{series_name}}/
      container_directory: '{{series_name}} - Season {{series_season}}'
    all_series:
      target: 720p
      timeframe: 72 hours
      upgrade: yes
      season_packs: only

```
#### To Do:

- [x] Verify conflicts do not occur when container_directory is used in conjunction with content_filename


flexget/plugins/clients/deluge.py
```diff
@@ -658,10 +660,14 @@ def rename(file, new_name):
                             entry['title'],
                             opts.get('main_file_ratio') * 100))
 
+                if config.get('container_directory'):
+                    log.info('Renaming Folder %s to %s' % (status['name'], pathscrub(entry.render(config.get('container_directory')))))
```
liiight: Let the logger do its own string interpolation, pass it the args directly

flexget/plugins/clients/deluge.py
```diff
@@ -423,6 +424,7 @@ def prepare_config(self, config):
         config.setdefault('magnetization_timeout', 0)
         config.setdefault('keep_subs', True)  # does nothing without 'content_filename' or 'main_file_only' enabled
         config.setdefault('hide_sparse_files', False)  # does nothing without 'main_file_only' enabled
+        config.setdefault('container_directory', '')
```
liiight: Setting an empty default seems redundant to me

andocromn: agreed, line removed

flexget/plugins/clients/deluge.py
andocromn: removed the pathscrub.  better?

tubedogg: Why not just render it once into a variable and then pass that variable to both the logger and the Deluge function? Saves a bit of processing.

tubedogg:
I haven't worked with Deluge's rename_folder before, but I wonder if it's smart enough to know if the something passed to it is not actually a folder, given that you can rename a complete folder structure by calling rename_file. If not, and the user sets container_directory with a single-file torrent that has no directory structure, it's going to end up renaming the file itself, since that's the torrent's name.

I don't think I'd rely on torrent name in either case. I'd split the (big?) file by os.sep and then call rename_folder on whichever portions you want, in your case just the top one I guess.

(I'd suggest revising the PR's name to "[add] deluge - Rename top-level torrent folder" so the changelog can pick it up easily. :))

flexget/plugins/clients/deluge.py
```diff
@@ -658,10 +659,17 @@ def rename(file, new_name):
                             entry['title'],
                             opts.get('main_file_ratio') * 100))
 
+                if config.get('container_directory'):
+                    if len(status['files']) > 1:
+                        log.info('Renaming Folder %s to %s', status['name'], entry.render(config.get('container_directory')))
```
liiight: I think info is a bit much for this. Maybe `verbose`?

flexget/plugins/clients/deluge.py
```diff
@@ -658,10 +659,17 @@ def rename(file, new_name):
...
+                        main_file_dlist.append(client.core.rename_folder(torrent_id, status['name'], pathscrub(entry.render(config.get('container_directory')))))
+                    else:
+                        log.info('container_directory specified however the number of files is not greater than 1 - skipping rename')
```
liiight: same here, info is too much, maybe even `debug` for this one

flexget/plugins/clients/deluge.py
```diff
@@ -660,16 +660,20 @@ def rename(file, new_name):
                             opts.get('main_file_ratio') * 100))
 
                 if config.get('container_directory'):
-                    if len(status['files']) > 1:
-                        log.info('Renaming Folder %s to %s', status['name'], entry.render(config.get('container_directory')))
-                        main_file_dlist.append(client.core.rename_folder(torrent_id, status['name'], pathscrub(entry.render(config.get('container_directory')))))
+                    if opts.get('content_filename') or opts.get('main_file_only'):
+                        folder_structure = big_file_name.split(os.sep)
```
tubedogg: This will throw the error ``local variable 'big_file_name' referenced before assignment`` if ``content_filename`` or ``main_file_only`` are set but there isn't one file above the "big" (main) file threshold, or if ``content_filename`` is not set ([line 626](https://github.com/Flexget/Flexget/pull/1790/files/197a4ed67009244f4c1d62fcb652d1b7b319cb20..35106e05fa88e48ca6c5d70012669b3c89bb8245#diff-a9b84b3d1f44f7e89a9560144236437eR626)). I'd add ``big_file_name = ''`` above [line 562](https://github.com/Flexget/Flexget/pull/1790/files/197a4ed67009244f4c1d62fcb652d1b7b319cb20..35106e05fa88e48ca6c5d70012669b3c89bb8245#diff-a9b84b3d1f44f7e89a9560144236437eR562) to ensure it's always assigned to something, and then just do ``if big_file_name`` here.

andocromn:
I've tested as many scenarios as i can think of.  I tested to make sure it doesn't break anything if used in conjunction with content_filename and main_file_only.  It's a bit kludgy because it ends up replacing the top level folder in content_filename with the container_directory, but it works.  Probably would put a note in the wiki recommending against using both, I can't really think of a real world scenario where both would be used 

I also also tested magnet links, there was an issue because the magnet has no files.  I added code to check that the array 'files' is populated and cancel the rename_folder if there are no files.  container_directory can be used with magnet links if convert_magnet: yes is used

liiight:
You guys tested the shit out of this right? Is it ready to be merged?

tubedogg:
Yeah it should be good to go.

andocromn:
Yes, I did a lot of testing.  there are some notes in my last comment about kludgy behaviour when using both container_directory and content_filename, it doesn't cause a problem or any errors.  I could change the way this is handled, but I don't think anyone would need to use both.  Just let me know if you'd like any further updates

PR #1792, opened 2017-04-17
Crossmatch: all_fields option, require all fields to match

cvium:
### Motivation for changes:
I want it
### Detailed changes:

- Added `all_fields` option that enforces all fields to match between two entries


liiight:
Looks good to me

PR #1800, opened 2017-04-19
[add] t411 - add the possibility to obtain results matching a query

endless22:
### Motivation for changes:
- Reduce useless entries 
- Reduce the api response size
- Increase the time available before a result is not longer in the results (100 results for a precise query vs 100 results for a full category)

### Detailed changes:
The api wrapper was already able to request results using a query.
I've just modified the plugin to be able to use the search query feature from the api.

### Config usage:
Added ability to set a query using `search`
```
  task:
    t411:
      search: <query>

```

### DIFF
--- a/flexget/plugins/sites/t411.py
+++ b/flexget/plugins/sites/t411.py
@@ -40,6 +40,7 @@ class T411InputPlugin(object):
       category: <see available categories on "flexget t411 list-cats">
       terms: <see available terms on "flexget t411 list-terms --category <category name>"
       max_resutls: XXX
+      search: <search query>
     """
 
     def __init__(self):
@@ -49,6 +50,7 @@ class T411InputPlugin(object):
                 'category': {'type': 'string'},
                 'terms': one_or_more({'type': 'string'}),
                 'max_results': {'type': 'number', 'default': 100}
+                'search': {'type': 'string'},
             },
             'additionalProperties': False
         }
@@ -64,6 +66,7 @@ class T411InputPlugin(object):
         query.category_name = config.get('category')
         query.term_names = list(config.get('terms', []))
         query.max_results = config.get('max_results')
+        query.expression = config.get('search')
         return query
 
     @plugin.internet(log)
--- a/flexget/plugins/sites/t411.py
+++ b/flexget/plugins/sites/t411.py
@@ -49,8 +49,8 @@ class T411InputPlugin(object):
             'properties': {
                 'category': {'type': 'string'},
                 'terms': one_or_more({'type': 'string'}),
-                'max_results': {'type': 'number', 'default': 100}
-                'search': {'type': 'string'},
+                'max_results': {'type': 'number', 'default': 100},
+                'search': {'type': 'string'}
             },
             'additionalProperties': False
         }


liiight:
@lildadou you wanna review this?

lildadou:
@liiight I will do my best. It's a bit old ...
@endless22 The search plugin is not enough? Can you show a use case that is not feasible as it is?

liiight:
@lildadou there's also two t411 plugins in flexget now, i have no idea why

liiight:
hmm, ok. I guess i'll remove it in the near future

endless22:
### Without query, default max results (100)

```  
task:
    t411:
      category: Série TV
    series:
        - The Expanse
    download: /tmp
```

`Accepted: 0 (Rejected: 0 Undecided: 94 Failed: 0)`

### Without query, max results 10000
```  
task:
    t411:
      category: Série TV
      max_results: 10000
    series:
        - The Expanse
    download: /tmp
```
`Accepted: 12 (Rejected: 5 Undecided: 9976 Failed: 0)`
### With query, default max results (100)
```  
newtask:
    t411:
       search: "The expanse"
    series:
        - The Expanse
    download: /tmp
```

`Accepted: 16 (Rejected: 27 Undecided: 57 Failed: 0)`

I just don't want to search through the whole category to find some specific results.

liiight:
but how do you use this dynamically? with `configure_series` for example

endless22:
Sorry but its a plugin that I have never used.
The documentation say :

> With this you don’t need to maintain series configuration in the configuration file.

I don't really understand what this plugin is for
 
Why did i need to use `configure_series` when i can use `series` ?
Can you show me an example of what you mean by using `configure_series`.
Can you show me an example of what my modification is not able to do that the plugin could already do.

liiight:
It's documentation should explain it, use a source like Trakt to configure your series. 
And the correct approach would be to look for relevant entry fields inside the search plugin probably.

endless22:
From the documentation : 
>This plugin can be used as a search or an input plugin for getting results from Torrent411.

I just want to add more options to allow the plugin to refine entries produced.
Its an input plugin and its always better if you can customize this input.
Actually without this modification its impossible to get the results you want without increase `max_results` to some ridiculous quantity.
I agree that its maybe not the best way to do it but alternatives ways are complicated and involved modification/usage of other plugins. Its not the best way but it's the simplest.

liiight:
It's not that it isn't the best way, it's a way that doesn't fit with how flexget does things at all, or more specifically, how series plugin works and how search sites should get more data. If you want this, change it to pass the data from the entry itself.

lildadou:
As @liiight explained, your approach runs counter to the flexget design.
I have provided 2 plugins for t411: one of type input and another one of type search. The first one is the one you use. It allows to discover the latest additions on the tracker and that is why it does not propose a search by expression. The second, as its name suggests, makes it possible to look for something specific.

On the documentation of the plugin, I illustrate with cases of use for each type of plugin: the input type allows to recover movies that pass criteria against IMDB while the search makes it possible to recover the series that you Look.

To search for new episodes, the search plugin is much more powerful than your approach. First, it will add the number of the episode to be searched (for t411, it uses season and episode filters). Secondly, it will only start looking for an episode if it has been aired!

PR #1811, opened 2017-05-03
 Multiples changes for t411 domain name change : from .ai to .al

poulpito:
According to t411 website : 
Il faudra donc tout faire passer du .ai au .al
L'adresse du site est https://www.t411.al/  (à mettre en favoris) 

Commit modify both plugins : api_t411 (+tests) and torrent411

### DIFF
--- a/flexget/plugins/internal/api_t411.py
+++ b/flexget/plugins/internal/api_t411.py
@@ -120,7 +120,7 @@ class FriendlySearchQuery(object):
         self.term_names.append("Episode %02d" % episode)
 
 
-T411API_DOMAIN_URL = "api.t411.ai"
+T411API_DOMAIN_URL = "api.t411.al"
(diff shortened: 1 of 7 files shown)


liiight:
Torrent411 is deprecated and will be removed in next minor release, so there's little point to fixing it...

poulpito:
Could you dispatch commit file or do i need to push a new PR without the old plugin ? (still in use btw)

liiight:
I really don't see a reason to have both of this plugins co exist. Is there functionality missing from the new plugin?

ProphecyKt:
@liiight i already answer to that question in my last pull request #1777.
The new plugins doesn't work properly and it's incomplete (exemple : differents categories with the same name)
The ancient plugin is just an urlrewritter but it work very well. It's simple.

liiight:
Right, forgot about that.

PR #1815, opened 2017-05-07
Rewrite Dockerfile #1542. Based on python:2.7-alpine

### LINKED ISSUE #1542: Docker support broken
Docker support is broken, this issue is a reminder for steve to fix it.

qk4l:
### Motivation for changes:

Have docker image from official git repo.

### Detailed changes:

- Based image on  python:2.7-alpine

### Addressed issues:

- Fixes #1542 

#### To Do:

1. The owner of flexget/flexget repo should to [create automated build](https://docs.docker.com/docker-hub/builds/) on hub.docker.com
2. Update _README.md_ to point to a new image name


qk4l:
Actually automated build is what we need to resolve #1542 . But it should be created by user who have permissions to git repo.

As a good option will be several tags based on branch
- develop > develop
- master > latest

liiight:
Thanks for this. 
I should look into docker-hub, I'm less than a noob with it currently.

liiight:
Merging this is the basis for creating an automated build right?

jawilson:
Been wanting a functioning Docker container for Flexget for some time. Thanks for this! I don't have access to the Flexget dockerhub account, but I'll test/verify this.

@liiight it's so that with the setup @qk4l mentioned above, docker hub automatically builds containers that are available to all at https://hub.docker.com/r/flexget/flexget/. I don't have the credentials to the dockerhub account, but I have a decent amount of experience with this sort of thing, so I could definitely help set it up if need be.

liiight:
Hmm @stevezau, did you setup that account?

qk4l:
> Merging this is the basis for creating an automated build right?

@liiight  Nope.

_Automated build_  required Docker Hub account with link to GitHub account and repo with Dockerfile.

When you create a docker repository with automated build option you choose a git repo from your Github`s repos  and then builder build an image  based _Dockerfile_ in root of repo.

README.md file also is used and it looks like this - https://hub.docker.com/r/qk4l/flexget/

Further docker hub starts to monitor the repo and will rebuild image if changes came. 

In this particular PR I slightly improve  _Dockerfile_:
* It based on smaller image with exactly python 2.7. Who knows what python is in alpine image now? =)
* CMD option allows to run docker image without additional options (in daemon mode). User of course can  redefine them.

liiight:
ok, i'm actually more familiar with this topic now, so some comments and questions:

1. I think it'll be preferable to use the highest supported version of python possible with this (currently that being 3.5.3).
2. I think it's best to keep log level defaults.
3. It'll be great to have several build types I think: master and develop branch, daemon and "regular" for each. Develop and master branches will install different dependencies.
4. I'm not sure if `.dockerignore` is up to date.
5. When i'll set up an automated build, it'll rely on this `Dockerfile` right? This won't go to waste or anything?

cvium:
1. Breaks compatibility with Deluge

liiight:
Right. I guess 2.7 is the best global match. We could go for both version
though, dunno if that's important though

cvium:
Latest Py2 and latest Py3 makes sense.

liiight:
Also, having a container that doesnt have a base `config.yml` is kinda weird. it'll just fail when trying to run it. I know you can create and copy into it, but that flow feels a bit off...

cvium:
This doesn't look right

qk4l:
I'm sorry. Just rebase my fork via Pycharm. Something gone wrong. (

PR #1817, opened 2017-05-10
Refactor config hash management

cvium:
### Motivation for changes:
It's ugly
### Detailed changes:

- Removed specific plugins hash table, hash table check is now global per task
- Config is checked after start phase so all changes on that phase can cause a new hash to be created 

### TODO
- [x] Think of a better solution that doesn't break shit...
- [x] Fix issue with remember rejected plugin 
- [x] Test all series since it changes config on metainfo phase 
- [x] Add hash specifc checks (include, configure_series and template plugins)

### DIFF
--- a/flexget/task.py
+++ b/flexget/task.py
@@ -545,8 +545,8 @@ class Task(object):
         try:
             merge_dict_from_to(new_config, self.config)
             self.check_config_hash()
-        except MergeException:
-            raise PluginError('Failed to merge include file to task %s, incompatible datatypes' % self.name)
+        except MergeException as e:
+            raise PluginError('Failed to merge configs for task %s: %s' % (self.name, e))
 
     def check_config_hash(self):
         """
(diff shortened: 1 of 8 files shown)


flexget/task.py
```diff
@@ -541,31 +541,33 @@ def config_changed(self):
...
 
-    def is_config_modified(self, last_hash):
-        """
-        Checks the task's config hash. Returns True/False depending on config has been modified and the config hash
+    def merge_config(self, new_config):
+        try:
+            merge_dict_from_to(new_config, self.config)
+            self.check_config_hash()
+        except MergeException:
+            raise PluginError('Failed to merge include file to task %s, incompatible datatypes' % self.name)
```
liiight: Copy pasta

cvium: ?

liiight: Include file

cvium:
This solution doesn't really allow detailed control, but if you can think of something...

PR #1824, opened 2017-05-16
New input plugin: trakt_calendar

cvium:
### Motivation for changes:
Dunno

```
  trakt_calendar:
    account: myaccount
    start_day: -7  # number of days from today
    days: 10
    strip_dates: yes
    type: episodes
```

### TODO
- [x] Test with `what` in discover plugin
~- [ ] Structural pytest?~
- [x] Make sure a "you missed an episode X days ago" task is feasible

### Example "you missed an episode X days ago" task
```YAML
tasks:
  notify_missed:
    disable: seen  # we do not want seen to interfere
    metainfo_series: yes
    notify: bla bla
    trakt_calendar:
      type: episodes
      account: asdas
      start_day: -7
      days: 6  # Grab episodes in the last 6 days (not including today)
    if:
      - "trakt_collected or trakt_watched": reject
      - "series_season == 0": reject
    trakt_lookup:
      account: dsadsa
    crossmatch:
      from:
        - trakt_list:
            account: dsadsa
            type: shows
            list: ddsadasdas
      action: accept
      fields: [trakt_series_name]
```

### DIFF
--- a/flexget/config_schema.py
+++ b/flexget/config_schema.py
@@ -12,6 +12,7 @@ from datetime import datetime
 
 import jsonschema
 from jsonschema.compat import str_types, int_types
+from dateutil.parser import parse as dateutil_parse
 
 from flexget.event import fire_event
 from flexget.utils import qualities, template
@@ -159,6 +160,12 @@ def parse_size(size_input):
         return int(1024 ** prefixes.index(unit) * value)
 
 
+def parse_date(date_input):
+    try:
+        return dateutil_parse(date_input)
+    except ValueError:
+        raise ValueError("should be in ISO 8601 format eg. YYYY-MM-DD")
+
 # Public API end here, the rest should not be used outside this module
 
 
@@ -268,6 +275,13 @@ def is_valid_template(instance):
     if not isinstance(instance, str_types):
         return True
     return get_template(instance) is not None
+
+
+@format_checker.checks('date', raises=ValueError)
+def is_valid_date(date_string):
+    if not isinstance(date_string, str_types):
+        return True
+    return parse_date(date_string)
 
 
 def set_error_message(error):
--- a/flexget/config_schema.py
+++ b/flexget/config_schema.py
@@ -12,7 +12,6 @@ from datetime import datetime
 
 import jsonschema
 from jsonschema.compat import str_types, int_types
-from dateutil.parser import parse as dateutil_parse
 
 from flexget.event import fire_event
 from flexget.utils import qualities, template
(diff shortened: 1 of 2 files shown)


flexget/config_schema.py
```diff
@@ -270,6 +277,13 @@ def is_valid_template(instance):
     return get_template(instance) is not None
 
 
+@format_checker.checks('date', raises=ValueError)
```
liiight: Json schema had a built in `date-time` format, can't we use that?

cvium: It doesn't conform to the same standard as dateutil, which means it validates a format that dateutil can't parse...

mfonville:
Thanks for working on this; I tried to get it working but already noticed the following things:
I had to set `"start_date: 2017-01-01"` with the `"` otherwise it is not interpreted as a string

Also I got:
```
Series `The Simpsons S28E22` with identified_by value `auto` are not supported.
```
which I had to solve by adding `identified_by: ep` but this is something I'd expect to be set by the plugin itself.

Also `days` seems to look ahead in time (so the next 7 days) instead of back in time, is that on purpose?

cvium:
> Also days seems to look ahead in time (so the next 7 days) instead of back in time, is that on purpose?

That's the API. I'm not sure if it supports negative numbers.

> which I had to solve by adding identified_by: ep but this is something I'd expect to be set by the plugin itself.

I don't think any other plugin automatically sets this...

liiight:
>Also days seems to look ahead in time (so the next 7 days) instead of back in time, is that on purpose?

The idea is to get shows with an upcoming broadcast date. @cvium maybe the default needs to be today?

mfonville:
> That's the API. I'm not sure if it supports negative numbers.
Ok, I understand. It is a bit counter-intuitive for users, so it will be important to document this in an accessible way.
It would make sense to then e.g. make the days to be '1' by default.

> I don't think any other plugin automatically sets this...
hmm, but it fails with `auto`, and calendar can only provides episode. imho it would make sense that the plugin at least sets the config-parts that it cannot do without and that are static.

tubedogg:
It's too bad Trakt doesn't offer through the API the [custom calendars](https://blog.trakt.tv/custom-calendars-9c6a8a224f01) that VIP members can create. It basically takes a list and makes a custom calendar that only includes shows on that list, instead of using watchlist/watched/collected.

That said, will this work with `crossmatch` to be able to kind of simulate that? I'd basically want to pull a list from Trakt and be able to use crossmatch to only return entries for episodes that match a series on that list.

cvium:
It should.

PR #1828, opened 2017-05-18
[fix] next_series_seasons - Limit size of emit in one run

tubedogg:
### Motivation for changes:
Given a season identifier of S2014 and backfill enabled, it is possible for `next_series_seasons` to emit all the way down to S01 in one run.

### Detailed changes:
This change prevents it from emitting if `begin` is not set and the group of seasons to review is more than 15, or if `begin` is set and the difference between the latest season to review and `begin` is more than 30. In both cases the user will be required to set or adjust `begin` before any emit will occur.

### Addressed issues:
- Fixes issue originally raised by @jawilson [here](https://github.com/Flexget/Flexget/commit/30751df7da16bdc12781411cdec9f51d9e02a890#commitcomment-22161686).

### DIFF
--- a/flexget/plugins/input/next_series_seasons.py
+++ b/flexget/plugins/input/next_series_seasons.py
@@ -14,6 +14,8 @@ from flexget.plugins.filter.series import get_latest_episode_release
 plugin_name = 'next_series_seasons'
 log = logging.getLogger(plugin_name)
 
+MAX_SEASON_DIFF_WITHOUT_BEGIN = 15
+MAX_SEASON_DIFF_WITH_BEGIN = 30
 
 class NextSeriesSeasons(object):
     """
@@ -71,8 +73,6 @@ class NextSeriesSeasons(object):
         if isinstance(config, bool):
             config = {}
 
-        DIFF_NO_SERIES_BEGIN = 15
-        DIFF_GAP_TOO_BIG = 30
         if task.is_rerun:
             # Just return calculated next eps on reruns
             entries = self.rerun_entries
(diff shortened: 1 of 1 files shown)


flexget/plugins/input/next_series_seasons.py
```diff
@@ -115,6 +115,19 @@ def on_task_input(self, task, config):
                 else:
                     latest_season = low_season + 1
 
+                if (latest_season - low_season > 15 and not series.begin) or (series.begin and
```
liiight: Maybe move these values to consts?

flexget/plugins/input/next_series_seasons.py
```diff
@@ -115,6 +115,19 @@ def on_task_input(self, task, config):
...
+                        log.error('Series `%s` has a begin episode set (`%s`), but the season currently being processed '
+                                  '(%s) is %s seasons later than it. To prevent emitting incorrect seasons, this ' 
+                                  'series will not emit unless the begin episode is adjusted to a season that is less '
+                                  'than 30 seasons from season %s.', series.name, series.begin.identifier, latest_season,
+                                  (latest_season - series.begin.season), latest_season)
+                    else:
+                        log.error('Series `%s` does not have a begin episode set and continuing this task would result '                                   'in more than 15 seasons being emitted. To prevent emitting incorrect seasons, this '
+                                  'series will not emit unless the begin episode is set in your series config or by '
+                                  'using the CLI subcommand `series begin "%s" <SxxExx>`.', series.name, series.name)
+                    continue
```
liiight: So if season is 2017 the user will get 2005 error messages in the log?

tubedogg: No, this is outside the loop that counts through the seasons. This should only happen once per series.

flexget/plugins/input/next_series_seasons.py
```diff
@@ -71,6 +71,8 @@ def on_task_input(self, task, config):
         if isinstance(config, bool):
             config = {}
 
+        DIFF_NO_SERIES_BEGIN = 15
```
liiight: These should ideally go at the beginning of the module, outside if the plugin class. The idea is that constant values like these can be easily viewed when editing the file and kit hidden internally. 
Also I'm not too wild about their names but can't think of a better idea...

PR #1836, opened 2017-05-23
[add] series CLI - --sort-by argument for series show

tubedogg:
### Motivation for changes:
I want to sort by entity ID.

### Detailed changes:
When using the `series show` CLI subcommand, you can now specify `--sort-by identifier` to sort by entity ID (aka season/episode) or `--sort-by age` for last seen date (default), and `--descending` or `--ascending` to specify the sort order.
`get_all_entities` now supports return entities sorted by `identifier` or age (default).
Renamed 'Entity ID' and 'Latest age' columns in `series show` output to 'Identifier' and 'Last seen', respectively.

### DIFF
--- a/flexget/plugins/cli/series.py
+++ b/flexget/plugins/cli/series.py
@@ -202,7 +202,7 @@ def display_details(options):
                        ' {}'.format(name, ', '.join(s.name for s in matches[1:])))
             if not options.table_type == 'porcelain':
                 console(warning)
-        header = ['Entity ID', 'Latest age', 'Release titles', 'Release Quality', 'Proper']
+        header = ['Identifier', 'Last seen', 'Release titles', 'Release Quality', 'Proper']
         table_data = [header]
         entities = get_all_entities(series, session=session, sort_by=sort_by, reverse=reverse)
         for entity in entities:
@@ -241,9 +241,9 @@ def display_details(options):
                        ' during this time.')
         else:
             footer += '\n Series uses `%s` mode to identify episode numbering (identified_by).' % series.identified_by
-        footer += ' \n See option `identified_by` for more information.\n'
+        footer += ' \n See option `identified_by` for more information.'
         if series.begin:
-            footer += ' Begin episode for this series set to `%s`.' % series.begin.identifier
+            footer += ' \n Begin episode for this series set to `%s`.' % series.begin.identifier
     try:
         table = TerminalTable(options.table_type, table_data, table_title, drop_columns=[4, 3, 1])
         console(table.output)
@@ -286,8 +286,8 @@ def register_parser_arguments():
 
     show_parser = subparsers.add_parser('show', parents=[series_parser, table_parser],
                           help='Show the releases FlexGet has seen for a given series')
-    show_parser.add_argument('--sort-by', choices=('seen', 'entity'), default='seen',
-                             help='Choose releases list sort: last seen date (default) or entity ID')
+    show_parser.add_argument('--sort-by', choices=('age', 'identifier'), default='age',
+                             help='Choose releases list sort: age (default) or identifier')
     show_order = show_parser.add_mutually_exclusive_group(required=False)
     show_order.add_argument('--descending', dest='order', action='store_true', help='Sort in descending order')
     show_order.add_argument('--ascending', dest='order', action='store_false', help='Sort in ascending order')
--- a/flexget/plugins/cli/series.py
+++ b/flexget/plugins/cli/series.py
@@ -183,6 +183,8 @@ def get_latest_status(episode):
 def display_details(options):
     """Display detailed series information, ie. series show NAME"""
     name = options.series_name
+    sort_by = options.sort_by
+    reverse = options.order
     with Session() as session:
         name = normalize_series_name(name)
         # Sort by length of name, so that partial matches always show shortest matching title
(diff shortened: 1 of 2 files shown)


flexget/plugins/cli/series.py
```diff
@@ -282,8 +284,14 @@ def register_parser_arguments():
...
-    subparsers.add_parser('show', parents=[series_parser, table_parser],
-                          help='Show the releases FlexGet has seen for a given series ')
+    show_parser = subparsers.add_parser('show', parents=[series_parser, table_parser],
+                          help='Show the releases FlexGet has seen for a given series')
+    show_parser.add_argument('--sort-by', choices=('seen', 'entity'), default='seen',
+                             help='Choose releases list sort: last seen date (default) or entity ID')
+    show_order = show_parser.add_mutually_exclusive_group(required=False)
+    show_order.add_argument('--descending', dest='order', action='store_true', help='Sort in descending order')
+    show_order.add_argument('--ascending', dest='order', action='store_false', help='Sort in ascending order')
+
```
liiight: I don't remember what happens if you don't use this group at all. What's the default?

tubedogg: The default is False (ascending) as set in filter/series.py. But your comment got me thinking, if it's left off, what does `options.order` end up being? It worked fine in my testing though.

flexget/plugins/cli/series.py
```diff
@@ -282,8 +284,14 @@ def register_parser_arguments():
     order.add_argument('--descending', dest='order', action='store_true', help='Sort in descending order')
     order.add_argument('--ascending', dest='order', action='store_false', help='Sort in ascending order')
 
-    subparsers.add_parser('show', parents=[series_parser, table_parser],
-                          help='Show the releases FlexGet has seen for a given series ')
+    show_parser = subparsers.add_parser('show', parents=[series_parser, table_parser],
+                          help='Show the releases FlexGet has seen for a given series')
+    show_parser.add_argument('--sort-by', choices=('seen', 'entity'), default='seen',
```
tubedogg: `age` is a good idea. I was trying to avoid using underscored suffixes in an attribute as I think it makes it easier to remember and use, but I can change it if you feel strongly about it.

liiight: You're right about underscore. Is `identifier` too much of an internal lingo?

flexget/plugins/cli/series.py
tubedogg: `store_true` stores a default of `False` (and the inverse for `store_false`). If neither is specified on the command line, `order` is given the default value of the first `add_argument` call.

```
import argparse
parser = argparse.ArgumentParser()
parser.add_argument('--sort-by', choices=('age', 'entity'), default='age')
order = parser.add_mutually_exclusive_group(required=False)
order.add_argument('--ascending', dest='order', action='store_false')
order.add_argument('--descending', dest='order', action='store_true')

>>> parser.parse_args('--sort-by age'.split())
Namespace(order=True, sort_by='age')
```

liiight: If i remember correctly, setting a store boolean type of action sets the default to the opposite action. So if your action is `store_true` the default will be `False`. 
However, since you've got two action on the same dest, the last one added takes precedence, and in this case sets `True` as the default.

flexget/plugins/cli/series.py
tubedogg: I think it might be - plus the column is headed as "Entity ID" in the resulting table, so I think `entity` or `entityid` makes it simpler.

tubedogg: Though I suppose "entity" is pretty internal, too. Maybe we change the column heading to "Identifier" and make the option the same?

liiight: How about `show-order`?

flexget/plugins/cli/series.py
tubedogg: The first one added takes precedence. I echoed it on the command line without specifying the argument and it was `False`. I then tried adding a third option, so the order was `store_true`, `store_false`, `store_true`, and it was still `False`.

liiight: Cool, good to know

flexget/plugins/cli/series.py
tubedogg: Any strong feelings on how to proceed? I think I like `identifier` or `series` best, and change the name of the column to "Identifier" or "Series ID", respectively.

flexget/plugins/filter/series.py
```diff
@@ -1136,10 +1136,13 @@ def show_seasons(series, start=None, stop=None, count=False, descending=False, s
     return seasons.slice(start, stop).from_self().all()
 
 
-def get_all_entities(series, session):
+def get_all_entities(series, session, sort_by='seen', reverse=False):
     episodes = show_episodes(series, session=session)
     seasons = show_seasons(series, session=session)
-    return sorted(episodes + seasons, key=lambda e: (e.first_seen or datetime.min, e.identifier))
+    if sort_by == 'entity':
```
liiight: Maybe something like this:
```python
    if sort_by == 'entity':
        key=lambda e: e.identifier
    else:
        key=lambda e: e.first_seen or datetime.min, e.identifier
    return sorted(episodes + seasons, key=key, reverse=reverse)
```

tubedogg: Addressed in 811bc8a2acb27c630396cd9d127f24b1499aacf0.

liiight:
This is done right?

tubedogg:
Yes sir.

PR #1838, opened 2017-05-23
[fix] newpct - Detect correct IDs

rsalas:
Take all the IDs of the content on the page but the valid are those are within "parametros".

### Detailed changes:

- Update the regular expression that detects the IDs.

[RegEx Test](https://regex101.com/r/vaGfdf/1)

Replace: (?:parametros\s*=\s*.*?) for check match.

### DIFF
--- a/flexget/plugins/sites/newpct.py
+++ b/flexget/plugins/sites/newpct.py
@@ -57,9 +57,7 @@ class UrlRewriteNewPCT(object):
         except requests.exceptions.RequestException as e:
             raise UrlRewritingError(e)
         try:
-            html = page.text
-            html = html.replace('\n', '').replace('\r', '')
-            soup = get_soup(html)
+            soup = get_soup(page.text)
         except Exception as e:
             raise UrlRewritingError(e)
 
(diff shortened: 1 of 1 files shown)


Aokromes:
This fixes the "- maybe a login page?)" thing i have noticed on the last day?

liiight:
If someone wanna confirm this fix ill merge it

rsalas:
@Aokromes No, it's for download pages.
It does not correctly detect the torrent ID and does not give a permanent redirect error.

malkavi:
We need to convert all "\n" and "\r" from a webpage? In my opinion is excessive.

I have fixed the error just by changing this line:
` torrent_id_prog = re.compile(",\s*'(?:torrentID|id)'\s*:\s*'(\d+)'")`
to avoid the 'id' from the ad.

But that solution also will work.
About the login page error, it's related to a false torrent caused by taking the 'id' from an ad. This PR or my fix will fix the error.

rsalas:
@malkavi I have tried your solution but can not find any ID.
If you check your solution with the link above, you will see that do not work.

In my example link the html is in one line, beause the "\n" is literal not is a new line
Need replace all line break, for parse multiple lines in one regular expression.
It is possible that I am confused, correct me if I am wrong.

malkavi:
Please try with non edited/replaced webpage. Copy and Paste a new one using control+u.
Example: https://regex101.com/r/vaGfdf/2

Maybe better this way? 
https://regex101.com/r/vaGfdf/4
or too complicated?
That way we can be sure that the id is from the struct 'parametros'.

rsalas:
@malkavi The second regular expression I like, because if the ad has a "," previous will take this id, with "parametros" is more safely.

I change the PR.

malkavi:
It's OK for me. I have tested this PR successfully. Added a new serie and downloaded the torrent.

Test:  "Los Pasajeros Del Tiempo"

`FlexGet has just downloaded 1 new entries for task mis_series:
- Los Pasajeros Del Tiempo - Temporada 1 [HDTV][Cap.101][Español Castellano] (http://www.newpct.com/torrents/091872.torrent)`

PR #1841, opened 2017-05-25
[change] series begin - SXX accepted as ID / CLI --forget option

tubedogg:
### Motivation for changes:
Be able to use `S02` as a `begin` identifier, and also forget a `begin` episode link without removing the history of the episode itself.

### Detailed changes:
Two changes relating to `begin`:
- Behind the scenes, appends ‘E01’ when an identifier of the type SXX is used as the begin episode. This can be done via CLI `series begin SXX`, config with series plugin, or API. To make this work, an additional config format validator of `episode_or_season_identifier` was added which calls `parse_episode_identifier` with the new parameter `identify_season` set to `True`. When this is not wanted, continue to use the existing `episode_identifier` config format validator.
- CLI `series begin <show> --forget` argument to remove begin episode association in series table. 

- feathub #[18](http://feathub.com/Flexget/Flexget/+18)

When using an identifier of the format `SXX` while setting begin via CLI:
```
`S05` was identified as a season.
Releases for `My Show` will be accepted starting with `S05E01`
```

When using the `--forget` argument:
```
The begin episode for `My Show` has been forgotten.
```

### DIFF
--- a/flexget/config_schema.py
+++ b/flexget/config_schema.py
@@ -263,8 +263,8 @@ def is_episode_identifier(instance):
     return parse_episode_identifier(instance) is not None
 
 
-@format_checker.checks('episode_or_season_identifier', raises=ValueError)
-def is_episode_or_season_identifier(instance):
+@format_checker.checks('episode_or_season_id', raises=ValueError)
+def is_episode_or_season_id(instance):
     if not isinstance(instance, (str_types, int)):
         return True
     return parse_episode_identifier(instance, identify_season=True) is not None
--- a/flexget/config_schema.py
+++ b/flexget/config_schema.py
@@ -263,6 +263,13 @@ def is_episode_identifier(instance):
     return parse_episode_identifier(instance) is not None
 
 
+@format_checker.checks('episode_or_season_identifier', raises=ValueError)
+def is_episode_or_season_identifier(instance):
+    if not isinstance(instance, (str_types, int)):
+        return True
+    return parse_episode_identifier(instance, identify_season=True) is not None
+
+
 @format_checker.checks('file_template', raises=ValueError)
 def is_valid_template(instance):
     if not isinstance(instance, str_types):
--- a/flexget/plugins/cli/series.py
+++ b/flexget/plugins/cli/series.py
@@ -2,7 +2,6 @@ from __future__ import unicode_literals, division, absolute_import
 from builtins import *  # noqa pylint: disable=unused-import, redefined-builtin
 
 import argparse
-import re
 from datetime import timedelta
 
 from colorclass.toggles import disable_all_colors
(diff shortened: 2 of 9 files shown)


flexget/plugins/filter/series.py
```diff
@@ -1000,6 +1002,13 @@ def set_series_begin(series, ep_id):
         session.flush()
     series.begin = episode
 
+def remove_series_begin(series):
```
liiight: This is completely redundant, just set to `None` where needed

flexget/config_schema.py
```diff
@@ -263,6 +263,13 @@ def is_episode_identifier(instance):
     return parse_episode_identifier(instance) is not None
 
 
+@format_checker.checks('episode_or_season_identifier', raises=ValueError)
+def is_episode_or_season_identifier(instance):
```
liiight: This is a terribly long name

flexget/plugins/filter/series.py
```diff
@@ -972,9 +972,11 @@ def set_series_begin(series, ep_id):
     # If identified_by is not explicitly specified, auto-detect it based on begin identifier
     # TODO: use some method of series parser to do the identifier parsing
     session = Session.object_session(series)
-    identified_by = parse_episode_identifier(ep_id)
+    identified_by = parse_episode_identifier(ep_id, identify_season=True)
     if identified_by == 'ep':
         ep_id = ep_id.upper()
+        if 'E' not in ep_id:
```
tubedogg: None of those would have gotten past `parse_episode_identifier`, which only recognizes SXXEXX (and now SXX).

tubedogg: That said I did change it to look for an `entity_type` of `season` returned from `parse_episode_identifier` instead of parsing the string directly, though that function still only recognizes SXXEXX or SXX.

flexget/utils/tools.py
```diff
@@ -492,6 +492,8 @@ def parse_episode_identifier(ep_id):
         identified_by = 'sequence'
     elif re.match(r'(?i)^S\d{1,4}E\d{1,3}$', ep_id):
         identified_by = 'ep'
+    elif re.match(r'(?i)^S\d{1,4}$', ep_id) and identify_season:
```
tubedogg: Addressed in 0497473bc9e334d7e7179afe33bdd4ed89d85b19.

flexget/plugins/cli/series.py
```diff
@@ -119,25 +120,38 @@ def display_summary(options):
...
+                series.name = series_name
+                session.add(series)
+            else:
+                series = series[0]
+            try:
+                set_series_begin(series, ep_id)
+            except ValueError as e:
+                console(e)
+            else:
+                if re.match(r'(?i)^S\d{1,4}$', ep_id):
```
liiight: Change this to take type as well

tubedogg: Okay, this should be addressed in 1240b73d0da475d3a02c050c0d73ce07f503b377.

tubedogg:
This should be ready to go.

PR #1842, opened 2017-05-25
[add] parser - Support for DD+ audio and more channels

cvium:
### Motivation for changes:
DD+ audio is becoming more common. It stands for Dolby Digital Plus aka Enhanced AC-3 and it is therefore better than both DD and AC-3. It supports up to 7.1 channels and a more efficient compression.
### Detailed changes:

- Added dd+/ddp to qualities.py
- Added tests for series and quality filter
- Added support for parsing x.{0,1} audio eg. 7.1, 5.0 etc.

### DIFF
--- a/flexget/tests/test_qualities.py
+++ b/flexget/tests/test_qualities.py
@@ -100,7 +100,10 @@ class TestQualityParser(object):
         ('Test.File.DTSHD', 'dtshd'),
         ('Test.File.DTS', 'dts'),
         ('Test.File.truehd', 'truehd'),
-        ('Test.File.DTSHDMA', 'dtshd')
+        ('Test.File.DTSHDMA', 'dtshd'),
+        ('Test.File.DD+5.1', 'dd+5.1'),
+        ('Test.File.DDP5.1', 'dd+5.1'),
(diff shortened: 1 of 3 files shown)


flexget/tests/test_series.py
```diff
@@ -2317,3 +2317,65 @@ def test_advanced_config(self, execute_task):
...
+          series:
+            - Channels:
+                quality: dd+5.1
+
+
+    """
+
+    @pytest.fixture()
+    def config(self):
+        """Overrides outer config fixture since season pack support does not work with guessit parser"""
```
liiight: Season_pack

flexget/tests/test_qualities.py
```diff
@@ -109,6 +111,23 @@ def test_quality_failures(self, parser, test_quality):
         ))
 
 
+class TestQualityInternalParser(object):
+    @pytest.mark.parametrize("test_quality", [
```
liiight: you can set it a two vars instead of one that'll create a tuple, might be a little easier to read imo

cvium: Just following the same format as the other tests.

PR #1850, opened 2017-05-29
[add] series CLI - use environment variables for defaults

tubedogg:
### Motivation for changes:
[Feature request](https://github.com/Flexget/Flexget/commit/dba8b0b37dc87316c67c754e98c491c2c015b920#commitcomment-22299675)

### Detailed changes:
`series show` and `series list` CLI subcommands now read environment variables to determine default parameters. These can be overridden on a per-run basis by utilizing the existing command-line arguments.

- `flexget series show`:
  - `FLEXGET_SERIES_SHOW_SORTBY_FIELD` can be set to `age` (default) or `identifier` to determine which field releases are sorted by. (`--sort-by` optional argument)
  - `FLEXGET_SERIES_SHOW_SORTBY_ORDER` can be set to `desc` to display results in descending order. (`--descending`/`--ascending` optional arguments)

- `flexget series list`:
  - `FLEXGET_SERIES_LIST_CONFIGURED` can be set to `configured` (default), `unconfigured`, or `all` to determine whether to only show series which are currently in the config. (positional argument)
  - `FLEXGET_SERIES_LIST_PREMIERES` can be set to `yes` to only show series which only have episode 1, or 1 and 2, downloaded. (`--premieres` optional argument)
  - `FLEXGET_SERIES_LIST_STATUS` can be set to `new` to only show series that have seen a release in the last 7 days, or `stale` to only show series that haven't seen a release in the last 365 days. (`--new`/`--stale` optional arguments)
  - `FLEXGET_SERIES_LIST_NUMDAYS` can be set to the number of days that `STATUS` limits to. (the `DAYS` option in `--new DAYS`/`--stale DAYS`)
  - `FLEXGET_SERIES_LIST_SORTBY_FIELD` can be set to `name` (default) or `age` to determine which field series are sorted by. (`--sort-by` optional argument)
  - `FLEXGET_SERIES_LIST_SORTBY_ORDER` can be set to `desc` to display results in descending order. (`--descending`/`--ascending` optional arguments)

Note there are presently bugs with the output of `series list --premieres` and `series list --sort-by age`, which already existed and are unrelated to these changes. Specifically, `series list --premieres` doesn't take season packs into account when determining which series to display, and `series list --sort-by age` seems to sort-of-but-not-completely sort by age. I'll be looking at these separately and submit another PR if I can resolve them.

### DIFF
--- a/flexget/plugins/cli/series.py
+++ b/flexget/plugins/cli/series.py
@@ -218,6 +218,8 @@ def display_details(options):
         table_data = [header]
         entities = get_all_entities(series, session=session, sort_by=sort_by, reverse=reverse)
         for entity in entities:
+            if not entity.releases:
+                continue
             if entity.identifier is None:
                 identifier = colorize(ERROR_COLOR, 'MISSING')
                 age = ''
--- a/flexget/plugins/cli/series.py
+++ b/flexget/plugins/cli/series.py
@@ -2,6 +2,7 @@ from __future__ import unicode_literals, division, absolute_import
 from builtins import *  # noqa pylint: disable=unused-import, redefined-builtin
 
 import argparse
+import os
 from datetime import timedelta
 
 from colorclass.toggles import disable_all_colors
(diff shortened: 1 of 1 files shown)


liiight:
Maybe move var names to consts? So you'd see them all at once?

tubedogg:
Addressed in 97acac7b6cb1a789ba06ddf712467d3c83b18730.

flexget/plugins/cli/series.py
```diff
@@ -270,29 +292,29 @@ def register_parser_arguments():
     list_parser = subparsers.add_parser('list', parents=[table_parser],
                                         help='List a summary of the different series being tracked')
     list_parser.add_argument('configured', nargs='?', choices=['configured', 'unconfigured', 'all'],
-                             default='configured',
                              help='Limit list to series that are currently in the config or not (default: %(default)s)')
-    list_parser.add_argument('--premieres', action='store_true',
+    list_parser.add_argument('--premieres', action='store_const', const='premieres',
```
liiight: why change this?

flexget/plugins/cli/series.py
```diff
@@ -270,29 +292,29 @@ def register_parser_arguments():
...
 
     show_parser = subparsers.add_parser('show', parents=[series_parser, table_parser],
                           help='Show the releases FlexGet has seen for a given series')
-    show_parser.add_argument('--sort-by', choices=('age', 'identifier'), default='age',
+    show_parser.add_argument('--sort-by', choices=('age', 'identifier'),
                              help='Choose releases list sort: age (default) or identifier')
     show_order = show_parser.add_mutually_exclusive_group(required=False)
-    show_order.add_argument('--descending', dest='order', action='store_true', help='Sort in descending order')
-    show_order.add_argument('--ascending', dest='order', action='store_false', help='Sort in ascending order')
+    show_order.add_argument('--descending', dest='order', action='store_const', const='desc',
```
liiight: you could probably reuse these args if you really really wanted to :-)
you can set a parent parser and inherit it, but it's really a minor issue.

flexget/plugins/cli/series.py
tubedogg: That's a good question. I think it was about eliminating the default `False` that `store_true` creates when the flag isn't used? Because for the `--descending`/`--ascending` flags I need to know if they were present to override envs (since both `--ascending` being present or `--descending` not being present result in a value of `False`, so no way to tell which is the case just based on the value). In this case, since there's only one, `False` would mean it's not present, so that would be fine. I can change it back.

liiight: it's fine by me, i was just curious

y:
Would it be possible to set these environment variables directly from the config? That would make it easier for users since they don't have to split logic between config and shell files and easier for devs to track down problems. You could add a new item to the issues template like "show the output of `env | grep FLEXGET`, but that wouldn't show anything for users who set the variables directly in their crontab.

liiight:
currently, no cli plugin even have access to the config. i'm not saying i'm necessarily against it, but making this change work generically and correctly is a large task IMO, and i'm not sure the added value is worth it. maybe someone else feels different though

PR #1854, opened 2017-05-31
[WIP] React Web UI

kristenmills:
Starting a WIP pull request for better visibility. 

### Motivation for changes
Nobody wants to write Angular 1. Now they don't have to! 💯 

### Detailed changes:
- Rewrite existing web ui using modern technologies (ES2015+, React, Redux, Webpack)

### To Do:
https://github.com/Flexget/Flexget/projects/4

### Wiki
https://flexget.com/Web-UI/v2


liiight:
Well I already see what you meant by react being more understandable than angular... 
Not sure I'd name the packages old and new ui. V1 and v2 fits better imo, we'll probably run them side by side for a long while until we nail down v2

kristenmills:
Yeah that makes sense. I'll make the change.

dev_tools.py
```diff
@@ -82,14 +83,32 @@ def build_webui():
     click.echo('running `npm install`')
     subprocess.check_call('npm install', cwd=cwd, shell=True)
 
-    # Build the ui
+    # Install bower packages
     click.echo('running `bower install`')
-    subprocess.check_call('bower install -p', cwd=cwd, shell=True)
+    subprocess.check_call('bower install', cwd=cwd, shell=True)
```
cvium: @kristenmills Maybe we should switch to Yarn? https://github.com/bower/bower/issues/2298

kristenmills: Bower was used on the old ui not the new one. We're using npm 5 on the new one.

stevezau:
Hey @kristenmills i'm having a look at this. Can you help me understand the layout of the folders etc? Also the __snapshots__ folders?

kristenmills:
I'd like to try to merge this and then start doing smaller, feature based pull requests. Anyone opposed?  
cc: @liiight @cvium @stevezau 

Also does someone mind reviewing the python portions to make sure i'm not doing anything ridiculous?

liiight:
Sounds good to me

flexget/api/app.py
```diff
@@ -21,7 +21,7 @@
 from flexget.webserver import User
 from . import __path__
 
-__version__ = '1.3.0'
+__version__ = '1.4.0'
```
liiight: For non breaking api changes we usually just bump patch. No biggie though

flexget/plugins/daemon/web_server.py
```diff
@@ -31,6 +32,9 @@
                 'ssl_private_key': {'type': 'string'},
                 'web_ui': {'type': 'boolean'},
                 'base_url': {'type': 'string'},
+                'version': {'type': 'integer',
```
liiight: I think that it'd be better to have an ability to run both uis simultaneously

stevezau:
yeah merge it in i'd say.. assuming v1 still works fine?

I've been learning reactjs-redux and am already ready to start work on this also.

liiight:
I think that until v1 can be completely deprecated, v2 should be able to run along side it and not replace it

stevezau:
v2 is not ready to take the place of v1 just yet. Just saying we need to keep v1 as default for now.

kristenmills:
Now the v1 ui runs at root. if you add `run_v2: true`, v2 will run at `/v2`. Sound good?

liiight:
Good enough for now but ideally I think having a list key will be nicer:
```yaml
web_ui:
- v1
- v2
```

stevezau:
agree, run side by side..

kristenmills:
Yeah key listing is probably better. I'll do that after work today.

PR #1856, opened 2017-06-01
[add] series CLI - manually add entities to series database

tubedogg:
### Motivation for changes:
Provide a better way to inject episodes and season packs into the series database.

### Detailed changes:
Implements a new CLI subcommand:
```
series add "<series_name>" <entity_id> [<entity_id> ...] [--quality <quality_string>]
```
- `<entity_id>`: One or more entity IDs recognized by FlexGet, either episodes or season packs.
- `--quality <quality_string>`: Set quality for all entities (otherwise will be shown as 'None' in `series show <series_name>`). Accepts any string that the series parser will recognize as a quality string.

- Feathub #[16](https://feathub.com/Flexget/Flexget/+16).

### DIFF
--- a/flexget/plugins/cli/series.py
+++ b/flexget/plugins/cli/series.py
@@ -262,7 +262,6 @@ def add(manager, options):
     series_name = options.series_name
     entity_ids = options.entity_id
     quality = options.quality
-    downloaded = not options.not_downloaded
     series_name = series_name.replace(r'\!', '!')
     normalized_name = normalize_series_name(series_name)
     with Session() as session:
@@ -276,7 +275,7 @@ def add(manager, options):
             series = series[0]
         for ent_id in entity_ids:
             try:
-                add_series_entity(session, series, ent_id, quality=quality, downloaded=downloaded)
+                add_series_entity(session, series, ent_id, quality=quality)
             except ValueError as e:
                 console(e.args[0])
             else:
@@ -335,8 +334,6 @@ def register_parser_arguments():
                                 help='Episode or season entity ID(s) to add')
     addshow_parser.add_argument('--quality', default=None, metavar='<quality>',
                                 help='Quality string to be stored for all entity ID(s)')
-    addshow_parser.add_argument('--not-downloaded', action='store_true', dest='not_downloaded',
-                                help='Do not mark entity ID(s) as downloaded')
 
     forget_parser = subparsers.add_parser('forget', parents=[series_parser],
                                           help='Removes episodes or whole series from the entire database '
--- a/flexget/plugins/cli/series.py
+++ b/flexget/plugins/cli/series.py
@@ -14,7 +14,7 @@ try:
     from flexget.plugins.filter.series import (Series, remove_series, remove_series_entity, set_series_begin,
                                                normalize_series_name, new_entities_after, get_latest_release,
                                                get_series_summary, shows_by_name, show_episodes, shows_by_exact_name,
-                                               get_all_entities)
+                                               get_all_entities, add_series_entity)
 except ImportError:
     raise plugin.DependencyError(issued_by='cli_series', missing='series',
                                  message='Series commandline interface not loaded')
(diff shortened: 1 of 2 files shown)


flexget/plugins/cli/series.py
```diff
@@ -256,6 +258,33 @@ def display_details(options):
         console(footer)
 
 
+def add(manager, options):
+    series_name = options.series_name
+    entity_ids = options.entity_id
+    quality = options.quality
+    downloaded = not options.not_downloaded
+    series_name = series_name.replace(r'\!', '!')
```
liiight: Isn't this being done in series name normalizer?

flexget/plugins/cli/series.py
```diff
@@ -299,6 +328,16 @@ def register_parser_arguments():
...
                               help='Episode ID to start getting the series from (e.g. S02E01, 2013-12-11, or 9, '
                                    'depending on how the series is numbered)')
+
+    addshow_parser = subparsers.add_parser('add', parents=[series_parser],
+                                           help='Add episode(s) and season(s) to series history')
+    addshow_parser.add_argument('entity_id', nargs='+', default=None, metavar='<entity_id>',
+                                help='Episode or season entity ID(s) to add')
+    addshow_parser.add_argument('--quality', default=None, metavar='<quality>',
+                                help='Quality string to be stored for all entity ID(s)')
+    addshow_parser.add_argument('--not-downloaded', action='store_true', dest='not_downloaded',
```
liiight: What's the use case of adding a not downloaded entry?

flexget/plugins/filter/series.py
```diff
@@ -1001,6 +1001,31 @@ def set_series_begin(series, ep_id):
...
+    """
+    Adds entity identified by `identifier` to series `name` in database.
+
+    :param series: Series in database to add entity to.
+    :param identifier: Series identifier to be added.
+    :param quality: If supplied, this will override the quality from the series parser.
+    """
+    name_to_parse = '{} {}'.format(series.name, identifier)
+    if quality:
+        name_to_parse = '{} {} {}'.format(series.name, identifier, quality)
```
liiight: Probably can depend on the previous formatting here, no biggie

tubedogg: You mean just pass quality straight to `store_parser` instead of appending it to the string to be parsed?

flexget/plugins/cli/series.py
tubedogg: Maybe if you're using it in a perverse way to track what you want to get by inputting undownloaded stuff? I know that sounds weird. I just figured since it was capable of displaying undownloaded releases, we should allow them to be added, but I don't have a solid use case in mind.

flexget/plugins/cli/series.py
tubedogg: Doesn't look like it, no.

```
TRANSLATE_MAP = {ord(u'&'): u' and '}
for char in u'\'\\':
    TRANSLATE_MAP[ord(char)] = u''
for char in u'_./-,[]():':
    TRANSLATE_MAP[ord(char)] = u' '


def normalize_series_name(name):
    """Returns a normalized version of the series name."""
    name = name.lower()
    name = name.replace('&amp;', ' and ')
    name = name.translate(TRANSLATE_MAP)  # Replaced some symbols with spaces
    name = u' '.join(name.split())
    return name
```

flexget/plugins/cli/series.py
liiight: YAGNI, best to drop it Imo.

flexget/plugins/cli/series.py
liiight: Maybe it should then, dunno. This doesn't feel right here

tubedogg: I dunno I just copied from `begin` :P

flexget/plugins/filter/series.py
liiight: I meant string concatenate quality to previously formatted name:
```python 
name_to_parse += ' {}'.format(quality) 
```

PR #1861, opened 2017-06-02
[change] api - Updated unsafe requests to use POST instead of GET

kristenmills:
### Motivation for changes:
So currently a bunch of unsafe requests are using `GET` instead of `POST`. Because all modern browsers support prefetching, it's not wise to use `GET` requests for things that do any sort of modification server side.  

### Detailed changes:
- Update `/logout`, `/server/reload`, `/server/shutdown`, `/database/vacuum`, `/database/cleanup`, and `/database/reset_plugin` to use `POST` instead of `GET`
- Update Web UI to account for these changed request methods.


liiight:
This is a good idea. I wonder though if we should move the action to the payload of the request instead of being in the url itself. Something like `{"operation": "vacuum"}`. What do you think? 
I know POST without body is valid but I feel like using a body fits better.

kristenmills:
I kind of like that too. We could combine the endpoints that way. Do you have a suggestion for the combined endpoint name? We could just do `POST /server` and `/database`? Maybe `/server/action` and `/database/action`?

liiight:
I think i like `/server` and `/database` best

flexget/api/core/database.py
```diff
@@ -17,53 +17,53 @@
...
+        data = request.json
+        operation = data['operation']
+        if operation == 'cleanup':
+            self.manager.db_cleanup(force=True)
+            msg = 'DB Cleanup finished'
+        elif operation == 'vacuum':
+            session.execute('VACUUM')
+            session.commit()
+            msg = 'DB VACUUM finished'
+        elif operation == 'list_plugins':
```
kristenmills: I think `list_plugins` should still be a GET request. It's idempotent and semantically that makes more sense.

liiight: True, i'll change it

kristenmills:
Sorry I haven't gotten around to doing the other endpoint yet...I'll try to get to it soon.

liiight:
You mean the server endpoint? I've been meaning to do that, been super busy...

liiight:
I went with `/server/manage`

kristenmills:
Alright I think this is good to go now. 👍

liiight:
Actually I think we missed one, `/user/token` better be post as well

kristenmills:
That one might more sense as a `put` but in general I agree.

liiight:
Awesome, thank for this change!

PR #1862, opened 2017-06-03
[add] bittorrent - expose torrent name in Torrent class

tubedogg:
### Motivation for changes:
The `name` attribute of the `info` dictionary in the torrent file is often used as the container folder name in various torrent programs. However, this is not exposed in FlexGet.

### Detailed changes:
- `name` is now a property of the `Torrent` class and returns the `name` attribute of the `info` dictionary from the torrent file.

### DIFF
--- a/flexget/utils/bittorrent.py
+++ b/flexget/utils/bittorrent.py
@@ -252,6 +252,11 @@ class Torrent(object):
         return files
 
     @property
+    def name(self):
+        """Return name of the torrent"""
+        return self.content['info'].get('name', False)
+
+    @property
     def size(self):
         """Return total size of the torrent"""
         size = 0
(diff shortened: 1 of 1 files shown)


flexget/utils/bittorrent.py
```diff
@@ -252,6 +252,11 @@ def get_filelist(self):
         return files
 
     @property
+    def name(self):
+        """Return name of the torrent"""
+        return self.content['info'].get('name', False)
```
cvium: Why `False`?

tubedogg: Because I was copying from `private`. It should be an empty string.

flexget/utils/bittorrent.py
```diff
@@ -252,6 +252,11 @@ def get_filelist(self):
         return files
 
     @property
+    def name(self):
+        """Return name of the torrent"""
+        return self.content['info'].get('name', '')
```
cvium: My point was rather, why return something specific at all? `get` will return `None` if the key isn't set.

PR #1864, opened 2017-06-03
[add] Allow any plugin to modify content_files

tubedogg:
### Motivation for changes:
`entry['content_files']` is currently generated only for torrents and is only actually used by the `deluge` plugin.
Additionally, the `transmission` and `deluge` plugins each do very similar renaming routines within the respective plugins, though `transmission` doesn't use `content_files`.
To allow wider use of the `content_files` key, even beyond just torrents, and to reduce redundant code in the above-listed torrent client plugins, the bulk of the renaming logic is being removed from those plugins. New modification plugins have been created which iterate over `content_files` and generate renamed files, download priorities, and other details (depending on the plugin).
To the extent that other plugins use lists of files, they can be converted to use this system as well. The source (i.e. torrent or `from_deluge` plugin) and output (i.e. `transmission` or `deluge` plugins) will be the content-specific parts of the process. The modify plugins in the middle are intended to be content-agnostic and generic.

Also, this (along with my release of [`app_deluge_find`](https://github.com/tubedogg/flexget-plugins/blob/master/app_deluge_find.py) as a separate plugin) is the final step in me not needing a custom Deluge plugin anymore. :D No, this isn't a huge part of the motivation, it's just a happy benefit.

### Detailed changes:
- Renaming logic removed from `deluge` and `transmission` plugins. If `content_files` exists in the entry, renaming operations and download priorities are run against the torrent once it's added to Deluge/Transmission.
- `content_files` is a list of dicts, currently generated by the `torrent_files` metainfo plugin:
  - The structure of each dict at creation is: `{'path': 'torrent-name/path/filename.mkv', 'size': 123456789}`.
  - Additional keys currently recognized by the torrent client plugins are:
    - `new_path`, which is what the file is to be renamed to
    - `download`, which determines whether the file is downloaded (1 or 0, default 1)
    - `priority` ([2,5,7] with 2 meaning low, 5 meaning normal, and 7 meaning high; default 5 if `download` is 1 but `priority` is not present)
    - further keys may be specified with no adverse effects, but the intended output plugin would need to be modified to take advantage of them
- New plugin `modify_content_files`. This is a cleaned-up version of my personal renaming plugin that I've been using for the past couple of years that works for movies, single-file TV episodes and multi-file TV season packs. See documentation below.
- The new plugin `deluge_rename` has been created and functions (though is not coded) identically to the old built-in logic. Simply moving the `deluge` plugin keys `content_filename`, `main_file_only`, `main_file_ratio`, `container_directory`, `hide_sparse_files`, and `keep_subs` under the `deluge_rename` plugin key should result in a seamless transition. _This is something that needs to be thoroughly tested by those who used the old renaming logic. I tested to the best of my ability but I haven't used it in my config in years, so I'm bound to have missed something._

```yaml
modify_content_files:
  content_filename: '{{series_id}}'
  fix_year: yes
  keep_container: no
  container_directory: '{% if season_pack %}{{series_name}} Season {{series_season}}{% else %}{{series_name}}{{" - " ~ series_id|lower}}{% endif %}'
  container_only_for_multi: yes
```
```yaml
deluge_rename:
  content_filename: '{%- if tvdb_series_name is defined -%}{{tvdb_series_name}}{%- else -%}{{series_name}}{%- endif -%}{{" - " ~ series_id|lower}}{%- if proper -%}{{" P" if proper == True else ""}}{%- endif -%}'
  main_file_only: yes
```

#### To Do:

- [x] Put old renaming logic from `deluge` plugin into its own plugin for people who want to continue using it
- [x] Change `torrent_rename_files` to not use `Torrent.get_filelist()` or `Torrent.name` if the input was `from_deluge` and not an actual torrent file.
- [x] Document `modified_content_files` key and structure that other plugins must use for `deluge` plugin to recognize that file renaming/priority setting needs to occur.
- [x] Document `torrent_rename_files` plugin
- [x] Document `deluge_rename` plugin
- [ ] Remove renaming logic from `transmission` plugin and convert it to use new `content_files` format

#### `torrent_rename_files` Documentation

| Option | Default value | Required | Value Type | Description |
| --- | --- | --- | --- | --- |
| main_file_only | `no` | no | boolean (yes/no) | If yes, all files but the main file inside the torrent (>90% of total torrent size) will be set to 'do not download'. |
| main_file_ratio | `0.9` | no | decimal | Sets the threshold value for *main_file_only*. Expects a number between 0 and 1 (i.e. `0.85` to change to 85%). |
| keep_subs | `no` | no | boolean (yes/no) | If yes, download the first subtitle file that is found inside the torrent. |
| keep_container | `yes` | no | boolean (yes/no) | If no, the directory structure inside the torrent (if there is one) will be removed.|
| container_directory | n/a | no | Jinja template | This is a template used to rename the main container directory within the torrent. Note that `keep_container` does not affect this, as that option is only concerned with removing any existing directory structure. Therefore `keep_container: no` and `container_directory: yes` would cause the existing structure to be removed and replaced with the structure from this option.
| container_only_for_multi | `yes` | no | boolean (yes/no) | If yes, only adds `container_directory` went the torrent contains more than one file. If no, always adds `container_directory` (if it's set).
| content_filename | n/a | no | Jinja template | This can be used to rename files within the torrent. |
| rename_main_file_only | `no` | no | boolean (yes/no) | If yes, rename only the main file inside the torrent (>90% of total torrent size) using the `content_filename` setting. |
| fix_year | `no` | no | boolean (yes/no) | If yes, if the last four characters of a series' name are numbers, add parentheses around them. e.g. The Show 2017 becomes The Show (2017). |
| unlisted_filetype_default | `yes` | no | boolean (yes/no) or dictionary | Boolean: If yes, renames all files (that are not explicitly specified under the `filetypes` setting below) using `content_filename`, unless `rename_main_file_only` is set to `yes`. Dictionary: can contain `download`, `priority`, `rename`, and `content_filename` keys. `download` can be 1 (download) or 0. `priority` can be 2, 5, 7 meaning low, regular, high priority. `rename` can be yes/no. `content_filename` can be a custom Jinja template for unlisted filetypes. |
| filetypes | n/a | no | boolean (yes/no) or dictionary | Specify rename settings by file type (file extension). Each type can be set to `yes` or `no` to determine whether or not to always rename them, or can be set to a dict (same structure as allowed for `unlisted_filetype_default`). |

```yaml
# filetypes example
torrent_rename_files:
  content_filename: '{{title}}'
  filetypes:
    # uses custom Jinja template for mp3 files
    mp3:
      content_filename: "{{title}}.mp3"
    # renames mkv files with `content_filename` in entry or general configuration
    mkv: yes
```

#### `deluge_rename` Documentation
See the following keys in the [`deluge`](https://flexget.com/Plugins/deluge) plugin: `content_filename`, `main_file_only`, `main_file_ratio`, `container_directory`, `hide_sparse_files`, `keep_subs`

### DIFF
--- a/flexget/plugins/clients/deluge.py
+++ b/flexget/plugins/clients/deluge.py
@@ -551,8 +551,7 @@ class OutputDeluge(DelugePlugin):
                     log.debug('Moving storage for %s to %s' % (entry['title'], move_now_path))
                     main_file_dlist.append(client.core.move_storage([torrent_id], move_now_path))
 
-                if entry.get('modified_content_files'):
-                    log.debug('entering modified_content_files loop')
+                if entry.get('content_files'):
                     def file_exists(filename):
                         # Checks the download path as well as the move completed path for existence of the file
                         if os.path.exists(os.path.join(status['save_path'], filename)):
@@ -583,16 +582,26 @@ class OutputDeluge(DelugePlugin):
                                                      [(file_info['index'], new_name)]))
                         log.debug('File `%s` in `%s` renamed to `%s`', file_info['path'], entry['title'], new_name)
 
+                    set_top_level = False
+                    if len(status['files']) > 1:
+                        top_levels = [f['new_path'].split(os.sep)[0] if f.get('new_path')
+                                     else f['path'].split(os.sep)[0] for f in entry['content_files']]
+                        if len(set(top_levels)) > 1:
+                            set_top_level = True
+                            log.warning('Appending top-level folder `%s` to all torrent paths as multiple top-level '
+                                        'folders were found, and this is not allowed per the torrent spec.')
                     file_priorities = []
                     for count, current_file in enumerate(status['files']):
                         try:
-                            modified_file = entry['modified_content_files'][count]
+                            modified_file = entry['content_files'][count]
                         except:
-                            log.debug('Cannot find file `%s` in modified_content_files', current_file['path'])
+                            log.error('Cannot find file `%s` in `content_files`', current_file['path'])
                             continue
-                        file_priorities.append(modified_file['download'] if modified_file.get('download') else 1)
-                        if modified_file['new_path'] and modified_file['new_path'] != current_file['path']:
-                            new_filename = unused_name(modified_file['new_path'])
+                        file_priorities.append(modified_file.get('priority', modified_file.get('download', 1)))
+                        if set_top_level or (modified_file.get('new_path')
+                                             and modified_file['new_path'] != current_file['path']):
+                            new_filename = unused_name(os.join(status['name'], top_levels[count]) if set_top_level
+                                                       else modified_file['new_path'])
                             rename(current_file, new_filename)
                         else:
                             log.verbose('Not renaming because `%s` is identical to the existing filename',
--- a/flexget/plugins/clients/deluge.py
+++ b/flexget/plugins/clients/deluge.py
@@ -590,7 +590,7 @@ class OutputDeluge(DelugePlugin):
                         except:
                             log.debug('Cannot find file `%s` in modified_content_files', current_file['path'])
                             continue
-                        file_priorities.append(1 if modified_file['download'] == 1 else 0)
+                        file_priorities.append(modified_file['download'] if modified_file.get('download') else 1)
                         if modified_file['new_path'] and modified_file['new_path'] != current_file['path']:
                             new_filename = unused_name(modified_file['new_path'])
                             rename(current_file, new_filename)
--- a/flexget/plugins/clients/deluge.py
+++ b/flexget/plugins/clients/deluge.py
@@ -236,7 +236,8 @@ class InputDeluge(DelugePlugin):
         'save_path': 'deluge_path',
         'label': 'deluge_label',
         'total_size': ('content_size', lambda size: size / 1024 / 1024),
-        'files': ('content_files', lambda file_dicts: [f['path'] for f in file_dicts])}
+        'files': ('content_files', lambda file_dicts: [f['path'] for f in file_dicts]),
+        'file_sizes': ('content_file_sizes', lambda file_sizes: [f['size'] for f in file_sizes])}
 
     extra_settings_map = {
         'active_time': ('active_time', lambda time: time / 3600),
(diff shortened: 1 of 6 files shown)


tubedogg:
It was suggested that `torrent_rename_files` be renamed `rename_content_files` or something along those lines.

Argument for: it modifies the files listed in `content_files` in the entry, so the name matches it's purpose.
Argument against: the user doesn't necessarily know what `content_files` is (nor should they need to), just that they're renaming files within the torrent, so it might be more confusing.

`content_files` is specific to torrents as it is added to the entry by plugins/metainfo/torrent_files.py.

Thoughts?

gazpachoking:
I've been thinking something like this should be done for a while, and I like the idea. Some thoughts:
Why is there both a separate `deluge_rename` and `torrent_rename_files`? If we go a generic route, I'm thinking it could work something like this:
- Some plugin populates a content_files field, (like torrent or deluge plugins can do now, maybe other things in the future). This should have all the info about the files (including size.)
- Renaming plugin(s) would either add info to the content_files field, or make some new field with the renaming info. These wouldn't be dependent on any plugins (just that proper entry field was there,) and wouldn't make any other changes but to entry_fields which compatible output plugins could read.
- Any output plugin where appropriate could listen to those rename entry fields, e.g. deluge, transmission, (decompress?), and make the actual calls to do the renaming.

tubedogg:
@gazpachoking `deluge_rename` and `torrent_rename_files` both exist as different renaming options. I wanted to provide an upgrade path for those who like and use the existing deluge renaming functionality without forcing them into a new renaming plugin.

The structure you describe is essentially what I am doing right now, with the exception that `torrent_rename_files` assumes the source is a torrent (either torrent file or info populated by `from_deluge`). I haven't made changes to any other output plugins for them to react to `modified_content_files`, though. I also haven't looked into whether the other torrent client plugins have analogs to `from_deluge`. I know nothing else currently populates `content_files`.

I do like the idea of having one key that has all file information. I guess with modified_content_files I was trying to keep a baseline available of what it was before changes, but there's no real reason it couldn't be stored in the dict alongside any updated information. For torrents, I'm thinking a base structure something like this that would be generated by e.g. `torrent_files`:
```python
content_files = [
  {'original_path': 'ABC123/abc123.mkv', size: 198574557}
]
```
From there, it would follow the route already established in this PR, where modify plugins would make whatever changes they want to, and output plugins would react based on the presence of certain keys in the dicts (i.e. `new_path` indicating a file to be renamed, `download` indicating whether or not to download a file, etc).

PR #1879, opened 2017-06-19
added urlrewrite plugin for rmz.cr (rapidmoviez.com|rapidmoviez.eu)

thawn:
### Motivation for changes:
rmz.cr needs an urlrewrite plugin for the rss feed (rmz.cr/feed) because the feed only provides links back to the rmz.cr release page.

### Detailed changes:
    rmz.cr (rapidmoviez.com) urlrewriter
    Version 0.1
    
    Configuration

    rmz:
      url_re:
        - domain\.com
        - domain2\.org

    Only add links that match any of the regular expressions listed under url_re.

    'urls' field of the entry.
    
    exec:

```
    rmz:
      url_re:
        - domain\.com
        - domain2\.org
```
```
DEBUG  rmz  TASK  Using links_re filters: <list of filter expressions>
DEBUG  rmz  TASK  Url: "<link found at entry url>" matched filter: <expression>

INFO      rmz  TASK  Found 2 links at <entry url>
```
#### To Do:

- maybe extend this to work also as a search plugin for rmz.cr

### DIFF
--- a/flexget/plugins/sites/rmz.py
+++ b/flexget/plugins/sites/rmz.py
@@ -14,11 +14,12 @@ from requests.exceptions import RequestException
 
 log = logging.getLogger('rmz')
 
+
 class UrlRewriteRmz(object):
     """
     rmz.cr (rapidmoviez.com) urlrewriter
     Version 0.1
-    
+
     Configuration
 
     rmz:
@@ -31,10 +32,10 @@ class UrlRewriteRmz(object):
     If more than one valid link is found, the url of the entry is rewritten to
     the first link found. The complete list of valid links is placed in the
     'urls' field of the entry.
-    
+
     Therefore, it is recommended, that you configure your output to use the
     'urls' field instead of the 'url' field.
-    
+
     For example, to use jdownloader 2 as output, you would use the exec plugin:
     exec:
       - echo "text={{urls}}" >> "/path/to/jd2/folderwatch/{{title}}.crawljob"
@@ -43,7 +44,7 @@ class UrlRewriteRmz(object):
     schema = {
         'type': 'object',
         'properties': {
-            'filehosters_re': {'type': 'array', 'items': {'format': 'regexp'} }
+            'filehosters_re': {'type': 'array', 'items': {'format': 'regexp'}}
         },
         'additionalProperties': False
     }
@@ -53,7 +54,7 @@ class UrlRewriteRmz(object):
         'filehosters_re': []
     }
 
-    #grab config
+    # grab config
     def on_task_start(self, task, config):
         self.config = config
 
@@ -75,11 +76,11 @@ class UrlRewriteRmz(object):
         except Exception as e:
             raise UrlRewritingError(str(e))
         link_elements = soup.find_all('pre', class_='links')
-        urls=[]
+        urls = []
         for element in link_elements:
             urls.extend(element.text.splitlines())
         regexps = self.config.get('filehosters_re', [])
-        filtered_urls=[]
+        filtered_urls = []
         for i, url in enumerate(urls):
             urls[i] = normalize_unicode(url)
             for regexp in regexps:
@@ -88,10 +89,11 @@ class UrlRewriteRmz(object):
                     log.debug('Url: "%s" matched filehoster filter: %s', urls[i], regexp)
                     break
             else:
-                log.debug('Url: "%s" does not match any of the given filehoster filters: %s', urls[i], str(regexps))
+                log.debug(
+                    'Url: "%s" does not match any of the given filehoster filters: %s', urls[i], str(regexps))
         if regexps:
             log.debug('Using filehosters_re filters: %s', str(regexps))
-            urls=filtered_urls
+            urls = filtered_urls
         else:
             log.debug('No filehoster filters configured, using all found links.')
         num_links = len(urls)
@@ -101,6 +103,7 @@ class UrlRewriteRmz(object):
             entry['url'] = urls[0]
         else:
             raise UrlRewritingError('No useable links found at %s' % entry['url'])
+
 
 @event('plugin.register')
 def register_plugin():
--- a/flexget/plugins/sites/rmz.py
+++ b/flexget/plugins/sites/rmz.py
@@ -11,6 +11,8 @@ from flexget.utils.soup import get_soup
 from flexget.entry import Entry
 from flexget.utils.search import normalize_unicode
 
+from requests.exceptions import RequestException
+
 log = logging.getLogger('rmz')
 
 class UrlRewriteRmz(object):
@@ -21,7 +23,7 @@ class UrlRewriteRmz(object):
     Configuration
 
     rmz:
-      link_re:
+      filehosters_re:
         - domain\.com
         - domain2\.org
 
@@ -42,14 +44,14 @@ class UrlRewriteRmz(object):
     schema = {
         'type': 'object',
         'properties': {
-            'links_re': {'type': 'array', 'items': {'format': 'regexp'} }
+            'filehosters_re': {'type': 'array', 'items': {'format': 'regexp'} }
         },
         'additionalProperties': False
     }
 
     # Since the urlrewriter relies on a config, we need to create a default one
     config = {
-        'links_re': None
+        'filehosters_re': None
     }
 
     #grab config
@@ -60,48 +62,42 @@ class UrlRewriteRmz(object):
     def url_rewritable(self, task, entry):
         url = entry['url']
         rewritable_regex = '^https?:\/\/(www.)?(rmz\.cr|rapidmoviez\.(com|eu))\/.*'
-        if re.match(rewritable_regex, url):
-            return True
-        else:
-            return False
+        return re.match(rewritable_regex, url) is not None
 
     @plugin.internet(log)
    # urlrewriter API
     def url_rewrite(self, task, entry):
         try:
             page = task.requests.get(entry['url'])
-        except Exception as e:
-            raise UrlRewritingError(e)
+        except RequestException as e:
+            raise UrlRewritingError(str(e))
         try:
             soup = get_soup(page.text)
         except Exception as e:
-            raise UrlRewritingError(e)
+            raise UrlRewritingError(str(e))
         link_elements = soup.find_all('pre', class_='links')
         urls=[]
         for element in link_elements:
             urls.extend(element.text.splitlines())
-        regexps = self.config.get('links_re', None)
+        regexps = self.config.get('filehosters_re', [])
         if regexps:
-            log.debug('Using links_re filters: %s', str(regexps))
+            log.debug('Using filehosters_re filters: %s', str(regexps))
         else:
-            log.debug('No link filters configured, using all found links.')
-        for i in reversed(range(len(urls))):
-            urls[i]=urls[i].encode('ascii','ignore')
-            if regexps:
-                accept = False
-                for regexp in regexps:
-                    if re.search(regexp, urls[i]):
-                        accept = True
-                        log.debug('Url: "%s" matched filter: %s', urls[i], regexp)
-                        break
-                if not accept:
-                    log.debug('Url: "%s" does not match any of the given filters: %s', urls[i], str(regexps))
-                    del(urls[i])
-        numlinks=len(urls)
-        log.info('Found %d links at %s.',numlinks, entry['url'])
-        if numlinks>0:
+            log.debug('No filehoster filters configured, using all found links.')
+        for i in reversed(enumerate(urls)):
+            urls[i] = urls[i].encode('ascii', 'ignore')
+            for regexp in regexps:
+                if re.search(regexp, urls[i]):
+                    log.debug('Url: "%s" matched filehoster filter: %s', urls[i], regexp)
+                    break
+            else:
+                log.debug('Url: "%s" does not match any of the given filehoster filters: %s', urls[i], str(regexps))
+                del(urls[i])
+        num_links=len(urls)
+        log.info('Found %d links at %s.',num_links, entry['url'])
+        if num_links:
             entry.setdefault('urls', urls)
-            entry['url']=urls[0]
+            entry['url'] = urls[0]
         else:
             raise UrlRewritingError('No useable links found at %s' % entry['url'])
 
--- a/flexget/plugins/sites/rmz.py
+++ b/flexget/plugins/sites/rmz.py
@@ -8,7 +8,6 @@ from flexget import plugin
 from flexget.event import event
 from flexget.plugins.internal.urlrewriting import UrlRewritingError
 from flexget.utils.soup import get_soup
-from flexget.entry import Entry
 from flexget.utils.search import normalize_unicode
 
 from requests.exceptions import RequestException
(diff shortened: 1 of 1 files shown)


flexget/plugins/sites/rmz.py
```diff
@@ -0,0 +1,110 @@
...
+    
+    For example, to use jdownloader 2 as output, you would use the exec plugin:
+    exec:
+      - echo "text={{urls}}" >> "/path/to/jd2/folderwatch/{{title}}.crawljob"
+    """
+
+    schema = {
+        'type': 'object',
+        'properties': {
+            'links_re': {'type': 'array', 'items': {'format': 'regexp'} }
```
liiight: 1. there's a mismatch between this value and the one in the doc (`url_re`).
2. I'm not sure why is this needed. Don't you know the TLD and format you need for the rewrite? why does the user need to input this? URL rewrite plugin do not need to rely on config.

flexget/plugins/sites/rmz.py
```diff
@@ -0,0 +1,110 @@
...
+        for i in reversed(range(len(urls))):
+            urls[i]=urls[i].encode('ascii','ignore')
+            if regexps:
+                accept = False
+                for regexp in regexps:
+                    if re.search(regexp, urls[i]):
+                        accept = True
+                        log.debug('Url: "%s" matched filter: %s', urls[i], regexp)
+                        break
+                if not accept:
```
cvium: Since you `break` when finding a match, you could just use `for-else` instead of the `accept` variable.

flexget/plugins/sites/rmz.py
```diff
@@ -0,0 +1,110 @@
...
+        urls=[]
+        for element in link_elements:
+            urls.extend(element.text.splitlines())
+        regexps = self.config.get('links_re', None)
+        if regexps:
+            log.debug('Using links_re filters: %s', str(regexps))
+        else:
+            log.debug('No link filters configured, using all found links.')
+        for i in reversed(range(len(urls))):
+            urls[i]=urls[i].encode('ascii','ignore')
```
cvium: Lacking some whitespace around equals and after comma.

flexget/plugins/sites/rmz.py
```diff
@@ -0,0 +1,110 @@
...
+                accept = False
+                for regexp in regexps:
+                    if re.search(regexp, urls[i]):
+                        accept = True
+                        log.debug('Url: "%s" matched filter: %s', urls[i], regexp)
+                        break
+                if not accept:
+                    log.debug('Url: "%s" does not match any of the given filters: %s', urls[i], str(regexps))
+                    del(urls[i])
+        numlinks=len(urls)
```
cvium: Space

flexget/plugins/sites/rmz.py
```diff
@@ -0,0 +1,110 @@
...
+        log.info('Found %d links at %s.',numlinks, entry['url'])
+        if numlinks>0:
```
cvium: Space

flexget/plugins/sites/rmz.py
```diff
@@ -0,0 +1,110 @@
...
+            entry.setdefault('urls', urls)
+            entry['url']=urls[0]
```
cvium: space

flexget/plugins/sites/rmz.py
```diff
@@ -0,0 +1,110 @@
...
+        else:
+            raise UrlRewritingError('No useable links found at %s' % entry['url'])
+
+@event('plugin.register')
+def register_plugin():
+    plugin.register(UrlRewriteRmz, 'rmz', interfaces=['urlrewriter', 'task'], api_ver=2)
```
cvium: There should be a newline at the end of the file.

flexget/plugins/sites/rmz.py
```diff
@@ -0,0 +1,110 @@
...
+
+    #grab config
+    def on_task_start(self, task, config):
+        self.config = config
+
+    # urlrewriter API
+    def url_rewritable(self, task, entry):
+        url = entry['url']
+        rewritable_regex = '^https?:\/\/(www.)?(rmz\.cr|rapidmoviez\.(com|eu))\/.*'
+        if re.match(rewritable_regex, url):
```
liiight: suggestion:
```python
return re.match(rewritable_regex, url) is not None
```

flexget/plugins/sites/rmz.py
```diff
@@ -0,0 +1,110 @@
...
+            raise UrlRewritingError(e)
+        try:
+            soup = get_soup(page.text)
+        except Exception as e:
+            raise UrlRewritingError(e)
+        link_elements = soup.find_all('pre', class_='links')
+        urls=[]
+        for element in link_elements:
+            urls.extend(element.text.splitlines())
+        regexps = self.config.get('links_re', None)
```
cvium: `None` is already the default value returned by `dict.get()`.

flexget/plugins/sites/rmz.py
```diff
@@ -0,0 +1,110 @@
...
+        if regexps:
+            log.debug('Using links_re filters: %s', str(regexps))
+        else:
+            log.debug('No link filters configured, using all found links.')
+        for i in reversed(range(len(urls))):
```
liiight: i'd use `enumerate` here

flexget/plugins/sites/rmz.py
```diff
@@ -0,0 +1,110 @@
...
+            urls[i]=urls[i].encode('ascii','ignore')
+            if regexps:
+                accept = False
+                for regexp in regexps:
```
liiight: this will crash if `regexps` is `None`

flexget/plugins/sites/rmz.py
liiight: use snake case `num_links`

flexget/plugins/sites/rmz.py
```diff
@@ -0,0 +1,110 @@
...
+            return True
+        else:
+            return False
+
+    @plugin.internet(log)
+   # urlrewriter API
+    def url_rewrite(self, task, entry):
+        try:
+            page = task.requests.get(entry['url'])
+        except Exception as e:
```
cvium: Too broad

flexget/plugins/sites/rmz.py
liiight: not pythonic, no need to check if higher than 0, just do:
```python
if numlinks:
```

flexget/plugins/sites/rmz.py
```diff
@@ -0,0 +1,110 @@
...
+            raise UrlRewritingError(e)
```
cvium: Probably need to wrap `e` in `str()`

flexget/plugins/sites/rmz.py
```diff
@@ -0,0 +1,110 @@
...
+        try:
+            soup = get_soup(page.text)
+        except Exception as e:
+            raise UrlRewritingError(e)
```
cvium: Probably need to wrap `e` in `str()`

flexget/plugins/sites/rmz.py
```diff
@@ -0,0 +1,110 @@
...
+    @plugin.internet(log)
+   # urlrewriter API
+    def url_rewrite(self, task, entry):
+        try:
+            page = task.requests.get(entry['url'])
+        except Exception as e:
+            raise UrlRewritingError(e)
+        try:
+            soup = get_soup(page.text)
+        except Exception as e:
```
cvium: This one is also too broad, but since BS4 can raise different exceptions it's fine in this case...

flexget/plugins/sites/rmz.py
thawn: Thanks for the tip, I'll try that.

flexget/plugins/sites/rmz.py
cvium: Look at line 90

flexget/plugins/sites/rmz.py
thawn: 1. I'll fix that asap.
2. The link_re config parameter works just like the link_re parameter in the html input plugin (thats why I renamed it).
The reason for using this parameter is that on the rmz.cr site, they often offer links to several filehosters. This parameter allows the user to choose which filehosters he prefers. For example, the user may have an account for one of the filehosters or he may prefer some filehosters because they offer superior download speed or no file size restrictions for free users.

flexget/plugins/sites/rmz.py
thawn: Could you please suggest an error class to use? I looked at the html plugin but they do not catch errors at all.
I tried to use requests.exceptions.RequestException but got an error.

flexget/plugins/sites/rmz.py
liiight: ah, gotcha. rmz.cr is an indexing site and the links are to external sites.
maybe name it `filehoster` instead then?

flexget/plugins/sites/rmz.py
liiight: that is the correct exception. What error did you get? did you import it correctly?

flexget/plugins/sites/rmz.py
liiight: ah ok, maybe just do `regexps = self.config.get('links_re', [])` and avoid the if statement

flexget/plugins/sites/rmz.py
```diff
@@ -60,48 +62,42 @@ def on_task_start(self, task, config):
...
+        for i in reversed(enumerate(urls)):
+            urls[i] = urls[i].encode('ascii', 'ignore')
+            for regexp in regexps:
+                if re.search(regexp, urls[i]):
+                    log.debug('Url: "%s" matched filehoster filter: %s', urls[i], regexp)
+                    break
+            else:
+                log.debug('Url: "%s" does not match any of the given filehoster filters: %s', urls[i], str(regexps))
+                del(urls[i])
+        num_links=len(urls)
```
cvium: space

flexget/plugins/sites/rmz.py
```diff
@@ -60,48 +62,42 @@ def on_task_start(self, task, config):
...
+        log.info('Found %d links at %s.',num_links, entry['url'])
```
cvium: space after comma

flexget/plugins/sites/rmz.py
thawn: I need to go through the array in reversed order because I am removing some elements within the for loop.

I just tried and enumerate did not work well with reversed:
when I tried to use `reversed(enumerate(urls))`, I got the error:
TypeError: argument to reversed() must be a sequence
Then I tried `reversed(list(enumerate(urls)))` (which is imho at least as ugly as what I used originally) but that gave me the error:
TypeError: list indices must be integers, not tuple

so I gave up and reverted to `reversed(range(len(urls)))`

flexget/plugins/sites/rmz.py
```diff
@@ -0,0 +1,106 @@
...
+            log.debug('No filehoster filters configured, using all found links.')
+        for i in reversed(range(len(urls))):
+            urls[i] = urls[i].encode('ascii', 'ignore')
+            for regexp in regexps:
+                if re.search(regexp, urls[i]):
+                    log.debug('Url: "%s" matched filehoster filter: %s', urls[i], regexp)
+                    break
+            else:
+                log.debug('Url: "%s" does not match any of the given filehoster filters: %s', urls[i], str(regexps))
+                del(urls[i])
```
cvium: It's a bad idea to iterate over a list and delete elements from it. You should either create a copy of the list or add to another list instead of deleting.

flexget/plugins/sites/rmz.py
thawn: I added a newline in my latest commit, but it does not show up here. Am I missing something?

flexget/plugins/sites/rmz.py
thawn: I never experienced any problem when I was traveling the list in reversed order. Are there any performance implications (although we are talking about 100 entries in the list tops)?

flexget/plugins/sites/rmz.py
```diff
@@ -0,0 +1,106 @@
...
+        num_links = len(urls)
+        log.info('Found %d links at %s.',num_links, entry['url'])
```
liiight: I think `log.verbose` level would fit here better

flexget/plugins/sites/rmz.py
thawn: good idea, done in the latest commit.

flexget/plugins/sites/rmz.py
thawn: o.k. got it to work like this.

flexget/plugins/sites/rmz.py
thawn: I added a newline a few commits ago and I just checked on my local machine and with the online editor and there is definitely a newline at the end of the file now.

Don't know why this review did not get marked as outdated...

flexget/plugins/sites/rmz.py
thawn: I got this fixed now. I had to add `from requests.exceptions import RequestException`. Now I am using RequestException instead of Exception.

flexget/plugins/sites/rmz.py
thawn: done

flexget/plugins/sites/rmz.py
thawn: I replaced this with [] to make sure it does not brake my for loop in line 86

flexget/plugins/sites/rmz.py
thawn: I read up on the performance implications of del.
switched to adding to a new list.

flexget/plugins/sites/rmz.py
thawn: changed that and added the space after the comma.

thawn:
@cvium @liiight 
Thank you so much for the kind code review. I learned a lot today!
I think I now managed to address all the issues you raised. Let me know if there is more that needs to be improved.

Cheers,

Thawn

flexget/plugins/sites/rmz.py
```diff
@@ -0,0 +1,107 @@
...
+            for regexp in regexps:
+                if re.search(regexp, urls[i]):
+                    filtered_urls.append(urls[i])
+                    log.debug('Url: "%s" matched filehoster filter: %s', urls[i], regexp)
+                    break
+            else:
+                log.debug('Url: "%s" does not match any of the given filehoster filters: %s', urls[i], str(regexps))
+        if regexps:
+            log.debug('Using filehosters_re filters: %s', str(regexps))
+            urls=filtered_urls
```
liiight: Missing spaces. See about running pep8 on your code. You can also use pycharm which is free

thawn: Thanks for the tip. I used autopep8 (via eclipse).

liiight:
Thanks, care to update the wiki?

thawn:
@liiight Sure, I'll write a wiki page and link to it under the "Data operations" section next to the generic urlrewrite plugin.

Thanks again for your help with improving the code!

thawn:
Wiki page is done:
https://www.flexget.com/Plugins/rmz

and linked from the plugins page (section Modification/Data Operations):
https://flexget.com/Plugins#modification

PR #1880, opened 2017-06-20
input/rss plugin: added ability to read the special content object created by feedparser

### LINKED ISSUE #656: Add support for `content` field grabbing by rss plugin
Hi there developers,

I have an issue that I cannot seem to resolve. I'm trying to extract links from a relatively detailed rss feed.

My current config looks like this:

```
tasks:
  Music Download:
    inputs:
      - rss:
          url: http://www.trancefix.nl/external.php?type=RSS2&forumids=210
          other_fields: ['content:encoded']  
          link: 'content:encoded'
    exec:
      - echo text="\"{{description}}\"" >> "/volume1/pyload/crawljob/{{title}}.crawljob"
      - echo downloadFolder="\"D:\\Downloads\\Trance\\\"" >> "/volume1/pyload/crawljob/{{title}}.crawljob"
```

However when I try and run this, flexget either doesn't recognise the field name or cannot read the rather long field "content:encoded" or it has a problem with the ":". I'm not sure but the output is the following:

```
2015-11-09 18:44 VERBOSE  task_queue                    There are 1 tasks to execute. Shutdown will commence when they have completed.
2015-11-09 18:44 WARNING  rss           Music Download  Skipped 2 RSS-entries without required information (title, link or enclosures)
2015-11-09 18:44 WARNING  inputs        Music Download  Input rss did not return anything
2015-11-09 18:44 WARNING  details       Music Download  Task didn't produce any entries. This is likely due to a mis-configured or non-functional input.
2015-11-09 18:44 VERBOSE  details       Music Download  Summary - Accepted: 0 (Rejected: 0 Undecided: 0 Failed: 0)
```

Any help would be appreciated!

thawn:
### Motivation for changes:
The rss plugin fails to import the 'content' field of a rss feed (in my case the field was named content:encoded) and logs an error: Cannot grab non text field content from rss.

This is because feedparser stores content in an object rather than a string: https://pythonhosted.org/feedparser/reference-entry-content.html

Since 'content' can be particularly useful, it may be worthwile to write a workaround.

This pull request fixed the issue for me.

### Detailed changes:
- added 10 lines of code that properly handle the `FeedParserDict` object which feedparser delivers for content elements in rss feeds
- since there may be more than one content element in atom feeds, the code loops over all content elements delivered by feedparser

### Addressed issues:
- Fixes #656 .

- Feathub #[25](http://feathub.com/Flexget/Flexget/+25).

```
WARNING  rss  Failed to decode entry `<title>` field `content`
DEBUG      rss  Field `content` set to `<content string>`  for `<title>`
```
#### To Do:
- [ ] similar routines for other rss elements for which feedparser does not return a string

### DIFF
--- a/flexget/plugins/input/rss.py
+++ b/flexget/plugins/input/rss.py
@@ -382,6 +382,16 @@ class InputRSS(object):
                 # fields dict may be modified during this loop, so loop over a copy (fields.items())
                 for rss_field, flexget_field in list(fields.items()):
                     if rss_field in entry:
+                        if rss_field == 'content':
+                            content_str = ''
+                            for content in entry[rss_field]:
+                                try:
+                                    content_str += decode_html(content.value)
+                                except UnicodeDecodeError:
+                                    log.warning('Failed to decode entry `%s` field `%s`', ea['title'], rss_field)
+                            ea[flexget_field] = content_str
+                            log.debug('Field `%s` set to `%s` for `%s`', rss_field, ea[rss_field], ea['title'])
+                            continue
                         if not isinstance(getattr(entry, rss_field), str):
                             # Error if this field is not a string
                             log.error('Cannot grab non text field `%s` from rss.', rss_field)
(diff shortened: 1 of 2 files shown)


thawn:
forgot to properly rebase

PR #1881, opened 2017-06-20
Add a threshold setting for number of episodes already downloaded to …

jawilson:
Adds a threshold to `next_series_seasons` to specify how many episodes may be downloaded for a particular season before emitting a season pack.

### Motivation for changes:

If several episodes of a season are already downloaded, I don't usually want to download the entire season pack.

### Detailed changes:
- Added `threshold` property to `next_series_seasons`
- If `series.episodes_for_season()` is greater than the threshold, skip the season

```
tasks:
  discover-tv:
    discover:
      release_estimations:
        optimistic: 30 days
      what:
        - next_series_seasons:
            backfill: yes
            from_start: yes
            threshold: 4
      from:
        - piratebay:
            category: "highres tv"
            sort_by: seeds
      interval: 1 day
```

### DIFF
--- a/flexget/plugins/input/next_series_seasons.py
+++ b/flexget/plugins/input/next_series_seasons.py
@@ -31,7 +31,8 @@ class NextSeriesSeasons(object):
                 'type': 'object',
                 'properties': {
                     'from_start': {'type': 'boolean', 'default': False},
-                    'backfill': {'type': 'boolean', 'default': False}
+                    'backfill': {'type': 'boolean', 'default': False},
+                    'threshold': {'type': 'integer', 'minimum': 0}
                 },
                 'additionalProperties': False
             }
@@ -80,6 +81,8 @@ class NextSeriesSeasons(object):
             return entries
         else:
             self.rerun_entries = []
+
+        threshold = config.get('threshold', None)
 
         entries = []
         impossible = {}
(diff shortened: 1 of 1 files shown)


flexget/plugins/input/next_series_seasons.py
```diff
@@ -81,6 +82,8 @@ def on_task_input(self, task, config):
         else:
             self.rerun_entries = []
 
+        threshold = config.get('threshold', None)
```
cvium: `None` is already the default value if key doesn't exist.

liiight:
You can already set threshold in series season pack config. It is set to 0 by default which means that if any episodes of a season were downloaded, season packs will be rejected.

How is this different?

jawilson:
@liiight the difference is that this configures the threshold for _emitting_ season from `next_series_seasons`, the `series` threshold only configures the threshold for _accepting_. This is the same situation for `backfill` and `from_start`.

The reason for this change is that I don't want to emit seasons for searching (causing extra calls to my search providers) for seasons that I would never end up accepting.

The other option would be to use the `series` config to control the `next_series_seasons` and `next_series_episodes`, which does make sense to me, but would be a change to existing config properties.

liiight:
Bah, having the same config setup in more than one plugin sucks, but I guess it's an edge case.

liiight:
Cool. Add to wiki?

jawilson:
Finally got around to adding it to the wiki.

PR #1884, opened 2017-06-21
Added urlrewrite plugin for Rlsbb

thawn:
### Motivation for changes:
rlsbb does not provide all possible download links (notably rapidgator links are missing) in their rss feed.
This urlrewrite plugin grabs additional links from the rlsbb.ru website and adds them to the urls field of the entry.

This plugin works similar to the rmz urlrewrite plugin which was just pulled recently:
#1879

### Detailed changes:
    rlsbb.ru urlrewriter
    Version 0.1

    Configuration

    rlsbb:
      filehosters_re:
        - domain\.com
        - domain2\.org
      link_text_re:
        - UPLOADGiG
        - NiTROFLARE
        - RAPiDGATOR
      parse_comments: no

    filehosters_re: Only add links that match any of the regular expressions 
      listed under filehosters_re.
    link_text_re: search for <a> tags where the text (not the url) matches 
      one of the given regular expressions. The href property of these <a> tags
      will be used as the url (or urls).
    parse_comments: whether the plugin should also parse the comments or only 
      the main post. Note that it is highly recommended to use filehosters_re
      if you enable parse_comments. Otherwise, the plugin may return too
      many and even some potentially dangerous links. 

    'urls' field of the entry.

    exec:

```
    rlsbb:
      filehosters_re:
        - domain\.com
        - domain2\.org
      link_text_re:
        - UPLOADGiG
        - NiTROFLARE
        - RAPiDGATOR
      parse_comments: no
```
```
DEBUG  rlsbb  TASK  Searching <entry url> for a tags where the text matches one of: <link_text_re>
DEBUG  rlsbb  TASK  Found link elements: <html <a> elements>
DEBUG  rlsbb  TASK  Comment parsing enabled: found <number> comments
WARNING  rlsbb  TASK  You have enabled comment parsing but you did not define any filehoster_re filter. You may get a lot of unwanted and potentially dangerous links from the comments.
DEBUG  rlsbb  TASK  Using filehosters_re filters: <list of filter expressions>
DEBUG  rlsbb  TASK  Url: "<link found at entry url>" matched filter: <expression>

VERBOSE  rlsbb  TASK  Found <number> links at <entry url>
```
#### To Do:

- [ ] Add ability to work as a search plugin.

### DIFF
--- a/flexget/plugins/sites/rlsbb.py
+++ b/flexget/plugins/sites/rlsbb.py
@@ -1,0 +1,174 @@
+from __future__ import unicode_literals, division, absolute_import
+from builtins import *  # noqa pylint: disable=unused-import, redefined-builtin
+
+import logging
+import re
+
+from flexget import plugin
+from flexget.event import event
+from flexget.plugins.internal.urlrewriting import UrlRewritingError
+from flexget.utils.soup import get_soup
+from flexget.utils.search import normalize_unicode
+
+from requests.exceptions import RequestException
+
+log = logging.getLogger('rlsbb')
+
+
+class UrlRewriteRlsbb(object):
+    """
+    rlsbb.ru urlrewriter
+    Version 0.1
+
+    Configuration
+
+    rlsbb:
+      filehosters_re:
+        - domain\.com
+        - domain2\.org
+      link_text_re:
+        - UPLOADGiG
+        - NiTROFLARE
+        - RAPiDGATOR
+      parse_comments: no
+
+    filehosters_re: Only add links that match any of the regular expressions 
+      listed under filehosters_re.
+    link_text_re: search for <a> tags where the text (not the url) matches 
+      one of the given regular expressions. The href property of these <a> tags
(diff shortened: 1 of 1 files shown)


flexget/plugins/sites/rlsbb.py
```diff
@@ -0,0 +1,174 @@
...
+        if num_links:
+            entry['urls'] = urls
+            entry['url'] = urls[0]
+        else:
+            raise UrlRewritingError('No useable links found at %s' % entry['url'])
+
+
+@event('plugin.register')
+def register_plugin():
+    plugin.register(UrlRewriteRlsbb, 'rlsbb', interfaces=[
```
liiight: Was this line messing pep8? Doesn't seen that long to me

thawn: I guess this was a relic from before I set pep8 line width to 120. Fixed it now.

thawn:
Thanks a lot!

PR #1886, opened 2017-06-22
rmz urlrewriter: fixed bug ...

thawn:
.. that may lose links if another plugin adds links to the urls field of an entry before the urlrewriter is executed.
So far this only occurred to me on another site (rlsbb) where the rss plugin already finds urls in enclosures. So this bug does not currently affect users but it might in the future, if rmz adds enclosures.

### Motivation for changes:
bugfix

### Detailed changes:
- the plugin now checks whether there is already an `urls` field in the entry and adds the contained links to its list of urls to process.


thawn:
Thanks :)

PR #1887, opened 2017-06-26
updated the nyaa plugin to use the new url.

razage:
nyaatorrents.se voluntarily took itself offline because of some rumors that the "police" were coming after them. a few alternate versions popped up shortly after, and this seems to be the de facto one. I simply took the old nyaa plugin and updated the urls and values wherever necessary.

### Motivation for changes:
The old nyaa plugin was using a url that no longer is online. 

### Detailed changes:
- I changed the category dictionary's values because they use different numbers.
- Their url is completely different, so I had to move around where the values get plugged in.
- I added a ".torrent" append to the url_rewrite function because the new site has a different url structure.

### DIFF
--- a/flexget/plugins/sites/nyaa.py
+++ b/flexget/plugins/sites/nyaa.py
@@ -16,10 +16,10 @@ log = logging.getLogger('nyaa')
 # TODO: Other categories
 CATEGORIES = {'all': '0_0',
               'anime': '1_0',
-              'anime eng': '1_37',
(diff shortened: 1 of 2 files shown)


liiight:
You need to update tests as well

razage:
Sorry about that. Got the test updated and looks like it passed.

paranoidi:
@razage AWESOME job. I was checking couple weeks ago that this would be the best thing to do but had no time ...

PR #1894, opened 2017-07-17
[add] Quality requirements per list item in movie lists

h0od:
### Motivation for changes:
Sometimes the quality requirements that you expect for a movie differs from movie to movie. To achieve this today, we could manage different lists and tasks for each quality requirement. However, it would be a lot easier if each movie list entry should hold the quality requirements for that movie title, only requiring one list to manage and match.

### Detailed changes:
- In the database, the entry's fields are now saved along with the list item for the movie_list plugin. The affected database table is ´movie_list_movies´ and the new column is called 'json' (string). This is the same behavior that we can observe in the `entry_list` plugin, where all the entry fields are serialized to a JSON string.
- Due to the changes in the database schema, upgrade logic for the database schema has been implemented and and version for the ´movie_list´ plugin has been bumped to "1". Any list items during upgrade will have its "json" column set to empty string which will basically mean that it will pass any quality requirement checks.
- Added property `check_quality` (default 'false') for `movie_list` that is only applicable when using it in a `list_match` context. If set to true in that context, the movie list item's entry fields are scanned for a field named `quality_req` that can hold one or several quality requirement strings. If such field exists, we only accept the entry if its quality satisfies any of quality requirements in the field. If no such field exists, we consider the quality requirement as 'any'.
- As most of the list plugins today do not add a ´quality_req´ field to the entries I felt the need of a way to force the quality requirements when adding an item to a movie list. For that very reason, I added a new `force_quality_req` property for `movie_list`. `force_quality_req` is only valid in the context of a `list_add` output action. Using this, a 'quality_req' field is added to the entry with the provided requirements (and overrides any present 'quality_req' field value if present)
- The movie-list CLI 'list' command also outputs the quality requirement for each list item in the terminal as a new column ("Quality requirement")
- Modified the `test_list_interface.py` test so that we test the new functionality

### Config usage
```
  # This task adds all entries from trakt to the list
  movie_list_add_new:
    priority: 1
    no_entries_ok: yes
    trakt_list:
      username: MYUSERNAME
      account: MYUSERNAME
      list: flexget_1080
      type: movies
    seen: local
    accept_all: yes
    list_add:
      - movie_list:
          list_name: wanted_movies
          force_quality_req:
            - 1080p bluray

  test_match:
    priority: 3
    text:
      url: <FILE>
      entry:
        title: title:\s(.*)
        url: url:\s(.*)
    imdb_lookup: yes
    list_match:
      from:
        - movie_list:
            list_name: wanted_movies
            check_quality: yes
```
### Up for discussion:

- Although all entry's fields are stored in the 'json' column for each list entry, we're only actually interested in the 'quality_req' field at this point. My motivation for this is: (1) to keep consistency with the behavior of entry_list and (2) to enable other fields to be filtered by the list_match+movie_list plugin in the future. Do you think this is a good idea or it is just wasteful/over-designed?
- Maybe this whole quality requirements thing should be implemented in a generic way for lists? For example by using the list interface so that we can enable all list plugins to provide quality requirements in a generic way and evaluate them in the list_match plugin.

### DIFF
--- a/flexget/plugins/cli/movie_list.py
+++ b/flexget/plugins/cli/movie_list.py
@@ -15,6 +15,30 @@ from flexget.plugins.list.movie_list import get_list_by_exact_name, get_movie_li
 from flexget.terminal import TerminalTable, TerminalTableError, table_parser, console
 from flexget.utils.tools import split_title_year
 
+def list_to_comma_separated_string(the_list):
+    """ Returns a comma separated string of a list """
+
+    # If the input is not a list, convert it to a list
+    if not isinstance(the_list, list):
+        return str(the_list)
+
+    return ', '.join(the_list)
+
+# These are custom fields that are printed during the "list" command.
+# The field is queried from the custom data entry stored with each
+# movie entry in the database.
+CUSTOM_FIELDS = [
+    # Structure of each tuple
+    #
+    # Index 0: Header name
+    # Index 1: Attribute name in the custom data entry
+    # Index 2: Default display value (if attribute is not found)
+    # Index 3: (optional) Callable that transform the value in 
+    #          the attribute to something more humanly readable
+    #          in the printed cell
+
+    ('Quality requirement', 'quality_req', '', list_to_comma_separated_string), 
+]
 
 def lookup_movie(title, session, identifiers=None):
     try:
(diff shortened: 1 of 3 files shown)


liiight:
We really appreciate the improvement suggestion and the fact that you went through the trouble of actually submitting a PR, but this is by design. The old implementation of `movie_list` (which was called `movie_queue`) did exactly that.

Lists are by design, metadata agnostic, relating only for the specific information they require in order to be customized (for example, with movie list it's `movie_name`, `movie_year` and optional additional 3rd party metadata like IMDB, TMDB, etc).

So unfortunately, we have to reject this.
HOWEVER, we have a bunch of other stuff we could use your help with, come hit us up in [gitter](https://gitter.im/Flexget/Flexget) 👍

PR #1895, opened 2017-07-18
Update sonarr_list.py

kjp4756:
configure_series_qulity, configure_series_qualities, and configure_series_path should NOT be altered unless INCLUDE_DATA is TRUE.

### Motivation for changes:
When NOT using INCLUDE_DATA configure_series quality settings do not work.  This is because sonarr_list clears configure_series_quality when not using INCLUDE_DATA
### Detailed changes:
- Add an if statement.


liiight:
Thanks for this. Nice to know it's even in use

PR #1897, opened 2017-07-22
[add] Managed list plugin for Radarr (radarr_list)

h0od:
### Motivation for changes:
A list plugin similar to the already existing Sonarr plugin but for Radarr which targets movies instead of TV shows.

The plugin acts both as an input plugin and as a managed list plugin.

### Detailed changes:
- A simple test is included which tests the input plugin functionality and all the list actions (add, remove, clear, match). A VCR cassette for the HTTP communication during test is also provided, so no live Radarr server is required during testing. **Be careful if running the test against a live server because all the movies will be cleared by the list_clear action**
- When matching entries, we compare in this order (1) IMDB and TMDB id's, (2) movie_name+movie_year, (3) title
- If setting `include_data: yes` when reading the list, all the entries generated will be decorated with attribute `quality_req`, which holds a set of quality requirements. These are transformed from the quality profile of the movie in Radarr to Fleget's own quality requirements. The simple mapping can be found [here](https://github.com/h0od/Flexget/commit/b8dc31cf5a9c0bd8b5797df0156d8f9bc21abc22#diff-10c57dd300d3be6f86a888887b303271R15)
- In Radarr, a quality profile can have a 'cut-off' quality. I realised as Flexget doesn't have a quality target/timeframe functionality for movies that the user might want to limit the quality requirements to just the cutoff quality. So by setting `only_use_cutoff_quality: yes`, the ´quality_req´ attribute is only set to the cut-off quality of the profile, and not all quality requirements.

#### Up for discussion:

- When using `list_add` to add entries to Radarr, the current solution provides no way to specify what quality profile the movie should be added with. We just add it to the default quality profile. Do we need such possibility?
- Is there a better way to map the [quality requirements](https://github.com/h0od/Flexget/commit/b8dc31cf5a9c0bd8b5797df0156d8f9bc21abc22#diff-10c57dd300d3be6f86a888887b303271R15)? 
- Because the entires in Radarr has a quality requirement, shouldn't we consider the quality of the entry when matching the entries in the list?

- Feathub #[8](http://feathub.com/Flexget/Flexget/+8).

### Config usage
```
        tasks:
          clear_and_add_to_radarr_list:
            list_clear:
              what:
                - radarr_list:
                    base_url:  http://...
                    api_key: ...
                    port: 7878
            mock:
              - { title: 'Despicable Me 2 (2013)', imdb_id: 'tt1690953', tmdb_id: 93456 }
              - { title: 'Sinister 2 (2015)', imdb_id: 'tt2752772', tmdb_id: 283445 }
              - { title: 'Crimson Peak (2015)', imdb_id: 'tt2554274', tmdb_id: 201085 }
              - { title: 'Deadpool (2016)', imdb_id: 'tt1431045', tmdb_id: 293660 }
            accept_all: yes
            list_add:
              - radarr_list:
                  base_url:  http://...
                  api_key: ...
                  port: 7878

          radarr_list_as_input_plugin:
            radarr_list:
              base_url:  http://...
              api_key: ...
              port: 7878
              include_data: True
            accept_all: yes

          remove_from_radarr_list:
            mock:
              - { title: "Ocean's Twelve (2004)", imdb_id: 'tt0349903', tmdb_id: 163 }
              - { title: 'Sinister 2 (2015)', imdb_id: 'tt2752772', tmdb_id: 283445 }
            accept_all: yes
            list_remove:
              - radarr_list:
                  base_url:  http://...
                  api_key: ...
                  port: 7878

          match_radarr_list:
            mock:
              - { title: 'Despicable.Me.2.2013.1080p.BluRay.x264-VeDeTT', imdb_id: 'tt1690953', tmdb_id: 93456 }
              - { title: 'Sinister.2.2015.720p.BluRay.x264-BLOW', imdb_id: 'tt2752772', tmdb_id: 283445 }
              - { title: 'Crimson.Peak.2015.720p.BluRay.x264-SPARKS', imdb_id: 'tt2554274', tmdb_id: 201085 }
              - { title: 'Deadpool.2016.1080p.BluRay.x264-SPARKS', imdb_id: 'tt1431045', tmdb_id: 293660 }
              - { title: 'Kung.Fu.Panda.3.2016.720p.BluRay.x264-SPARKS', imdb_id: 'tt2267968', tmdb_id: 140300 }
            list_match:
              from:
                - radarr_list:
                    base_url:  http://...
                    api_key: ...
                    port: 7878
````

### DIFF
--- a/flexget/plugins/list/radarr_list.py
+++ b/flexget/plugins/list/radarr_list.py
@@ -1,0 +1,378 @@
+from __future__ import unicode_literals, division, absolute_import
+from builtins import *
+
+import logging
+from collections import MutableSet
+
+from flexget import plugin
+from flexget.entry import Entry
+from flexget.event import event
+from flexget.utils.qualities import Requirements
+from flexget.utils.radarr import RadarrAPIService
+
+# Maps (lowercase) Radarr qualities to flexget
+# quality reqirement strings
+QUALITIES_MAP = {
+    'workprint': 'workprint',
+    'cam': 'cam',
+    'telesync': 'ts',
+    'telecine': 'tc',
+    'dvdscr': 'dvdscr',
+    'sdtv': 'sdtv',
+    'dvd': 'dvdrip', # not completely correct
+    'dvd-r': 'dvdrip', # not completely correct
+    'webdl-480p': 'webdl 480p',
+    'bluray-480p': 'bluray 480p',
+    'bluray-576p': 'bluray 576p',
+    'hdtv-720p': 'hdtv 720p',
+    'webdl-720p': 'webdl 720p',
+    'bluray-720p': 'bluray 720p',
+    'hdtv-1080p': 'hdtv 1080p',
+    'webdl-1080p': 'webdl 1080p',
+    'bluray-1080p': 'bluray 1080p',
+    'remux-1080p': 'remux 1080p',
+    'hdtv-2160p': 'hdtv 2160p',
+    'webdl-2160p': 'webdl 2160p',
+    'bluray-2160p': 'bluray 2160p',
+    'remux-2160p': 'remux 2160p',
+    'br-disk': 'remux', # not completely correct
+    'raw-hd': 'remux', # not completely correct
+
+    # No idea of how to map these:
+    # 'regional': 'UNKNOWN'
+}
+
+log = logging.getLogger('radarr_list')
+
+def radarr_quality_to_flexget_quality_req(radarr_quality):
+    """
+    Translates the provided Radarr quality string to a Flexget Requirement instance.
+    Returns None if translation is unsuccessful
+    """
+
+    # QUALITIES_MAP has its keys in lower case
+    radarr_quality = radarr_quality.lower()
+    if not radarr_quality in QUALITIES_MAP:
+        log.warning(
+            'Did not find a suitible translation for Radarr quality \'%s\'',
+            radarr_quality)
+        return None
+
+    flexget_quality_req_string = QUALITIES_MAP[radarr_quality]
+
+    try:
+        return Requirements(flexget_quality_req_string)
+    except ValueError:
+        log.error(
+            'Failed to convert %s into a valid quality requirement',
+            flexget_quality_req_string)
+        return None
+def get_flexget_qualities(profile, cutoff_only=False):
+    quality_requirements = []
+    if cutoff_only:
+        name = profile['cutoff']['name']
+        quality_req = radarr_quality_to_flexget_quality_req(name)
+        if not quality_req is None:
+            quality_requirements.append(quality_req)
+    else:
+        for quality in profile['items']:
+            if quality['allowed']:
+                name = quality['quality']['name']
+                quality_req = radarr_quality_to_flexget_quality_req(name)
+                if not quality_req is None:
+                    quality_requirements.append(quality_req)
+
+    return quality_requirements
+
+
+
+class RadarrSet(MutableSet):
+    """ Accesses the Radarr movies using the provided the config """
+
+    def __init__(self, config):
+        self.config = config
+        self.service = RadarrAPIService(
+            config['api_key'],
+            config['base_url'],
+            config['port']
+            )
+
+        # Class member used for caching the items to avoid
+        # unnecessary calls to the Radarr API.
+        # We use the self.items property to access it.
+        # Just set this to None again later to invalidate the cache.
+        self._movie_entries = None
+
+    def __iter__(self):
+        return iter(self.items)
+
+    def __len__(self):
+        return len(self.items)
+
+    def discard(self, value):
+        if not value:
+            return
+
+        if not isinstance(value, Entry):
+            raise TypeError()
+
+        matching_entry = self.__find_matching_entry(value)
+        if matching_entry:
+            movie_id = matching_entry['radarr_id']
+            response = self.service.delete_movie(movie_id)
+            log.verbose('Removed movie %s from Radarr', matching_entry['title'])
+            # Clear the cache
+            self._movie_entries = None
+        else:
+            log.debug('Could not find any matching movie to remove for entry %s', value)
+
+    def __ior__(self, other):
+        for entry in other:
+            if not isinstance(entry, Entry):
+                raise TypeError()
+
+            self.add(entry)
+
+    def __contains__(self, entry):
+        if not value:
+            return False
+
+        if not isinstance(value, Entry):
+            raise TypeError()
+
+        matching_entry = self.__find_matching_entry(value)
+
+        return not matching_entry is None
+
+    def add(self, entry):
+
+        if not isinstance(entry, Entry):
+            raise TypeError()
+
+        # The easiest way to add a movie to Radarr is to
+        # first use the lookup API. Using that we will get
+        # a json response which gives us most of the input
+        # we need for the PUT request.
+
+        result = None
+
+        # If the entry has a IMDB id, use that for lookup
+        if not result and 'imdb_id' in entry:
+            imdb_id = entry['imdb_id']
+            result = self.service.lookup_by_imdb(imdb_id)
+            if len(result) == 0:
+                result = None
+
+        # If the entry has a TMDB id, use that for lookup
+        if not result and 'tmdb_id' in entry:
+            tmdb_id = entry['tmdb_id']
+            result = self.service.lookup_by_tmdb(tmdb_id)
+            if len(result) == 0:
+                result = None
+
+        # Could not lookup by id. Try to use the title.
+        # However, we can only accept any results if it's
+        # one item, otherwise we don't know which to select.
+        if not result:
+            title = entry['title']
+            results = self.service.lookup_by_term(title)
+            if len(result) > 1:
+                log.debug(
+                    'The lookup for \'%s\' returned %d results. Using the first result \'%s\'.',
+                    title,
+                    len(results),
+                    results[0]['title']
+                )
+                result = results[0]
+            elif len(result) == 0:
+                result = None
+
+        if result:
+            rootfolders = self.service.get_rootfolders()
+            rootFolderPath = rootfolders[0]['path']
+
+            # TODO: should we let the user affect this one, 
+            # or try to parse the 'quality' value somehow?
+            qualityProfileId = 1
+
+            try:
+                add_result = self.service.add_movie(
+                    result['title'],
+                    qualityProfileId,
+                    result['titleSlug'],
+                    result['images'],
+                    result['tmdbId'],
+                    rootFolderPath)
+                log.verbose('Added movie %ls to Radarr list', result['title'])
+            except RadarrMovieAlreadyExistsError:
+                log.warning(
+                    'Could not add movie %ls because it already exists on Radarr',
+                    result['title']
+                    )
+        else:
+            log.verbose(
+                'The lookup for entry %s did not return any results.'
+                'Can not add the movie in Radarr.',
+                entry)
+
+    def _from_iterable(self, it):
+        # The following implementation is what's done in every other
+        # list plugin. Does not makes sense.
+        # As I understand it, it aims to be an override to base class Set._from_iterable.
+        # However, thats a classmethod and this does not even match its signature
+        # https://blog.devzero.com/2013/01/28/how-to-override-a-class-method-in-python/
+
+        # TODO: is this the right answer? the returned object won't have our custom __contains__ logic
+        return set(it)
+
+    def get(self, entry):
+        # Here we must return a matching entry using the provided the "hint" entry
+        # In Radarr's case we will check different movie ids and then resort to name/year.
+        return self.__find_matching_entry(entry)
+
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

File: flexget/plugins/list/radarr_list.py [hunk 1]
```diff
@@ -0,0 +1,378 @@
...
+    """
+    Translates the provided Radarr quality string to a Flexget Requirement instance.
+    Returns None if translation is unsuccessful
+    """
+
+    # QUALITIES_MAP has its keys in lower case
+    radarr_quality = radarr_quality.lower()
+    if not radarr_quality in QUALITIES_MAP:
+        log.warning(
+            'Did not find a suitible translation for Radarr quality \'%s\'',
```
liiight: Weird indentation...

flexget/plugins/list/radarr_list.py
```diff
@@ -0,0 +1,378 @@
...
+            radarr_quality)
+        return None
```
liiight: No need to explicitly return 'None` in python, just, `return`

flexget/plugins/list/radarr_list.py
```diff
@@ -0,0 +1,378 @@
...
+        log.error(
+            'Failed to convert %s into a valid quality requirement',
+            flexget_quality_req_string)
+        return None
+def get_flexget_qualities(profile, cutoff_only=False):
+    quality_requirements = []
+    if cutoff_only:
+        name = profile['cutoff']['name']
+        quality_req = radarr_quality_to_flexget_quality_req(name)
+        if not quality_req is None:
```
liiight: If you wanna look for the opposite of `None` you could just do `if quality_req`. That's the more pythonic approach to it

flexget/plugins/list/radarr_list.py
```diff
@@ -0,0 +1,378 @@
...
+
+    def __len__(self):
+        return len(self.items)
+
+    def discard(self, value):
+        if not value:
+            return
+
+        if not isinstance(value, Entry):
+            raise TypeError()
```
liiight: Is this condition even possible? Also, if you want to raise exception, see if `plugin.PluginError` and `plugin.PluginWarning` can suit you since they are handles elegantly by the plugin system (PluginError will abort task)

flexget/plugins/list/radarr_list.py
```diff
@@ -0,0 +1,378 @@
...
+    def __contains__(self, entry):
+        if not value:
+            return False
+
+        if not isinstance(value, Entry):
+            raise TypeError()
+
+        matching_entry = self.__find_matching_entry(value)
+
+        return not matching_entry is None
```
liiight: Pep8 dictates `return matching_entry is not None`

flexget/plugins/list/radarr_list.py
```diff
@@ -0,0 +1,378 @@
...
+
+        if not isinstance(entry, Entry):
+            raise TypeError()
+
+        # The easiest way to add a movie to Radarr is to
+        # first use the lookup API. Using that we will get
+        # a json response which gives us most of the input
+        # we need for the PUT request.
+
+        result = None
```
liiight: I gather from the next few lines that this is expected to be a list, so init it as one.

flexget/plugins/list/radarr_list.py
```diff
@@ -0,0 +1,378 @@
...
+
+        # If the entry has a IMDB id, use that for lookup
+        if not result and 'imdb_id' in entry:
```
liiight: I don't see point on checking for result here as it was init a line before. Also, I'd do the check for imdb and tmdb ids with an `elif`

flexget/plugins/list/radarr_list.py
```diff
@@ -0,0 +1,378 @@
...
+            imdb_id = entry['imdb_id']
+            result = self.service.lookup_by_imdb(imdb_id)
+            if len(result) == 0:
+                result = None
+
+        # If the entry has a TMDB id, use that for lookup
+        if not result and 'tmdb_id' in entry:
+            tmdb_id = entry['tmdb_id']
+            result = self.service.lookup_by_tmdb(tmdb_id)
+            if len(result) == 0:
```
liiight: No need for this, checking `if list_name` will return False if list is empty

flexget/plugins/list/radarr_list.py
```diff
@@ -0,0 +1,378 @@
...
+                    title,
+                    len(results),
+                    results[0]['title']
+                )
+                result = results[0]
+            elif len(result) == 0:
+                result = None
+
+        if result:
+            rootfolders = self.service.get_rootfolders()
```
liiight: Camel case

flexget/plugins/list/radarr_list.py
```diff
@@ -0,0 +1,378 @@
...
+            rootFolderPath = rootfolders[0]['path']
+
+            # TODO: should we let the user affect this one, 
+            # or try to parse the 'quality' value somehow?
+            qualityProfileId = 1
```
liiight: This is if be for now. Let's discuss online how and if it should be handled differently

flexget/utils/radarr.py
```diff
@@ -0,0 +1,208 @@
...
+    if error_message.lower() == 'this movie has already been added':
+        return RadarrMovieAlreadyExistsError()
+
+    return None
+
+def request_get_json(url, headers):
+    """ Makes a GET request and returns the JSON response """
+    try:
+        response = requests.get(url, headers=headers)
+        if response.status_code == 200:
```
liiight: You can replace this with `response.raise_for_status()`

flexget/utils/radarr.py
```diff
@@ -0,0 +1,208 @@
...
+        response = requests.post(url, headers=headers, data=data)
+        if response.status_code == 201:
+            return response.json()
+        else:
+            errorMessage = None
+            try:
+                json_response = response.json()
+                if len(json_response) > 0:
+                    errorMessage = json_response[0]['errorMessage']
+            except:
```
liiight: Never do this... At the very minimum add some logs here, but try to avoid the generic except / pass paradigm like an STD...

flexget/utils/radarr.py
```diff
@@ -0,0 +1,208 @@
...
+            raise RadarrRequestError(
+                'Invalid response received from Radarr: %s' % 
+                response.content,
+                log,
+                status_code=response.status_code,
+                error_message=errorMessage)
+
+    except RequestException as e:
+        raise RadarrRequestError('Unable to connect to Radarr at %s. Error: %s' % (url, e))
+def request_put_json(url, headers):
```
liiight: Maybe a base class for radarr api calls would work well here, look into `api_tvdb` plugin for an example

flexget/plugins/list/radarr_list.py
h0od: Yep, figured that out a bit to late. This one I forgot to change.

flexget/plugins/list/radarr_list.py
h0od: Do you mean the arguments of the log.warning() method? Do you rather want to keep the first argument on line 56?

flexget/plugins/list/radarr_list.py
h0od: Interesting. Didn't know this. Thought a function returned 'undefined' if I just 'return'.

flexget/plugins/list/radarr_list.py
h0od: Well, I thought, because nothing in Python is really typed I would need to make sure that the item provided is of the correct type in runtime to avoid unhandled exceptions later. But I can't see any other plugin doing these checks so I will just remove them in terms of consistency with the rest of the code base.

flexget/plugins/list/radarr_list.py
h0od: That's strange, because __find_matching_entry can return None if nothing is found. And I want the contains method to return a boolean value. Is this a false positive from Pep8?

flexget/plugins/list/radarr_list.py
h0od: No actually not, if you read a bit further you will see that result is set to a single result entry :)
lookup_by_imdb() returns a single result item
lookup_by_tmdb() returns a single result item
lookup_by_term() returns a list, but as you can see the code assigns `result = results[0]`

flexget/plugins/list/radarr_list.py
h0od: You're right about the 'not result'.
But the risk of doing an elif here is that if the IMDB service is down we will skip the lookup in TMDB.

flexget/plugins/list/radarr_list.py
cvium: It's just a matter of ordering. `matching_entry is not None` is easier to read than `not matching_entry is None`.

flexget/utils/radarr.py
h0od: Valid point :) json() apparently raises ValueError if content could not be decoded.

flexget/plugins/list/radarr_list.py
h0od: @cvium Ah, makes sense!

flexget/utils/radarr.py
h0od: Can you please clarify what you mean with this?

flexget/plugins/list/radarr_list.py
liiight: Yeah. If it doesn't validate pep8 line length (108 chars if I'm not mistaken, which I very well may be) no need to indent

flexget/plugins/list/radarr_list.py
liiight: Yep. No such thing as undefined in python

flexget/plugins/list/radarr_list.py
liiight: Defensive is a valid approach, but it feels a little like overkill in an internal plugin method.

flexget/plugins/list/radarr_list.py
liiight: Then why are you checking its length?

h0od: Because if no match is found the API return an empty json response == empty dictionary. Can I check for False instead? With other words, will `if {}:` evaluate False?

h0od: Turns out it does :) Changing to `if result:`!

flexget/plugins/list/radarr_list.py
```diff
@@ -0,0 +1,378 @@
...
+class RadarrList(object):
+    """ List plugin for Radarr that also works as an input plugin """
+
+    schema = {
+        'type': 'object',
+        'properties': {
+            'base_url': {'type': 'string'},
+            'port': {'type': 'number', 'default': 80},
+            'api_key': {'type': 'string'},
+            'only_monitored': {'type': 'boolean', 'default': True},
```
jacaru: Good enhancements here, that would need to be ported to sonarr_list also, would be:

- monitored, type string, values yes, no, any, default any (replacing only_monitored)
- profile, type string, no default (default add profile)

It never made sense to me to handle the quality profiles as an input plugin but I guess somebody has a use for that.

flexget/plugins/list/radarr_list.py
```diff
@@ -0,0 +1,378 @@
...
+            try:
+                add_result = self.service.add_movie(
+                    result['title'],
+                    qualityProfileId,
+                    result['titleSlug'],
+                    result['images'],
+                    result['tmdbId'],
+                    rootFolderPath)
+                log.verbose('Added movie %ls to Radarr list', result['title'])
+            except RadarrMovieAlreadyExistsError:
```
jacaru: The sonarr_list plugin checks if the entry is already added by first fetching the list from sonarr and does not fail if that is this case.  Do we really want to behave different here?

jacaru: Exceptions not imported.

cvium:
Status on this @h0od?

flexget/tests/test_radarr_list.py
```diff
@@ -0,0 +1,117 @@
...
+        task = execute_task('match_radarr_list')
+        assert task.find_entry('accepted', title='Despicable.Me.2.2013.1080p.BluRay.x264-VeDeTT'), \
+          "movie should have been matched but it wasn't"
+        assert task.find_entry('accepted', title='Crimson.Peak.2015.720p.BluRay.x264-SPARKS'), \
+          "movie should have been matched but it wasn't"
+        assert task.find_entry('accepted', title='Deadpool.2016.1080p.BluRay.x264-SPARKS'), \
+          "movie should have been matched but it wasn't"
+        assert task.find_entry('undecided', title='Sinister.2.2015.720p.BluRay.x264-BLOW'), \
+          "movie should not have been matched but it was"
+        assert task.find_entry('undecided', title='Kung.Fu.Panda.3.2016.720p.BluRay.x264-SPARKS'), \
```
paranoidi: No release names in the tests

paranoidi:
Forked and opened new PR#2293 with corrections.

PR #1898, opened 2017-07-23
Symlink: Added support for hardlinking files and dirs

cvium:
### Motivation for changes:
I like hardlinks and Plex doesn't
### Detailed changes:
- Changed Symlink plugin to take a 'type' param ('soft' or 'hard').

### DIFF
--- a/flexget/plugins/output/symlink.py
+++ b/flexget/plugins/output/symlink.py
@@ -88,6 +88,7 @@ class Symlink(object):
             for f in files:
                 src_file = os.path.join(root, f)
                 dst_file = os.path.join(dest_dir, f)
+                log.debug('Hardlinking %s to %s', src_file, dst_file)
                 os.link(src_file, dst_file)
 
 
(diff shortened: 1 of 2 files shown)


flexget/plugins/output/symlink.py
```diff
@@ -19,37 +19,88 @@ class Symlink(object):
...
+            for f in files:
+                src_file = os.path.join(root, f)
+                dst_file = os.path.join(dst_dir, f)
+                log.debug('Hardlinking %s to %s', src_file, dst_file)
+                try:
+                    os.link(src_file, dst_file)
+                except OSError as e:
+                    log.debug('Failed to create hardlink for file %s: %s', f, e)
+                    if existing == 'fail':
+                        raise  # reraise to fail the entry in the calling function
```
liiight: Maybe raise PluginError?

cvium: Dunno, failing the entry feels more appropriate

PR #1899, opened 2017-07-24
New plugin: abort_exists

cvium:
### Motivation for changes:
I wants it. When using LFTP or another program to download from a remote location, it can ruin a lot if FlexGet moves files before the transfer is completed. A lot of it can be handled by using "incomplete" filenames in LFTP, but there are situations where I want the transfer to be fully completed before moving.

### Detailed changes:
- Added new plugin

### DIFF
--- a/flexget/plugins/filter/abort_exists.py
+++ b/flexget/plugins/filter/abort_exists.py
@@ -29,6 +29,7 @@ class PluginAbortExists(object):
             # if pattern matches any entry
             if field not in entry:
                 log.debug('Field %s not found. Skipping.', field)
+                continue
             if abort_re.search(entry[field]):
                 task.abort('An entry contained %s in field %s. Abort!' % (config['regexp'], field))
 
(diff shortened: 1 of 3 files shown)


flexget/plugins/filter/abort_exists.py
```diff
@@ -0,0 +1,37 @@
...
+
+    # Execute as the first thing in filter phase
+    @plugin.priority(255)
+    def on_task_filter(self, task, config):
+        abort_re = re.compile(config['regexp'], re.IGNORECASE)
+        field = config['field']
+        for entry in task.all_entries:
+            # if pattern matches any entry
+            if field not in entry:
+                log.debug('Field %s not found. Skipping.', field)
```
liiight: Continue?

flexget/plugins/filter/abort_exists.py
```diff
@@ -0,0 +1,37 @@
...
+            if abort_re.search(entry[field]):
+                task.abort('An entry contained %s in field %s. Abort!' % (config['regexp'], field))
```
liiight: You can silently abort, bit sure if it fits here though

cvium: Dunno. I kind of want logging but I also don't want to add a separate log entry.

PR #1903, opened 2017-08-01
[add] urlrewrite plugin for allyoulike.com

thawn:
### Motivation for changes:
Make allyoulike.com work with the html plugin.

### Detailed changes:
    allyoulike.com urlrewriter
    Version 0.1

    Rewrites urls for allyoulike.com

    On allyoulike, each link points to a page which contains the links to the actual files. 
    Often these pages contain links to more than one quality (for example 1080p, 720p etc.).
    The plugin chooses the set of links with the most files which should translate into the
    Best quality movie.

    There is no configuration for this plugin.

    'urls' field of the entry.

    exec:

    The plugin is intended to be used in conjunction with the html plugin.

    Example configuration for the html plugin:

    html:
      url: "http://www.allyoulike.com/category/movies/"
      title_from: link
      links_re:
        - allyoulike.com/\d*/(?!renew-or-purchase|for-vip-members-only)[^/]*/$

```
    rlsbb:
      filehosters_re:
        - domain\.com
        - domain2\.org
      link_text_re:
        - UPLOADGiG
        - NiTROFLARE
        - RAPiDGATOR
      parse_comments: no
```
```
VERBOSE  rlsbb  TASK  Found <number> links at <entry url>
```
#### To Do:

- [ ] Add ability to work as a search plugin.
- [ ] Add configuration option to choose quality (ideally would grab the configuration from the quality plugin and use that).

### DIFF
--- a/flexget/plugins/sites/allyoulike.py
+++ b/flexget/plugins/sites/allyoulike.py
@@ -1,0 +1,107 @@
+from __future__ import unicode_literals, division, absolute_import
+from builtins import *  # noqa pylint: disable=unused-import, redefined-builtin
+
+import logging
+import re
+
+from flexget import plugin
+from flexget.event import event
+from flexget.plugins.internal.urlrewriting import UrlRewritingError
+from flexget.utils.soup import get_soup
+from flexget.utils.search import normalize_unicode
+
+from requests.exceptions import RequestException
+
+log = logging.getLogger('rlsbb')
+
+
+class UrlRewriteAllyoulike(object):
+    """
+    allyoulike.com urlrewriter
+    Version 0.1
+
+    Rewrites urls for allyoulike.com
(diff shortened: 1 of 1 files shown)


thawn:
thanks :-)

liiight:
Thank you. Care to update the wiki?

thawn:
The plugin itself does not have any configuration options. Should I 
write a documentation for the plugin itself or should I write a working 
example configuration on how to get allyoulike.com to work with the html 
and exec plugins (using jdownloader to handle the actual downloads).

liiight:
i lean towards the latter

thawn:
Here is a documentation of an example configuration:
https://flexget.com/Cookbook/Urlrewrite/allyoulike

PR #1904, opened 2017-08-01
[fix] bug in rlsbb.py that occurred when the entry had no urls field

thawn:
### Motivation for changes:
A bug occurred in rlsbb.py when an `entry` had no `urls` field.

### Addressed issues:
- Fixes a bug I just recently found.


thawn:
thanks :-)

PR #1905, opened 2017-08-01
[fix] bug in rmz.py that caused an error if the entry does not contai…

thawn:
### Motivation for changes:
A bug occurred in rmz.py when an `entry` had no `urls` field.

### Addressed issues:
- Fixes a bug I just recently found.


thawn:
thanks :)

PR #1967, opened 2017-09-15
Torrent match plugin (for cross-seeding)

cvium:
### Motivation for changes:
I want more automation in my life

### Example config
```
  auto:
    # get some torrents
    discover:
      release_estimations: ignore
      what:
        - entry_list: blabla
      from:
        - list of trackers
    # filters
    if:
      - "some condition": accept
    ...
    other_filters...
    ...
    # match entries with local files
    torrent_match:
      what:
        - entry_list: blabla
        - filesystem: blabla
```
### TODO
- [x] Search for `torrent_name`-dir in local dir recursively. Eg. `Some.Torrent.Name` -> Search in this structure:
```
some_dir:
  some_other_dir:
    Some.Torrent.Name:
      my.files.mkv
```
- [x] tests?


flexget/plugins/filter/torrent_match.py
```diff
@@ -0,0 +1,185 @@
...
+
+    def prepare_config(self, config):
+        if not isinstance(config['max_size_difference'], float):
+            config['max_size_difference'] = float(config['max_size_difference'].rstrip('%'))
+
+        return config
+
+    # Run first, this is not really an output plugin though, but we need 'torrent' field, which is set in modify
+    @plugin.priority(255)
+    def on_task_output(self, task, config):
```
cvium: Should this run _last_ in modify phase instead?

flexget/plugins/filter/torrent_match.py
```diff
@@ -0,0 +1,185 @@
...
+                total_size = 0
+                for f in torrent_files:
+                    for local_file in local_files:
+                        # remove the root dir from the local path
+                        local_path = local_file.path
+                        if skip_root_dir:
+                            local_path = os.path.relpath(local_path, local_entry['files_root'])
+                        # if f == local_file, break out of inner loop and increment match counter
+                        # we force sizes to be exact, TODO is this correct?
+                        # TODO allow root dir to differ? Requires torrent clients have "do not add name to path"
```
cvium: I don't think any of our client plugins support the "do not add name to path" and to keep things simple I'd prefer not supporting it anyway.

flexget/plugins/filter/torrent_match.py
```diff
@@ -0,0 +1,185 @@
...
+    """TODO"""
+
+    schema = {
+        'type': 'object',
+        'properties': {
+            'what': {'type': 'array', 'items': {
+                'allOf': [{'$ref': '/schema/plugins?phase=input'}, {'maxProperties': 1, 'minProperties': 1}]
+            }},
+            'max_size_difference': {'type': 'string', 'format': 'percent', 'default': '0%'},
+            'name_in_path': {'type': 'boolean', 'default': True}
```
cvium: See other comment about "do not add name to path"

flexget/plugins/filter/torrent_match.py
liiight: Yes

PR #1975, opened 2017-09-26
[fix] Update email.py

Soljia:
Handle unknown exceptions, and try connecting again.
Only try sending max_attempts amount of times. Default 5
Also, please double check this. This is my first time dabbling in python.

### DIFF
--- a/flexget/plugins/notifiers/email.py
+++ b/flexget/plugins/notifiers/email.py
@@ -179,7 +179,9 @@ class EmailNotifier(object):
             self.connect_to_smtp_server(config)
 
         connection_error = None
-        while True:
+        counter = 0
+        max_attempts = config.get('max_attempts') if config.get('max_attempts') else 5
+        for counter in range(max_attempts):
             try:
                 self.mail_server.sendmail(email['From'], config['to'], email.as_string())
                 break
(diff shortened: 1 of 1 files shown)


liiight:
I'm not sure I want this. what triggered you to do this? what issues did you encounter?

Soljia:
I kept getting this error when my notify plugin tried sending out an email.

unbound method sendmail() must be called with SMTP instance as first argument

I did some googling, and it seemed to be an issue with calling method from a class vs calling method from an instance of a class. I thought maybe the smtplib was changed for whatever reason, but I double checked their documentation and what Flexget's current code was...everything checked out. I was trying to figure out what was going wrong, and saw while(true) which is always suspect to me. Unless it's meant to be running 100% of the time, it shouldn't be forever looped (opinion). Now, it does seem to fix my issue, as it sends emails now. My assumption is that self.connect_to_smtp_server(config) was not fully processed?

Here is my config for the email, from what I can gather there is nothing wrong with it.
```
    notify:
      task:
        template: html
        via:
          - email:
              smtp_host: smtp.gmail.com
              smtp_port: 587
              smtp_username: *****************
              smtp_password: ****
```*************
              smtp_tls: yes
              from: *****************@gmail.com
              to: *****************@gmail.com
              html: yes

**Comment by liiight** [comment]
The whole true isn't endless, it breaks after the first success or second failure. Not sure this fix will do the trick, need to dig in some more.

If you could provide the exact crash log if your error that'll be great.

File: flexget/plugins/notifiers/email.py [hunk 1]
```diff
@@ -189,7 +191,11 @@ def notify(self, title, message, config):
                     connection_error = e
                 else:
                     raise PluginWarning('Could not connect to SMTP server: %s' % str(e))
-
+            except:
+                log.debug( "Unexpected error:", sys.exc_info()[0])
+                self.mail_server.quit() #Lets clear the connection, and try again
+                self.connect_to_smtp_server(config)
+                raise
```
cvium: Raising here doesn't make a lot of sense

PR #1977, opened 2017-09-29
Allow config to specify date format preferences for exists_series

OmgOhnoes:
### Motivation for changes:
Series identified by date are stored using YYYY-mm-dd format but when this is processed using default datetime parameters within exists_series it reads files from disc as YYYY-dd-mm thus creating false comparisons. 

This change allows the user to tell exists_series to assume dates are in a specific format on disc. Useful if you rename your series using series_id.

NOTE: this only works on the built in parser, guessit parsing is not influenced by this change

### Detailed changes:
- Added date_dayfirst and date_yearfirst to exists_series plugin

### Addressed issues:
- Fixes exists_series parsing dates as YYYY-dd-mm instead of YYYY-mm-dd

```
    exists_series:
      path: >
        /media/TV/{{ tvdb_series_name|default(series_name,True)|replace('/', '_')|replace(':', ' -')|replace(',', '') }}
      allow_different_qualities: better
      date_dayfirst: False
      date_yearfirst: True
```

### DIFF
--- a/flexget/plugins/filter/exists_series.py
+++ b/flexget/plugins/filter/exists_series.py
@@ -91,8 +91,8 @@ class FilterExistsSeries(object):
                     try:
                         disk_parser = get_plugin_by_name('parsing').instance.parse_series(data=filename.name,
                                                                                           name=series_parser.name,
-                                                                                          date_dayfirst=config['date_dayfirst'],
-                                                                                          date_yearfirst=config['date_yearfirst'])
+                                                                                          date_dayfirst=config.get('date_dayfirst'),
+                                                                                          date_yearfirst=config.get('date_yearfirst'))
                     except ParseWarning as pw:
                         disk_parser = pw.parsed
                         log_once(pw.value, logger=log)
--- a/flexget/plugins/filter/exists_series.py
+++ b/flexget/plugins/filter/exists_series.py
@@ -33,7 +33,7 @@ class FilterExistsSeries(object):
                 'type': 'object',
                 'properties': {
                     'path': one_or_more({'type': 'string', 'format': 'path'}),
-                    'allow_different_qualities': {'enum': ['better', True, False], 'default': False}
+                    'allow_different_qualities': {'enum': ['better', True, False], 'default': False},
                     'date_dayfirst': {'type': 'boolean', 'default': False},
                     'date_yearfirst': {'type': 'boolean', 'default': False}
                 },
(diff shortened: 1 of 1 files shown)


flexget/plugins/filter/exists_series.py
```diff
@@ -88,7 +90,9 @@ def on_task_filter(self, task, config):
                     # run parser on filename data
                     try:
                         disk_parser = get_plugin_by_name('parsing').instance.parse_series(data=filename.name,
-                                                                                          name=series_parser.name)
+                                                                                          name=series_parser.name,
+                                                                                          date_dayfirst=config['date_dayfirst'],
+                                                                                          date_yearfirst=config['date_yearfirst'])
```
cvium: You need to add defaults to `prepare_config` or use dict.get().

flexget/plugins/filter/exists_series.py
```diff
@@ -33,7 +33,9 @@ class FilterExistsSeries(object):
                 'type': 'object',
                 'properties': {
                     'path': one_or_more({'type': 'string', 'format': 'path'}),
-                    'allow_different_qualities': {'enum': ['better', True, False], 'default': False}
+                    'allow_different_qualities': {'enum': ['better', True, False], 'default': False},
```
cvium: Is this pep8-compliant? It looks like it's indented 1 more space than the line above it.

flexget/plugins/filter/exists_series.py
```diff
@@ -88,7 +90,9 @@ def on_task_filter(self, task, config):
                     # run parser on filename data
                     try:
                         disk_parser = get_plugin_by_name('parsing').instance.parse_series(data=filename.name,
-                                                                                          name=series_parser.name)
+                                                                                          name=series_parser.name,
```
cvium: This should be indented to just past the parenthesis in order to be pep8-compliant.

liiight:
Please add tests for this if possible

flexget/plugins/filter/exists_series.py
```diff
@@ -33,7 +33,9 @@ class FilterExistsSeries(object):
                 'type': 'object',
                 'properties': {
                     'path': one_or_more({'type': 'string', 'format': 'path'}),
-                    'allow_different_qualities': {'enum': ['better', True, False], 'default': False}
+                    'allow_different_qualities': {'enum': ['better', True, False], 'default': False},
+                    'date_dayfirst': {'type': 'boolean', 'default': False},
```
cvium: Not sure if this PR is still relevant, but setting these defaults is a bad idea. The underlying code in the parsers will try different pairs of `True` and `False` if these options are _not_ set.

paranoidi:
Not elegant to have two different options, instead we should have ONE option.

Perhaps "dateformat: yyyy-mm-dd" type of configuration?

Default value should be ISO 8610 ...

PR #2001, opened 2017-11-04
Decouple v2 webui

stevezau:
### Motivation for changes:

Start the process to de-couple the WebUI into https://github.com/Flexget/webui

### TODO

- [x]  Remove v2 React Code
- [x]  Cleanup handling when running from git
- [x]  Add readme's in each webui version
- [x]  Fix Release in Jenkins


kristenmills:
I think you forgot to remove the stuff from the .travis.yml related to v2. That's why the travis build is failing

flexget/ui/v1/__init__.py
```diff
@@ -5,7 +5,7 @@
 import os
 import fnmatch
 
-from flask import send_from_directory, Flask
+from flask import send_from_directory, Flask, abort
```
liiight: why add `abort`?

stevezau: Ah, i was using it to send a 404 but figured out a different way to handle running from github. Removed it.

flexget/ui/v2/index.html
```diff
@@ -0,0 +1,13 @@
...
+<html lang="en">
+<head>
+    <meta charset="UTF-8">
+    <title>Failed to load Flexget Web UI</title>
+</head>
+<body>
+<h1>Failed to load Flexget Web UI</h1>
+
+<p>It looks like you are running flexget from the GitHub source. If you want to use the webui you'll have to run the WebUI build steps by following
+    <a href="http://flexget.com/wiki/Web-UI">http://flexget.com/wiki/Web-UI/v2</a></p>
```
kristenmills: href points to the v1 wiki. Should be changed to v2

kristenmills: Also I need to update the v2 wiki. I'll do that today.

flexget/ui/v2/__init__.py
```diff
@@ -14,10 +14,11 @@
 
 manager = None
 debug = False
-app_base = None
+exists = True
 
 ui_base = os.path.dirname(os.path.realpath(__file__))
```
kristenmills: Can we make this configurable? If someone wants to run flexget from git, I don't think the solution should be to copy the ui folders. Maybe have some way to specify in the config where the dist folder is located?

stevezau: You can always setup a symlink. If you are really doing dev work on the WebUI you would run it separately anyway.

PR #2003, opened 2017-11-04
Gitmask Anonymous PR

gitmask-anonymous:
This is an anonymous PR submitted via Gitmask - https://www.gitmask.com

### DIFF
--- a/flexget/plugins/sites/passthepopcorn.py
+++ b/flexget/plugins/sites/passthepopcorn.py
@@ -86,6 +86,34 @@ RELEASE_TYPES = {
     'golden popcorn': 2
 }
 
+RESOLUTION_MAP = {
+    '1080p': '1080p',
+    '720p': '720p',
+    '576p': '576p',
+    '480p': '480p',
+}
+
+SOURCE_MAP = {
+    'CAM': 'cam',
+    'TS': 'ts',
+    'R5': 'r5',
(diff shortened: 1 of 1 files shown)


flexget/plugins/sites/passthepopcorn.py
```diff
@@ -275,6 +303,15 @@ def search(self, task, entry, config):
...
                     e['uploaded_at'] = dateutil_parse(torrent['UploadTime'])
+                    e['imdb_id'] = 'tt%s' % str(movie['ImdbId']).zfill(7)
+
+                    resolution = self._lookup_quality(torrent, 'Resolution', RESOLUTION_MAP)
+                    source = self._lookup_quality(torrent, 'Source', SOURCE_MAP)
+                    codec = self._lookup_quality(torrent, 'Codec', CODEC_MAP)
+                    quality = ' '.join([resolution, source, codec]).strip()
+
+                    if quality:
+                        e['quality'] = quality
```
cvium: Plugins are not really supposed to set the `quality` field themselves. That's the parser's job.

gazpachoking: I'm not sure this is really a problem. If a plugin can provide a better quality than the parser could produce I think they should. On the other hand, it should be a `Quality` instance, not a string in that case.

PR #2007, opened 2017-11-12
Added fast_resume option for rtorrent

cvium:
### Motivation for changes:
Fast resume saves a lot of CPU cycles

### Detailed changes:
- Added `fast_resume` option that adds resume data to a .torrent. **Note: it only works if flexget has direct access to the files**

### DIFF
--- a/flexget/plugins/clients/rtorrent.py
+++ b/flexget/plugins/clients/rtorrent.py
@@ -638,12 +638,13 @@ class RTorrentOutputPlugin(RTorrentPluginBase):
                 files = []
 
                 for f in entry['torrent'].get_filelist():
+                    file_path = os.path.join(base, os.path.join(f['path'], f['name']))
                     # TODO should it simply add the torrent anyway?
-                    if not os.path.exists(f['path']):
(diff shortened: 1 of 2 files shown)


flexget/plugins/clients/rtorrent.py
```diff
@@ -621,6 +627,29 @@ def add_entry(self, client, entry, options, start=True, mkdir=False):
...
+                base = options.get('directory')
+                if not base:
+                    base = client.get_directory()
+
+                piece_size = entry['torrent'].piece_size
+                chunks = int((entry['torrent'].size + piece_size - 1) / piece_size)
+                files = []
+
+                for f in entry['torrent'].get_filelist():
+                    # TODO should it simply add the torrent anyway?
```
cvium: Rutorrent will add the torrent even if fast resume fails.

stevezau:
what’s the use case for fast_resume? if it’s a new .tor wouldn’t it be empty anyway?

cvium:
I also said this in Gitter, but I will repeat it here for posterity:

@stevezau The use case for fast_resume is when you're adding a .torrent from another tracker and you already have the files

PR #2014, opened 2017-11-14
[WIP] Custom qualities plugin

cvium:
### Motivation for changes:
Quality is a complex thing. Being able to customize it without altering the parser sounds like a nice idea.

```
  some_task:
    series:
      - Some Show
    custom_quality:
      release_groups:
        - somegroup:
            score: 20
    mock:
      - {title: 'Some.Show.S01E01.1080p.hdtv-othergroup'}
      - {title: 'Some.Show.S01E01.1080p.HDTV-somegroup'}
```

### TODO

- [ ] How to handle requirements?
- [ ] How to handle quality ordering?
- [ ] Tests

### DIFF
--- a/flexget/utils/qualities.py
+++ b/flexget/utils/qualities.py
@@ -24,7 +24,7 @@ class QualityComponent(object):
         :param defaults: An iterable defining defaults for other quality components if this component matches.
         """
 
-        if type not in ['resolution', 'source', 'codec', 'audio'] + _custom_components.keys():
+        if type not in ['resolution', 'source', 'codec', 'audio'] + list(_custom_components.keys()):
             raise ValueError('%s is not a valid quality component type.' % type)
         self.type = type
         self.value = value
--- a/flexget/utils/qualities.py
+++ b/flexget/utils/qualities.py
@@ -8,6 +8,8 @@ import logging
 
 log = logging.getLogger('utils.qualities')
 
+_custom_components = {}
+
 
 class QualityComponent(object):
     """"""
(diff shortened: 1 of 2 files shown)


flexget/utils/qualities.py
```diff
@@ -253,18 +260,24 @@ def _find_best(self, qlist, default=None, strip_all=True):
...
-        name = ' '.join(str(p) for p in (self.resolution, self.source, self.codec, self.audio) if p.value != 0)
+        name = ' '.join(str(p) for p in (self.resolution, self.source, self.codec, self.audio) +
+                        tuple(self.custom_components) if p.value != 0)
         return name or 'unknown'
 
     @property
     def components(self):
-        return [self.resolution, self.source, self.codec, self.audio]
+        # TODO lexicographical ordering maybe doesn't make sense anymore?
+        return self.custom_components + [self.resolution, self.source, self.codec, self.audio]
```
cvium: I'm a little unsure about this one. Apparently we use lexicographical ordering when comparing two qualities. That means we must put the custom components somewhere that makes sense, which begs the question; do we still want lexicographical ordering?

gazpachoking: lexicographical ordering? All of the components define their own ordering based on their numeric value. Not sure what you mean here.

cvium: I don't remember 100%, but iirc the comparison between two qualities compares the qualities in order, so if the resolution is not equal to or greater than the other, then it's a "worse" quality.

gazpachoking: Yeah, it does do that. Came up in chat as well, where should the custom component go in that order, and does it need to be configurable. Not sure I have any good answers.

PR #2017, opened 2017-11-19
Add plugins for Gazelle-based sites (including RED and NWCD)

pR0Ps:
### Motivation for changes:
Add capability to search Gazelle-based sites

### Detailed changes:
- Use the old what.cd plugin as a base (what.cd was Gazelle-based)
- Use a base class to do most of the work
- Implement specific sites (NWCD and RED included in this PR) in subclasses
- Add 2FA detection (2FA is currently not supported)

#### To Do:
- [x] Implement `search` interface
- [x] Testing
- [ ] Wiki documentation

### DIFF
--- a/flexget/plugins/input/gazelle.py
+++ b/flexget/plugins/input/gazelle.py
@@ -1,0 +1,495 @@
+from __future__ import unicode_literals, division, absolute_import
+from builtins import *  # noqa pylint: disable=unused-import, redefined-builtin
+
+from datetime import datetime
+import logging
+
+from flexget import plugin, db_schema
+from flexget.config_schema import one_or_more
+from flexget.entry import Entry
+from flexget.event import event
+from flexget.manager import Session
+from flexget.plugin import PluginError
+from flexget.utils.database import json_synonym
+from flexget.utils.requests import TokenBucketLimiter
+from flexget.utils.search import normalize_unicode
+from flexget.utils.tools import parse_filesize
+
+from sqlalchemy import Column, Unicode, String, DateTime
+
+
+DETECT_2FA = ("Authenticator Code", "TOTP code")
+log = logging.getLogger('gazelle')
+Base = db_schema.versioned_base('gazelle_session', 0)
+
+
+class GazelleSession(Base):
+    __tablename__ = 'gazelle_session'
+
+    username = Column(Unicode, primary_key=True)
+    base_url = Column(String, primary_key=True)
+
+    authkey = Column(String)
+    passkey = Column(String)
+    _cookies = Column('cookie', Unicode)
+    cookies = json_synonym('_cookies')
+    expires = Column(DateTime)
+
+
+class InputGazelle(object):
+    """A generic plugin that searches a Gazelle-based website
+
+    Limited functionality but should work for almost all of them.
+    """
+
+    def __init__(self):
+        """Set up a plugin that only has the ability to do about basic search"""
+        self.base_url = None
+
+        # Aliases for config -> api params
+        # Extended in subclasses
+        self.aliases = {
+            "search": "searchstr"
+        }
+
+        # API parameters
+        # None means a raw value entry (no validation other than schema)
+        # A dict means an enum with a config -> api mapping
+        # A list is an enum with no mapping
+        # Extended in subclasses
+        self.params = {
+            "searchstr": None
+        }
+
+    @property
+    def schema(self):
+        """The schema of the plugin
+
+        Subclasses should extend this to implement more params
+        """
+        schema = {
+            'type': 'object',
+            'properties': {
+                'base_url': {'type': 'string'},
+                'username': {'type': 'string'},
+                'password': {'type': 'string'},
+                'max_pages': {'type': 'integer'},
+                'search': {'type': 'string'},
+            },
+            'required': ['username', 'password'],
+            'additionalProperties': False
+        }
+        # base_url is only required if the subclass doesn't set it
+        if not self.base_url:
+            schema['required'].append('base_url')
+        return schema
+
+    def _key(self, key):
+        """Gets the API key name from the entered key"""
+        if key in self.aliases:
+            return self.aliases[key]
+        return key
+
+    def _opts(self, key):
+        """Gets the options for the specified key"""
+        return self.params[self._key(key)]
+
+    def _getval(self, key, val):
+        """Gets the value for the specified key based on a config option"""
+        opts = self._opts(key)
+        if isinstance(opts, dict):
+            # Translate the input value to the API value
+            # The str cast converts bools to 'True'/'False' for use as keys
+            # This allows for options that have True/False/Other values
+            return opts[str(val)]
+        elif isinstance(val, list):
+            # Fix yaml parser making a list out of a string
+            return ",".join(val)
+        return val
+
+    def params_from_config(self, config):
+        """Filter params and map config values -> api values"""
+        ret = {}
+        for k, v in config.items():
+            key = self._key(k)
+            if key in self.params:
+                ret[key] = self._getval(k, v)
+        return ret
+
+    def setup(self, task, config):
+        """Set up a session and log in"""
+        self._session = task.requests
+        base_url = config.get('base_url', "").rstrip("/")
+        if base_url:
+            if self.base_url and self.base_url != base_url:
+                log.warning("Using plugin designed for %s on %s - things may break", self.base_url, base_url)
+            self.base_url = base_url
+
+        if not self.base_url:
+            raise PluginError("No 'base_url' configured")
+
+        # Any more than 5 pages is probably way too loose of a search
+        self.max_pages = config.get('max_pages', 5)
+
+        # The consistent request limiting rule seems to be:
+        # "Refrain from making more than five (5) requests every ten (10) seconds"
+        self._session.add_domain_limiter(TokenBucketLimiter(self.base_url, 2, '2 seconds'))
+
+        self.username = config['username']
+        self.password = config['password']
+
+        # Login
+        self.authenticate()
+
+        # Logged in successfully, it's ok if nothing matches now
+        task.no_entries_ok = True
+
+    def resume_session(self):
+        """Resume an existing session from the datebase
+
+        Return True on successful recovery, False otherwise
+        """
+        log.debug("Attempting to find an existing session in the DB")
+        with Session() as session:
+            db_session = session.query(GazelleSession).filter(
+                GazelleSession.base_url == self.base_url,
+                GazelleSession.username == self.username
+            ).one_or_none()
+            if db_session and db_session.expires and db_session.expires >= datetime.utcnow():
+                # Found a valid session in the DB - use it
+                self._session.cookies.update(db_session.cookies)
+                self.authkey = db_session.authkey
+                self.passkey = db_session.passkey
+                return True
+        return False
+
+    def save_current_session(self):
+        """Store the current session in the database so it can be resumed later"""
+        log.debug("Storing session info in the DB")
+        with Session() as session:
+            expires = None
+            for c in self._session.cookies:
+                if c.name == "session":
+                    expires = datetime.utcfromtimestamp(c.expires)
+            db_session = GazelleSession(username=self.username,
+                                        base_url=self.base_url,
+                                        cookies=dict(self._session.cookies),
+                                        expires=expires, authkey=self.authkey,
+                                        passkey=self.passkey)
+            session.merge(db_session)
+
+    def authenticate(self, force=False):
+        """Log in and store auth data from the server
+
+        Adapted from https://github.com/isaaczafuta/whatapi
+        """
+        # clean slate before creating/restoring cookies
+        self._session.cookies.clear()
+
+        if not force and self.resume_session():
+            log.info("Logged into %s using cached session", self.base_url)
+            return
+
+        # Forcing a re-login or no session in DB - log in using provided creds
+        url = "{}/login.php".format(self.base_url)
+        data = {
+            'username': self.username,
+            'password': self.password,
+            'keeplogged': 1,
+        }
+        r = self._session.post(url, data=data, allow_redirects=False, raise_status=True)
+        if not r.is_redirect or r.next.url != "{}/index.php".format(self.base_url):
+            msg = "Failed to log into {}".format(self.base_url)
+            for otp_text in DETECT_2FA:
+                # TODO: Find a better signal that 2FA is enabled
+                if otp_text in r.text:
+                    msg += " - Accounts using 2FA are currently not supported"
+                    break
+            raise PluginError(msg)
+
+        account_info = self.request(no_login=True, action='index')
+        self.authkey = account_info['authkey']
+        self.passkey = account_info['passkey']
+        log.info("Logged in to %s", self.base_url)
+
+        # Store the session so we can resume it later
+        self.save_current_session()
+
+    def request(self, no_login=False, **params):
+        """Make an AJAX request to the API
+
+        If `no_login` is True, logging in will not be attempted if the request
+        is redirected to the login page.
+
+        Adapted from https://github.com/isaaczafuta/whatapi
+        """
+        if 'action' not in params:
+            raise ValueError("An 'action' is required when making a request")
+
+        ajaxpage = "{}/ajax.php".format(self.base_url)
+        r = self._session.get(ajaxpage, params=params, allow_redirects=False, raise_status=True)
+        if not no_login and r.is_redirect and r.next.url == "{}/login.php".format(self.base_url):
+            log.warning("Redirected to login page, reauthenticating and trying again")
+            self.authenticate(force=True)
+            return self.request(no_login=True, **params)
+
+        if r.status_code != 200:
+            raise PluginError("{} returned a non-200 status code".format(self.base_url))
+
+        try:
+            json_response = r.json()
+            if json_response['status'] != "success":
+
+                # Try to deal with errors returned by the API
+                error = json_response.get('error', json_response.get('status'))
+                if not error or error == "failure":
+                    error = json_response.get('response', str(json_response))
+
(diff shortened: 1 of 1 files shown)


flexget/plugins/input/gazelle.py
```diff
@@ -0,0 +1,434 @@
...
+        """Set up a session and log in"""
+        self._session = Session()
+        base_url = config.get('base_url', "").rstrip("/")
+        if base_url:
+            if self.base_url and self.base_url != base_url:
+                log.warning("Using plugin designed for %s on %s - "
+                            "things may break", self.base_url, base_url)
+            self.base_url = base_url
+
+        if not self.base_url:
```
liiight: could replace this to `else`

flexget/plugins/input/gazelle.py
```diff
@@ -0,0 +1,434 @@
...
+            raise PluginError("No 'base_url' configured")
+
+        # The consistent request limiting rule seems to be:
+        # "Refrain from making more than five (5) requests every ten (10) seconds"
+        self._session.add_domain_limiter(TokenBucketLimiter(self.base_url, 2, '2 seconds'))
+
+        # Custom user agent
+        user_agent = config.get('user_agent', "Flexget [Gazelle plugin]")
+        if user_agent:
```
liiight: this check is redundant since you provided a default

flexget/plugins/input/gazelle.py
```diff
@@ -0,0 +1,434 @@
...
+        # clean slate before creating/restoring cookies
+        self._session.cookies.clear()
+
+        if not force:
+            with FlexgetSession() as session:
+                db_session = session.query(GazelleSession).filter(
+                    GazelleSession.base_url == self.base_url,
+                    GazelleSession.username == self.username
+                ).one_or_none()
+                if (db_session and db_session.expires and
```
liiight: parentheses are redundant here

flexget/plugins/input/gazelle.py
```diff
@@ -0,0 +1,434 @@
...
+                        db_session.expires >= datetime.utcnow()):
+                    # Found a valid session in the DB - use it
+                    self._session.cookies.update(db_session.cookies)
+                    self.authkey = db_session.authkey
+                    self.passkey = db_session.passkey
+                    log.info("Logged in to %s using cached session", self.base_url)
+                    return
```
liiight: no need to return if you dont care about return value

flexget/plugins/input/gazelle.py
```diff
@@ -0,0 +1,434 @@
...
+
+        # Forcing a re-login or no session in DB - log in using provided creds
+        url = "{}/login.php".format(self.base_url)
+        data = {
+            'username': self.username,
+            'password': self.password,
+            'keeplogged': 1,
+        }
+        r = self._session.post(url, data=data, allow_redirects=False)
```
liiight: need to do `r.raise_for_response()`and catch `RequestException` here and in any other relevant call

flexget/plugins/input/gazelle.py
```diff
@@ -0,0 +1,434 @@
...
+        if not r.is_redirect or r.next.url != "{}/index.php".format(self.base_url):
+            msg = "Failed to log into {}".format(self.base_url)
+            for x in DETECT_2FA:
```
liiight: avoid using meaningless vars please

flexget/plugins/input/gazelle.py
```diff
@@ -0,0 +1,434 @@
...
+        If `no_login` is True, logging in will not be attempted if the request
+        is redirected to the login page.
+
+        Adapted from https://github.com/isaaczafuta/whatapi
+        """
+        if 'action' not in params:
+            raise ValueError("An 'action' is required when making a request")
+
+        ajaxpage = "{}/ajax.php".format(self.base_url)
+        r = self._session.get(ajaxpage, params=params, allow_redirects=False)
```
liiight: handle request exceptions

flexget/plugins/input/gazelle.py
```diff
@@ -0,0 +1,434 @@
...
+        if (not no_login and r.is_redirect and
```
liiight: redundant parentheses

flexget/plugins/input/gazelle.py
```diff
@@ -0,0 +1,434 @@
...
+                r.next.url == "{}/login.php".format(self.base_url)):
+            log.warning("Redirected to login page, reauthenticating and trying again")
+            self.authenticate(force=True)
+            return self.request(no_login=True, **params)
+
+        if r.status_code != 200:
```
liiight: should be moved to exception handling

flexget/plugins/input/gazelle.py
```diff
@@ -0,0 +1,434 @@
...
+                # Try to deal with errors returned by the API
+                error = json_response.get('error', json_response.get('status'))
+                if not error or error == "failure":
+                    error = json_response.get('response', str(json_response))
+
+                raise PluginError("{} gave a failure response of '{}'"
+                                  "".format(self.base_url, error))
+            return json_response['response']
+        except (ValueError, TypeError, KeyError):
+            raise PluginError("{} returned an invalid response"
```
liiight: avoid redundant line breaks

flexget/plugins/input/gazelle.py
```diff
@@ -0,0 +1,434 @@
...
+        },
+        "vanityhouse": {
+            "False": 0,
+            "True": 1,
+        }
+    }
+
+    def __init__(self):
+        """Init client and set up the schema"""
+        super().__init__()
```
liiight: this only works for python 3 IMO

flexget/plugins/input/gazelle.py
```diff
@@ -0,0 +1,434 @@
...
+            "silver": 101,
+            "gold/silver": 100,
+            "lineage": -5,
+            "unscored": -1,
+            "missing lineage": -6,
+            "missing dr score": -7,
+            "missing sample rate": -8,
+            "missing description": -9
+        }
+        super().__init__()
```
liiight: change to be python 2 compatible

flexget/plugins/input/gazelle.py
pR0Ps: The return is so the rest of the function isn't executed.

flexget/plugins/input/gazelle.py
pR0Ps: I'm assuming this is more about the wrapping and less about the method of doing it? I was keeping to 80 chars out of habit but have just noticed that 120 is what's configured in setup.py - I'll unwrap the line since it's only 100 unwrapped.

flexget/plugins/input/gazelle.py
pR0Ps: Using `else` would break in the case where no `base_url` is in the config but one is set up on the subclass so it doesn't need to be.

flexget/plugins/input/gazelle.py
```diff
@@ -0,0 +1,434 @@
...
+    _cookies = Column('cookie', Unicode)
+    cookies = json_synonym('_cookies')
+    expires = Column(DateTime)
+
+
+class InputGazelle(object):
+    """A plugin that searches a Gazelle-based website"""
+
+    # Aliases for config -> api params
+    ALIASES = {
```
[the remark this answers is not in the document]
pR0Ps: I agree with you, however, these are more like instance constants. They're changed during the initial setup of the class but should stay static after that point.

flexget/plugins/input/gazelle.py
pR0Ps: There's no real recovery from an error response here. If it's not the expected value, then an exception should be raised and the task stopped. Should I catch the `RequestException`, swallow it, and raise a `PluginError` instead? Or just let the `RequestException` bubble up?

flexget/plugins/input/gazelle.py
pR0Ps: This response returns a 200 in a success case. Any other 3xx or 2xx response, while not considered an error to `requests`, is considered an error at this point in execution.

flexget/plugins/input/gazelle.py
```diff
@@ -0,0 +1,461 @@
...
+        self._session.headers.update({"User-Agent": user_agent})
+        self.username = config['username']
+        self.password = config['password']
+
+        # Login
+        self.authenticate()
+
+        # Logged in successfully, it's ok if nothing matches now
+        task.no_entries_ok = True
+
```
cvium: One too many linebreaks

flexget/plugins/input/gazelle.py
```diff
@@ -0,0 +1,461 @@
...
+        """
+        Log in and store auth data from the server
+        Adapted from https://github.com/isaaczafuta/whatapi
+        """
+
+        # clean slate before creating/restoring cookies
+        self._session.cookies.clear()
+
+        if not force:
+            with FlexgetSession() as session:
```
cvium: We prefer if `task.requests` is used whenever possible.

flexget/plugins/input/gazelle.py
```diff
@@ -0,0 +1,461 @@
...
+                            for k in ('torrentId',)))
+
+                yield Entry(
+                    title="{groupName} ({groupId} - {torrentId}).torrent".format(**temp),
+                    url="{}/torrents.php?action=download&id={}&authkey={}&torrent_pass={}"
+                        "".format(self.base_url, temp['torrentId'], self.authkey, self.passkey),
+                    torrent_seeds=tor['seeders'],
+                    torrent_leeches=tor['leechers'],
+                    # Size is returned in bytes, convert to MB for compat with the content_size plugin
+                    content_size=tor['size'] / (1024 ** 2)
```
cvium: We have a util for this

flexget/plugins/input/gazelle.py
```diff
@@ -0,0 +1,461 @@
...
+    @plugin.internet(log)
+    def on_task_input(self, task, config):
+        """Task input interface"""
+        self.setup(task, config)
+
+        params = self.params_from_config(config)
+        return list(self.get_entries(self.search_results(params)))
+
+
+class InputRedacted(InputGazelle):
```
cvium: The plugins that inherit from `InputGazelle` should be in separate files. Not sure where to put `InputGazelle`... maybe utils?

liiight: Yeah, we don't really have a precedence for this. Making it part of core feels weird so I think it's best that all plugins be places together to avoid import issues

flexget/plugins/input/gazelle.py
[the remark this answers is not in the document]
liiight: i dont think it matters. even if they're changed on inheritence, they're still instance variables which is the equivilant of const, and should be lowercase. no biggie though

flexget/plugins/input/gazelle.py
liiight: right, my bad

flexget/plugins/input/gazelle.py
pR0Ps: I did see that but didn't think it was appropriate. That function takes a string, guesses the correct units, and converts it to bytes. To use it I would have to do something like:
```python
Entry(
    content_size=parse_filesize(str(tor['size']) + "b")
)
```
Is that preferred over just converting it manually?

flexget/plugins/input/gazelle.py
pR0Ps: There's a lot of coupling between them and I worry that if they were split up it would be really easy to make a change in one of them that would break the others. Keeping them in the same file would help make it easier for people unfamiliar with the code to understand the implications of changing a parent class.

Ex: `InputNotWhat` inherits from `InputRedacted`. To set up the params and schema it just uses the parent classes values and adds/changes things. If RED makes a change and adds a `releasetype` and someone updates the plugin to include it without taking into account the child, it could break the NWCD plugin. Granted this is kinda messy anyway, but alternative was redefining the entire `PARAMS` structure which seemed like overkill since it's 95% the same.

This also holds true for things like cookies. If the `RedactedInput` plugin switched from using a cookie called `session` to one called `red_session` and the plugin is updated, `NotWhatInput` will break

To be honest I'm not totally happy with having `NotWhatInput` inherit from `RedactedInput` anyway so I'll look into solving this with a better inheritance model.

flexget/plugins/input/gazelle.py
cvium: It might be a good idea to create a function that does that. It's not the first time we've needed it.

flexget/plugins/input/gazelle.py
```diff
@@ -0,0 +1,498 @@
...
+from flexget import plugin, db_schema
+from flexget.config_schema import one_or_more
+from flexget.entry import Entry
+from flexget.event import event
+from flexget.manager import Session
+from flexget.plugin import PluginError
+from flexget.utils.database import json_synonym
+from flexget.utils.requests import TokenBucketLimiter
+from flexget.utils.search import normalize_unicode
+from sqlalchemy import Column, Unicode, String, DateTime
```
cvium: A nitpick: We divide imports into four groups:

1. future
2. builtin
3. flexget
4. 3rd party.

flexget/plugins/input/gazelle.py
```diff
@@ -0,0 +1,498 @@
...
+            "missing sample rate": -8,
+            "missing description": -9
+        })
+
+
+@event('plugin.register')
+def register_plugin():
+    plugin.register(InputGazelle, 'gazelle', interfaces=['task', 'search'], api_ver=2)
+    plugin.register(InputGazelleMusic, 'gazellemusic', interfaces=['task', 'search'], api_ver=2)
+    plugin.register(InputRedacted, 'redactedch', interfaces=['task', 'search'], api_ver=2)
```
cvium: We don't typically include the TLD in the plugin name unless it's part of the site name... It should just be `redacted`

liiight:
I think it's fine now. Don't worry about the commits, we squash the entire pr on merge

flexget/plugins/input/gazelle.py
```diff
@@ -0,0 +1,499 @@
...
+
+        # Any more than 5 pages is probably way too loose of a search
+        self.max_pages = config.get('max_pages', 5)
+
+        # The consistent request limiting rule seems to be:
+        # "Refrain from making more than five (5) requests every ten (10) seconds"
+        self._session.add_domain_limiter(TokenBucketLimiter(self.base_url, 2, '2 seconds'))
+
+        # Custom user agent
+        user_agent = config.get('user_agent', "Flexget [Gazelle plugin]")
```
cvium: Is this necessary? We already set a user_agent in `task.requests` (https://github.com/Flexget/Flexget/blob/develop/flexget/utils/requests.py#L184)

flexget/plugins/input/gazelle.py
```diff
@@ -0,0 +1,499 @@
...
+                GazelleSession.base_url == self.base_url,
+                GazelleSession.username == self.username
+            ).one_or_none()
+            if db_session and db_session.expires and db_session.expires >= datetime.utcnow():
+                # Found a valid session in the DB - use it
+                self._session.cookies.update(db_session.cookies)
+                self.authkey = db_session.authkey
+                self.passkey = db_session.passkey
+                return True
+        return False
```
cvium: You don't _need_ to return False. No return statement is the same as `return None`

flexget/plugins/input/gazelle.py
```diff
@@ -0,0 +1,499 @@
...
+        r = self._session.post(url, data=data, allow_redirects=False, raise_status=True)
+        if not r.is_redirect or r.next.url != "{}/index.php".format(self.base_url):
+            msg = "Failed to log into {}".format(self.base_url)
+            for otp_text in DETECT_2FA:
+                if otp_text in r.text:
+                    msg += " - Accounts using 2FA are currently not supported"
+                    break
+            raise PluginError(msg)
+
+        accountinfo = self.request(no_login=True, action='index')
```
cvium: Nitpick: Shouldn't this be called `account_info`?

flexget/plugins/input/gazelle.py
```diff
@@ -0,0 +1,499 @@
...
+                break
+            for x in result['results']:
+                yield x
+
+            pages = result.get('pages', pages)
+            page += 1
+
+        if page > self.max_pages:
+            log.warning("Stopped after %d pages (out of %d total pages)",
+                        self.max_pages, pages)
```
cvium: I don't think this linebreak is necessary. We allow 120 chars.

pR0Ps:
@cvium Fixed up style issues and switched to using `parse_filesize` as per your original comment.

flexget/plugins/input/gazelle.py
pR0Ps: The reason I explicitly return `False` is that having it return `True`/`False` vs. `True`/`None` is a little more logical IMO. Same reason why I don't return `True`/`0` even though it would behave the same way when used in truthyness comparisons.

flexget/plugins/input/gazelle.py
```diff
@@ -0,0 +1,499 @@
...
+        # Any more than 5 pages is probably way too loose of a search
+        self.max_pages = config.get('max_pages', 5)
+
+        # The consistent request limiting rule seems to be:
+        # "Refrain from making more than five (5) requests every ten (10) seconds"
+        self._session.add_domain_limiter(TokenBucketLimiter(self.base_url, 2, '2 seconds'))
+
+        # Custom user agent
+        if 'user_agent' in config:
+            self._session.headers.update({"User-Agent": config['user_agent']})
```
cvium: We already have a plugin for this though https://flexget.com/Plugins/headers

pR0Ps: That's useful, I didn't know about that - I've removed the custom UA support from the plugin

liiight:
Thanks. Wanna update the wiki?

pR0Ps:
Added a wiki page: https://www.flexget.com/Searches/gazelle

liiight:
Looks great

PR #2018, opened 2017-11-20
RTorrent Plugin Tweaks

stevezau:
### Motivation for changes:
Make rtorrent plugin more efficient especially when connecting to rtorrent's over higher latency networks.

### Detailed changes:
- Remove unneeded version check(most trackers force the latest version anyway)
- Only connect to rtorrent IF we are going to do something..


flexget/plugins/clients/rtorrent.py
```diff
@@ -513,52 +496,56 @@ def on_task_download(self, task, config):
...
-                    continue
-                self.delete_entry(client, entry)
-
-            if config['action'] == 'update':
-                if task.options.test:
-                    log.info('Would update %s (%s) in rTorrent', entry['title'], entry['torrent_info_hash'])
-                    continue
-                self.update_entry(client, entry, config)
+        except IOError as e:
+            raise plugin.PluginError("Couldn't connect to rTorrent: %s" % str(e))
```
cvium: I don't think `str` is necessary.

stevezau: I didn't change this, it's a move of the code. I think we don't do str the message comes out differently

stevezau: maybe you are right, i just tested it. Are we sure this is not a py3 thing?

PR #2020, opened 2017-11-21
Series performance improvements

stevezau:
### Motivation for changes:
If 100's of series exist flexget's performance can suffer when parsing. 

### Detailed changes:
- [x] Prefetch Series Data
- [x] Better Logic to parsing, right now it loops every entry * configured series..

### DIFF
--- a/flexget/plugins/filter/series.py
+++ b/flexget/plugins/filter/series.py
@@ -16,7 +16,7 @@ from sqlalchemy import (
 )
 from sqlalchemy.exc import OperationalError
 from sqlalchemy.ext.hybrid import Comparator, hybrid_property
-from sqlalchemy.orm import relation, backref, object_session
+from sqlalchemy.orm import relation, backref, object_session, joinedload
 
 from flexget import db_schema, options, plugin
 from flexget.config_schema import one_or_more
@@ -678,6 +678,28 @@ class SeriesTask(Base):
 
     def __init__(self, name):
         self.name = name
+
+
+def map_names_to_series(names, session, preload=None):
+    """
+    Return a dict with key=name and value=series_object for all series matching the names.
+
+    :param names: list of series names
+    :param session: db session
+    :return:
+    """
+    # Normalize the names.
+    series_names_normalized = dict(map(lambda n: (n, normalize_series_name(n)), names))
+
+    query = session.query(Series).filter(Series._name_normalized.in_(series_names_normalized.values()))
+    if preload:
+        for j in preload:
+            query = query.options(joinedload(j))
+
+    # Return a dict of key=orig_name and value=series_object
+    found_series = dict(map(lambda s: (s._name_normalized, s), query.all()))
+
+    return dict(map(lambda n: (n, found_series.get(series_names_normalized[n])), names))
 
 
 @with_session
@@ -1599,6 +1621,8 @@ class FilterSeries(FilterSeriesBase):
             if entry.get('series_name') and entry.get('series_id') is not None and entry.get('series_parser'):
                 found_series.setdefault(entry['series_name'], []).append(entry)
 
+        # Prefetch Series
+
         for series_item in config:
             with Session() as session:
                 series_name, series_config = list(series_item.items())[0]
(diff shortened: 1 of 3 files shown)


stevezau:
@gazpachoking @liiight cau you take a look at this? not sure other ways we can improve the speeds?

gazpachoking:
Hmm. Is all the manual normalization needed? `Series.name` is meant to automatically compare with normalized version of the name. How much of a speed difference does this make?

stevezau:
@gazpachoking it didn't work with name.in_. Do you know how to handle that in https://github.com/Flexget/Flexget/blob/develop/flexget/plugins/filter/series.py#L236-L238

stevezau:
With pre-fetch it cut down the process time by 50% for me.

It's still very slow though. if you have an rss feed that has 100 entires and series with 100-200 entries it loops over the 100 entries times by the amount of series you have.. Can you think of a better way? I couldn't.

gazpachoking:
Perhaps we need to implement this on the comparator? http://docs.sqlalchemy.org/en/latest/orm/internals.html#sqlalchemy.orm.properties.ColumnProperty.Comparator.in_

stevezau:
ok, i'll look into that.

Any other idea how we can speed it up. Seems crazy to have to loop over each entry times by the number of series.

If 200 series and 100 entry rss feed that's 20,000 loops

cvium:
Could we not sort the configured series alphabetically and put them in a dict keyed on first letter like this
```
series:
  - a series
  - another series
  - other series
  - third series
```
-->
```
series_map = {
    'a': ['a series', 'another series'],
    'o': ['other series'],
    't': ['third series']
}
```

stevezau:
How will that help @cvium?

Won’t we still need to loop over them all right?

cvium:
We can guess the series name from the title, grab the first letter and look up in the dict and try to match with the (hopefully much shorter) list

stevezau:
Hmm is that safe to do? I guess we need to also check alt names..

cvium:
All possible names would need to have its own node in the tree

stevezau:
@gazpachoking i'm getting lost when trying to figure out how to properly impl series.name.in_ 

Do you understand how Comparator work? I think we need to impl def in_() but i'm not sure what to return?

gazpachoking:
@stevezau Maybe try this for the comparator:
```python
def in_(other):
    return super().in_([normalize_series_name(e) for e in other])
```

stevezau:
@gazpachoking that's what i thought yesterday but it still calls operate https://github.com/Flexget/Flexget/blob/series_speedup/flexget/plugins/filter/series.py#L238

I can fix it by checking if it's a list in operate but that does not seem right?

flexget/plugins/filter/series.py
```diff
@@ -1579,13 +1592,36 @@ def auto_exact(self, config):
...
-            log.trace('series_name: `%s`, series_config: `%s`', series_name, series_config)
+
+        parser = get_plugin_by_name('parsing').instance
+
+        # Sort Entries into data model similar to https://en.wikipedia.org/wiki/Trie
+        # Only process series if both the entry title and series title first letter match
+        entries_map = {}
+        for entry in task.entries:
+            parsed = parser.parse_series(entry['title'])
+            if parsed.name:
```
liiight: maybe also check `parsed.valid`, not sure if it'll have a `name` property if it isn't valid

flexget/plugins/filter/series.py
```diff
@@ -1579,13 +1592,36 @@ def auto_exact(self, config):
...
         self.auto_exact(config)
-        for series_item in config:
-            series_name, series_config = list(series_item.items())[0]
-            log.trace('series_name: `%s`, series_config: `%s`', series_name, series_config)
+
+        parser = get_plugin_by_name('parsing').instance
+
+        # Sort Entries into data model similar to https://en.wikipedia.org/wiki/Trie
+        # Only process series if both the entry title and series title first letter match
+        entries_map = {}
```
liiight: you can set it to `defaultdict(list)`, makes for a neater implementation later

flexget/plugins/filter/series.py
```diff
@@ -1579,13 +1592,36 @@ def auto_exact(self, config):
...
+            parsed = parser.parse_series(entry['title'])
+            if parsed.name:
+                entries_map.setdefault(parsed.name[:1].lower(), []).append(entry)
+
+        with Session() as session:
+            # Preload series
+            # str() added to make sure number shows (e.g. 24) are turned into strings
+            series_names = [str(s.keys()[0]) for s in config]
+            existing_db_series = session.query(Series).filter(Series.name.in_(series_names))
+            existing_db_series = dict([(s.name_normalized, s) for s in existing_db_series])
```
liiight: `existing_db_series = {s.name_normalized: s for s in existing_db_series}`

flexget/utils/tools.py
```diff
@@ -482,6 +482,19 @@ def get_config_hash(config):
...
 
+def get_config_as_array(config, key):
+    """
+    Return configuration key as array, even if given as a single string
+    :param dict config: Configuration
+    :param string key: Configuration
+    :return: Array
+    """
+    v = config.get(key, [])
+    if isinstance(v, str):
```
liiight: this should probably be `if isinstance(v, basestring)`

stevezau:
@gazpachoking @liiight @cvium tests are passing. Ready for review.

flexget/plugins/filter/series.py
```diff
@@ -2158,16 +2218,27 @@ def on_task_start(self, task, config):
                     db_series.alternate_names = [alt for alt in db_series.alternate_names if alt.alt_name in alts]
                     # Add/update the possibly new alternate names
                 else:
+                    # TODO: Remove, added for debugging
```
cvium: Forget something @stevezau ?

PR #2030, opened 2017-11-30
Parser cache

stevezau:
### Motivation for changes:
Reduce compute by caching internal parser using lru cache

Tests on 100 entries show a speedup of 3x with multiple tasks.


paranoidi:
Does this actually speed up parsing with caching or just by reducing instantiation? If later then this is overcomplicated solution ..

cvium:
I tested this on a single task with the yappi profiler and the parser instantiation was fairly inconsequential wrt. performance.

I don't think this is worth pursuing and it doesn't seem like you want to finish it anyway 😛

PR #2032, opened 2017-12-07
[symlink] Support Jinja2 templating for "to" field

dolohow:
This is on par with "decompress" plugin.

### Motivation for changes:
I would like to inject data into path while making hard links.

```
symlink:
      to: '/home/user/{{ variable }}'
```

Tested ✅

### DIFF
--- a/flexget/plugins/output/symlink.py
+++ b/flexget/plugins/output/symlink.py
@@ -6,6 +6,7 @@ import os
 
 from flexget import plugin
 from flexget.event import event
+from flexget.utils.template import render_from_entry, RenderError
 
 log = logging.getLogger('symlink')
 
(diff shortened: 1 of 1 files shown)


dolohow:
Any plans on merging?

@liiight

flexget/plugins/output/symlink.py
```diff
@@ -50,7 +51,12 @@ def on_task_output(self, task, config):
                 continue
             lnkfrom = entry['location']
             name = os.path.basename(lnkfrom)
-            lnkto = os.path.join(config['to'], name)
+            try:
+                lnkto = render_from_entry(os.path.join(config['to'], name), entry)
+            except RenderError as error:
+                log.error('Could not render path: %s', to)
```
cvium: What is `to` in this context? Doesn't look like it's initialized anywhere.

dolohow: Yes, it should be `lnkto`.  Will fix it in few hours when I get back home.

Nice catch.

dolohow: Fixed

dolohow:
Is it good now? @liiight @cvium

PR #2040, opened 2017-12-15
Trakt to Sonarr plugin: Add couple more options.

JourneyOver:
### Motivation for changes:

Adding more options to the sonarr plugin for Trakt to sonarr

### Detailed changes:
- Added option to set quality profile ID (defaults to 1/any) -- You can find quality profile/profile ID at http://localhost:8989/api/profile?apikey=APIKEY
- Added option to set season folder -- True/False (Default is False)
- Added option to set monitored status of whole series -- True/False (Default is True)

*** NOTE
One thing I don't understand is the why there is a profileId and a qualityProfileId as it seems only the profileId actually does anything, but I'm leaving it in as it seems it's part of the API call anyways versus profileId not being any part of the Sonarr API..

EDIT:

- Added option to set rootfolderpath -- The way I had to set this up was that it uses the ID instead of a string path (as I couldn't get string paths to work for some reason) and it would take the number from "rootFolderPath" config and subtract it by 1 due to how things are currently programs in the sonarr.py script, so 1 = 0, 2 = 1, so on and so forth. (Defaults to 1) You can find the rootFolder ID here http://localhost:8989/api/rootfolder?apikey=APIKEY

Sadly the way I had to setup rootFolderPath means you have to already have 1 or more rootFolderPaths already created in sonarr for this option to work.

EDIT2:

- Added Option for seriesType, string value of standard, daily, or anime. (default value is standard)

### DIFF
--- a/flexget/plugins/list/sonarr_list.py
+++ b/flexget/plugins/list/sonarr_list.py
@@ -33,7 +33,7 @@ class SonarrSet(MutableSet):
             'profile_id': {'type': 'integer', 'default': 1},
             'season_folder': {'type': 'boolean', 'default': False},
             'monitored': {'type': 'boolean', 'default': True},
-            'root_folder_path': {'type': 'integer', 'default': 1}, # Users are passing actual rootFolderPath ID (number (integer) starting from 1 and going up) as apposed to an number (integer) system starting from 0. Read comment on show['rootFolderPath'] for more information.
+            'root_folder_path': {'type': 'integer', 'default': 1}, # Users are passing actual rootFolderPath ID (number (integer) starting from 1 and going up) as apposed to an number (integer) system starting from 0. Passing root_folder_path as a string seems to make the script fail, so that's why it's coded as an integerinstead. Read comment on show['rootFolderPath'] for more information.
             'series_type': {'type': 'string', 'enum': ['standard', 'daily', 'anime'], 'default': 'standard'},
             'tags': {'type': 'array', 'items': {'type': 'integer'}, 'default': [0]}
         },
--- a/flexget/plugins/list/sonarr_list.py
+++ b/flexget/plugins/list/sonarr_list.py
@@ -29,7 +29,10 @@ class SonarrSet(MutableSet):
             'include_data': {'type': 'boolean', 'default': False},
             'search_missing_episodes': {'type': 'boolean', 'default': True},
             'ignore_episodes_without_files': {'type': 'boolean', 'default': False},
-            'ignore_episodes_with_files': {'type': 'boolean', 'default': False}
+            'ignore_episodes_with_files': {'type': 'boolean', 'default': False},
+            'profile_id': {'type': 'number', 'default': 1},
+            'season_folder': {'type': 'boolean', 'default': False},
+            'monitored': {'type': 'boolean', 'default': True}
         },
         'required': ['api_key', 'base_url'],
         'additionalProperties': False
@@ -193,8 +196,10 @@ class SonarrSet(MutableSet):
         rootfolder = self.get_json(rootfolder_series_url, headers=rootfolder_series_headers)
 
         # Setting defaults for Sonarr
-        show['profileId'] = 1
-        show['qualityProfileId '] = 1
+        show['profileId'] = self.config.get('profile_id')
+        show['qualityProfileId'] = self.config.get('profile_id')
+        show['seasonFolder'] = self.config.get('season_folder')
+        show['monitored'] = self.config.get('monitored')
         show['rootFolderPath'] = rootfolder[0]['path']
         show['addOptions'] = {"ignoreEpisodesWithFiles": self.config.get('ignore_episodes_with_files'),
                               "ignoreEpisodesWithoutFiles": self.config.get('ignore_episodes_without_files'),
--- a/flexget/plugins/list/sonarr_list.py
+++ b/flexget/plugins/list/sonarr_list.py
@@ -32,7 +32,8 @@ class SonarrSet(MutableSet):
             'ignore_episodes_with_files': {'type': 'boolean', 'default': False},
             'profile_id': {'type': 'number', 'default': 1},
             'season_folder': {'type': 'boolean', 'default': False},
-            'monitored': {'type': 'boolean', 'default': True}
+            'monitored': {'type': 'boolean', 'default': True},
+            'rootFolderPath': {'type': 'number', 'default': 1}
         },
         'required': ['api_key', 'base_url'],
         'additionalProperties': False
@@ -200,7 +201,7 @@ class SonarrSet(MutableSet):
         show['qualityProfileId'] = self.config.get('profile_id')
         show['seasonFolder'] = self.config.get('season_folder')
         show['monitored'] = self.config.get('monitored')
-        show['rootFolderPath'] = rootfolder[0]['path']
+        show['rootFolderPath'] = rootfolder[self.config.get('rootFolderPath') - 1]['path']
         show['addOptions'] = {"ignoreEpisodesWithFiles": self.config.get('ignore_episodes_with_files'),
                               "ignoreEpisodesWithoutFiles": self.config.get('ignore_episodes_without_files'),
                               "searchForMissingEpisodes": self.config.get('search_missing_episodes')}
--- a/flexget/plugins/list/sonarr_list.py
+++ b/flexget/plugins/list/sonarr_list.py
@@ -33,7 +33,7 @@ class SonarrSet(MutableSet):
             'profile_id': {'type': 'number', 'default': 1},
             'season_folder': {'type': 'boolean', 'default': False},
             'monitored': {'type': 'boolean', 'default': True},
-            'rootFolderPath': {'type': 'number', 'default': 1}
+            'root_folder_path': {'type': 'number', 'default': 1}
         },
         'required': ['api_key', 'base_url'],
         'additionalProperties': False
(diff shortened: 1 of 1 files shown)


JourneyOver:
If you need me to explain a little bit better on how the rootFolderPath stuff is setup let me know but it should be kinda easy to understand in some form.

1 = 0 
2 = 1
3 = 2
and so on

Let's say you have 3 rootFolderPaths setup for an example.
"D:\Media\TV Shows\"
"D:\Media\Anime\"
"F:\Media\Anime\"

the first rootFolderPath in the API is set as 1, but the current iteration of the trakt to sonarr plugin sees/makes the first one act like a 0 instead, second one would be a 2 but plugin sees it as a 1 and third would be 3 but again plugin sees it as a 2. So this is where my subtract 1 comes into play. Say you want the rootFolderPath to be "F:\Media\Anime\" for example, well the ID for it in reality is 3 so you can put 3 in your config for "rootFolderPath" and when the trakt to sonarr plugin runs it'll push the correct rootFolderPath since I am subtracting 1 from that 3 so the plugin sees it as 2 which would be correct in the plugins eyes.

flexget/plugins/list/sonarr_list.py
```diff
@@ -29,7 +29,11 @@ class SonarrSet(MutableSet):
             'include_data': {'type': 'boolean', 'default': False},
             'search_missing_episodes': {'type': 'boolean', 'default': True},
             'ignore_episodes_without_files': {'type': 'boolean', 'default': False},
-            'ignore_episodes_with_files': {'type': 'boolean', 'default': False}
+            'ignore_episodes_with_files': {'type': 'boolean', 'default': False},
+            'profile_id': {'type': 'number', 'default': 1},
+            'season_folder': {'type': 'boolean', 'default': False},
+            'monitored': {'type': 'boolean', 'default': True},
+            'rootFolderPath': {'type': 'number', 'default': 1}
```
cvium: We usually use snake case and not camel case even if it's a "keyword".

JourneyOver: @cvium Never even heard of snake case before.. Guessing snake case is like "profile_id" for example?

cvium: Yes

JourneyOver: This has now been taken care of.

JourneyOver:
And with the seriesType option I believe that is now everything that is possible. I am waiting on an answer about tags, but I believe it's not possible during the phase of adding a series so I guess we can call this finished.

flexget/plugins/list/sonarr_list.py
```diff
@@ -29,7 +29,12 @@ class SonarrSet(MutableSet):
             'include_data': {'type': 'boolean', 'default': False},
             'search_missing_episodes': {'type': 'boolean', 'default': True},
             'ignore_episodes_without_files': {'type': 'boolean', 'default': False},
-            'ignore_episodes_with_files': {'type': 'boolean', 'default': False}
+            'ignore_episodes_with_files': {'type': 'boolean', 'default': False},
+            'profile_id': {'type': 'number', 'default': 1},
+            'season_folder': {'type': 'boolean', 'default': False},
+            'monitored': {'type': 'boolean', 'default': True},
+            'root_folder_path': {'type': 'number', 'default': 1},
+            'series_type': {'type': 'string', 'default': 'standard'}
```
liiight: This is a closed list right? Please limit it with an `enum`

flexget/plugins/list/sonarr_list.py
```diff
@@ -29,7 +29,12 @@ class SonarrSet(MutableSet):
             'include_data': {'type': 'boolean', 'default': False},
             'search_missing_episodes': {'type': 'boolean', 'default': True},
             'ignore_episodes_without_files': {'type': 'boolean', 'default': False},
-            'ignore_episodes_with_files': {'type': 'boolean', 'default': False}
+            'ignore_episodes_with_files': {'type': 'boolean', 'default': False},
+            'profile_id': {'type': 'number', 'default': 1},
+            'season_folder': {'type': 'boolean', 'default': False},
+            'monitored': {'type': 'boolean', 'default': True},
+            'root_folder_path': {'type': 'number', 'default': 1},
```
liiight: I think this should be an `integer`

liiight:
Thanks for this, I plan to give this a closer look and maybe even test this on Monday

flexget/plugins/list/sonarr_list.py
JourneyOver: yes and done.

flexget/plugins/list/sonarr_list.py
JourneyOver: it doesn't seem to matter between it being a number or an integer typing as both ways work just fine, but I went ahead and switched it and profile_id to be an integer type instead of number type (as it's an int as well going by https://github.com/Sonarr/Sonarr/wiki/Series#post).

JourneyOver:
@liiight not a problem at all :p didn't know if/when you where going to do my requests so I just took it upon myself to get them done, and along the way I added options I didn't even request as well xD I hope that I made the root_folder_path change somewhat understandable in my explanation on how it all works now.

btw don't forget to test my #2041 bug while your at it ;) since it's an issue even with the default plugin.

JourneyOver:
Just a note: I got confirmation about if it's possible to push tags in during adding a show phase, and it is indeed possible [sonarr forums](https://forums.sonarr.tv/t/is-it-possible-to-add-tags-during-adding-a-show-to-sonarr-through-the-api/17100/2?u=eightpixels) but this is where it gets hard and I don't know if it's possible to do with flexget.. The way you have to pass in tags is like the profile ID and rootfolder path but, 
( Tag ID's can be found here http://localhost:8989/api/tag?apikey=APIKEY )

1. You have to pre-create whatever tags you want to use (just like with rootfolder path).
2. You have to pass them in as an integer and you have to do it like so 1,2,3,4,5

so here is the predicament, I have gotten it somewhat working but with only pushing 1 tag, but if I try to add pushing more tags ex.1,2 instead of just 1 then flexget throws an error about expecting an integer, if I try to switch the type to a number flexget then complains about expecting a number instead, if I try to switch the type to a list flexget gets a little crazy but still ends up throwing an error in the end, and if I try switch the type to a string then sonarr throws an error making flexget throw an error in the process as well. 

So I am currently at a standstill on how to proceed with possibly getting tags added in, right now I'm leaning towards just not adding the option at all right now as all the other options I have added are good enough to make the plugin better anyways but at the same time I do want to get tags added as it's literally the last thing that can be passed in during the series creation phase.

I'm all for getting this merged whenever you feel like without having tags @liiight and if down the road I/you/someone else finds a way to add tags that allow for more than just 1 tag then so be it.

liiight:
what was the issue with specifying a path for rootFolderPath? that int thingy is weird...

JourneyOver:
@liiight for some reason I couldn't get a string value to work exactly for rootFolderPath, I think it's due to the way the script is setup (plus I don't know enough python to really dive into the the script), and part of the reason for setting up using int is for the default value.

liiight:
I can see you can also send `path` directly. Might be better just to use that (or maybe keep both)

JourneyOver:
Actually path was already in there by default (in the default script), you need to send both rootfolderpath and path together as path appends to the end of rootfolderpath. so instead of it just being for example 

> D:\Media\TV Shows

it turns into 

> D:\Media\TV Shows\{nameofshow}

if you take out path or only use path, then it would fail since you'd only be passing in {nameofshow}

liiight:
ah, my bad. ill see about that failing test though and merge this

JourneyOver:
Alright thanks ^^ btw for the most part the int thing for the rootFolderPath shouldn't be hard on any users as they will actually be using the actual ID of the rootFolderPath, it may just seem confusing to people who look at it from the script point of things xD

liiight:
maybe add a comment in the code then

JourneyOver:
Yea I can do that while your working on seeing what the issue is with #2041 should have a comment on the code up in a jiffy

JourneyOver:
@liiight added comments, idk if it makes it any easier really as I tend to make comments confusing sometimes just how my brain works :x but let me know if it's any better.

flexget/plugins/list/sonarr_list.py
```diff
@@ -29,7 +29,13 @@ class SonarrSet(MutableSet):
             'include_data': {'type': 'boolean', 'default': False},
             'search_missing_episodes': {'type': 'boolean', 'default': True},
             'ignore_episodes_without_files': {'type': 'boolean', 'default': False},
-            'ignore_episodes_with_files': {'type': 'boolean', 'default': False}
+            'ignore_episodes_with_files': {'type': 'boolean', 'default': False},
+            'profile_id': {'type': 'integer', 'default': 1},
+            'season_folder': {'type': 'boolean', 'default': False},
+            'monitored': {'type': 'boolean', 'default': True},
+            'root_folder_path': {'type': 'integer', 'default': 1}, # Users are passing actual rootFolderPath ID (number (integer) starting from 1 and going up) as apposed to an number (integer) system starting from 0. Passing root_folder_path as a string seems to make the script fail, so that's why it's coded as an integerinstead. Read comment on show['rootFolderPath'] for more information.
```
liiight: That's way too long, keep line limit to 108 chars please

JourneyOver: Okay, I'll try to condense it down in some way.

liiight: Feel free to use multi line comments

PR #2044, opened 2017-12-18
Create directories recursively for hardlinks

dolohow:
### Motivation for changes:
I want to be able to create hardlinks for files in directories which not exists.


cvium:
What about hardlinks for dirs? This seems unfinished.

PR #2046, opened 2017-12-20
Changes how root_folder_path option is setup

JourneyOver:
### Motivation for changes:
None really lol
### Detailed changes:
- Changes how root_folder_path config option is setup. Now uses string value instead of an integer. Defaults to using an integer if no value is passed in config.
- Removed comments on old root_folder_path setup.

### DIFF
--- a/flexget/plugins/list/sonarr_list.py
+++ b/flexget/plugins/list/sonarr_list.py
@@ -33,10 +33,7 @@ class SonarrSet(MutableSet):
             'profile_id': {'type': 'integer', 'default': 1},
             'season_folder': {'type': 'boolean', 'default': False},
             'monitored': {'type': 'boolean', 'default': True},
-            # Users are passing actual RootFolderPath ID instead of starting from 0 and counting up.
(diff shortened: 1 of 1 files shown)


flexget/plugins/list/sonarr_list.py
```diff
@@ -208,10 +205,7 @@ def add_show(self, entry):
         show['monitored'] = self.config.get('monitored')
         show['seriesType'] = self.config.get('series_type')
         show['tags'] = self.config.get('tags')
-        # Requires both rootFolder and path otherwise it fails.
-        # takes root_folder_path ID from config and subtracts 1 from it due to how script is setup
-        # since users are passing actual ID it won't be confusing to users
-        show['rootFolderPath'] = rootfolder[self.config.get('root_folder_path') - 1]['path']
+        show['rootFolderPath'] = self.config.get('root_folder_path', rootfolder[0]['path'])
```
liiight: if the user supplies `root_folder_path` we should probably not query for `rootfolder`, no?

JourneyOver: explained in gitter.

PR #2047, opened 2017-12-20
sonarr list refactor

### LINKED ISSUE #2041: BUG: Invalid response received from Sonarr: This series has already been added
### Expected behaviour:

Skip any shows that are already added on sonarr.

### Actual behaviour:

Doesn't skip the shows that are already added on sonarr so it keeps throwing the error -
`Invalid response received from Sonarr: b'[\r\n {\r\n "propertyName": "TvdbId",\r\n "errorMessage": "This series has already been added",\r\n "attemptedValue": 295829,\r\n "formattedMessageArguments": [],\r\n "formattedMessagePlaceholderValues": {\r\n "propertyName": "Tvdb Id",\r\n "propertyValue": 295829\r\n }\r\n }\r\n]'` whenever the task is executed and done.

The "attemptedValue" and the "propertyValue" are always different, so far it has hit on every show that I already have in sonarr.. It's acting like the shows don't exist at all on sonarr when in reality they do.

### Steps to reproduce:
- Step 1:  Add a show to sonarr
- Step 2: Add same show to a trakt list
- Step 3: Try to import said list into sonarr with the Trakt to Sonarr plugin.

#### Config:

```
variables: secrets.yml

web_server:
  bind: 0.0.0.0
  port: 9091
  web_ui: yes
  
schedules:
##### Import Trakt list "Sonarr-TV" to sonarr
  - tasks: ['import-trakt-to-sonarr']
    interval:
      weeks: 1

tasks:

##### Import Trakt list "Sonarr-TV" to sonarr
  import-trakt-to-sonarr:
    trakt_list:
      username: "{? trakt.account ?}"
      account: "{? trakt.account ?}"
      list: "{? trakt.series ?}"
      type: shows
    accept_all: yes
    list_add:
       - sonarr_list:
          base_url: http://localhost
          port: 8989
          api_key: "{? sonarr.api_key ?}"
          ignore_episodes_with_files: true
          ignore_episodes_without_files: true
          search_missing_episodes: false
```

#### Log:
Running daemon with webUI --
https://pastebin.com/LUV6Jzyq

Running without daemon and not on the webUI  (running through CLI) --
https://pastebin.com/mc3C1FsD

### Additional information:

- FlexGet version: Flexget: 2.11.11  
- Python version: 3.6.3 but have tried with python 2.7 as well
- Installation method: pip
- Using daemon (yes/no): yes but have tried with not running the daemon as well.
- OS and version: Windows 10 Pro x64 bit, 1709 build 16299.64
- Link to crash log:

liiight:
### Motivation for changes:
sonarr_list code was old and sucky. Also a bugfix
### Detailed changes:
- Added ability to bypass filters when listing all entries. It was needed when adding or removing a show in order to get the unfiltered list of shows
- Made `base_url` default to `'http://localhost` and made it not required
- Refactor and tweaks

### Addressed issues:
- Fixes #2041

### DIFF
--- a/flexget/plugins/list/sonarr_list.py
+++ b/flexget/plugins/list/sonarr_list.py
@@ -19,6 +19,12 @@ PROFILE_ENDPOINT = 'profile'
 ROOTFOLDER_ENDPOINT = 'Rootfolder'
 DELETE_ENDPOINT = 'series/{}'
 
+# Sonarr qualities that do no exist in Flexget
+QUALITY_MAP = {
+    'Raw-HD': 'remux',
+    'DVD': 'dvdrip'
+}
+
 
 class SonarrSet(MutableSet):
     supported_ids = ['tvdb_id', 'tvrage_id', 'tvmaze_id', 'imdb_id', 'slug', 'sonarr_id']
@@ -54,7 +60,9 @@ class SonarrSet(MutableSet):
             url += '?term={}'.format(term)
         try:
             rsp = requests.request(method, url, headers=headers, json=data)
-            return rsp.json()
+            data = rsp.json()
+            log.trace('sonarr response: %s', data)
+            return data
         except RequestException as e:
             base_msg = 'Sonarr returned an error. {}'
             if e.response is not None:
@@ -66,17 +74,13 @@ class SonarrSet(MutableSet):
 
     def translate_quality(self, quality_name):
         """
-        Translate Sonnar's qualities to ones recognize by Flexget
+        Translate Sonarr's qualities to ones recognize by Flexget
         """
-        if quality_name == 'Raw-HD':  # No better match yet in Flexget
-            return 'remux'
-        elif quality_name == 'DVD':  # No better match yet in Flexget
-            return 'dvdrip'
-        else:
-            return quality_name.replace('-', ' ').lower()
+        if quality_name in QUALITY_MAP:
+            return QUALITY_MAP[quality_name]
+        return quality_name.replace('-', ' ').lower()
 
     def quality_requirement_builder(self, quality_profile):
-
         allowed_qualities = [self.translate_quality(quality['quality']['name']) for quality in quality_profile['items']
                              if quality['allowed']]
         cutoff = self.translate_quality(quality_profile['cutoff']['name'])
@@ -85,10 +89,10 @@ class SonarrSet(MutableSet):
 
     def list_entries(self, filters=True):
         shows = self._sonarr_request(SERIES_ENDPOINT)
-
+        profiles = None
         # Retrieves Sonarr's profile list if include_data is set to true
         if self.config.get('include_data'):
-            profiles_json = self._sonarr_request(PROFILE_ENDPOINT)
+            profiles = self._sonarr_request(PROFILE_ENDPOINT)
 
         entries = []
         for show in shows:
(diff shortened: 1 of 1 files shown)


flexget/plugins/list/sonarr_list.py
```diff
@@ -37,69 +41,23 @@ class SonarrSet(MutableSet):
...
+        if term:
+            url += '?term={}'.format(term)
         try:
-            response = requests.get(url, headers=headers)
-            if response.status_code == 200:
-                return response.json()
-            else:
-                raise plugin.PluginError('Invalid response received from Sonarr: %s' % response.content)
+            rsp = requests.request(method, url, headers=headers, json=data)
+            rsp.raise_for_status()
```
cvium: Why not use https://github.com/Flexget/Flexget/blob/develop/flexget/utils/requests.py#L214 ?

flexget/plugins/list/sonarr_list.py
```diff
@@ -37,69 +41,23 @@ class SonarrSet(MutableSet):
...
-        return url, headers
-
-    def get_json(self, url, headers):
+    def _sonarr_request(self, endpoint, term=None, method='get', data=None):
+        base_url = self.config['base_url']
+        port = self.config['port']
+        url = '{}:{}/api/{}'.format(base_url, port, endpoint)
+        headers = {'X-Api-Key': self.config['api_key']}
+        if term:
+            url += '?term={}'.format(term)
```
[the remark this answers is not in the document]
liiight: I had no choice with that. I had to build url like that since the param can contain `:`which requests escapes and sonarr doesn't like

flexget/plugins/list/sonarr_list.py
liiight: Dunno. Since I didn't need to use our custom session features I saw no reason using our own implementation

cvium: It's just a wrapper. I'd rather we were consistent on the matter. You can just use this https://github.com/Flexget/Flexget/blob/develop/flexget/utils/requests.py#L253

flexget/plugins/list/sonarr_list.py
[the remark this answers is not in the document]
liiight: I just now get what you meant. yeah, it's kinda odd but i think it's clear, so...

flexget/plugins/list/sonarr_list.py
```diff
@@ -102,11 +106,12 @@ def list_entries(self, filters=True):
                 # Checks if to retrieve ended shows
                 if show['status'] == 'ended' and not self.config.get('include_ended'):
                     continue
-            if self.config.get('include_data') and profiles_json:  # Check if to retrieve quality & path
+            if profiles:  # Check if to retrieve quality & path
```
cvium: Why not put the loop here instead?

flexget/plugins/list/sonarr_list.py
```diff
@@ -155,17 +120,16 @@ def list_entries(self):
...
+                          sonarr_id=show.get('id'))
             if self.config.get('include_data'):
                 if len(fg_qualities) > 1:
                     entry['configure_series_qualities'] = fg_qualities
                 elif len(fg_qualities) == 1:
                     entry['configure_series_quality'] = fg_qualities[0]
-                else:
-                    entry['configure_series_quality'] = fg_qualities
                 if path:
                     entry['configure_series_path'] = path
```
cvium: Why not just put `show.get('path')` here?

PR #2053, opened 2018-01-02
IFTTT notifications

kroozo:
Notification plugin to send notification to ifttt.com's webhook service.
(https://ifttt.com/maker_webhooks)

### Motivation for changes:
Adds a simple notification plugin to send the notification to IFTTT webhook trigger that then can be used
in an IFTTT applet.

```
notify:
  task:
    via:
      - ifttt:
          keys:
            - some api key
            - other api key
          event: test
```
```
2018-01-02 13:53 INFO     ifttt         notify_download Sent notification to key: some_valid_api_key
2018-01-02 13:53 ERROR    ifttt         notify_download Error sending to https://maker.ifttt.com/trigger/test/with/key/bad_key: 401 Client Error: Unauthorized for url: https://maker.ifttt.com/trigger/test/with/key/bad_key
```

### DIFF
--- a/flexget/plugins/notifiers/ifttt.py
+++ b/flexget/plugins/notifiers/ifttt.py
@@ -1,6 +1,5 @@
 from __future__ import unicode_literals, division, absolute_import
 from builtins import *  # noqa pylint: disable=unused-import, redefined-builtin
-from future.utils import text_to_native_str
 
 import logging
 
--- a/flexget/plugins/notifiers/ifttt.py
+++ b/flexget/plugins/notifiers/ifttt.py
@@ -10,7 +10,6 @@ from flexget.utils.requests import Session
 from flexget.plugin import PluginWarning
 
 from requests.exceptions import RequestException
-from types import StringTypes
 
 plugin_name = 'ifttt'
 log = logging.getLogger(plugin_name)
@@ -83,7 +82,7 @@ class IFTTTNotifier(object):
             raise PluginWarning("Failed to send notifications")
 
     def prepare_config(self, config):
-        if isinstance(config['keys'], StringTypes):
+        if not isinstance(config['keys'], list):
             config['keys'] = [config['keys']]
         return config
 
--- a/flexget/plugins/notifiers/ifttt.py
+++ b/flexget/plugins/notifiers/ifttt.py
@@ -7,7 +7,10 @@ from flexget import plugin
 from flexget.config_schema import one_or_more
 from flexget.event import event
 from flexget.utils.requests import Session
-from requests.exceptions import RequestError
+from flexget.plugin import PluginWarning
+
+from requests.exceptions import RequestException
+from types import StringTypes
 
 plugin_name = 'ifttt'
 log = logging.getLogger(plugin_name)
@@ -65,14 +68,24 @@ class IFTTTNotifier(object):
         :param str title: message subject
         :param dict config: plugin config
         """
+        config = self.prepare_config(config)
         notification_body = {'value1': title, 'value2': message}
+        errors = False
         for key in config['keys']:
             url = self.url_template.format(config['event'], key)
             try:
-                res = self.session.post(url, json=notification_body)
+                self.session.post(url, json=notification_body)
                 log.info("Sent notification to key: {}".format(key))
-            except RequestError as e:
-                log.error("Error sending to {}: {}".format(url, e))
+            except RequestException as e:
+                log.error("Error sending notification to key {}: {} ".format(key, e))
+                errors += 1
+        if errors:
+            raise PluginWarning("Failed to send notifications")
+
+    def prepare_config(self, config):
+        if isinstance(config['keys'], StringTypes):
+            config['keys'] = [config['keys']]
+        return config
 
 
 @event('plugin.register')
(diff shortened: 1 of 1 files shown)


flexget/plugins/notifiers/ifttt.py
```diff
@@ -0,0 +1,85 @@
...
+        with the values for the config, with a json body setting 'value1' to the message title,
+        and 'value2' to the message body.
+
+        If multiple keys are provided the event will be triggered for all of them.
+
+        :param str message: message body
+        :param str title: message subject
+        :param dict config: plugin config
+        """
+        notification_body = dict()
```
liiight: Just declare the dict with the data directly

flexget/plugins/notifiers/ifttt.py
```diff
@@ -0,0 +1,85 @@
...
+        notification_body['value1'] = title
+        notification_body['value2'] = message
+        for key in config['keys']:
```
liiight: This won't work if `keys` is a string. You need to make sure it's a list. Look into other notifiers with `prepare_data` methods

flexget/plugins/notifiers/ifttt.py
```diff
@@ -0,0 +1,85 @@
...
+            url = self.url_template.format(config['event'], key)
+            try:
+                res = self.session.post(url, json = notification_body)
```
liiight: Remove spaces around key word calls.
Also, using `Session()` will raise on errors by default, so you'll never reach that else clause. Just remove the conditionals and assume that the notification was successful. Handle the failure in the except section

flexget/plugins/notifiers/ifttt.py
```diff
@@ -0,0 +1,85 @@
...
+                if res.ok:
+                    log.info("Sent notification to key: {}".format(key))
+                else:
+                    log.error("Error sending to: {}: {} {} -- {}".format(url, res.status_code, res.reason, res.text))
+            except HTTPError as e:
+                log.error("Error sending to {}: {}".format(url, e))
```
liiight: Please raise `plugin.PluginWarning` here. Look into other notifiers

flexget/plugins/notifiers/ifttt.py
```diff
@@ -0,0 +1,85 @@
...
+        notification_body['value2'] = message
+        for key in config['keys']:
+            url = self.url_template.format(config['event'], key)
+            try:
+                res = self.session.post(url, json = notification_body)
+                if res.ok:
+                    log.info("Sent notification to key: {}".format(key))
+                else:
+                    log.error("Error sending to: {}: {} {} -- {}".format(url, res.status_code, res.reason, res.text))
+            except HTTPError as e:
```
liiight: I think it's better to catch all RequestError here, just in case

flexget/plugins/notifiers/ifttt.py
```diff
@@ -0,0 +1,93 @@
...
+                self.session.post(url, json=notification_body)
+                log.info("Sent notification to key: {}".format(key))
+            except RequestException as e:
+                log.error("Error sending notification to key {}: {} ".format(key, e))
+                errors += 1
+        if errors:
+            raise PluginWarning("Failed to send notifications")
+
+    def prepare_config(self, config):
+        if isinstance(config['keys'], StringTypes):
```
cvium: `if not isinstance(config['keys'], list):`

kroozo:
@liiight Fixed it up based on your comments, thanks.

Few questions:
- I've moved the PluginWarning after the loop so that a bad endpoint does not kill the rest. However, would it make sense to raise a PluginError instead if _all_ of them fail?
- Tests: does it make any sense for you to add any? The whole thing is pretty straightforward. I can add unit tests around some of these, like key is string, key is list, bad key at the front, whatever, but for me that mostly just sounds like a burden that will need to be changed if anyone touches this anyway. The only thing that would be beneficial is checking if the IFTTT endpoint is breaking, but I would be hesitant to hook some random account of theirs into your CI.

Also, while I was scratching my own itch I realized that you actually have a feature request for this: https://feathub.com/Flexget/Flexget/+19

flexget/plugins/notifiers/ifttt.py
```diff
@@ -0,0 +1,93 @@
...
+        :param dict config: plugin config
+        """
+        config = self.prepare_config(config)
+        notification_body = {'value1': title, 'value2': message}
+        errors = False
+        for key in config['keys']:
+            url = self.url_template.format(config['event'], key)
+            try:
+                self.session.post(url, json=notification_body)
+                log.info("Sent notification to key: {}".format(key))
```
liiight: let the logger handle its own interpolation:
`log.info("Sent notification to key: %s", key)`

flexget/plugins/notifiers/ifttt.py
kroozo: list it is then :)
I am not a big fan of str with all the unicode hassle, that is why I usually resort to StringTypes. And testing for a list has other caveats (they end up being tuples sometimes for eg) but if it is more consistent with the rest of your code base, then sure. jsonschema makes it moot anyway most probably here.

flexget/plugins/notifiers/ifttt.py
kroozo: fixed

PR #2065, opened 2018-01-25
Fix unicode handling for urlopen in Python 2

### LINKED ISSUE #1965: filesystem plugin fails when encountering unusual characters in file name
I see there is a bug reported about unicode characters leading to crash, but I don't think it's the same issue. This isn't about python3, and it's not about a crash, just a mishandling of (possibly unicode?) characters in a plugin. I think.

### Expected behaviour:
Expect filesystem plugin to handle files with untypical characters in file name

### Actual behaviour:
filesystem plugin fails to process files with certain characters in file name. In my experince: Apostroph (’) and japanese unicode characters. Probably quite a few other characters, too.

One might say that such characters shouldn't be used in file names (I'd agree with that!), but that's the file name provided for files from some trackers. Such trackers should perhaps be more conservative in their data output, but flexget should also be liberal in what data input it accepts.

### Steps to reproduce:
- Step 1: A torrent file exists in watchdir processed by filesystem plugin contains this character:  --> ’  <-- , say something like Artist - Artist's Album.torrent
- Step 2. Entry failes: <urlopen error [Errno 2] No such file or directory: '/home/username/watchdir/Artist - Artist\xc3\xa2\xc2\x80\xc2\x99s Album.torrent'>

#### Config:
```
PARTIAL config:
tasks:
  music-watchdir:
    filesystem:
      path: /home/username/watchdir
      mask: '*.torrent'
    set:
      path: /home/username/temp_downloads
      movedone: /home/username/downloads
      label: music
    accept_all: yes
    deluge:
      port: 6666


```
  
#### Log:
```
Too much private info in log
```

### Additional information:

- FlexGet version: 2.10.89 You are on the latest release.
- Python version: Python 2.7.9
- Installation method: pip
- Using daemon (yes/no): no
- OS and version: Debian stable
- Link to crash log:

cvium:
### Detailed changes:
- Encoding to utf-8 in Py2

### Addressed issues:
- Fixes #1965

### DIFF
--- a/flexget/utils/requests.py
+++ b/flexget/utils/requests.py
@@ -2,6 +2,7 @@ from __future__ import unicode_literals, division, absolute_import
 from builtins import *  # noqa pylint: disable=unused-import, redefined-builtin
 from future.moves.urllib.request import urlopen
 from future.moves.urllib.parse import urlparse
+from future.utils import PY2
 
 import time
 import logging
@@ -139,7 +140,11 @@ def _wrap_urlopen(url, timeout=None):
 
     """
     try:
-        raw = urlopen(url, timeout=timeout)
+        if PY2:
+            encoded_url = url.encode('utf-8')
+        else:
+            encoded_url = url
+        raw = urlopen(encoded_url, timeout=timeout)
     except IOError as e:
         msg = 'Error getting %s: %s' % (url, e)
         log.error(msg)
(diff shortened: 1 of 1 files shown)


cvium:
This one I'm not too sure about. Problem is that unicode characters seem to be a problem for urlopen in Python 2, but in Python 3 it expects a str and not a byte string.

PR #2069, opened 2018-01-30
Allow to set website URL for piratebay plugin

### LINKED ISSUE #2068: Allow PirateBay proxies/mirrors via the config file
The PirateBay is blocked in some countries like the UK and The Netherlands.

At the moment the URL for the piratebay is (relatively) hardcoded in the plugin, it would be useful for users from those countries if they could specify a proxy/mirror in their config file that they would be used instead of the regular domain name.

mfonville:
Fixes #2068 for piratebay reverse-proxies

~~Cassettes for testing should be updated, but I can't run those tests against the original domain, because for me it is blocked~~

### DIFF
--- a/flexget/plugins/sites/piratebay.py
+++ b/flexget/plugins/sites/piratebay.py
@@ -4,6 +4,7 @@ from future.moves.urllib.parse import quote
 
 import re
 import logging
+import urlparse
 
 from flexget import plugin
 from flexget.entry import Entry
@@ -15,8 +16,7 @@ from flexget.utils.tools import parse_filesize
 
 log = logging.getLogger('piratebay')
 
-PROTOCOL = 'https'
-DOMAIN = 'thepiratebay.org'
+URL = 'https://thepiratebay.org'
 
 CATEGORIES = {
     'all': 0,
@@ -48,8 +48,7 @@ class UrlRewritePirateBay(object):
             {
                 'type': 'object',
                 'properties': {
-                    'domain': {'type': 'string', 'default': DOMAIN},
-                    'protocol:': {'type': 'string', 'default': PROTOCOL},
+                    'url': {'type': 'string', 'default': URL},
                     'category': {
                         'oneOf': [
                             {'type': 'string', 'enum': list(CATEGORIES)},
@@ -64,8 +63,15 @@ class UrlRewritePirateBay(object):
         ]
     }
 
-    url_match = re.compile('^%s://(?:torrents\.)?(%s)/.*$' % (re.escape(PROTOCOL), re.escape(DOMAIN)))
-    url_search = re.compile('^%s://(%s)/search/.*$' % (re.escape(PROTOCOL), re.escape(DOMAIN)))
+    url = URL
+
+    def on_task_start(self, task, config=None):
+        if not isinstance(config, dict):
+            config = {}
+        self.url = config.get('url', URL)
+        parsed_url = urlparse(self.url)
+        self.url_match = re.compile('^%s://(?:torrents\.)?(%s)/.*$' % (re.escape(parsed_url.scheme), re.escape(parsed_url.netloc)))
+        self.url_search = re.compile('^%s/search/.*$' % (re.escape(self.url)))
 
     # urlrewriter API
     def url_rewritable(self, task, entry):
@@ -115,12 +121,6 @@ class UrlRewritePirateBay(object):
         """
         if not isinstance(config, dict):
             config = {}
-
-        protocol = config.get('protocol', PROTOCOL)
-        domain = config.get('domain', DOMAIN)
-        self.url_match = re.compile('^%s://(?:torrents\.)?(%s)/.*$' % (re.escape(protocol), re.escape(domain)))
-        self.url_search = re.compile('^%s://(%s)/search/.*$' % (re.escape(protocol), re.escape(domain)))
-
         sort = SORT.get(config.get('sort_by', 'seeds'))
         if config.get('sort_reverse'):
             sort += 1
@@ -138,7 +138,7 @@ class UrlRewritePirateBay(object):
             query = query.replace('-', ' ').replace("'", " ")
 
             # urllib.quote will crash if the unicode string has non ascii characters, so encode in utf-8 beforehand
-            url = '%s://%s/search/%s%s' % (protocol, domain, quote(query.encode('utf-8')), filter_url)
+            url = '%s/search/%s%s' % (self.url, quote(query.encode('utf-8')), filter_url)
             log.debug('Using %s as piratebay search url' % url)
             page = task.requests.get(url).content
             soup = get_soup(page)
@@ -148,7 +148,7 @@ class UrlRewritePirateBay(object):
                 if not entry['title']:
                     log.error('Malformed search result. No title or url found. Skipping.')
                     continue
-                entry['url'] = '%s://%s%s' % (protocol, domain, link.get('href'))
+                entry['url'] = self.url + link.get('href')
                 tds = link.parent.parent.parent.find_all('td')
                 entry['torrent_seeds'] = int(tds[-2].contents[0])
                 entry['torrent_leeches'] = int(tds[-1].contents[0])
--- a/flexget/plugins/sites/piratebay.py
+++ b/flexget/plugins/sites/piratebay.py
@@ -62,13 +62,19 @@ class UrlRewritePirateBay(object):
         ]
     }
 
+    def __init__(self):
+        self.set_urls(URL)
+
     def on_task_start(self, task, config=None):
         if not isinstance(config, dict):
             config = {}
-        self.url = config.get('url', URL)
-        parsed_url = urlparse(self.url)
+        self.set_urls(config.get('url', URL))
+
+    def set_urls(self, url):
+        self.url = url
+        parsed_url = urlparse(url)
         self.url_match = re.compile('^%s://(?:torrents\.)?(%s)/.*$' % (re.escape(parsed_url.scheme), re.escape(parsed_url.netloc)))
-        self.url_search = re.compile('^%s/search/.*$' % (re.escape(self.url)))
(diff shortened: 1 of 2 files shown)


flexget/plugins/sites/piratebay.py
```diff
@@ -65,20 +64,23 @@ class UrlRewritePirateBay(object):
         ]
     }
 
+    url_match = re.compile('^%s://(?:torrents\.)?(%s)/.*$' % (re.escape(PROTOCOL), re.escape(DOMAIN)))
```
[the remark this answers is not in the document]
mfonville: Yes, but I wasn't sure if the config would have been initialized if the url-rewrite functions would be used. So that is why I initalize them here for the default thepiratebay.org-value

flexget/plugins/sites/piratebay.py
```diff
@@ -113,6 +115,12 @@ def search(self, task, entry, config=None):
         """
         if not isinstance(config, dict):
             config = {}
+
+        protocol = config.get('protocol', PROTOCOL)
+        domain = config.get('domain', DOMAIN)
+        self.url_match = re.compile('^%s://(?:torrents\.)?(%s)/.*$' % (re.escape(protocol), re.escape(domain)))
```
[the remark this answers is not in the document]
mfonville: See previous comment, in the 'search' I am sure that there is a config-value passed and that they could override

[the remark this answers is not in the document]
cvium: This belongs in `on_task_start` if anything

flexget/plugins/sites/piratebay.py
```diff
@@ -51,6 +48,8 @@ class UrlRewritePirateBay(object):
             {
                 'type': 'object',
                 'properties': {
+                    'domain': {'type': 'string', 'default': DOMAIN},
```
cvium: Couldn't it just be URI instead? I don't see the benefit in splitting it up.

mfonville: We could, but probably would have to drop the support for the `torrents.`-subdomain support in the `url_match` statement (or do some dirty hack there)

flexget/plugins/sites/piratebay.py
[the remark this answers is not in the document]
mfonville: @cvium thanks for your tip, do you have an example of the on_task_start somewhere?

mfonville:
@cvium I pushed an update that uses on_task_start to set it, but it does not get called for the search-plugin; could you arrange that it does get called?

mfonville:
hmm, I might be wrong, and that it is not that `on_task_start` doesn't get called, but that the config-key is not read properly?

mfonville:
@cvium I'll be needing your help how to add this plugin to the task interface, and also how we could update the tests accordingly

cvium:
Add `'task'` to `interfaces`: https://github.com/Flexget/Flexget/pull/2069/files#diff-d8847fd4a4b4d1b34e2270b81426a26eR180

mfonville:
I did add it to the interfaces, but still this `on_task_start` method doesn't seem to get called that the `self.`-instance variables are stored.

e.g. I now tried with a dirty hack that it calls the method from `search`, but later the `url_rewritable` spawns a similar error:
```
2018-02-02 22:00 CRITICAL task          Series-Trakt    BUG: Unhandled error in plugin urlrewriting: 'UrlRewritePirateBay' object has no attribute 'url_match'
  File "/usr/local/lib/python2.7/dist-packages/flexget/task.py", line 486, in __run_plugin
    return method(*args, **kwargs)
  File "/usr/local/lib/python2.7/dist-packages/flexget/event.py", line 23, in __call__
  File "/usr/local/lib/python2.7/dist-packages/flexget/plugins/internal/urlrewriting.py", line 33, in on_task_urlrewrite
    self.url_rewrite(task, entry)
  File "/usr/local/lib/python2.7/dist-packages/flexget/plugins/internal/urlrewriting.py", line 55, in url_rewrite
    while self.url_rewritable(task, entry) and entry.accepted:
  File "/usr/local/lib/python2.7/dist-packages/flexget/plugins/internal/urlrewriting.py", line 46, in url_rewritable
    if urlrewriter.instance.url_rewritable(task, entry):
  File "/usr/local/lib/python2.7/dist-packages/flexget/plugins/sites/piratebay.py", line 78, in url_rewritable
    return bool(self.url_match.match(url))
AttributeError: 'UrlRewritePirateBay' object has no attribute 'url_match'
```

cvium:
I guess adding it to the interfaces is not enough. The config must be at a task level to trigger `on_task_start`. I haven't had a chance to test it, sorry.

I think the best option is to keep it in `on_task_start` and then include it in `search` if the class property hasn't been set.

mfonville:
but should we then also calll `on_task_start` from the `url_rewritable`-method? because it ain't triggered there either according to the stacktrace

cvium:
Three scenarios:

1. No config: Need to resort to a default regexp
2. Task level config: `on_task_start` is triggered
3. `discover`: `on_task_start` is _not_ triggered

mfonville:
@cvium could you check if these changes would suffice for the proper workflow? The regexeps are now in their own function and set via a constructor and overriden from the methods that accept a config to be passed.

Still the tests need to be updated with someone that actually connect to the piratebay.org website (which is impossible for me in NL)

flexget/plugins/sites/piratebay.py
```diff
@@ -65,20 +62,34 @@ class UrlRewritePirateBay(object):
...
+
+    def on_task_start(self, task, config=None):
+        if not isinstance(config, dict):
+            config = {}
+        self.set_urls(config.get('url', URL))
+
+    def set_urls(self, url):
+        self.url = url
+        parsed_url = urlparse(url)
+        self.url_match = re.compile('^%s://(?:torrents\.)?(%s)/.*$' % (re.escape(parsed_url.scheme), re.escape(parsed_url.netloc)))
```
cvium: You're still compiling these once per entry.

mfonville:
Hey, wondering if this PR can be accepted, or if it'd require more work?

flexget/plugins/sites/piratebay.py
```diff
@@ -65,20 +62,37 @@ class UrlRewritePirateBay(object):
...
+    def on_task_start(self, task, config=None):
+        if not isinstance(config, dict):
+            config = {}
+        self.set_urls(config.get('url', URL))
+
+    def set_urls(self, url):
+        if self.url != url:
+            self.url = url
+            parsed_url = urlparse(url)
+            self.url_match = re.compile('^%s://(?:torrents\.)?(%s)/.*$' % (re.escape(parsed_url.scheme), re.escape(parsed_url.netloc)))
```
cvium: Line too long. 120 is max.

flexget/plugins/sites/piratebay.py
```diff
@@ -51,6 +47,7 @@ class UrlRewritePirateBay(object):
             {
                 'type': 'object',
                 'properties': {
+                    'url': {'type': 'string', 'default': URL},
```
cvium: Add `'format': 'url'`

flexget/plugins/sites/piratebay.py
```diff
@@ -65,20 +62,37 @@ class UrlRewritePirateBay(object):
...
+        self.set_urls(URL)
+
+    def on_task_start(self, task, config=None):
+        if not isinstance(config, dict):
+            config = {}
+        self.set_urls(config.get('url', URL))
+
+    def set_urls(self, url):
+        if self.url != url:
+            self.url = url
```
cvium: You should probably `rstrip` forward slashes here since you seem to make the assumption it doesn't have any later on.

mfonville:
I implemented the suggestions from the review

flexget/plugins/sites/piratebay.py
```diff
@@ -15,11 +15,7 @@
 
 log = logging.getLogger('piratebay')
 
-CUR_TLD = 'org'
-TLDS = 'com|org|sx|ac|pe|gy|to|se|gd|vg|%s' % CUR_TLD
```
cvium: What about the old TLDs? We used to match them all and rewrite to the CUR_TLD.

mfonville: This is not applicable anymore, because the piratebay (as far as I know) does not provide domain-specific URLs anymore for torrents (only magnet links) and the only rewriting happening is to match the domain of the reverse-proxy (or the original website) that is used to access tpb

flexget/plugins/sites/piratebay.py
```diff
@@ -65,20 +62,39 @@ class UrlRewritePirateBay(object):
...
+            config = {}
+        self.set_urls(config.get('url', URL))
+
+    def set_urls(self, url):
+        url = url.rstrip('/')
+        if self.url != url:
+            self.url = url
+            parsed_url = urlparse(url)
+            self.url_match = re.compile('^%s://(?:torrents\.)?(%s)/.*$' % (re.escape(parsed_url.scheme),
+                                                                           re.escape(parsed_url.netloc)))
```
cvium: Is this pep8-compliant? I'd expect it to align with the re.escape above.

mfonville: Yes, this is compliant. Alignment is based on the parentheses, which is the python-preferred way using the opening delimiter as anchor.

cvium: Hmm, when I was looking at this at home it was definitely not compliant. I can see now (at work) that it is. I'll chalk it up to ninja-edits or my monospace font acting up.

cvium:
I have a few reservations about this change, but I also can't really come up with a better alternative.

Will you update the wiki?

mfonville:
Thanks for merging. I understand the reservations, but I also don't see any alternative. Maybe with a future larger redesign/overhaul of flexget we could improve the codebase to adapt to this feature.

I updated the wiki: https://flexget.com/Searches/piratebay

PR #2070, opened 2018-01-31
Adapt to latest newpct1 changes

kir4h:
Fix URL for Newpct1 and modify regular expression to capture the new torrent ID

### Motivation for changes:

Plugin started failing when downloading torrents from newpct1 feeds

### Detailed changes:
- Changed URL rewrite prefix to adapt to newpct1 changes
- Changed regular expression capturing torrent ID to adapt to newpct/newpct1 changes

```
Tested the changes locally and the plugin is working again for newpct1
```
#### To Do:

- [ ] Ensure newpct.com is also working and if not fix it. Probably not working either before this patch as torrent id seems to have changed.

### DIFF
--- a/flexget/plugins/sites/newpct.py
+++ b/flexget/plugins/sites/newpct.py
@@ -23,7 +23,7 @@ requests.add_domain_limiter(TimedLimiter('newpct1.com', '2 seconds'))
 requests.add_domain_limiter(TimedLimiter('newpct.com', '2 seconds'))
 
 NEWPCT_TORRENT_FORMAT = 'http://www.newpct.com/torrents/{:0>6}.torrent'
-NEWPCT1_TORRENT_FORMAT = 'http://www.newpct1.com/torrents/{:0>6}.torrent'
+NEWPCT1_TORRENT_FORMAT = 'http://www.newpct1.com/download/{:0>6}.torrent'
 
 class UrlRewriteNewPCT(object):
     """NewPCT urlrewriter and search."""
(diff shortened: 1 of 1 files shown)


cvium:
This plugin breaks a lot. Maybe it's time to redesign it to take a regexp via config.

kabracity:
Agreed, but given that a regexp config would make it more complicated for the end user, this regexp should be updated somewhere (so that at some point a user can go there and paste it into his config file). 
On the other hand, this time it was not only the regexp, but also the rewritten url that changed its path.

cvium:
I agree it might be too complicated. Just be prepared to fix it again :smile:

Aokromes:
Tried after today's changes, still fails for me.

cvium:
Today's changes?

Aokromes:
yeah it was working this night, now fails.

cvium:
How is it related to this PR? It hasn't been merged yet...

Aokromes:
I manually applied this pr and it still fails.

kabracity:
@Aokromes  which link failed to be downloaded in your case? I tried a random one from today's feed (Arrow 612) and the same logic is still working for me. Is it newpct1 or newpct?

Aokromes:
flexget failed list
Chicago Fire - Temporada 3 [HDTV][Cap.318][Español Castellano]                               x 2          x        
for example using http://feeds2.feedburner.com/newpctorrent so, newpct only.

kabracity:
Checked it and seems the torrent itself is not available, so the issue is likely to be related to that

PR #2072, opened 2018-02-01
add encoding field to regexp_parse

stanionascu:
### Motivation for changes:
Sometimes servers are not aware of the content encoding which they serve, thus it is necessary to give a hint to Flexget about it.

### Detailed changes:
- Field **encoding** is added as _optional_ to the schema of **regexp_parse** plugin. If the value is set, it will be given to requests.Response.encoding as a hint about content encoding.

### Addressed issues:
N/A

N/A

**regexp_parse** plugin schema is updated with an optional **encoding** field and documentation is updated.
```
    regexp_parse:
      source: http://username:password@ezrss.it/feed/
      encoding: "utf-8"
      sections:
        - {regexp: "(?<=<item>).*?(?=</item>)", flags: "DOTALL,IGNORECASE"}
```

N/A

#### To Do:

- [ ] **Optional**: make e.g. encoding == _auto_, use requests.Response.apparent_encoding

### DIFF
--- a/flexget/plugins/input/regexp_parse.py
+++ b/flexget/plugins/input/regexp_parse.py
@@ -43,6 +43,7 @@ class RegexpParse(object):
 
     regexp_parse:
       source: http://username:password@ezrss.it/feed/
+      encoding: "utf-8"
       sections:
         - {regexp: "(?<=<item>).*?(?=</item>)", flags: "DOTALL,IGNORECASE"}
 
@@ -88,6 +89,7 @@ class RegexpParse(object):
 
         root.accept('url', key='source', required=True)
         root.accept('file', key='source', required=True)
+        root.accept('text', key='encoding')
 
         # sections to divied source into
         sections_regexp_lists = root.accept('list', key='sections')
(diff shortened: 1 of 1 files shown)


flexget/plugins/input/regexp_parse.py
```diff
@@ -152,13 +154,19 @@ def isvalid(self, entry):
     @plugin.internet(log)
     def on_task_input(self, task, config):
         url = config['source']
+        encoding = config.get('encoding')
 
         # if it's a file open it and read into content (assume utf-8 encoding)
         if os.path.isfile(os.path.expanduser(url)):
-            content = codecs.open(url, 'r', encoding='utf-8').read()
+            if encoding is None:
```
cvium: Should the encoding be set to `utf-8` if `encoding` is `None`?

stanionascu: In my opinion, `encoding` should not be set at all if it's `None` (to preserve whatever defaults), but the patch tries to not break the current behavior (when encoding for open is always set to utf-8).

flexget/plugins/input/regexp_parse.py
```diff
@@ -152,13 +154,19 @@ def isvalid(self, entry):
     @plugin.internet(log)
     def on_task_input(self, task, config):
         url = config['source']
+        encoding = config.get('encoding')
```
cvium: You can just add the default here `config.get('encoding', 'utf-8')`. I'm not sure why we assume utf-8, but let's not break anything.

stanionascu: Done as suggested. Thanks!

paranoidi:
Seems innocent enough, merge and update the documentation?

gazpachoking:
I think I agree with @stanionascu original theory that we shouldn't be forcing the encoding to be utf-8 when the user has not specified an encoding in the config. Requests uses headers/chardet to determine the best encoding, and always overriding this with utf-8 seems less desirable.

gazpachoking:
I made the appropriate changes and merged this.

PR #2078, opened 2018-02-12
Add jinja2 templates to ftp_download and fix infinite loop bug.

TomyMMX:
### Motivation for changes:
ftp_download is really not useful without the ability to use jinja2 templates.
Also made a quick fix for cases when we run into download problems and the download loop starts to loop forever.

### Detailed changes:
- Really simple fix and feature so all is explained above.

### DIFF
--- a/flexget/plugins/output/ftp_download.py
+++ b/flexget/plugins/output/ftp_download.py
@@ -90,10 +90,22 @@ class OutputFtp(object):
             except ftplib.all_errors as e:
                 entry.fail("Unable to connect to server : %s" % (e))
                 break
-            to_path = pathscrub(entry.render(config['ftp_tmp_path']))
-            if not os.path.isdir(to_path):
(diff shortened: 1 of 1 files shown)


flexget/plugins/output/ftp_download.py
```diff
@@ -88,25 +90,25 @@ def on_task_download(self, task, config):
             except ftplib.all_errors as e:
                 entry.fail("Unable to connect to server : %s" % (e))
                 break
-
-            if not os.path.isdir(config['ftp_tmp_path']):
-                log.debug('creating base path: %s' % config['ftp_tmp_path'])
-                os.mkdir(config['ftp_tmp_path'])
+            to_path = pathscrub(entry.render(config['ftp_tmp_path']))
```
liiight: better to catch `RenderError` here

TomyMMX: @liiight fixed as per your comment.

PR #2082, opened 2018-02-25
Update cache control in API to better handle dynamic content

stevezau:
Update ETag Cache params for the WebUI.

### Motivation for changes:
The etag for content such as getting the entries within a list should have a cache max-age of 0 otherwise you may see weird behaviour within the WebUI.

### DIFF
--- a/flexget/api/core/server.py
+++ b/flexget/api/core/server.py
@@ -164,7 +164,7 @@ class ServerPIDAPI(APIResource):
 
 @server_api.route('/config/')
 class ServerConfigAPI(APIResource):
-    @etag(cache_age=0)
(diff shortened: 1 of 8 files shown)


flexget/api/core/server.py
```diff
@@ -164,7 +164,7 @@ def get(self, session=None):
 
 @server_api.route('/config/')
 class ServerConfigAPI(APIResource):
-    @etag
+    @etag(cache_age=0)
```
cvium: This is the default value, so there's no need to explicitly set it here. It could be argued that `cache_age` shouldn't have a default at all though.

PR #2085, opened 2018-02-26
Tasks api disabled tasks

### LINKED ISSUE #2084: Tasks API should no return hidden tasks
### Expected behaviour:

Tasks API should only return visible tasks
### Actual behaviour:
Return all tasks, including those starting with `-`

liiight:
### Motivation for changes:
Hidden tasks should be returned by the API
### Detailed changes:
- Filtered out hidden tasks for `/tasks/` endpoint

### Addressed issues:
- Fixes #2084

### DIFF
--- a/flexget/api/app.py
+++ b/flexget/api/app.py
@@ -21,7 +21,7 @@ from flexget.utils.database import with_session
 from flexget.webserver import User
 from . import __path__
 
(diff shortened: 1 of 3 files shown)


stevezau:
Looks fine to me.. However, i've never heard of tasks starting with _ before so taking your lead on it.

PR #2087, opened 2018-02-28
Added ability to filter pending list entries via API

liiight:
### Motivation for changes:
@stevezau asked
### Detailed changes:
- Added `/pending_list/<list_id>/entries?search=<string>` endpoint

### DIFF
--- a/flexget/plugins/list/pending_list.py
+++ b/flexget/plugins/list/pending_list.py
@@ -208,9 +208,11 @@ def delete_list_by_id(list_id, session=None):
 
 @with_session
 def get_entries_by_list_id(list_id, start=None, stop=None, order_by='title', descending=False, approved=False,
-                           session=None):
+                           filter=None, session=None):
     log.debug('querying entries from pending list with id %d', list_id)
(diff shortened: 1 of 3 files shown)


flexget/plugins/list/pending_list.py
```diff
@@ -208,9 +208,11 @@ def delete_list_by_id(list_id, session=None):
 
 @with_session
 def get_entries_by_list_id(list_id, start=None, stop=None, order_by='title', descending=False, approved=False,
-                           session=None):
+                           filter=None, session=None):
     log.debug('querying entries from pending list with id %d', list_id)
     query = session.query(PendingListEntry).filter(PendingListEntry.list_id == list_id)
+    if filter:
+        query = query.filter(func.lower(PendingListEntry.title).contains(filter.lower()))
```
stevezau: What happens if i search for "Some Series" but the titles are Some.Series?

liiight: It won't find it. Just a substring search currently

PR #2094, opened 2018-03-07
Ombi list plugin

bagobones:
### Motivation for changes:
Flexget lacks a UI for requesting content to download. Ombi implements a great multi user request system.

As of Ombi v3 is has a very detailed API and runs well in a docker.

ttps://github.com/tidusjar/Ombi

### Detailed changes:
- New basic ombi plugin that implements basic move, shows, seasons, episodes list functions 

- Feathub #[38](https://feathub.com/Flexget/Flexget/+38). "This was my own request, found some time to work on it"

```
    schema = {
        'type': 'object',
        'properties': {
            'base_url': {'type': 'string'},
            'port': {'type': 'number', 'default': 3579},
            'api_key': {'type': 'string'},
            'type': {'type': 'string', 'enum': ['shows', 'seasons', 'episodes', 'movies']},            
            'only_approved': {'type': 'boolean', 'default': True},
            'include_available': {'type': 'boolean', 'default': False},
            'strip_year': {'type': 'boolean', 'default': False}
        },
        'required': ['api_key', 'base_url', 'type'],
        'additionalProperties': False
    }
```

#### To Do:

- Adding (add) and Deleting (discard) items from Ombi via the API is not implemented.


flexget/plugins/list/ombi_list.py
```diff
@@ -0,0 +1,256 @@
...
+                log.debug('Found series: %s', request["title"])
+                if config.get('type') == 'shows':
+                    entry = Entry(title=generate_title(request, config.get('strip_year')),
+                                  url='http://www.imdb.com/title/' + request.get('imdbId') +'/',
+                                  series_name=generate_title(request, config.get('strip_year')),
+                                  tvdb_id=request.get('tvDbId'),
+                                  imdb_id=request.get('imdbId'),
+                                  ombi_status=request.get('status'),
+                                  ombi_request_id=request.get('id'))
+                    if entry.isvalid():
```
liiight: consider flipping the logic for better looking indentation:
```python
if not entry.isvalid():
    log.error('Invalid entry created? %s', entry)
    continue
log.debug('Valid entry %s', entry)
entries.append(entry)
```

flexget/plugins/list/ombi_list.py
```diff
@@ -0,0 +1,256 @@
...
+                        log.debug('Valid entry %s', entry)
+                        entries.append(entry)
+                    else:
+                        log.error('Invalid entry created? %s', entry)
+                    continue                
+                else:
```
liiight: all of the else clauses are way too long and cumbersome. please break them into separate methods, and also try to take the repeated parts to separate method as well

flexget/plugins/list/ombi_list.py
```diff
@@ -0,0 +1,256 @@
...
+import requests
+from requests import RequestException
+
+from flexget import plugin
+from flexget.entry import Entry
+from flexget.event import event
+
+log = logging.getLogger('ombi_list')
+
+def generate_title(item, strip):
```
liiight: why is this not another static class method? seems inconsistant

flexget/plugins/list/ombi_list.py
```diff
@@ -0,0 +1,256 @@
...
+                                  ombi_status=request.get('status'),
+                                  ombi_request_id=request.get('id'))
+                    if entry.isvalid():
+                        log.debug('Valid entry %s', entry)
+                        entries.append(entry)
+                    else:
+                        log.error('Invalid entry created? %s', entry)
+                    continue                
+                else:
+                    for childRequest in request["childRequests"]:
```
liiight: please use snake_case and not camel case for variable names

flexget/plugins/list/ombi_list.py
```diff
@@ -0,0 +1,256 @@
...
+            'only_approved': {'type': 'boolean', 'default': True},
+            'include_available': {'type': 'boolean', 'default': False},
+            'strip_year': {'type': 'boolean', 'default': False}
+        },
+        'required': ['api_key', 'base_url', 'type'],
+        'additionalProperties': False
+    }
+
+    @property
+    def requests(self):
```
liiight: requests isn't the best name for this, and while it technically doesn't overshadow the `request` lib used in the outer scope, it's still not a great idea to reuse the name. Maybe call it entities?

flexget/plugins/list/ombi_list.py
```diff
@@ -0,0 +1,256 @@
...
+        return (entry for entry in self.requests)
+
+    def __len__(self):
+        return len(self.requests)
+
+    def __contains__(self, entry):
+        return self._find_entry(entry) is not None
+
+    def add(self, entry):
+        log.verbose('List adding not yet implemented')
```
liiight: don't forget about this

flexget/plugins/list/ombi_list.py
```diff
@@ -0,0 +1,256 @@
...
+
+    def discard(self, entry):
+        log.verbose('List item removal not yet implemented')
```
liiight: don't forget about this

flexget/plugins/list/ombi_list.py
```diff
@@ -0,0 +1,256 @@
...
+                                                  ombi_request_id=request.get('id'),
+                                                  ombi_childrequest_id=childRequest["id"],
+                                                  ombi_season_id=season["id"],
+                                                  ombi_episode_id=episode["id"],
+                                                  ombi_approved=episode.get('approved'),
+                                                  ombi_available=episode.get('available'),
+                                                  ombi_requested=episode.get('requested'))
+                                    if entry.isvalid():
+                                        log.debug('Valid entry %s', entry)
+                                        if config.get('only_approved') and not entry.get('ombi_approved'):
```
liiight: you should check for these conditions before building the entry

flexget/plugins/list/ombi_list.py
```diff
@@ -0,0 +1,256 @@
...
+                                            log.verbose('Request not approved skipping: %s', entry.get('title'))
+                                        else:
+                                            if not config.get('include_available') and entry.get('ombi_available'):
```
liiight: you should check for these conditions before building the entry

flexget/plugins/list/ombi_list.py
```diff
@@ -0,0 +1,256 @@
...
+        try:
+            return requests.get(url).json()
+        except RequestException as e:
+            raise plugin.PluginError('Unable to connect to Ombi at %s. Error: %s' % (url, e))
+
+    @staticmethod
+    def list_requests(config):
+        log.debug('Connecting to Ombi to retrieve request list.')
+        if config.get('type')  in ['movies']:
+            request_url = OmbiBase.movie_list_request(config.get('base_url'), config.get('port'), config.get('api_key'))
```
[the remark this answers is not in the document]
bagobones: This plugin is me bashing together couchpotoato_list and trakt_list ideas to learn python and how flexget plugins work at the same time.

Feedback noted.

flexget/plugins/list/ombi_list.py
bagobones: This one is likely the result of replacing the word movies with requests. As requests are what I am pulling from Ombi not movies like couchpotato.

I could see how the name could cause problems however.

flexget/plugins/list/ombi_list.py
```diff
@@ -0,0 +1,256 @@
...
+            return entries
+ 
+class OmbiSet(MutableSet):
+    supported_ids = ['imdb_id', 'tmdb_id', 'ombi_id']
+    schema = {
+        'type': 'object',
+        'properties': {
+            'base_url': {'type': 'string'},
+            'port': {'type': 'number', 'default': 3579},
+            'api_key': {'type': 'string'},
```
cvium: Consider adding username/password support. I started an ombi_list plugin and I have implemented both in a simple function. I can post it here later if you want.

flexget/plugins/list/ombi_list.py
```diff
@@ -0,0 +1,256 @@
...
+
+    @property
+    def requests(self):
+        if not self._requests:
+            self._requests = OmbiBase.list_requests(self.config)
+        return self._requests
+        
+    def show_match(self, entry1, entry2):
+        if any(entry1.get(ident) is not None and entry1[ident] == entry2.get(ident) for ident in
+               ['series_name', 'tvdb_id', 'imdb_id']):
```
cvium: Isn't the point to use supported_ids here?

flexget/plugins/list/ombi_list.py
```diff
@@ -0,0 +1,256 @@
...
+        return (self.show_match(entry1, entry2) and entry1.get('series_season') is not None and
+                entry1['series_season'] == entry2.get('series_season'))
+
+    def episode_match(self, entry1, entry2):
+        return (self.season_match(entry1, entry2) and entry1.get('series_episode') is not None and
+                entry1['series_episode'] == entry2.get('series_episode'))
+
+    def movie_match(self, entry1, entry2):
+        if any(entry1.get(id) is not None and entry1[id] == entry2[id] for id in
+               ['imdb_id', 'tmdb_id']):
```
cvium: Same id comment here. You end up with string duplication.

flexget/plugins/list/ombi_list.py
```diff
@@ -0,0 +1,256 @@
...
+from requests import RequestException
+
+from flexget import plugin
+from flexget.entry import Entry
+from flexget.event import event
+
+log = logging.getLogger('ombi_list')
+
+def generate_title(item, strip):
+    if item.get('releaseDate') and not strip:
```
cvium: releaseDate for an episode is the air date I take it? Won't make a whole lot of sense to include the air date in the title.

flexget/plugins/list/ombi_list.py
bagobones: Went with the URI api key for easy and similarity to other plugins. Also integration is the only reason Ombi generates a api_key so I am not sure why username and password would be better. There is a another api_key method using post methods as well I believe.

flexget/plugins/list/ombi_list.py
bagobones: To be honest i don't completely understand this logic as it was pulled over from the trakt plugin.

flexget/plugins/list/ombi_list.py
cvium: The api key gives the user admin rights. Username/password uses the built-in role administration.

flexget/plugins/list/ombi_list.py
bagobones: When taken from Ombi, it refers ether to the Movie release date or a Series Release date.

It is not used for an episode title nor does it refer to an episode airdate (maybe the first airdate of a series), at least in regards to ombi.

Most of the time thetvdb or imbdb it will be used for matching but when using the title for search of a TV series it is important to differentiate Doctor Who (2005) and Doctor Who (1963) when the list is used with discover and only the title is used.

flexget/plugins/list/ombi_list.py
bagobones: Dealing with a python noob who hasn't done any real coding in probably 10 years. Also another example of taking concepts directly from the trakt_list plugin.

Wasn't expecting to get high style points.. Figured there would be a lot to clean up.

liiight: I hope you take this feedback in the spirit it's given, which is a scrutinising as possible 😄 It's best to be as nit-picky as possible at the time of the creation since maintenance of all of the 3rd party plugins is scarce. You did a great job and I honestly didn't feel that a python noob did this.

cvium: In other words: We will probably be very rough, but you shouldn't take it personally, we just want high quality code.

bagobones: I appreciate the feedback.

flexget/plugins/list/ombi_list.py
```diff
@@ -0,0 +1,248 @@
...
+                          ombi_approved=episode.get('approved'),
+                          ombi_available=episode.get('available'),
+                          ombi_requested=episode.get('requested'))
+        else:
+            raise plugin.PluginError('Error: Unknown list type %s.' % (config.get('type')))
+            
+        if config.get('type') in ['shows','seasons']:
+            # shows/seasons do not have approval or available status so return
+            return entry
+        elif config.get('only_approved') and not entry.get('ombi_approved'):
```
cvium: These should be at the top. No need to create an entry that we won't use. The "fail fast" principle sort of applies here (it's not a failure though).

flexget/plugins/list/ombi_list.py
```diff
@@ -0,0 +1,248 @@
...
+        try:
+            return requests.get(url).json()
+        except RequestException as e:
+            raise plugin.PluginError('Unable to connect to Ombi at %s. Error: %s' % (url, e))
+
+    @staticmethod
+    def generate_entry(config, parent_request, child_request=False, season=False, episode=False):
+        log.debug('Generating Entry')
+        entry = Entry()
+        if config.get('type') == 'movies':
```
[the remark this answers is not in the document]
bagobones: This function does ONE thing, it builds an entry for the list, each type has fairly different fields, passing the config and using the type is very convenient. I see no need for additional functions.

flexget/plugins/list/ombi_list.py
```diff
@@ -0,0 +1,265 @@
...
+
+    @staticmethod
+    def get_json(url):
+        try:
+            return requests.get(url).json()
+        except RequestException as e:
+            raise plugin.PluginError('Unable to connect to Ombi at %s. Error: %s' % (url, e))
+
+    @staticmethod
+    def generate_entry(config, parent_request, child_request=False, season=False, episode=False):
```
cvium: The default value for non-booleans should be `None`.

flexget/plugins/list/ombi_list.py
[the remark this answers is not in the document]
cvium: It's convenient, but you very quickly end up with a mess. I already think it's messy. It may be convenient, but it's easier to read if you create separate functions eg. `generate_episode_entry`, where you pass the proper arguments. Just by looking at the function and its arguments, you can easily see what it does without reading through the code.

This is a prime example of the mess this approach can create: https://github.com/Flexget/Flexget/blob/develop/flexget/plugins/internal/api_trakt.py#L859-L1105

flexget/plugins/list/ombi_list.py
```diff
@@ -0,0 +1,265 @@
...
+            return entry
+        
+            
+    @staticmethod
+    def list_requests(config):
+        log.debug('Connecting to Ombi to retrieve request list.')
+        connection_url = OmbiBase.construct_url(config)
+        
+        log.debug('URL %s', connection_url)
+        json = OmbiBase.get_json(connection_url)
```
liiight: if you need to access other class method then don't make this a `staticmethod`

flexget/plugins/list/ombi_list.py
```diff
@@ -0,0 +1,265 @@
...
+                          ombi_requested=episode.get('requested'))
+        else:
+            raise plugin.PluginError('Error: Unknown list type %s.' % (config.get('type')))
+            
+        if config.get('type') in ['shows','seasons']:
+            # shows/seasons do not have approval or available status so return
+            return entry
+        elif config.get('only_approved') and not entry.get('ombi_approved'):
+            log.verbose('Request not approved skipping: %s', entry.get('title'))
+            return False
```
liiight: Don't return False, just return, which will return `None`

flexget/plugins/list/ombi_list.py
bagobones: Hmm I will look into it.

flexget/plugins/list/ombi_list.py
bagobones: Thanks, was going to go back and look a the python docks, was thinking null initially.

flexget/plugins/list/ombi_list.py
```diff
@@ -117,144 +149,77 @@ def generate_entry(config, parent_request, child_request=False, season=False, ep
...
-                if entry:
+            log.debug('URL %s', connection_url)
+            json = self.get_json(connection_url)
+            
+            self._items = []
+
+            for parent_request in json:            
+                if self.config.get('type') == 'movies':
+                    # check that the request is approved unless user has selected to include everything
+                    if (self.config.get('only_approved') and not parent_request.get('approved')) or parent_request.get('approved'):
```
cvium: Parenthesis not necessary

flexget/plugins/list/ombi_list.py
```diff
@@ -117,144 +149,77 @@ def generate_entry(config, parent_request, child_request=False, season=False, ep
...
+                        # Always include items that are not available and only include available items if user has selected to do so
+                        if (self.config.get('include_available') and parent_request.get('available') or not parent_request.get('available')):
```
cvium: Parenthesis not necessary

flexget/plugins/list/ombi_list.py
```diff
@@ -117,144 +149,77 @@ def generate_entry(config, parent_request, child_request=False, season=False, ep
...
-        return len(self.requests)
-
-    def __contains__(self, entry):
-        return self._find_entry(entry) is not None
+                                    log.debug('Entry %s', entry)
+                                    self._items.append(entry)
+                            else:
+                                for episode in season['episodes']:
+                                    # check that the request is approved unless user has selected to include everything
+                                    if (self.config.get('only_approved') and not episode.get('approved') or episode.get('approved')):
```
cvium: Parenthesis not necessary

flexget/plugins/list/ombi_list.py
```diff
@@ -117,144 +149,77 @@ def generate_entry(config, parent_request, child_request=False, season=False, ep
...
+                                        # Always include items that are not available and only include available items if user has selected to do so
+                                        if (self.config.get('include_available') and episode.get('available')) or not episode.get('available'):
```
cvium: Parenthesis not necessary

flexget/plugins/list/ombi_list.py
bagobones: User name and password support has been added.

bagobones:
likely looking into the add and remove functions next as well as error handling.

flexget/plugins/list/ombi_list.py
```diff
@@ -231,17 +230,36 @@ def items(self):
                                             self._items.append(entry)
         return self._items
 
+    def add_movie(self, entry):
+        log.verbose('Adding entry %s', entry.get('title'))
+        auth_header = self.ombi_auth()
+
+        parsedurl = urlparse(self.config.get('base_url'))
+        url = '%s://%s:%s%s/api/v1/Request/movie' % (parsedurl.scheme, parsedurl.netloc, self.config.get('port'), parsedurl.path)
```
cvium: I notice you've hardcoded this in a lot of places and I apologize for not bringing it up sooner.

1. Use global/module variables for the endpoints similar to this https://github.com/Flexget/Flexget/blob/develop/flexget/plugins/list/sonarr_list.py#L16-L20. You can then use a function like this https://github.com/Flexget/Flexget/blob/develop/flexget/plugins/internal/api_trakt.py#L200-L211 to generate a proper URL.
2. You use urlparse a lot but always on the same url. You should only do it once and re-use the values. You don't have to use it though. We have a json schema validator `"format": "url"` that you can use and then it's a simple matter of `url.rstrip('/')` to remove any trailing slashes.

flexget/plugins/list/ombi_list.py
```diff
@@ -231,17 +230,36 @@ def items(self):
                                             self._items.append(entry)
         return self._items
 
+    def add_movie(self, entry):
+        log.verbose('Adding entry %s', entry.get('title'))
```
cvium: We prefer to use brackets for values we _know_ will be present. A valid entry will _always_ have a title and url, so this should be `entry['title']`.

flexget/plugins/list/ombi_list.py
```diff
@@ -0,0 +1,281 @@
...
+            'api_key': {'type': 'string'},
+            'username': {'type': 'string'},
+            'password': {'type': 'string'},
+            'type': {'type': 'string', 'enum': ['shows', 'seasons', 'episodes', 'movies']},
+            'only_approved': {'type': 'boolean', 'default': True},
+            'include_available': {'type': 'boolean', 'default': False},
+            'include_year': {'type': 'boolean', 'default': False},
+            'include_ep_title': {'type': 'boolean', 'default': False}
+        },
+        'required': ['base_url', 'type'],
```
cvium: You should add this to enforce api_key _or_ username+password:

```
'oneOf': [
  {'required': ['username', 'password']},
  {'required': ['api_key']}
]
```

flexget/plugins/list/ombi_list.py
```diff
@@ -0,0 +1,281 @@
...
+        url = '%s://%s:%s%s/api/v1/Token' % (parsedurl.scheme, parsedurl.netloc, self.config.get('port'), parsedurl.path)
+        data = {'username': self.config.get('username'),
+                'password': self.config.get('password')
+        }
+        headers = { 'Content-Type': 'application/json',
+                    'Accept': 'application/json'
+        }
+        try:
+            log.debug('Logging in with username and password to get access token')
+            log.debug('URL %s', url)
```
cvium: You should import our requests util instead https://github.com/Flexget/Flexget/blob/develop/flexget/utils/requests.py, which incidentally debug logs _all_ requests, so you can remove this debug line.

flexget/plugins/list/ombi_list.py
```diff
@@ -0,0 +1,281 @@
...
+    def ombi_auth(self):
+        if self.config.get('api_key'):
+            log.debug('api_key auth selected')
+            api_key = self.config.get('api_key')
+            header = {'apikey': api_key}
+            return header
+        elif self.config.get('username') and self.config.get('password'):
+            log.debug('username and password auth selected')
+            access_token = self.get_access_token()
+            return {"Authorization": "Bearer %s" %access_token}
```
cvium: Missing space.

flexget/plugins/list/ombi_list.py
```diff
@@ -0,0 +1,281 @@
...
+            return access_token
+        except RequestException as e:
+            raise plugin.PluginError('Ombi username and password login failed: %s' % e)
+
+    def ombi_auth(self):
+        if self.config.get('api_key'):
+            log.debug('api_key auth selected')
+            api_key = self.config.get('api_key')
+            header = {'apikey': api_key}
+            return header
```
cvium: Might as well just return the dict instead of saving it in a variable first.

flexget/plugins/list/ombi_list.py
```diff
@@ -0,0 +1,281 @@
...
+            access_token = self.get_access_token()
+            return {"Authorization": "Bearer %s" %access_token}
+        else:
+            raise plugin.PluginError('Error: an api_key or username and password must be configured')
+
+    def get_request_list(self):
+        auth_header = self.ombi_auth()
+
+        parsedurl = urlparse(self.config.get('base_url'))
+        url=''
```
cvium: Unused. `url` will always be set or a plugin error is raised.

flexget/plugins/list/ombi_list.py
```diff
@@ -0,0 +1,281 @@
...
+        if self.config.get('type') in ['movies']:
+            log.debug('Received movie list request')
+            url = '%s://%s:%s%s/api/v1/Request/movie' % (parsedurl.scheme, parsedurl.netloc, self.config.get('port'), parsedurl.path)
+        elif self.config.get('type') in ['shows', 'seasons', 'episodes']:
+            log.debug('Received TV list request')
```
cvium: I don't think these log lines are necessary. If we need to debug it, we always ask for a user's config.

flexget/plugins/list/ombi_list.py
```diff
@@ -0,0 +1,281 @@
...
+            json = self.get_request_list()
+
+            self._items = []
+
+            for parent_request in json:
+                if self.config.get('type') == 'movies':
+                    # check that the request is approved unless user has selected to include everything
+                    if self.config.get('only_approved') and not parent_request.get('approved') or parent_request.get('approved'):
+                        # Always include items that are not available and only include available items if user has selected to do so
+                        if (self.config.get('include_available') and parent_request.get('available') or not parent_request.get('available')):
```
cvium: You have three nested `if`s here. It hurts readability. Consider putting them in variables if you don't want a multi-line `if` statement.

flexget/plugins/list/ombi_list.py
```diff
@@ -0,0 +1,281 @@
...
+                            entry = self.generate_movie_entry(parent_request)
+                            log.debug('Entry %s', entry)
+                            self._items.append(entry)
+                elif self.config.get('type') == 'shows':
+                    # Shows do not have approvals or available flags so include them all
+                    entry = self.generate_tv_entry(parent_request)
+                    log.debug('Valid entry %s', entry)
+                    self._items.append(entry)
+                else:
```
cvium: Consider splitting the contents of the outer-most if-elif-else into separate function

flexget/plugins/list/ombi_list.py
```diff
@@ -0,0 +1,281 @@
...
+            if entry.get('tmdb_id'):
+                result = self.add_movie(entry)
+                if result.get('message'):
+                    log.verbose('%s', result.get('message'))
+                elif result.get('errorMessage'):
+                    log.verbose('%s', result.get('errorMessage'))
+                else:
+                    log.verbose('Unknown result')
+            else:
+                log.verbose('Skipping %s ombi requires an tmdb_id to add movies, try using the tmdb_lookup plugin', entry.get('title'))
```
cvium: If it's a requirement, we should help the user out and use the tmdb_lookup plugin for them. 

```
from flexget.plugin import get_plugin_by_name

get_plugin_by_name('tmdb_lookup').instance.lookup(entry)
```

flexget/plugins/list/ombi_list.py
```diff
@@ -0,0 +1,281 @@
...
+    def generate_title(self, item, season=None, episode=None):
+        temptitle = item.get('title')
+
+        if item.get('releaseDate') and self.config.get('include_year'):
+            temptitle = temptitle + ' (' + str(item.get('releaseDate')[0:4]) +')'
+
+        if season or episode:
+            temptitle = temptitle + ' ' + self.generate_series_id(season, episode)
+            if episode:
+                if episode.get('title') and self.config.get('include_ep_title'):
```
cvium: Unnecessary nesting

bagobones:
Apparently the Ombi API for TV requests may be changing soon so I don't expect to commit anything more for a while.

I would LOVE to see the list functions available even without the add and remove functions since I think it is very useful.

liiight:
Without add or remove this is just an input plugin. Just remove the 2 lines that make the lost interface part, remove the word `list`from the plugin interface declarations and that should be it.

cvium:
I don't think there's any point in merging an incomplete plugin.

liiight:
Might still be useful as an input

bagobones:
Personally I only wanted it as an input from the beginning. I only went with a list since the API existed. 

If you want me to rename it and move it to the input folders I can. I just figured since I had add/remove movies working I should leave the code there for future completion. 

I’ll be honest here  I just do not have the time to keep plugging away at this at this time.

I think Ombi is a fairly unique input as it is self hosted and multi user.

liiight:
I got no issues with merging this as is

cvium:
The git history looks kinda fucked. I'm also a bit wary of merging incomplete stuff. It's just adding to the already huge technical debt.

bagobones:
Going to just purge this, and resubmit some other time, maybe only as an input.

The history is a mess mostly because I switched tools and ended up with some oddities syncing up my fork to upstream..

PR #2102, opened 2018-03-16
Deluge client switch

gazpachoking:
### Motivation for changes:
Stop requiring a deluge install to connect to deluge daemon. This allows easier configuration on flexget installs which connect to remote deluge daemons, as well as python 3 support in the deluge plugin.

### Detailed changes:
- Switch to https://github.com/JohnDoee/deluge-client as deluge client library.

#### To Do:

- [x] Test it more.
- [x] Implement get_localhost_auth ourselves?
- [x] Autodetect deluge version? https://github.com/JohnDoee/deluge-client/issues/11
- [x] Code can probably be cleaned up a bit more, mostly just a copy/paste from the old callback version still.


JohnDoee:
https://github.com/JohnDoee/deluge-client/issues/5 and other timeout issues might cause problems but will see if I can't get that working properly while adding auto-detection.

JohnDoee:
I made something like get_localhost_auth already, the resolution should be very similar to the existing Deluge.
https://github.com/JohnDoee/autotorrent/blob/develop/autotorrent/clients/deluge.py#L55-L96

movalex:
Tried this on MacOs 10.13 with no success.
First I've got this error:
    
```
2018-05-01 00:40 CRITICAL plugin                        Plugin `flexget.plugins.clients.deluge2` failed to import dependencies
  File "/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/site-packages/flexget/plugin.py", line 396, in _load_plugins_from_dirs
    __import__(module_name)
  File "/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/site-packages/flexget/plugins/clients/deluge2.py", line 15, in <module>
    from utils.pathscrub import pathscrub
ImportError: No module named utils.pathscrub
```

So I had to change import in `deluge2.py`:
`from utils.pathscrub import pathscrub` to `from flexget.utils.pathscrub import pathscrub` 
as well as `from utils.template import RenderError` to  `from flexget.utils.template import RenderError`

Then I've got this error. 
```
2018-05-01 00:44 DEBUG    deluge        tv1             Error importing deluge: dynamic module does not define init function (init_openssl)
2018-05-01 00:44 CRITICAL task          tv1             Plugin `deluge` cannot be used because dependency `deluge` is missing.
2018-05-01 00:44 DEBUG    task          tv1             Deluge >=1.2 module and it's dependencies required. ImportError: dynamic module does not define init function (init_openssl)
```
Any ideas?

movalex:
Sorry, I realized I had to use `deluge3` plugin in config. I've created a [pull request](https://github.com/Flexget/Flexget/pull/2135) with working solution. 
Initially I had an error with Deluge 1.3.15 installed : 
```
TypeError: __init__() got an unexpected keyword argument 'deluge_version'
```
So I had to omit `deluge_version` in `DelugeRPCClient` setup and everything worked like a charm.

cvium:
I will try to clean the code a bit when I get time

gazpachoking:
> Ideally, I'd like it if we made a wrapper around DelugeRPCClient to get rid of the direct procedure calls in the plugin(s)

@cvium Not sure what you mean by this?

I think probably the one thing we should do is implement a very simple version of the get_localhost_auth, so as to not break everybodies setup who is relying on this when we release it.

cvium:
@gazpachoking Just that deluge-client is very barebones and I don't think calling `client.call('core.get_torrents_status', {}, ['name'])` directly in the plugin is very nice. If the API changes, it's just easier to maintain if the Deluge plugins just do something like `client.get_torrents_status('name', 'path')`.

gazpachoking:
@cvium Not sure I think wrapping is worth it. What about if we just enable a nicer call format in the new client? I think this PR would essentially make the deluge-client calls the same as the native deluge client calls.

https://github.com/JohnDoee/deluge-client/pull/19

Superchicken:
Specifying a filter under from_deluge in config.yml does not filter the entries.

the `self.get_torrents_status` call on [line 220](https://github.com/Flexget/Flexget/blob/823f69bb8bdae5e0c8c3ffe1035240ec5901fafa/flexget/plugins/clients/deluge.py#L220) only sends one argument, it should send the filter as the second argument.

PR #2104, opened 2018-03-19
Refactor decompress plugin to be a bit cleaner

ianstalk:
### Motivation for changes:
Clean up logic in the decompress plugin to make it cleaner and easier to follow. I've also been mulling over possibly making `decompress` an entry list plugin in the future, this should make that a bit easier.

### Detailed changes:
- Added ArchiveInfo class to `utils/archive.py`, which now does most of the heavy listing for decompressing files
- Moved additional logic into standalone functions

### Addressed issues:
- Fixes # .

```
paste_config_here
```
```
flexget/tests/test_decompress.py .....                                                                                                                                                               [100%]

============================================================================================= warnings summary =============================================================================================
flexget/tests/test_decompress.py::TestExtract::()::test_rar
  /Users/ieaston/Flexget/flexget/tests/conftest.py:212: DeprecationWarning: use of getfuncargvalue is deprecated, use getfixturevalue
    dst = dst.replace('__tmp__', request.getfuncargvalue('tmpdir').strpath)
  /Users/ieaston/Flexget/flexget/tests/conftest.py:66: DeprecationWarning: use of getfuncargvalue is deprecated, use getfixturevalue
    config = config.replace('__tmp__', request.getfuncargvalue('tmpdir').strpath)

-- Docs: http://doc.pytest.org/en/latest/warnings.html
=================================================================================== 5 passed, 2 warnings in 2.54 seconds ===================================================================================
```

### DIFF
--- a/flexget/plugins/output/decompress.py
+++ b/flexget/plugins/output/decompress.py
@@ -53,7 +53,7 @@ def get_output_path(to, entry):
             return render_from_entry(to, entry)
         else:
             return os.path.dirname(entry.get('location'))
-    except RenderError as error:
+    except RenderError:
         raise PluginError('Could not render path: %s', to)
 
 
--- a/flexget/plugins/output/decompress.py
+++ b/flexget/plugins/output/decompress.py
@@ -28,13 +28,14 @@ def open_archive_entry(entry):
     archive = None
 
     try:
-        archive_path = entry['location']
-        if not archive_path:
-            raise ValueError('Entry does not appear to represent a local file.')
+        archive_path = entry.get('location')
 
-        archive = archiveutil.open_archive(archive_path)
-    except (KeyError, ValueError):
-        log.error('Entry does not appear to represent a local file.')
+        if not os.path.exists(archive_path):
+            log.error('File no longer exists: %s', entry['location'])
+        elif not archive_path:
+            log.error('Entry does not appear to represent a local file.')
+        else:
+            archive = archiveutil.open_archive(archive_path)
     except archiveutil.BadArchive as error:
         fail_entry_with_error(entry, 'Bad archive: %s (%s)' % (archive_path, error))
     except archiveutil.NeedFirstVolume:
@@ -47,10 +48,13 @@ def open_archive_entry(entry):
 
 def get_output_path(to, entry):
     """Determine which path to output to"""
-    if to:
-        return render_from_entry(to, entry)
-    else:
-        return os.path.dirname(entry.get('location'))
+    try:
+        if to:
+            return render_from_entry(to, entry)
+        else:
+            return os.path.dirname(entry.get('location'))
+    except RenderError as error:
+        raise PluginError('Could not render path: %s', to)
 
 
 def extract_info(info, archive, to, keep_dirs):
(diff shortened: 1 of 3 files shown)


flexget/plugins/output/decompress.py
```diff
@@ -25,53 +25,60 @@ def open_archive_entry(entry):
...
     """
-    arch = None
+    archive = None
 
     try:
         archive_path = entry['location']
-        arch = archive.open_archive(archive_path)
-    except KeyError:
+        if not archive_path:
+            raise ValueError()
```
liiight: Please add a message

flexget/plugins/output/decompress.py
```diff
@@ -25,53 +25,60 @@ def open_archive_entry(entry):
...
-    """
-    Tests whether the file descibed in info is a directory
-    """
+    destination = get_destination_path(info, to, keep_dirs)
+
+    log.debug('Attempting to extract: %s to %s', info.filename, destination)
+    try:
+        info.extract(archive, destination)
+    except archiveutil.FSError as error:
+        log.error('OS error while creating file: %s (%s)' % (destination, error))
```
liiight: Should this/these raise PluginError instead?

flexget/plugins/output/decompress.py
```diff
@@ -25,53 +25,60 @@ def open_archive_entry(entry):
...
-        return not base
 
+def get_destination_path(info, to, keep_dirs):
+    """Generate the destination path for a given file"""
 
-def makepath(path):
-    if not os.path.exists(path):
-        log.debug('Creating path: %s', path)
-        os.makedirs(path)
+    if keep_dirs:
```
liiight: style suggestion:
```python
path_suffix = info.path if keep_dirs else os.path.basename(info.path)
```

cvium: It's not. I find it more readable in this case though, but I don't care either way.

flexget/plugins/output/decompress.py
ianstalk: Is PluginError fatal for the plugin? There's a chance that a FSError is related to a single file within an archive and the rest will extract successfully.

cvium: PluginError will abort the task

flexget/plugins/output/decompress.py
ianstalk: That's two votes in favor of it, so I'll do it. :)

flexget/plugins/output/decompress.py
```diff
@@ -28,13 +28,14 @@ def open_archive_entry(entry):
...
     try:
-        archive_path = entry['location']
-        if not archive_path:
-            raise ValueError('Entry does not appear to represent a local file.')
+        archive_path = entry.get('location')
 
-        archive = archiveutil.open_archive(archive_path)
-    except (KeyError, ValueError):
-        log.error('Entry does not appear to represent a local file.')
+        if not os.path.exists(archive_path):
```
cvium: This will raise an exception if archive_path is None

ianstalk: Addressed with my latest commit

cvium: Wouldn't it be better to reverse the if-else?
```
if not archive_path:
elif not os.path.exists
```

ianstalk: Not a bad idea. Let me change it.

ianstalk:
@liiight Fixed.

PR #2118, opened 2018-04-10
Fix: imdb_list does not support 'list: watchlist'

jeroenl:
The list exported when using 'list: watchlist' omits a number of columns compared to the regular export. These changes make these columns optional, keeping the fields empty if they were not provided.

### DIFF
--- a/flexget/plugins/list/imdb_list.py
+++ b/flexget/plugins/list/imdb_list.py
@@ -225,8 +225,8 @@ class ImdbEntrySet(MutableSet):
                     item_type = row['title type'].lower()
                     name = row['title']
                     year = int(row['year']) if row['year'] != '????' else None
-                    created = datetime.strptime(row['created'], '%Y-%m-%d') if row.get('created', '') else None
-                    modified = datetime.strptime(row['modified'], '%Y-%m-%d') if row.get('modified', '') else None
+                    created = datetime.strptime(row['created'], '%Y-%m-%d') if row.get('created') else None
+                    modified = datetime.strptime(row['modified'], '%Y-%m-%d') if row.get('modified') else None
                     entry = Entry({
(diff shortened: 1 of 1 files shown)


flexget/plugins/list/imdb_list.py
```diff
@@ -225,17 +225,17 @@ def items(self):
...
                         'title': '%s (%s)' % (name, year) if year != '????' else name,
                         'url': row['url'],
                         'imdb_id': row['const'],
                         'imdb_url': row['url'],
-                        'imdb_list_position': int(row['position']),
+                        'imdb_list_position': int(row['position']) if 'position' in row else None,
                         'imdb_list_created': created,
                         'imdb_list_modified': modified,
-                        'imdb_list_description': row['description'],
+                        'imdb_list_description': row.get('description', None),
```
cvium: None is already the default value

flexget/plugins/list/imdb_list.py
```diff
@@ -225,17 +225,17 @@ def items(self):
                     item_type = row['title type'].lower()
                     name = row['title']
                     year = int(row['year']) if row['year'] != '????' else None
-                    created = datetime.strptime(row['created'], '%Y-%m-%d') if row['created'] else None
-                    modified = datetime.strptime(row['modified'], '%Y-%m-%d') if row['modified'] else None
+                    created = datetime.strptime(row['created'], '%Y-%m-%d') if row.get('created', '') else None
```
cvium: No need for a default value

flexget/plugins/list/imdb_list.py
```diff
@@ -225,17 +225,17 @@ def items(self):
...
+                    modified = datetime.strptime(row['modified'], '%Y-%m-%d') if row.get('modified', '') else None
```
cvium: See above

PR #2119, opened 2018-04-10
Series bug SQLAlchemy

### LINKED ISSUE #2058: Unhandled error in plugin series: expected string or bytes-like object
### Expected behaviour:
Plugin series should properly parse qualities.

### Actual behaviour:
Crash at each run

### Steps to reproduce:
- Step 1: ...

#### Config:
https://pastebin.cedricgirard.fr/view/15f3af31
  
#### Log:
https://pastebin.cedricgirard.fr/view/f8c4b3c9

### Additional information:

- FlexGet version: 2.11.24
- Python version: 3.6.4
- Installation method: Distrib package built from git
- Using daemon (yes/no): yes
- OS and version: ArchLinux (Rolling, Up to Date)

liiight:
### Motivation for changes:
SQLAlchemy update to 1.2.0 breaks some stuff for us

### Detailed changes:
- Querying quality column when it has no value now returns `InstrumentedProperty` object which causes `Quality` to crash. Changed to check if returned value is actually a string before trying to return `Quality` via the `quality_property` util.

### Addressed issues:
- Fixes #2058


gazpachoking:
Replaced by #2120

PR #2120, opened 2018-04-11
[fix] Make sqlalchemy 1.1 and 1.2 behave the same. fix #2058

### LINKED ISSUE #2058: Unhandled error in plugin series: expected string or bytes-like object

gazpachoking:
### Motivation for changes:
SQLAlchemy update to 1.2.0 breaks some stuff for us

### Detailed changes:
[This](http://docs.sqlalchemy.org/en/latest/changelog/migration_12.html#hybrid-attributes-support-reuse-among-subclasses-redefinition-of-getter) change in sqlalchemy 1.2 broke our custom comparators for hybrid properties we were using. Since calling 'comparator' method on hybrid properties no longer mutates the property, we have to re-assign the result back to the property name.

Replaces #2119, fixing the issue at the root.

### Addressed issues:
- Fixes #2058

### DIFF
--- a/flexget/plugins/filter/series.py
+++ b/flexget/plugins/filter/series.py
@@ -256,22 +256,22 @@ class AlternateNames(Base):
     _alt_name_normalized = Column('alt_name_normalized', Unicode, index=True, unique=True)
     series_id = Column(Integer, ForeignKey('series.id'), nullable=False)
 
-    def name_setter(self, value):
+    @hybrid_property
+    def alt_name(self):
+        return self._alt_name
+
+    @alt_name.setter
+    def alt_name(self, value):
         self._alt_name = value
         self._alt_name_normalized = normalize_series_name(value)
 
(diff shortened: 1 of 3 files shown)


gazpachoking:
Codacy, I think you on crack. Nothing is shadowed here, we define a (settable) property, then set that property in `__init__`.

PR #2134, opened 2018-05-08
add base_path support to sonarr_list plugin

deansheather:
If Sonarr is behind a reverse proxy with a custom `URL Base` setting, the `sonarr_list` plugin wouldn't work because it doesn't support a custom base path. This should fix that.

### Motivation for changes:
- Can't use `sonarr_list` with my Sonarr setup (custom base path with reverse proxy)

### Detailed changes:
- Added `base_path` config schema option to `sonarr_list` with default value of `''` (should be backwards compatible)

```yml
tasks:
  task-name:
    sonarr_list:
      base_url: https://server.example.com
      port: 8443
      base_path: /sonarr
      api_key: ...
```

This example config will use `https://server.example.com:8443/sonarr/api/` as the base API path.

### DIFF
--- a/flexget/plugins/list/sonarr_list.py
+++ b/flexget/plugins/list/sonarr_list.py
@@ -32,6 +32,7 @@ class SonarrSet(MutableSet):
         'type': 'object',
         'properties': {
             'base_url': {'type': 'string', 'default': 'http://localhost'},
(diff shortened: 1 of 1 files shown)


liiight:
thanks!

PR #2136, opened 2018-05-14
Recognise "p60"

lingfish:
Resolutions can contain p60, like p50.

### Motivation for changes:
Qualities aren't matching.

### Detailed changes:
- Recognise p60 rates


cvium:
Please add another 'test case' similar to this https://github.com/Flexget/Flexget/blob/develop/flexget/tests/test_qualities.py#L72

lingfish:
Done :)

cvium:
Thank you.

PR #2140, opened 2018-05-19
update get_localhost_auth

movalex:
since auth level is not used, this should be enough...

### Detailed changes:
- 

### Addressed issues:
- Fixes # .

```
paste_config_here
```
```
paste output here
```
#### To Do:

- [ ] Stuff..


gazpachoking:
Yep, looks good to me, thanks.

PR #2142, opened 2018-05-20
Add metacritic score lookup functionality

raJorgeLuisG:
Added filtering based on Metacritic score parsed from IMDB page.

### Motivation for changes:
I want to be able to filter movies based on metacritic score as well as IMDB score

### Detailed changes:
- 

### Addressed issues:
- Fixes # .

```
paste_config_here
```
```
paste output here
```
#### To Do:

- [ ] Stuff..

### DIFF
--- a/flexget/plugins/metainfo/imdb_lookup.py
+++ b/flexget/plugins/metainfo/imdb_lookup.py
@@ -63,6 +63,7 @@ class Movie(Base):
 
     score = Column(Float)
     votes = Column(Integer)
+    meta_score = Column(Integer)
     year = Column(Integer)
     plot_outline = Column(Unicode)
     mpaa_rating = Column(String, default='')
@@ -222,6 +223,7 @@ class ImdbLookup(object):
         'imdb_plot_outline': 'plot_outline',
         'imdb_score': 'score',
         'imdb_votes': 'votes',
+        'imdb_meta_score': 'meta_score'
         'imdb_year': 'year',
         'imdb_genres': lambda movie: [genre.name for genre in movie.genres],
         'imdb_languages': lambda movie: [lang.language.name for lang in movie.languages],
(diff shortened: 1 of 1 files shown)


raJorgeLuisG:
Sorry,  new to this whole thing. The error the circleci tests are given are due to flexget/utils/imdb.py needing the changes that I made in pull request #2143. How do I get the tester to test against that version of imdbParser?

raJorgeLuisG:
I figured out to just put all of my changes into this branch. It looks like all tests pass now. Will remove other two pull requests

cvium:
Since you have made changes to a database table, you need to update the upgrade function here https://github.com/raJorgeLuisG/Flexget/blob/05c96a308758b5887ae0bcfd55e60836079716cd/flexget/plugins/metainfo/imdb_lookup.py#L195-L203

raJorgeLuisG:
Thanks, I made the updates per your suggestion. Let me know if you see anything else that needs to be addressed.

flexget/utils/imdb.py
```diff
@@ -321,7 +322,13 @@ def parse(self, imdb_id):
         else:
             log.debug('No score found for %s' % self.imdb_id)
 
-        # get director(s)
+        meta_score_elem = title_overview.find(attrs={'class': 'metacriticScore'})
+        if meta_score_elem:
+            self.meta_score = str_to_int(meta_score_elem.text)
+        else:
+            log.debug('No Metacritic score found for %s' % self.imdb_id)
```
cvium: This line should be changed to `log.debug('No Metacritic score found for %s', self.imdb_id)`

raJorgeLuisG: Does that mean that all of the other element debug statements need to be corrected as well? (e.g. score_elem, votes_elem, original_name_elem, photo_elem, mpaa_rating_elem, etc.) They are all formatted as I have it for meta_score_elem.  If this is the case I can fix those as well.

cvium: Yes, it's best to let the logger handle the string interpolation.

raJorgeLuisG:
Updated debug statements from 
log.debug('No storyline found for %s' % self.imdb_id)
to
log.debug('No storyline found for %s', self.imdb_id)
unless it looks like it is doing something more complicated. Lines omitted are 99, 140, 208, 232, 284. 
I'm not very familiar with what formatting debug can do, nor with exactly what % does, so left the more complex statements as they are.

PR #2143, opened 2018-05-20
Add metacritic score parsing

raJorgeLuisG:
Add metacritic score parsing from IMDB movie pages. 

Note: I was not able to test this. I think this should work, but the HTML class is defined as follows in IMDB:
<div class="metacriticScore score_favorable titleReviewBarSubItem">

I'm not sure if the class statement on line 325 needs to include all three classes to make the match.

### Detailed changes:
- 

### Addressed issues:
- Fixes # .

```
paste_config_here
```
```
paste output here
```
#### To Do:

- [ ] Stuff..

### DIFF
--- a/flexget/utils/imdb.py
+++ b/flexget/utils/imdb.py
@@ -251,6 +251,7 @@ class ImdbParser(object):
         self.writers = {}
         self.score = 0.0
         self.votes = 0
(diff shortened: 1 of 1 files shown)


raJorgeLuisG:
Combined into #2142 Closing.

PR #2147, opened 2018-05-29
TorrentLeech - Updated Categories from website

blastcodem:
Added new naming for new categories
Added to option to new categories name (Kept old one for backward compatiblity)

I can update wiki when pulled


cvium:
> Added to option to new categories name (Kept old one for backward compatiblity)

Are they still using the old categories?

flexget/plugins/sites/torrentleech.py
```diff
@@ -23,12 +23,23 @@
     # Movies
     'Cam': 8,
     'TS': 9,
+    'TS/TC': 9,
+#    R5 is deprecated
     'R5': 10,
     'DVDRip': 11,
+	'DVDRip/DVDScreener': 11,
```
cvium: What's with all the random indentations?

blastcodem: Copy paste, Error, will fix

flexget/plugins/sites/torrentleech.py
```diff
@@ -48,9 +59,10 @@ class UrlRewriteTorrentleech(object):
           password: xxxxxxxx  (required)
           category: HD
 
-          Category is any combination of: all, Cam, TS, R5,
-          DVDRip, DVDR, HD, BDRip, Movie Boxsets, Documentaries,
-          Episodes, TV BoxSets, Episodes HD
+          Category is any combination of: all, Cam, TS, TS/TC, DVDRip,
+				DVDRip/DVDScreener, WEBRip': 37, HDRip': 43, BDRip, DVDR,
```
cvium: This doesn't look right either

cvium:
I see no reason to keep any categories they no longer support

PR #2157, opened 2018-06-20
change irc daemon task_re configuration structure to enable AND conditions

aidan-:
### Motivation for changes:
The current `task_re` functionality in the `irc_daemon` plugin allows users to define one or more regex that when matched will cause that task to execute.  When more than one regex is defined for a task, the task is executed when ANY of them are a match.  If you want to only execute a task when regex1 matches AND regex2 matches, this cannot currently be done.

This pull request restructures the format of the `task_re` config so both AND conditions and OR conditions can be achieved.

### Detailed changes:
Change format of `task_re` config to allow a task multiple times with different regexes (OR condition) and multiple regexes to be defined for a single task (AND condition).

In this example, the tasks will only execute when both of their respective patterns are a match:
```
task_re:
  - task: get_movie_entry
    patterns:
      - regexp: (MovieHD)
        field: irc_category
      - regexp: 'true'
        field: irc_scene
  - task: get_tv_entry
    patterns:
      - regexp: (TvHD)
        field: irc_category
      - regexp: 'true'
        field: irc_scene
```
In this example, the `get_tv_entry` will execute if either regex matches (or both):
```
task_re:
  - task: get_tv_entry
    patterns:
      - regexp: (TvSD)
        field: irc_category
  - task: get_tv_entry
    patterns:
      - regexp: (TvHD)
        field: irc_category
```

#### To Do:

- [ ] Update documentation

### DIFF
--- a/flexget/plugins/daemon/irc.py
+++ b/flexget/plugins/daemon/irc.py
@@ -55,23 +55,32 @@ schema = {
                     'task': one_or_more({
                         'type': 'string'
                     }),
-                    'task_re': one_or_more({
-                        'type': 'object',
-                        'properties': {
-                            'task': {
-                                'type': 'string'
-                            },
-                            'patterns': one_or_more({
-                                'type': 'object',
-                                'properties': {
-                                    'regexp': {'type': 'string'},
-                                    'field': {'type': 'string'}
+                    'task_re': {
+                        'type': 'array',
+                        'items': {
+                            'type': 'object',
+                            'properties': {
+                                'task': {
+                                    'type': 'string'
                                 },
-                                'required': ['regexp', 'field'],
-                                'additionalProperties': False
-                            })
-                        },
-                    }, unique_items=True),
+                                'patterns': {
+                                    'type': 'array',
+                                    'items': {
+                                        'type': 'object',
+                                        'properties': {
+                                            'regexp': {'type': 'string'},
+                                            'field': {'type': 'string'}
+                                        },
+                                        'required': ['regexp', 'field'],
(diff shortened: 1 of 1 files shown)


aidan-:
I've had this code running for a week or so now, and it has been working as intended for that period.  Before I update the documentation to match, can I get some feedback on the actual code changes?  Whats the likely hood of this being accepted?

Thanks!

flexget/plugins/daemon/irc.py
```diff
@@ -375,15 +380,19 @@ def run_tasks(self):
...
-                        config = [config]
-                    for c in config:
-                        if re.search(c['regexp'], entry.get(c['field'], ''), re.IGNORECASE):
-                            matched = True
-                            if not tasks_entry_map.get(task):
-                                tasks_entry_map[task] = []
-                            tasks_entry_map[task].append(entry)
+                for task_config in tasks_re:
+                    pattern_match = 0
+                    for pattern in task_config.get('patterns'):
```
cvium: You've specified `one_or_more` in the schema, so this will fail if it's not a list.

cvium:
@aidan- is it done then?

aidan-:
Yep! I am happy with this.  How/when should I update the documentation?  Should there be a notice about the change in configuration structure?

cvium:
There will be a minor version bump to 2.14 and a notice (which you can write) on https://flexget.com/UpgradeActions as well as an update to its wiki page. You can do it now.

aidan-:
Both pages on the wiki have been updated to reflect the changes!

PR #2162, opened 2018-07-01
3.7 regex fix

tobinjt:
### Motivation for changes:

Fix breakage with Python 3.7.

### Detailed changes:

https://docs.python.org/3.7/library/re.html#re.sub
In Python 3.6 onwards, the replacement text can't contain \w or any similar regex sequences.  Use re.split and join instead.

### Addressed issues:
Without this change this crash occurs:

```
  File "/usr/local/lib/python3.7/site-packages/flexget/task.py", line 486, in __run_plugin
    return method(*args, **kwargs)
  File "/usr/local/lib/python3.7/site-packages/flexget/event.py", line 23, in __call__
  File "/usr/local/lib/python3.7/site-packages/flexget/plugins/filter/all_series.py", line 48, in on_task_metainfo
    if guess_entry(entry, config=group_settings):
  File "/usr/local/lib/python3.7/site-packages/flexget/plugins/metainfo/series.py", line 47, in guess_entry
    allow_seasonless=allow_seasonless)
  File "/usr/local/lib/python3.7/site-packages/flexget/plugins/parsers/plugin_parsing.py", line 74, in parse_series
    return parser.parse_series(data, name=name, **kwargs)
  File "/usr/local/lib/python3.7/site-packages/flexget/plugins/parsers/parser_internal.py", line 47, in parse_series
    parser.parse(data)
  File "/usr/local/lib/python3.7/site-packages/flexget/utils/titles/series.py", line 225, in parse
    name_to_re(name, self.ignore_prefixes, self) for name in [self.name] + self.alternate_names)
  File "/usr/local/lib/python3.7/site-packages/flexget/utils/tools.py", line 205, in __init__
    list.__init__(self, *args, **kwargs)
  File "/usr/local/lib/python3.7/site-packages/flexget/utils/titles/series.py", line 225, in <genexpr>
    name_to_re(name, self.ignore_prefixes, self) for name in [self.name] + self.alternate_names)
  File "/usr/local/lib/python3.7/site-packages/flexget/plugins/parsers/parser_common.py", line 85, in name_to_re
    res = re.sub(' +', blank + '*', res, re.UNICODE)
  File "/usr/local/Cellar/python/3.7.0/Frameworks/Python.framework/Versions/3.7/lib/python3.7/re.py", line 192, in sub
    return _compile(pattern, flags).sub(repl, string, count)
  File "/usr/local/Cellar/python/3.7.0/Frameworks/Python.framework/Versions/3.7/lib/python3.7/re.py", line 309, in _subx
    template = _compile_repl(template, pattern)
  File "/usr/local/Cellar/python/3.7.0/Frameworks/Python.framework/Versions/3.7/lib/python3.7/re.py", line 300, in _compile_repl
    return sre_parse.parse_template(repl, pattern)
  File "/usr/local/Cellar/python/3.7.0/Frameworks/Python.framework/Versions/3.7/lib/python3.7/sre_parse.py", line 1024, in parse_template
    raise s.error('bad escape %s' % this, len(this))
re.error: bad escape \w at position 5
```

#### To Do:

3 tests still fail with similar errors, but tracking them down is hard because an exception is raised when handling the exception :(


serhiy-storchaka:
An alternate fix is duplicating backslashes in the replacement string:
```python
res = re.sub(' +', blank.replace('\\', r'\\') + '*', res, flags=re.UNICODE)
```

Note also that the third positional argument of re.sub() is a maximal count of replacements, not flags. Passing re.UNICODE as a third positional argument you just order re.sub() to make not more than 128 replacements. There is the same bug in other places.

tobinjt:
I can do that instead - would that be preferred?

Thanks,

paranoidi:
@tobinjt 
Hmm, I think duplicate \\'s are preferred, it's proper escaping after all ..

JohnDoee:
This is currently blocking 3.7 support and the only issue left discussed seems to be how to style the line.

The proposed fix works as far as I can see and the `flags` issue highlighted by @serhiy-storchaka should be moved to a different issue.

tobinjt:
I've changed it to use replace() instead of split() and join().
There are still a small number of test failures but this resolves the vast majority.

tobinjt:
Added commits fixing another 2 failing tests, 1 failing test remains.

flexget/tests/test_rtorrent.py
```diff
@@ -67,7 +67,9 @@ def test_load(self, mocked_proxy):
 
         fields = [p for p in called_args[2:]]
         assert len(fields) == 3
-        assert 'd.directory.set=\\/data\\/downloads' in fields
+        assert ('d.directory.set=\\/data\\/downloads' in fields
+                # Python 3.7+.
+                or 'd.directory.set=/data/downloads' in fields)
```
JohnDoee: Isn't that additional slash intentional ?

tobinjt: In Python <= 3.6 more characters were escaped by re.escape(), including '/', but in Python >= 3.7 re.escape() only escapes regex metacharacters like '$' and '.'.  To make the test pass with Python >= 3.7 we need to look for `'d.directory.set=/data/downloads'`, with earlier versions we need to look for `'d.directory.set=\\/data\\/downloads'`.

I've pushed a new commit that expands the comment, let me know if it's clear enough or it needs more details.

flexget/tests/test_rtorrent.py
```diff
@@ -67,7 +67,10 @@ def test_load(self, mocked_proxy):
 
         fields = [p for p in called_args[2:]]
         assert len(fields) == 3
-        assert 'd.directory.set=\\/data\\/downloads' in fields
+        assert ('d.directory.set=\\/data\\/downloads' in fields
```
serhiy-storchaka: You could use implementation independent test:
```python
assert ('d.directory.set=' + re.escape('/data/downloads')) in fields
```

tobinjt: Good idea, done.

tobinjt:
Hmm, I don't think those failing tests are my fault:
```
Build-agent version 0.1.437-d180d4d7 (2018-08-22T16:07:19+0000)
Starting container flexget/cci-python:2.7
  image cache not found on this host, downloading flexget/cci-python:2.7

  Error pulling image flexget/cci-python:2.7: Error response from daemon: received unexpected HTTP status: 503 Service Unavailable... retrying
  image cache not found on this host, downloading flexget/cci-python:2.7

  Error pulling image flexget/cci-python:2.7: Error response from daemon: received unexpected HTTP status: 503 Service Unavailable... retrying
  image cache not found on this host, downloading flexget/cci-python:2.7
```

How can I trigger those tests again?

JohnDoee:
@tobinjt The rtorrent test case isn't doing anything regexp related, it is testing the exact data sent to the rtorrent API. That's why I find it all a bit weird and question the slash change.

cvium:
What's the status on this?

JohnDoee:
@cvium To me it looks like it can be merged as is.

The rtorrent regex escape stuff is just plain strange and completely unrelated to how rtorrent API works. Might be smart to fix correctly in a different pull request if the tests doesn't pass (although I think they were fixed).

tobinjt:
@JohnDoee @cvium As far as I know it can be merged - there are no outstanding comments and it fixes a huge number of test failures under Python 3.7.
There are other problems with Python 3.7 but they will need to be fixed separately.

Thanks,

cvium:
@tobinjt @JohnDoee I had to revert these changes. I should've looked at it more carefully. The changes in this PR are definitely _not_ the correct way to handle this.

> Changed in version 3.7: Unknown escapes in repl consisting of '\' and an ASCII letter now are errors.

The keyword here is *unknown escapes*. There are quite a few *known escapes* that are affected by simply replacing `\` with `\\` in the _repl_ string.

asm0dey:
@cvium should I report bug here or IRC is enough?

cvium:
There is no bug (anymore).

serhiy-storchaka:
As the author of this change in Python 3.7 I found this PR correct. Have I missed something?

cvium:
Well, it broke something that worked previously. This is an excerpt from IRC:

```
<removed> Hey guys
<removed> Suddenly my replace patterns are broken
<removed> they look like this: replace:
<removed> regexp: 'e(\d{2}-)?(\d{2})'
<removed> format: 'e\2'
<removed> and suddenly I see \2 in names!
```

I also added a unit test that verifies that that capture groups could be referenced with `\1` prior to the changes, but not after.

serhiy-storchaka:
Ah, if `replace_config['format']` can contain group references, it shouldn't be escaped.

asm0dey:
@serhiy-storchaka oh, I see, it was notice for yourself, not for me.

tobinjt:
Sorry for the breakage :(

@serhiy-storchaka Is there a function in Python 3.7 that escapes the necessary characters but doesn't touch others?  I wasn't able to find one.

@cvium It would be great if there were tests for that functionality.  Are there other places where this is likely to break?

PR #2170, opened 2018-07-13
NPO Watchlist website update

mfonville:
NPO updated their website, to make the plugin work again we need this change.
Also fixed an error that could occur if `_get_series_info` failed to fetch info


flexget/plugins/input/npo_watchlist.py
```diff
@@ -181,9 +181,10 @@ def _get_series_info(self, task, config, mediaId):
                            'npo_description': series.find('div', id='metaContent').find('p').text,
                            'npo_language': 'nl'}  # hard-code the language as if in NL, for lookup plugins
             log.debug('Parsed series info for: %s (%s)', series_info['npo_name'], mediaId)
+            return series_info
         except RequestException as e:
             log.error('Request error: %s' % str(e))
-        return series_info
+        return  # if it failed, return empty
```
cvium: You could just initialize series_info to None at the top of the function

mfonville: yes, that was also an option I considered, I will change it

Issue #2071, opened 2018-01-31
2.11.22 broke some series parsing


Jonybat:
### Expected behaviour:
In the task fill-series-list from my config, it used to fetch series from a trakt list, lookup for the tvdb name for those series, and add the names to a list

### Actual behaviour:
After 2.11.22, this doesnt work anymore, because the parsing fails for some reason, and is unable to resolve the tvdb_series_name. Either that or some execution order change...

### Steps to reproduce:
- Just run the config below or see the logs

#### Config:
```
tasks:
  fill-series-list:
    metainfo_series: yes
    thetvdb_lookup: yes
    list_clear:
      what:
        - entry_list: trakt-series
    trakt_list:
      username: "jonybat"
      list: "test"
      type: shows
    accept_all: yes
    set:
      title: "{{tvdb_series_name}}"
    list_add:
      - entry_list: trakt-series
```
  
#### Log:
2.11.21:
```
2018-01-31 21:09 VERBOSE  details       fill-series-list Produced 1 entries.
2018-01-31 21:09 DEBUG    api_tvdb      fill-series-list Looking up tvdb information for 'Westworld (2016)'. TVDB ID: 296762
2018-01-31 21:09 DEBUG    api_tvdb      fill-series-list Series <name=Westworld (2016),tvdb_id=296762> information restored from cache.
2018-01-31 21:09 VERBOSE  task          fill-series-list ACCEPTED: `Westworld` by accept_all plugin
2018-01-31 21:09 DEBUG    urlrewriter   fill-series-list Checking 1 entries
2018-01-31 21:09 VERBOSE  details       fill-series-list Summary - Accepted: 1 (Rejected: 0 Undecided: 0 Failed: 0)
2018-01-31 21:09 VERBOSE  list_add      fill-series-list adding accepted entries into entry_list - trakt-series
2018-01-31 21:09 DEBUG    api_tvdb      fill-series-list Looking up tvdb information for 'Westworld (2016)'. TVDB ID: 296762
2018-01-31 21:09 DEBUG    api_tvdb      fill-series-list Series <name=Westworld (2016),tvdb_id=296762> information restored from cache.
2018-01-31 21:09 DEBUG    api_tvdb      fill-series-list Looking up tvdb information for 'Westworld (2016)'. TVDB ID: 296762
2018-01-31 21:09 DEBUG    api_tvdb      fill-series-list Series <name=Westworld (2016),tvdb_id=296762> information restored from cache.
[...]

PR #2172, opened 2018-07-13
metainfo_series: don't use title if series_name is set

cvium:
... probably not a  very good solution

Fixes https://github.com/Flexget/Flexget/issues/2071

### DIFF
--- a/flexget/plugins/metainfo/series.py
+++ b/flexget/plugins/metainfo/series.py
@@ -43,7 +43,14 @@ class MetainfoSeries(object):
         identified_by = 'auto'
         if config and 'identified_by' in config:
             identified_by = config['identified_by']
-        title = entry.get('series_name') or entry['title']
+        if entry.get('series_name'):
+            title = entry['series_name']
+            if entry.get('series_year'):
+                title += ' (%s)'.format(entry['series_year'])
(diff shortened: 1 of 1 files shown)


flexget/plugins/metainfo/series.py
```diff
@@ -43,7 +43,15 @@ def guess_entry(self, entry, allow_seasonless=False, config=None):
         identified_by = 'auto'
         if config and 'identified_by' in config:
             identified_by = config['identified_by']
-        parsed = get_plugin_by_name('parsing').instance.parse_series(data=entry['title'], identified_by=identified_by,
+        if entry.get('series_name'):
+            title = entry['series_name']
+            if entry.get('series_year'):
+                title += ' ({})'.format(entry['series_year'])
+            if entry.get('series_id'):
+                title += ' {}'.format(entry['series_id'])
```
liiight: what's the use case for this? i don't remember what's series_id...

cvium: I think it's the value of the ep identifier

cvium:
@liiight @stevezau Where are we on this? Should we just push this?

liiight:
I guess

PR #2174, opened 2018-07-15
[rtorrent] Adding a torrent can sometimes take a while, so verification fails

cvium:
### Motivation for changes:
Loading torrents into rtorrent can take a while for me, so when adding multiple torrents it cannot verify the subsequent additions, but it's _never_ happened for me that it wasn't added.
### Detailed changes:
- Changed entry.fail to log.warning


cvium:
If there are no comments I will merge this later today

PR #2190, opened 2018-08-14
Refactor newznab plugin

rbraunschweig:
### Motivation for changes:
The newznab plugin was having problems with season packs, so I cleaned it up a bit... and fixed a bug.

### Detailed changes:
- Schema only supports implemented searches
- Use `tvdb_id` instead of `tvrage_id`
- `series_season` and `series_episode` are now optional

### Addressed issues:
- Skipped search if `season_episode` was not specified

#### To Do:
- [ ] Implement `book` and `music` search categories
- [ ] Edit wiki

### DIFF
--- a/flexget/plugins/sites/newznab.py
+++ b/flexget/plugins/sites/newznab.py
@@ -60,7 +60,6 @@ class Newznab(object):
     def fill_entries_for_url(self, url, task):
         entries = []
         try:
-            log.verbose('Fetching %s' % url)
             response = task.requests.get(url)
         except RequestException as e:
             log.error("Failed fetching '%s': %s" % (url, e))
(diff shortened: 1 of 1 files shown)


flexget/plugins/sites/newznab.py
```diff
@@ -46,41 +46,35 @@ class Newznab(object):
...
-
         return config
 
     def fill_entries_for_url(self, url, task):
         entries = []
-        log.verbose('Fetching %s' % url)
-
         try:
-            r = task.requests.get(url)
+            log.verbose('Fetching %s' % url)
```
cvium: `task.requests.get` should already log the url and headers as DEBUG

PR #2193, opened 2018-08-19
Python3.7 support

JohnDoee:
### Motivation for changes:

Add Python 3.7 support, some of the stuff not covered in #2162

### Detailed changes:
- Updated pyyaml & rpyc version
- Added CircleCi support
- Added 3.7 to setup.py

#### To Do:

- [x] Push 3.7 image to docker hub
- [ ] Finish up #2162
- [ ] Make tests pass (e.g. ensure rpyc works) - This is a bit hard to see as long as #2162 is still in works.
- [ ] Add tests that covers the rpyc code(?)


gazpachoking:
Hmm, all tests seem to be passing on this PR, should we merge it as-is, and open another ticket to track further needs?

JohnDoee:
@gazpachoking Sadly, this one basically just tracks all the other needed items for py3.7 and tests last ran when the #2162 was merged.

The regex building logic is a bit of a mess and might require more than meets the eye to get working correctly.

gazpachoking:
Gotcha. Maybe I'll try to tackle the regex stuff then. I'm thinking we'll just need to fix our usage of re.sub, and require users to escape their own regex substitutions if they are specifying a replacement in the config.

gazpachoking:
This was replaced by #2225, which has now been merged.

PR #2195, opened 2018-08-22
Upgrade rpyc and pip-tools.

tobinjt:
### Motivation for changes:

Fix Python 3.7 breakage.
Make updating dependencies work.

### Detailed changes:
Upgrade rpyc and pip-tools.

- rpyc < 4.0 is broken with Python 3.7 because 'async' is now a keyword: https://github.com/tomerfiliba/rpyc/issues/253
- pip-tools < 2.0.1 is incompatible with pip >= 10 so pip-compile fails: https://github.com/jazzband/pip-tools/issues/648

### Addressed issues:
- None.

- None.

None.

Cannot test properly until https://github.com/Flexget/Flexget/pull/2162 is merged because that's also a fix for Python 3.7.

#### To Do:

None.


JohnDoee:
I already got #2193 going, we should probably merge these two.

liiight:
> I already got #2193 going, we should probably merge these two.

I agree

Issue #2160, opened 2018-06-27
support of guessit 3.0 and python-dateutil 2.7.3


snickers2k:
### Expected behaviour:
flexget 2.14.x to support guessit 3.0

### Actual behaviour:
```
flexget 2.14.1 has requirement guessit==2.1.4, but you'll have guessit 3.0.0 which is incompatible.
flexget 2.14.1 has requirement python-dateutil==2.6.1, but you'll have python-dateutil 2.7.3 which is incompatible.
```

### Steps to reproduce:
- Install newest flexget with guessit 3.0  and python-dateutil 2.7.3

### Logs:
```
2018-06-27 21:53 DEBUG    manager                       Figuring out config load paths
2018-06-27 21:53 DEBUG    manager                       Found config: /home/snikay/.flexget/config.yml
2018-06-27 21:53 DEBUG    manager                       Config file /home/snikay/.flexget/config.yml selected
2018-06-27 21:53 DEBUG    manager                       sys.defaultencoding: ascii
2018-06-27 21:53 DEBUG    manager                       sys.getfilesystemencoding: UTF-8
2018-06-27 21:53 DEBUG    manager                       flexget detected io encoding: utf-8
2018-06-27 21:53 DEBUG    manager                       os.path.supports_unicode_filenames: False
2018-06-27 21:53 DEBUG    plugin                        Trying to load plugins from: [u'/home/snikay/.flexget/plugins', '/usr/local/lib/python2.7/dist-packages/flexget/plugins']
2018-06-27 21:53 DEBUG    plugin                        (u'Plugin `%s` requires `%s` to load.', u'memusage', u'ext lib `guppy`')
2018-06-27 21:53 CRITICAL plugin                        Exception while loading plugin flexget.plugins.parsers.parser_guessit
Traceback (most recent call last):
  File "/usr/local/lib/python2.7/dist-packages/flexget/plugin.py", line 396, in _load_plugins_from_dirs
    __import__(module_name)
  File "/usr/local/lib/python2.7/dist-packages/flexget/plugins/parsers/parser_guessit.py", line 38, in <module>
    guessit_api = GuessItApi(rebulk_builder().rebulk(_id_regexps))
TypeError: rebulk_builder() takes exactly 1 argument (0 given)
2018-06-27 21:53 DEBUG    manager                       Removed /home/snikay/.flexget/.config-lock
Traceback (most recent call last):
  File "/usr/local/bin/flexget", line 11, in <module>
    sys.exit(main())
[...]

PR #2197, opened 2018-08-28
Guessit 3.0

sarendsen:
### Motivation for changes:
Upgrade to guessit 3.0 so we can take advantage of new features and bugfixes.

### Detailed changes:
- Rewrite initiation of GuessItApi for 3.0
- Added mapping for new guessit namings
- Bump requirements versions

- https://github.com/Flexget/Flexget/issues/2160

### DIFF
--- a/flexget/plugins/parsers/parser_guessit.py
+++ b/flexget/plugins/parsers/parser_guessit.py
@@ -42,16 +42,20 @@ def rules_builder(config):
 
 
 guessit_api = GuessItApi()
-guessit_api.configure({}, rules_builder=rules_builder)
+guessit_api.configure(
+    options={},
+    rules_builder=rules_builder,
+    force=True)
 
 
 def normalize_component(data):
     if data is None:
         return []
-    if isinstance(data, list):
-        return [d.lower().replace('-', '') for d in data]
 
-    return [data.lower().replace('-', '')]
+    if not isinstance(data, list):
+        data = [data]
+
+    return [d.lower().replace('-', '') for d in data]
 
 
 class ParserGuessit(object):
@@ -93,10 +97,38 @@ class ParserGuessit(object):
         """Generate a FlexGet Quality from a guessit result."""
         resolution = normalize_component(guessit_result.get('screen_size'))
         other = normalize_component(guessit_result.get('other'))
-        if not resolution and 'hr' in other:
+        if not resolution and 'high resolution' in other:
             resolution.append('hr')
 
         source = normalize_component(guessit_result.get('source'))
+
+        # some source renaming happened in guessit 3.0. This is here so quality lookups keeps working
+        # see https://github.com/guessit-io/guessit/blob/3.0.0/docs/migration2to3.rst
+        if 'digital tv' in source:
+            source[0] = 'dvb'
+        elif any(s in source for s in ['camera', 'hd camera']):
+            source[0] = 'cam'
+        elif 'hd telesync' in source:
+            source[0] = 'telesync'
+        elif 'payperview' in source:
+            source[0] = 'ppv'
+        elif 'digital tv' in source:
+            source[0] = 'dvb'
+        elif 'video on demand' in source:
+            source[0] = 'vod'
+        elif 'web' in source:
+            if 'rip' in other:
+                source[0] = 'webrip'
+            else:
+                source[0] = 'webdl'
+        elif 'analog hdtv' in source:
+            source[0] = 'ahdtv'
+        elif 'ultra hdtv' in source:
+            source[0] = 'uhdtv'
+        elif 'hd telecine' in source:
+            source[0] = 'hdtc'
+        # end renaming
+
         if 'preair' in other:
             source.append('preair')
         if 'screener' in other:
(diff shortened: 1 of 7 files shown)


sarendsen:
So, there is one unit test failing. See https://circleci.com/gh/Flexget/Flexget/4569
Which seems to be a bug in guessit. It doesnt parse the episode range correctly on that one.
Wondering howto deal whit this. Remove it or make work of it (meaning putting in a ticket at the guessit repo and waiting for a new release)

Fun fact. It works when not giving an expected_title to guessit. 

```
guessit 'Something.4x05-06'
For: Something.4x05-06
GuessIt found: {
    "title": "Something",
    "season": 4,
    "episode": [
        5,
        6
    ],
    "type": "episode"
}
```
but fails with expected title 
```
guessit 'Something.4x05-06' -T Something
For: Something.4x05-06
GuessIt found: {
    "title": "Something",
    "season": 4,
    "episode": 5,
    "release_group": "06",
    "type": "episode"
}
```

cvium:
I tried upgrading a few months ago and I ran into the same problem. I just gave up. Last time I opened a few issues on their github repo and they were quick to fix them, but their release schedule is slow.

sarendsen:
Exactly, even if we get this fixed on there side we will be waiting for a release. As this only breaks one unit test, might the maintainers consider this to be a tradeoff? Im open for ideas

JohnDoee:
GuessIt development is basically dead so waiting for it to get fixed might not be viable.

snickers2k:
> GuessIt development is basically dead so waiting for it to get fixed might not be viable.

Why you think that? 3.0.0 is released on  2018-05-22 and there is a Version 3.0.1 in developement. But since we are still using 2.1.4  it would be great to have flexget support for 3.0

cvium:
> we are still using 2.1.4 it would be great to have flexget support for 3.0

The bug (https://github.com/Flexget/Flexget/pull/2197#issuecomment-416686350) needs to be fixed before that can happen.

gazpachoking:
Did someone open an issue at guessit? Hard for them to fix it, or for us to track it without that.

snickers2k:
i just did that with referring to this issue here https://github.com/guessit-io/guessit/issues/577
we'll see if they even take a look at it. Maybe JohnDoee was right with guessit developement being dead. I Don't know.  
I guess i'll just stick with internal for now and see if guessit goes forward anytime. For now, i think sarendsen is right - guessit maybe should be just removed at this state and point. Maybe with just linking the guessit string to internal, so most people don't  have to change anything. but thats on you.

Toilal:
Hey, guessit development is not dead 😊, I just released a new release 3.0.2 and will try to have a look to this use case in the next few days. Stay tuned.

Toilal:
guessit has just been released with the fix (`3.0.3`).

Note that [series keyword has been removed](https://github.com/guessit-io/guessit/commit/12dc120e2c3a8f3114c190053d165bbede1808a0#diff-eba58ec782971c34eb1c1b92476130f0L156) from the default configuration, so [FlexGet.Series.2013.14.of.21.Title.Here.720p.HDTV.AAC5.1.x264-NOGRP use case](https://github.com/Flexget/Flexget/blob/80fdc98886c70082de5efab699f1d4afa4f82870/flexget/tests/test_seriesparser.py#L605) will probably fail ... 

This keyword was removed because of [this](https://github.com/guessit-io/guessit/issues/550) and [this](https://github.com/guessit-io/guessit/issues/561#issuecomment-417865808).

You can still apply an [advanced configuration](https://github.com/guessit-io/guessit/blob/develop/docs/configuration.rst) to bring back this keyword in guessit, but i'll vote against it.

sarendsen:
Updated to 3.0.3, and worked out some kinks. Ready for review i guess.

flexget/plugins/parsers/parser_guessit.py
```diff
@@ -229,7 +263,7 @@ def parse_series(self, data, **kwargs):
             episode = episode[0]
         date = guess_result.get('date')
         quality = self._quality(guess_result)
-        proper_count = self._proper_count(guess_result)
+        proper_count = self._proper_count(guess_result, )
```
cvium: Why the comma?

sarendsen: Nice catch. Small typo. Committed a fix.

sarendsen:
@cvium Bump :)

snickers2k:
+1
too many errors with 2.x :(

cvium:
Seems like there's a lot of errors with 3.x as well. 

https://github.com/Flexget/Flexget/issues/2249
https://discuss.flexget.com/t/unhandled-error-in-lazy-lookup-plugin-unhashable-type-list/4384/

Toilal:
There are options to enforce single value or list.

```
-s, --single-value Keep only first value found for each property
-l, --enforce-list Wrap each found value in a list even when property has
a single value
```


Le mar. 13 nov. 2018 à 07:25, Rémi Alvergnat <toilal.dev@gmail.com> a
écrit :

sarendsen:
@Toilal: Thanks for your taste response
@cvium: Based on the log and Toilal's response this should be fixable on our side. Ill see if i can put out a fix today.

PR #2203, opened 2018-09-11
Fixing 1337x.to scraping

tverona1:
### Motivation for changes:
1337x.to scraping has been broken. I'm not entirely sure when it broken, but the scraping login does not match current site content.

### Detailed changes:
In site_1337x.py, 1) Fixed url_rewrite() logic to properly match on torrent and magnet urls, 2) Fixed search() to properly match on download entries and corresponding columns (seeds, leeches, size).

### DIFF
--- a/flexget/plugins/sites/site_1337x.py
+++ b/flexget/plugins/sites/site_1337x.py
@@ -66,8 +66,8 @@ class Site1337x(object):
 
         soup = get_soup(page.content)
 
-        magnet_url = str(soup.find("a", href=re.compile(r"^magnet:\?")).get('href')).lower()
-        torrent_url = str(soup.find("a", href=re.compile(r"\.torrent$")).get('href')).lower()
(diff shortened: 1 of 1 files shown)


flexget/plugins/sites/site_1337x.py
```diff
@@ -66,8 +66,8 @@ def url_rewrite(self, task, entry):
 
         soup = get_soup(page.content)
 
-        magnet_url = str(soup.find('a', id='magnetdl').get('href')).lower()
-        torrent_url = str(soup.find('a', id='torrentdl').get('href')).lower()
+        magnet_url = str(soup.find("a", href=re.compile(r"^magnet:\?")).get('href')).lower()
```
cvium: We use single quotes for strings

tverona1: Fixed

PR #2216, opened 2018-10-07
Use HTTPS endpoint of Betaseries

chipho80:
Betaseries.com recommends the use of the HTTPS endpoint. Edited the plugin accordingly.

### Motivation for changes:
https://www.betaseries.com/api/requetes states the HTTPS is preferred over plain-text HTTP.

### Detailed changes:
- changed API_URL_PREFIX to use HTTPS
- also changed the links to betaseries.com documentation to https.


cvium:
Thanks

PR #2218, opened 2018-10-07
Add option to escape unescaped ampersands

asm0dey:
### Motivation for changes:

Right now some trackers (e.g. newstudio) do not eascape ampersands with `&amp;`

### Detailed changes:

This commits adds logics to fix that. Option is disabled by default because it adds some processing overhead and needs to go thru whole XML content and then rebuild new XML.

```
tasks:
  Newstudio:
    rss:
      url: '{? newstudio.address ?}'
      escape: yes
```

### DIFF
--- a/flexget/plugins/input/rss.py
+++ b/flexget/plugins/input/rss.py
@@ -192,7 +192,7 @@ class InputRSS(object):
 
         for idx, char in enumerate(content):
 
-            if type(char) is int:
+            if isinstance(char, int):
                 char = tobytes(chr(char))
             if not in_cdata_block and char == b'&':
                 if not content[idx:idx + 7].startswith(valid_escapes):
--- a/flexget/plugins/input/rss.py
+++ b/flexget/plugins/input/rss.py
@@ -186,22 +186,19 @@ class InputRSS(object):
         log.critical('I have saved the invalid content to %s for you to view', filepath)
 
     def escape_content(self, content):
-        valid_escapes = ['&quot;', '&apos;', '&lt;', '&gt;', '&amp;']
+        valid_escapes = ('&quot;', '&apos;', '&lt;', '&gt;', '&amp;')
         future_result = []
         in_cdata_block = False
         
         for idx, char in enumerate(content):
             char = chr(char)
             if not in_cdata_block and char == '&':
-                for escape in valid_escapes:
-                    if content[idx:idx+len(escape)] == escape:
-                        break
-                else:
+                if not str(content[idx:idx+8]).startswith(valid_escapes):
                     char = '&amp;'
             elif not in_cdata_block and char == '<' and content[idx:idx+9] == '<![CDATA[':
                 in_cdata_block = True
-            elif in_cdata_block and char == ']' and content[idx:idx+3] == ']]>':
-                should_change = False
+            elif in_cdata_block and char == ']' and content[idx-2:idx+1] == ']]>':
+                in_cdata_block = False
             future_result.append(char)
         return ''.join(future_result)
 
--- a/flexget/plugins/input/rss.py
+++ b/flexget/plugins/input/rss.py
@@ -190,23 +190,19 @@ class InputRSS(object):
         future_result = []
         in_cdata_block = False
         
-        for idx, char in enumerate(letter for letter in content):
-            if not in_cdata_block and chr(char) == '&':
+        for idx, char in enumerate(content):
+            char = chr(char)
+            if not in_cdata_block and char == '&':
                 for escape in valid_escapes:
                     if content[idx:idx+len(escape)] == escape:
                         break
                 else:
-                    future_result.append('&amp;')
-            elif not in_cdata_block and chr(char) == '<':
-                if content[idx:idx+9] == '<![CDATA[':
-                    should_change = False
-                future_result.append(chr(char))
-            elif in_cdata_block and chr(char) == ']':
-                if content[idx:idx+3] == ']]>':
-                    should_change = True
-                future_result.append(chr(char))
-            else:
-                future_result.append(chr(char))
+                    char = '&amp;'
+            elif not in_cdata_block and char == '<' and content[idx:idx+9] == '<![CDATA[':
+                in_cdata_block = True
+            elif in_cdata_block and char == ']' and content[idx:idx+3] == ']]>':
+                should_change = False
+            future_result.append(char)
         return ''.join(future_result)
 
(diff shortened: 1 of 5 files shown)


flexget/plugins/input/rss.py
```diff
@@ -184,6 +185,26 @@ def process_invalid_content(self, task, data, url):
...
+            if not in_cdata_block and char == '&':
+                for escape in valid_escapes:
+                    if content[idx:idx+len(escape)] == escape:
+                        break
+                else:
+                    char = '&amp;'
+            elif not in_cdata_block and char == '<' and content[idx:idx+9] == '<![CDATA[':
+                in_cdata_block = True
+            elif in_cdata_block and char == ']' and content[idx:idx+3] == ']]>':
+                should_change = False
```
cvium: This doesn't really do anything (you probably forgot to change it).  You probably want to like this instead, since it becomes a little more complex to skip ahead 3 characters while enumerating:

```
elif in_cdata_block and char == ']' and content[idx-2:idx+1] == ']]>':
     in_cdata_block= False
```

asm0dey: Thank you for comment and finding performance issue

flexget/plugins/input/rss.py
```diff
@@ -184,6 +185,23 @@ def process_invalid_content(self, task, data, url):
...
+        in_cdata_block = False
+        
+        for idx, char in enumerate(content):
+            char = chr(char)
+            if not in_cdata_block and char == '&':
+                if not str(content[idx:idx+8]).startswith(valid_escapes):
+                    char = '&amp;'
+            elif not in_cdata_block and char == '<' and content[idx:idx+9] == '<![CDATA[':
+                in_cdata_block = True
+            elif in_cdata_block and char == ']' and content[idx-2:idx+1] == ']]>':
```
cvium: Should be `char == '>'`

asm0dey: Maybe it's better to test on ']' because it can be seen in XML not such often as '>'?

cvium: You're already inside a CDATA block though.

asm0dey: But it's common practice to put XML inside CDATA

cvium:
It would be nice to have a couple of tests though.

asm0dey:
I would love to write some, but have no idea how to do it

asm0dey:
I can write tests in naive form, but don't know how to write them in python

we should test that 

```
escape('') == ''
escape('<![CDATA[]]>') == '<![CDATA[]]>'
escape('<![CDATA[&]]>') == ''<![CDATA[&]]>
escape('&amp;') == '&amp;'
escape('&gt;') == '&gt;'
escape('&&gt;&') == '&amp;&gt;&amp;'
escape('<![CDATA[&amp;]]>') == ''<![CDATA[&amp;]]>
```

asm0dey:
I do not understand ho they work :( How do they get content on localhost?

cvium:
The relative path points to https://github.com/Flexget/Flexget/blob/develop/flexget/tests/rss.xml

cvium:
You can just create a new text XML-file and write a few tests in the same vein as the existing ones.

flexget/plugins/input/rss.py
```diff
@@ -184,6 +185,25 @@ def process_invalid_content(self, task, data, url):
...
 
+    def escape_content(self, content):
+        valid_escapes = ('&quot;', '&apos;', '&lt;', '&gt;', '&amp;')
+        future_result = []
+        in_cdata_block = False
+
+        for idx, char in enumerate(content):
+
+            if type(char) is int:
+                char = chr(char)
```
gazpachoking: Not sure what these lines are for, and I don't think they should be here.

flexget/plugins/input/rss.py
```diff
@@ -184,6 +185,25 @@ def process_invalid_content(self, task, data, url):
             f.write(data)
         log.critical('I have saved the invalid content to %s for you to view', filepath)
 
+    def escape_content(self, content):
+        valid_escapes = ('&quot;', '&apos;', '&lt;', '&gt;', '&amp;')
```
gazpachoking: These (and every string literal in this function) should be changed to byte literals. i.e. `b'&quot;'`

flexget/plugins/input/rss.py
```diff
@@ -184,6 +185,25 @@ def process_invalid_content(self, task, data, url):
...
+        future_result = []
+        in_cdata_block = False
+
+        for idx, char in enumerate(content):
+
+            if type(char) is int:
+                char = chr(char)
+            if not in_cdata_block and char == '&':
+                if not str(content[idx:idx+7]).startswith(valid_escapes):
```
gazpachoking: This `str()` call should be removed. Once the literals you are comparing to are bytes, it won't be a problem.

asm0dey:
@gazpachoking is there anything else I can do to get this commit merged?

flexget/plugins/input/rss.py
```diff
@@ -184,6 +185,25 @@ def process_invalid_content(self, task, data, url):
...
 
+    def escape_content(self, content):
+        valid_escapes = (b'&quot;', b'&apos;', b'&lt;', b'&gt;', b'&amp;')
+        future_result = []
+        in_cdata_block = False
+
+        for idx, char in enumerate(content):
+
+            if isinstance(char, int):
+                char = tobytes(chr(char))
```
gazpachoking: After some investigating, the integers happen when iterating bytes in python 3. I think this way to convert to a length one bytestring is cleaner though:
`char = bytes([char])`

gazpachoking: I'm actually thinking remove the isinstance check, as the above will work fine on python2 as well.

asm0dey: `char = bytes([char])` doesn't work  in python2:

```
  File "/home/finkel/work_self/Flexget/flexget/plugin.py", line 118, in wrapped_func
    return func(*args, **kwargs)
  File "/home/finkel/work_self/Flexget/flexget/plugins/input/rss.py", line 311, in on_task_input
    content = self.escape_content(content)
  File "/home/finkel/work_self/Flexget/flexget/plugins/input/rss.py", line 195, in escape_content
    char = bytes([char])
  File "/usr/lib/python2.7/site-packages/future/types/newbytes.py", line 115, in __new__
    raise ValueError('bytes must be in range(0, 256)')
```

PR #2222, opened 2018-10-11
Allow variables plugin to replace arbitrary sections of config

### LINKED ISSUE #1765: Cannot use variables with integer-only values
### Expected behaviour:
Variables should be able to handle integer-only values.

### Actual behaviour:
Configuration parsing error (when using ``'{? deluge.port ?}'``): ``Got `50586`, expected: integer``
Configuration parsing error (when using ``{? deluge.port ?}``):
```
while parsing a flow mapping in "<unicode string>", line 16, column 13: port: {? deluge.port ?} ^ expected ',' or '}', but got '?' in "<unicode string>", line 16, column 28: port: {? deluge.port ?} ^
```

### Steps to reproduce:
- Step 1: Try to use below config.

#### Config:
```
    from_deluge:
      host: '{? deluge.host ?}'
      port: '{? deluge.port ?}'

```
  
#### Log:
See above.

### Additional information:

- Flexget Version: 2.10.24
- Python Version: 2.7.9
- Installation method: pip
- OS and version: macOS El Capitan 10.11.6
- Link to crash log: n/a

gazpachoking:
### Motivation for changes:

Variables plugin could not be used for anything except for string fields in config (or parts of string fields.)

### Detailed changes:
- Allows variables plugin to replace complex bits of config, like numbers, lists, ore whole dictionaries.
- Allows variables to be defined directly in the main config.

### Addressed issues:
- Fixes #1765

```yaml
variables:
  mock_entry_list:
  - title: a
  - title: b
  integer: 2
tasks:
  test_int_var:
    limit_new: "{? integer ?}"
  test_var_mock:
    mock: "{? mock_entry_list ?}"
```
#### To Do:

- [x] Update upgrade actions. Users that were relying on the string conversion will have to add some quotes to achieve old behavior.
- [x] Make sure there aren't any other side effects. (adding extra quotes in replacement should guarantee old behavior, as noted in upgrade actions)


liiight:
So you can basically use this now instead of yaml anchors ?
Also you had that native environment idea back in the day, would that fit this PR as well?

gazpachoking:
> So you can basically use this now instead of yaml anchors ?
Yeah, I think.
> Also you had that native environment idea back in the day, would that fit this PR as well?
This is that idea. 😉 It turns out it can do more than I thought.

liiight:
> This is that idea. 😉 It turns out it can do more than I thought.

Cool 😄

snickers2k:
for some reason, this isn't working for me.
```
    content_size:
      min:  "'{? Minimum_Movie_Size_Main ?}'"
      max:  "'{? Maximum_Movie_Size_Main ?}'"
```

```
2019-08-13 12:51 CRITICAL manager                       [/templates/enable-Movie_Size_Main/content_size/max] Got `'{? Maximum_Movie_Size_Main ?}'`, expected: number
2019-08-13 12:51 CRITICAL manager                       [/templates/enable-Movie_Size_Main/content_size/min] Got `'{? Minimum_Movie_Size_Main ?}'`, expected: number
```

tried both "'{ and "{ ..

PR #2225, opened 2018-10-12
Update regex usage for python 3.7

### LINKED ISSUE #2189: Flexget is incompatible with python 3.7
### Expected behaviour:

I run `flexget --cron execute`, and it does its job and exits silently. It however fails to start because of a dependency error.

### Actual behaviour:

The program crashes during the import phase, since a dependency, `rpyc`, uses a newly reserved word in Python 3.7.

The actual fix is simple: bump the `rpyc` requirement to at least 4.0.0. In my own set-up, this appeared to fix it, but it did complain about package versions. It might be that some plugins do suffer, my config isn't that special.

### Steps to reproduce:
- Step 1: Run flexget with Python 3.7
- Stop 2: Observe crash

#### Config:
Config isn't relevant to this issue, running Python 3.7 is. Either way:
```
tasks:
    Arch Linux ISOs:
        rss: https://www.archlinux.org/feeds/releases/
        accept_all: yes
        transmission: SECRET CONFIG

schedules:
        - tasks: '*'
          interval:
                  hours: 1
```
  
#### Log:
```
  File "/var/lib/flexget/.local/bin/flexget", line 7, in <module>
    from flexget import main
  File "/var/lib/flexget/.local/lib/python3.7/site-packages/flexget/__init__.py", line 12, in <module>                                                                                          
    from flexget.manager import Manager
  File "/var/lib/flexget/.local/lib/python3.7/site-packages/flexget/manager.py", line 38, in <module>                                                                                           
    from flexget.ipc import IPCClient, IPCServer  # noqa
  File "/var/lib/flexget/.local/lib/python3.7/site-packages/flexget/ipc.py", line 10, in <module>                                                                                               
    import rpyc
  File "/var/lib/flexget/.local/lib/python3.7/site-packages/rpyc/__init__.py", line 50
    from rpyc.utils.helpers import async, timed, buffiter, BgServingThread, restricted
                                       ^
SyntaxError: invalid syntax
```

### Additional information:

- FlexGet version: 2.14.15
- Python version: 3.7.0
- Installation method: `pip install --user`
- Using daemon (yes/no): no
- OS and version: Arch Linux ARM

gazpachoking:
### Motivation for changes:
Fix our usage of re.sub to work on python 3.7

### Detailed changes:
- Escapes any needed escapes in re.sub replacement patterns
- Also includes circleci test enabling from #2193

### Addressed issues:
- Fixes #2189

#### To Do:
- [ ] Add UpgradeActions, any users specifying regex replacements in manipulate plugin which contain unknown escape patterns will have to double escape them. (I think it's unlikely many users are doing this though.)
- [ ] Figure out proper fix for rtorrent client. Currently it works inconsistently across python versions and contains a hack to get test to pass.


JohnDoee:
Can you elaborate on the rtorrent problems ? I'm unsure why Flexget escapes at all.

gazpachoking:
@JohnDoee The rtorrent issues are here: https://github.com/Flexget/Flexget/blob/8709b7d9b89ee123d7a0e2c0bc15f3267ea9bb79/flexget/plugins/clients/rtorrent.py#L314-L316
It's using `re.escape` to escape, which changed functionality on python 3.7+. No idea what really needs to be escaped there, so I didn't know how to fix it.

JohnDoee:
So I looked through a bit of the rtorrent command argument parsing code.

My C++ isn't the best so it might be a bit wrong, anyways:

Argument parsing starts after first = found
https://github.com/rakshasa/rtorrent/blob/f0207ce6548026853ec57ab26f2e57a872f223bb/src/rpc/parse_commands.cc#L148

Variable parsing: https://github.com/rakshasa/rtorrent/blob/f0207ce6548026853ec57ab26f2e57a872f223bb/src/rpc/parse.cc#L175

Some helper commands: https://github.com/rakshasa/rtorrent/blob/f0207ce6548026853ec57ab26f2e57a872f223bb/src/rpc/parse.h#L53

Special characters in the given parsing that should probably be escaped for these use-cases:

* `,` List separator
* `"` Quote / Grouping an item
* `\` Escape characters
* `{}` For a list ?
* `()` Some grouping ? can be nested ?

Alternatively, throw quotes around, escape quotes and strip `\`

chestm007:
I have the diff of this PR currently being applied to my AUR build [flexget-git](https://aur.archlinux.org/packages/flexget-git/)  and its working fine for me. Perhaps we could get this PR in and make a sub/another task to fix rtorrent?

paranoidi:
I got some complaints with Python 3.6, committed fixes for those.

PR #2231, opened 2018-10-18
Add more entry fields to PassThePopcorn search plugin

jtpavlock:
### Motivation for changes:

Ability for more reliable and finer control over passthepopcorn search results.

### Detailed changes:
- Added entry fields from search results that are already present in the request, but are currently ignored and not saved. 

#### To Do:

- If possible, I could try to map passthepopcorn qualities to current flexget qualities e.g. Blu-ray:bluray. These mapped qualities would take precedence over parsed qualities from torrent title which would result in more accurate quality mapping. However, I'm not sure how viable or even possible this would be.

### DIFF
--- a/flexget/plugins/sites/passthepopcorn.py
+++ b/flexget/plugins/sites/passthepopcorn.py
@@ -291,6 +291,7 @@ class SearchPassThePopcorn(object):
                     e['torrent_id'] = int(torrent['Id'])
                     e['golden_popcorn'] = torrent['GoldenPopcorn']
                     e['checked'] = torrent['Checked']
+                    e['scene'] = torrent['Scene']
                     e['uploaded_at'] = dateutil_parse(torrent['UploadTime'])
 
                     e['url'] = self.base_url + 'torrents.php?action=download&id={}&authkey={}&torrent_pass={}'.format(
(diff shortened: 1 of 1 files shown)


flexget/plugins/sites/passthepopcorn.py
```diff
@@ -288,8 +291,16 @@ def search(self, task, entry, config):
...
                     e['checked'] = torrent['Checked']
+                    e['scene'] = torrent['Scene']
                     e['uploaded_at'] = dateutil_parse(torrent['UploadTime'])
 
+                    e['ptp_remaster_title'] = torrent['RemasterTitle'] # tags such as remux, 4k remaster, etc.
+                    e['ptp_quality'] = torrent['Quality'] # high, ultra high, or standard definition
+                    e['ptp_resolution'] = torrent['Resolution'] # 1080p, 720p, etc.
+                    e['ptp_source'] = torrent['Source'] # blu-ray, dvd, etc.
+                    e['ptp_container'] = torrent['Container'] # mkv, vob ifo, etc.
+                    e['ptp_codec'] = torrent['Codec'] # x264, XviD, etc.
```
cvium: Can we be certain that all these keys will exist in every response?

jtpavlock: Actually the only one we need to worry about is remastertitle which I missed the first time around. The rest are all require and will always be there. Regardless, I switched the key access to use dict.get() which should prevent any issues.

PR #2236, opened 2018-10-21
Fix non-critical PassThePopcorn entry field comparison error

jtpavlock:
### Motivation for changes:

Fixes a non-critical error occurring when accessing one of the entry fields that is None-type. This was occurring for me when comparing a string to type None for example. 

### Detailed changes:
- Sets the default value for entries without values to be an empty string instead of None. This will remove any errors when trying to compare a string in the entry field.

### DIFF
--- a/flexget/plugins/sites/passthepopcorn.py
+++ b/flexget/plugins/sites/passthepopcorn.py
@@ -294,12 +294,12 @@ class SearchPassThePopcorn(object):
                     e['scene'] = torrent['Scene']
                     e['uploaded_at'] = dateutil_parse(torrent['UploadTime'])
 
-                    e['ptp_remaster_title'] = torrent.get('RemasterTitle') # tags such as remux, 4k remaster, etc.
-                    e['ptp_quality'] = torrent.get('Quality') # high, ultra high, or standard definition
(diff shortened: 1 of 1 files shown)


paranoidi:
Hmm, if the value is empty do we want to omit the field instead, I think this is how we generally have operated? Looks good otherwise.

jtpavlock:
on the configuration side, there is no way to know if a field will be empty or not ahead of time. If you omit the field entirely and the someone attempts to access it in their config (I’m thing via regexp or if plug-in) wouldn’t that cause issues?

liiight:
We generally not so this. Any entry field can be none and you should check for this in your jinja usage. I'm not saying its ideal, but it's what we use all around.

jtpavlock:
I see. What’s a good way to check if an entry field is none before attempting to access it then?

gazpachoking:
You can check if a field has a value in jinja with `{% if somefield %}` This will be false if somefield is None, or an empty string. If you want to also check if the field exists at all, you can do `{% if somefield is defined and somefield %}`. If you are just trying to turn a None into an empty string, you can do `{{ somefield|default('') }}`

cvium:
Based on the discussion, I'll close this.

PR #2241, opened 2018-10-24
PassThePopcorn ignore movie year discrepancies if we have imdb id

jtpavlock:
### Motivation for changes:

Sometimes if there is a year discrepancy in a movie's release year between trakt (or other input sites that originally set movie_year), imdb, and passthepopcorn, it will reject the movie on passthepopcorn based on this discrepancy. This change will allow to ignore year discrepancies if we are dealing specifically with an imdb_id (no possibility of mismatches).

### Detailed changes:
- Ignores year matching check in PassThePopcorn plugin if using imdb_lookup and thus dealing with imdb ids i.e. there will be no mismatches so checking the year is not necessary. 

Note that if whatever plugin produces the imdb_id produces the wrong or incorrect id, PassThePopcorn will download that movie with the imdb_id without failsafe procedure.

### DIFF
--- a/flexget/plugins/sites/passthepopcorn.py
+++ b/flexget/plugins/sites/passthepopcorn.py
@@ -266,10 +266,14 @@ class SearchPassThePopcorn(object):
             passkey = result['PassKey']
 
             for movie in result['Movies']:
-                # skip movies that are irrelevant
-                if entry.get('movie_year') and int(movie['Year']) != int(entry['movie_year']):
-                    log.debug('Movie year %s does not match %s', movie['Year'], entry['movie_year'])
+                # skip movies with wrong year
+                # don't consider if we have imdb_id (account for year discrepancies if we know we have
+                # the right movie)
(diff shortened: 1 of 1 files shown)


flexget/plugins/sites/passthepopcorn.py
```diff
@@ -266,10 +266,14 @@ def search(self, task, entry, config):
             passkey = result['PassKey']
 
             for movie in result['Movies']:
-                # skip movies that are irrelevant
-                if entry.get('movie_year') and int(movie['Year']) != int(entry['movie_year']):
-                    log.debug('Movie year %s does not match %s', movie['Year'], entry['movie_year'])
+                # skip movies with wrong year
+                # don't consider if we have imdb_id (account for year discrepancies if we know we have
+                # the right movie)
+                if (not entry.get('imdb_id') and entry.get('movie_year') and
```
paranoidi: Mpre pythonic way to check if field is set:

'movie_year' in entry

PR #2245, opened 2018-11-03
Add proxy support for rutracker plugin

definitio:
### Motivation for changes:
I need it because rutracker is blocked in my country.

### Detailed changes:
- Use requests from task instead of importing in plugin

### DIFF
--- a/flexget/plugins/sites/rutracker.py
+++ b/flexget/plugins/sites/rutracker.py
@@ -91,8 +91,8 @@ class RutrackerAuth(AuthBase):
                 sleep(3)
         raise PluginError('unable to obtain cookies from rutracker')
 
-    def __init__(self, task, login, password, cookies=None, db_session=None):
-        self.requests = task.requests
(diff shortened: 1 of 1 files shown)


flexget/plugins/sites/rutracker.py
```diff
@@ -87,15 +84,15 @@ def update_base_url():
...
+            self.requests.post('{}/forum/login.php'.format(self.base_url), data=payload)
+            if self.requests.cookies and len(self.requests.cookies) > 0:
+                return self.requests.cookies
             else:
                 sleep(3)
         raise PluginError('unable to obtain cookies from rutracker')
 
-    def __init__(self, login, password, cookies=None, db_session=None):
+    def __init__(self, task, login, password, cookies=None, db_session=None):
+        self.requests = task.requests
```
cvium: If you don't use the task for anything other than grabbing its requests session, why not cut out the middle man and just pass it directly?

definitio: Fixed

Issue #2249, opened 2018-11-10
BUG: list_match crashes when an RSS entry contains two years


RicardoVelaC:
### Expected behaviour:

When in a RSS task parse an entry of a movie uses `imdb_lookup` if found an coincidence from 
`list_match` add task to `transmissionrcp`
It's work with different RSS providers with same config.

### Actual behaviour:

- When one entry of RSS contains two year `list_match` and `imdb_lookup` plugins make flexget crashes.

- RSS have a format like this: 
`[MicroHD] Some title (estreno 2018)(2016/MHD/1080P/AC3 5.1/DUAL/SUB) [SEEDERS (29)/LEECHERS (2)]`

- It's fail with `internal` and `guessit` parser. If title only contain one year it's work ok with `guessit` parser. If contains two years fail.

### Steps to reproduce:
- Step 1: Create task including an RSS source with two years in the title

#### Config:
```
web_server:
  run_v2: true

templates:
  global:
    torrent_alive: no #number of seeders needed to accept

  transmit-movies:
    transmission:
      host: 192.X.X.X
      port: 9091
      username: admin
      password: password
      path: /var/lib/transmission-daemon/downloads/movies
    metainfo_movie: yes
    set:
      rename_like_files: yes
      content_filename: "{{ movie_name }} ({{ movie_year }})"

schedules:
  - tasks: '*'
    interval:
      minutes: 10

  mov_xbytes task:
    priority: 3
    rss: http://domain.com/advanced_rss.php?cats=16&tpc=5&auth=AUTH-PERSONAL-CODE
    parsing:
      movie: guessit
    imdb_lookup: yes
    list_match:
      from:
        - movie_list: movies
      action: accept
    template: transmit-movies
```
  
#### Log:
<details>
 <summary>(click to expand)</summary>

```
2018-11-10 13:01 DEBUG    manager                       Config file /config/config.yml selected
2018-11-10 13:02 INFO     scheduler                     Starting scheduler
2018-11-10 13:02 INFO     web_server_daemon                 Running web server at IP 0.0.0.0:5050
2018-11-10 13:02 INFO     web_server_daemon                 Initiating API
2018-11-10 13:02 INFO     web_server_daemon                 Registering WebUI v2
2018-11-10 13:02 INFO     web_server_daemon                 Registering WebUI v1
2018-11-10 13:02 INFO     web_server                    Web interface available at http://172.17.0.4:5050
[...]

PR #2252, opened 2018-11-13
Force single values from guessit

sarendsen:
### Motivation for changes:

Guessit would return a list of property values; found when parsing instead of a string.
This resulted in unexpected behaviour and errors on flexget side. Most noticable erroring out in the guessit parser mappings.

### Addressed issues:
- #2249
- https://discuss.flexget.com/t/unhandled-error-in-lazy-lookup-plugin-unhashable-type-list/4384/
- https://github.com/Flexget/Flexget/pull/2197


gazpachoking:
Looks good to me, thanks!

PR #2253, opened 2018-11-17
Add tags to my_anime_list plugin entries

AnthonyGuerreiro:
### Motivation for changes:
Allow user to filter by entry tags with the `if` plugin - example in config_usage
### Detailed changes:
- Add `mal_tags` field to the entries produced by my_anime_list plugin, populated by `tags` provided by MyAnimeList API

```
      if:
        - '"ignored" not in mal_tags.split(", ")': accept
```

### DIFF
--- a/flexget/plugins/input/my_anime_list.py
+++ b/flexget/plugins/input/my_anime_list.py
@@ -32,6 +32,7 @@ ANIME_TYPE = [
     'unknown'
 ]
 
+
 class MyAnimeList(object):
     """" Creates entries for series and movies from MyAnimeList list
     Syntax:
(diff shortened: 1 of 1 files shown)


flexget/plugins/input/my_anime_list.py
```diff
@@ -90,7 +91,8 @@ def on_task_input(self, task, config):
                         url="https://myanimelist.net" + anime["anime_url"],
                         mal_name=anime["anime_title"],
                         mal_poster=anime["anime_image_path"],
-                        mal_type=anime["anime_media_type_string"]
+                        mal_type=anime["anime_media_type_string"],
+                        tags=anime["tags"]
```
cvium: This should probably be prefixed with `mal_`

AnthonyGuerreiro: Good point, I'll change that.

PR #2259, opened 2018-11-23
Transmission refactor

AnthonyGuerreiro:
### Motivation for changes:
Try to make transmission plugin somewhat more maintainable

### Detailed changes:
- too many.. overall just a refactor but the following configuration was changed:
Not accepting the short form anymore, given by:
```
 'anyOf': [
            {'type': 'boolean'},
            (.. advanced config follows..)
```
reason: I couldn't find a proper way to have default values with it other than checking one by one.
However, this should work with all default values, I think it's acceptable?:

```
tasks:
   thetask:
      transmission: {}
```
As tested by TestTransmission (test_transmission.py)

### DIFF
--- a/flexget/plugins/clients/transmission/clean.py
+++ b/flexget/plugins/clients/transmission/clean.py
@@ -9,7 +9,7 @@ from flexget.utils.tools import parse_timedelta
 
 from flexget.plugins.clients.transmission.client import create_rpc_client
 from flexget.plugins.clients.transmission.transmission import TransmissionBase, log
-from flexget.plugins.clients.transmission.utils import torrent_info, check_seed_limits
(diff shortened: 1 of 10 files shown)


gazpachoking:
Sorry for the lack of response here @AnthonyGuerreiro . I recently did some other refactoring of the transmission plugins and didn't remember this was in the queue. If there is stuff in here that is still applicable maybe you could break it down into smaller PRs that are easier to evaluate and get merged? As for your question about defaults, we normally handle that by creating a function to turn the user config into a standard format, something like:
```python
def prepare_config(config):
    if isinstance(config, bool):
        config = {'enabled': config}
    config.setdefault('someoption', 'somevalue')
    return config
```
then call that from every function that gets passed the user's config.

PR #2263, opened 2018-12-05
Erichlf torrentday

### LINKED ISSUE #2105: No results found on torrentday
### Expected behaviour:
When using the torrentday as a source with discover, results should display

### Actual behaviour:
Always returns "No results from torrentday"

### Steps to reproduce:
Setup to search as per config

#### Config:
https://pastebin.com/6xWSD9i2
  
#### Log:
https://pastebin.com/uqa0bDea

### Additional information:
torrentday appearts to have updated their browse/search page https://www.torrentday.com/browse.php

erichlf:
### Motivation for changes: 
- torrentday always resulted in no entry found for searches

### Detailed changes: 
- The tags on torrentday changed, so I changed the plugin tags to match torrentday's current tags

### Addressed issues:
- Fixes #2105

### DIFF
--- a/flexget/plugins/sites/torrentday.py
+++ b/flexget/plugins/sites/torrentday.py
@@ -127,26 +127,29 @@ class UrlRewriteTorrentday(object):
         entries = set()
         for search_string in entry.get('search_strings', [entry['title']]):
 
-            url = 'https://www.torrentday.com/browse.php'
-            params['search'] = normalize_unicode(search_string).replace(':', '')
+            url = 'https://www.torrentday.com/t'
+            params['q'] = normalize_unicode(search_string).replace(':', '')
             cookies = { 'uid': config['uid'], 'pass': config['passkey'], '__cfduid': config['cfduid'] }
 
             try:
-                page = requests.get(url, params=params, cookies=cookies).content
+                page = requests.get(url, params=params, cookies=cookies)
             except RequestException as e:
                 raise PluginError('Could not connect to torrentday: %s' % e)
 
-            soup = get_soup(page)
-
-            for tr in soup.find_all('tr', { 'class': 'browse' }):
+            soup = get_soup(page.content)
+            form = soup.find('form', id='torrents')
+            table = form.find('table', { 'id': 'torrentTable' })
+            # the first row is the header so skip it
+            for tr in table.find_all('tr')[1:-1]:
                 entry = Entry()
                 # find the torrent names
-                title = tr.find('a', { 'class': 'torrentName' })
+                td = tr.find('td', { 'class': 'torrentNameInfo' })
+                title = td.find('a')
                 entry['title'] = title.contents[0]
                 log.debug('title: %s', title.contents[0])
 
                 # find download link
-                torrent_url = tr.find('td', { 'class': 'dlLinksInfo' })
+                torrent_url = tr.find_all('td', { 'class': 'ac' })[0]
                 torrent_url = torrent_url.find('a').get('href')
 
                 # construct download URL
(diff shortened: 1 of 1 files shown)


flexget/plugins/sites/torrentday.py
```diff
@@ -123,30 +123,39 @@ def search(self, task, entry, config=None):
...
 
-            url = 'https://www.torrentday.com/browse.php'
-            params['search'] = normalize_unicode(search_string).replace(':', '')
+            url = 'https://www.torrentday.com/t'
+            params['q'] = normalize_unicode(search_string).replace(':', '')
             cookies = { 'uid': config['uid'], 'pass': config['passkey'], '__cfduid': config['cfduid'] }
 
             try:
-                page = requests.get(url, params=params, cookies=cookies).content
+                page = requests.get(url, params=params, cookies=cookies)
```
gazpachoking: Grabbing `.content` from a request can still cause network exceptions, like timeout errors, which is probably why that bit was included in the try/except block. There should probably be another try/except around that if it gets moved out of this one.

flexget/plugins/sites/torrentday.py
```diff
@@ -123,30 +123,39 @@ def search(self, task, entry, config=None):
...
+                raise PluginError('Could not connect to torrentday: {}'.format(e))
 
-            soup = get_soup(page)
-
-            for tr in soup.find_all('tr', { 'class': 'browse' }):
+            soup = get_soup(page.content)
+            form = soup.find('form', id='torrents')
+            table = form.find('table', { 'id': 'torrentTable' })
+            # the first row is the header so skip it
+            for tr in table.find_all('tr')[1:-1]:
```
gazpachoking: Is this meant to skip the last row too? If you just want to skip the first row, you probably want `[1:]`

flexget/plugins/sites/torrentday.py
```diff
@@ -123,30 +123,39 @@ def search(self, task, entry, config=None):
...
                 entry = Entry()
                 # find the torrent names
-                title = tr.find('a', { 'class': 'torrentName' })
+                td = tr.find('td', { 'class': 'torrentNameInfo' })
+                if not td:
+                    raise PluginError('Could not find entry torrentNameInfo for {}.'.format(search_string))
```
gazpachoking: This will abort causing zero results. Is it possible we just want to log this, and continue checking for other results that might be valid? (I don't know the answer to this, it could be appropriate as is.)

Same question for the other PluginErrors within this loop.

flexget/plugins/sites/torrentday.py
```diff
@@ -123,30 +123,39 @@ def search(self, task, entry, config=None):
...
+                title = td.find('a')
+                if not title:
+                    raise PluginError('Could not determine title for {}.'.format(search_string))
                 entry['title'] = title.contents[0]
                 log.debug('title: %s', title.contents[0])
 
                 # find download link
-                torrent_url = tr.find('td', { 'class': 'dlLinksInfo' })
+                torrent_url = tr.find_all('td', { 'class': 'ac' })[0]
```
gazpachoking: Is there a reason to do `.find_all()`, but then just pick the first result? calling plain `.find()` will get the first result but then not bother searching the rest of the document.

flexget/plugins/sites/torrentday.py
```diff
@@ -155,7 +164,7 @@ def search(self, task, entry, config=None):
                 entry['url'] = torrent_url
 
                 # us tr object for seeders/leechers
-                seeders, leechers = tr.find_all('td', { 'class': ['seedersInfo', 'leechersInfo']})
+                seeders, leechers = tr.find_all('td', { 'class': ['ac seedersInfo', 'ac leechersInfo']})
```
gazpachoking: I'd probably split this into 2 find calls explicitly for seeders, then leechers.

flexget/plugins/sites/torrentday.py
erichlf: Yeah, that is what I originally did, but it caused a runtime error. I was so confused by that.

flexget/plugins/sites/torrentday.py
erichlf: Sure

flexget/plugins/sites/torrentday.py
erichlf: I was just doing what was already there, but I can split it.

erichlf: Splitting this cause an error, so I am going to leave this how it was.

gazpachoking:
Looks good to me, thanks!

PR #2292, opened 2019-01-02
[fix] upgrade plugin propers and target quality.

stevezau:
### Motivation for changes:
The upgrade plugin was not properly checking for propers or check if the target quality has been hit.

### Detailed changes:
- If target quality hit, stop upgrading unless proper found
- if propers not set, don't accept propers

### DIFF
--- a/flexget/plugins/filter/upgrade.py
+++ b/flexget/plugins/filter/upgrade.py
@@ -68,31 +68,54 @@ class FilterUpgrade(object):
         config.setdefault('propers', True)
         return config
 
-    def filter_entries(self, entries, existing, target, action_on_lower):
+    def filter_entries(self, entries, existing, target, propers, action_on_lower):
 
         target_requirement = qualities.Requirements(target) if target else None
+        target_quality = qualities.Quality(target) if target else None
         filtered = []
 
         for entry in entries:
-            # Filter out entries within target
             if target:
+                # Filter out entries not within target range
                 if not target_requirement.allows(entry['quality']):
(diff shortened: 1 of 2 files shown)


flexget/plugins/filter/upgrade.py
```diff
@@ -68,31 +68,54 @@ def prepare_config(self, config):
         config.setdefault('propers', True)
         return config
 
-    def filter_entries(self, entries, existing, target, action_on_lower):
+    def filter_entries(self, entries, existing, target, propers, action_on_lower):
 
         target_requirement = qualities.Requirements(target) if target else None
+        target_quality = qualities.Quality(target) if target else None
```
gazpachoking: Hmm. Is 'target' a specific quality, or a quality requirement? It doesn't really make sense for it to be both.

flexget/plugins/filter/upgrade.py
```diff
@@ -68,31 +68,54 @@ def prepare_config(self, config):
...
         filtered = []
 
         for entry in entries:
-            # Filter out entries within target
             if target:
+                # Filter out entries not within target range
                 if not target_requirement.allows(entry['quality']):
-                    log.debug('Skipping %s as does not meet upgrade quality requirements', entry['title'])
+                    if entry['quality'] > target_quality:
```
liiight: style suggestion:

```suggestion
                    reason = '(higher quality)' if entry['quality'] > target_quality else '(lower quality)'
```

gazpachoking:
Isn't the proper_movies plugin supposed to handle getting propers? Is this duplicating that functionality?

PR #2295, opened 2019-01-04
Make disable plugin work on all plugins

gazpachoking:
### Motivation for changes:
Right now, it's hard or impossible to disable certain plugins like estimators or urlrewriters which don't get configured directly in a task.

### Detailed changes:
- Creates a new property on the task class called disabled_plugins. A new method, disable_plugin can be called to add to it.
- Adds a parameter to get_plugins, exclude, which can be used to easily get a list of plugins with task.disable_plugins being passed as the exclude argument.
- Just realized there was disable_urlrewriters plugin. Deprecated that in favor of this
- There is a bit of creep here because I realized plugins registered within the test functions were staying registered for every test. Added a new way to register plugins for specific tests using `pytest.mark.register_plugin(<args to flexget.plugin.register here>)`

TODO:
- [ ] Make this work for estimators
- [ ] Make it work for parsers
- [ ] Make it work for the movie_metainfo plugins?

### DIFF
--- a/flexget/plugins/internal/change_warn.py
+++ b/flexget/plugins/internal/change_warn.py
@@ -31,10 +31,6 @@ class ChangeWarn(object):
             log.critical('Plugin nzb_size is deprecated, use content_size instead')
             found_deprecated = True
 
-        if 'disable_urlrewriters' in task.config:
-            log.critical('Plugin disable_urlrewriters is deprecated, use disable instead')
-            found_deprecated = True
-
         if found_deprecated:
             task.manager.shutdown(finish_queue=False)
             task.abort('Deprecated config.')
--- a/flexget/plugins/internal/change_warn.py
+++ b/flexget/plugins/internal/change_warn.py
@@ -31,6 +31,10 @@ class ChangeWarn(object):
             log.critical('Plugin nzb_size is deprecated, use content_size instead')
             found_deprecated = True
 
+        if 'disable_urlrewriters' in task.config:
+            log.critical('Plugin disable_urlrewriters is deprecated, use disable instead')
+            found_deprecated = True
+
         if found_deprecated:
             task.manager.shutdown(finish_queue=False)
             task.abort('Deprecated config.')
--- a/flexget/plugin.py
+++ b/flexget/plugin.py
@@ -469,7 +469,7 @@ def load_plugins(extra_dirs=None):
     log.debug('Plugins took %.2f seconds to load. %s plugins in registry.', took, len(plugins.keys()))
 
 
-def get_plugins(phase=None, interface=None, category=None, name=None, min_api=None):
+def get_plugins(phase=None, interface=None, category=None, name=None, min_api=None, exclude=None):
     """
     Query other plugins characteristics.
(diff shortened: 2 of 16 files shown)


flexget/plugins/internal/change_warn.py
```diff
@@ -31,6 +31,10 @@ def on_task_start(self, task, config):
             log.critical('Plugin nzb_size is deprecated, use content_size instead')
             found_deprecated = True
 
+        if 'disable_urlrewriters' in task.config:
+            log.critical('Plugin disable_urlrewriters is deprecated, use disable instead')
```
liiight: Wouldn't a warning fit here better? Also, why disable the old way right away? You can set a depreciation message on that plugin for now.

gazpachoking: Well, this change warn was silly turns out. We won't ever get here if I leave it removed (because schema validation will fail.) The main reason I wanted to delete it right now is because we don't have a good system to follow up on deleting deprecated things.

gazpachoking: Re-added this plugin which just calls the disable plugin methods now and has a deprecation warning.

gazpachoking:
> Keeping task specific flags in PluginInfo seems like a bad idea from single responsibility principle point of view.

Yeah, I agree this isn't the 'right' way to do it. I just took the easy way out here and continued doing it the same way the disable plugin currently works. Not exactly sure a nice way to do it properly.

paranoidi:
> > Keeping task specific flags in PluginInfo seems like a bad idea from single responsibility principle point of view.
> 
> Yeah, I agree this isn't the 'right' way to do it. I just took the easy way out here and continued doing it the same way the disable plugin currently works. Not exactly sure a nice way to do it properly.

What if task did maintain list of disabled plugins and did not call them at all?

gazpachoking:
> What if task did maintain list of disabled plugins and did not call them at all?

I guess that's the thing to do. The only complication is that any other plugin that uses plugins also needs to check that list. e.g.
https://github.com/Flexget/Flexget/blob/80fdc98886c70082de5efab699f1d4afa4f82870/flexget/plugins/parsers/plugin_parsing.py#L25
https://github.com/Flexget/Flexget/blob/80fdc98886c70082de5efab699f1d4afa4f82870/flexget/plugins/internal/urlrewriting.py#L41
https://github.com/Flexget/Flexget/blob/80fdc98886c70082de5efab699f1d4afa4f82870/flexget/plugins/estimators/est_released.py#L24
https://github.com/Flexget/Flexget/blob/80fdc98886c70082de5efab699f1d4afa4f82870/flexget/plugins/modify/urlrewrite_search.py#L48
https://github.com/Flexget/Flexget/blob/80fdc98886c70082de5efab699f1d4afa4f82870/flexget/plugins/list/movie_list.py#L34

gazpachoking:
Thinking the best way is adding a Task.disabled_plugins property and just checking that where appropriate manually. Closing this for now and will write up that way if it becomes important or I get some motivation for it again.

PR #2299, opened 2019-01-12
Update `get_marker` to `get_closest_marker` to fix pytest

mfonville:
### Motivation for changes:
Without this change the tests fail if pytest is up to date, because `get_marker` has been removed


gazpachoking:
LGTM, thanks!

PR #2301, opened 2019-01-15
Update torrentz.py

Daeymien:
### Motivation for changes:
correct invalid/incorrect urls/domains

invalid feeds left commented for ease should they eventually be re-introduced at a later date


gazpachoking:
Thanks!

PR #2302, opened 2019-01-17
Dependency cleanups

paranoidi:
### Motivation for changes:

Plugin imports are messy

### Detailed changes:
- One method in all places except rewriters 
- Updated some dependencies which crashed
- Added missing dependency (progressbar)


gazpachoking:
I'm 👍 on this

PR #2303, opened 2019-01-18
Decouple core and plugin parsers

paranoidi:
### Motivation for changes:

Parsers are depending on core and other way around

### Detailed changes:
- Moved generic stuff, renamed things ..


flexget/plugins/parsers/parser_guessit.py
```diff
@@ -1,22 +1,23 @@
 from __future__ import unicode_literals, division, absolute_import
-from builtins import *  # noqa pylint: disable=unused-import, redefined-builtin
-from future.utils import native
 
 import logging
 import re
 import sys
 import time
+from builtins import *  # noqa pylint: disable=unused-import, redefined-builtin
```
gazpachoking: We had been leaving all the python2/3 compatibility stuff up in the top block with the `__future__` import. (`builtins` is a module from future.) Do you disagree with that, or was this just a result of isort automatically sorting? In either case, `builtins` isn't from stdlib, so it doesn't go in this block.

paranoidi: Pycharm sorting, manual sorting is not really worth the trouble IMO ..

gazpachoking: It looks like isort has an option for us which should work
```
known_future_library = builtins,future,past
```
for some reason it doesn't work for the `builtins` import. (It does work for future and past imports though)

gazpachoking: Okay, I figured out how to get builtins sorting up with the future section and updated our `.isort.cfg`

PR #2306, opened 2019-01-24
added speed limit support for qBittorrent

monokoo:
1. maxupspeed: support to set torrent upload speed limit. Unit in KiB/s  --same as transmission
2. maxdownspeed: support to set torrent download speed limit. Unit in KiB/s --same as transmission


cvium:
Thanks!

PR #2310, opened 2019-01-25
Add a 'sort' option to series plugin

gazpachoking:
Allows changing or disabling series plugin sorting

### Motivation for changes:
Series plugin has inbuilt sorting, preventing user from specifying which release is preferred when there are multiple copies of the same episode available.

### Detailed changes:
- Add the 'sort' option to series plugin. Allows sorting similar to sort_by plugin, but also per series.
- If the sort option is set to [], sorting is turned off, and task order is respected (including sort_by plugin ordering, if included in the task)

```yaml
series:
- My Show:
    sort:
      - sort_field_1
      - field: sort_field_2
        reverse: yes
```

#### To Do:

- [ ] Allow `sort: no` format instead of `sort: []` format to disable sorting? If so, do we allow `sort: yes`? Does that mean sort by the default (quality, descending)?
- [ ] Allow a single field name or dict instead of requiring list format?
- [ ] Just thought about it, should the option be called sort_by?
- [ ] For a future PR, should we modify sort_by plugin to match whatever config format we allow here? (currently sort_by can only take a single field to sort by)

### DIFF
--- a/flexget/plugins/filter/series.py
+++ b/flexget/plugins/filter/series.py
@@ -1388,6 +1388,20 @@ class FilterSeriesBase(object):
                 'timeframe': {'type': 'string', 'format': 'interval'},
                 'upgrade': {'type': 'boolean'},
                 'target': {'type': 'string', 'format': 'quality_requirements'},
+                # Sort order, a list of either field names, or dicts with field and reverse key
+                'sort': {
+                    'type': 'array',
+                    'items': {'anyOf': [
+                        {'type': 'string'},
+                        {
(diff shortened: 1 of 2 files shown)


gazpachoking:
Thinking now that we should remove the internal sorting entirely from series plugin. sort_by could be used at task level to do any sorting. Won't allow sorting per series, but that was probably overkill anyway.

gazpachoking:
#2311 will allow similar functionality, closing this in favor of that.

PR #2312, opened 2019-01-26
Remove internal sorting from series plugin

gazpachoking:
### Motivation for changes:
The series plugin internal sorting means you are currently unable to prioritize certain entries when there are multiple in the task that are acceptable. Removing the internal sorting allows the user to use sort_by plugin to prioritize based on other criteria than best quality. e.g. Grab worst (acceptable) quality, grab from one tracker over another, from one input plugin over another.

This should wait for #2311, to allow duplication of current behavior.

#### To Do:

- [x] Figure out propers. The series plugin used to sort by proper count after quality, to make sure it grabbed a proper version over the nuked one. (because it will download the proper anyway on the next run)
- [x] Double (or any multi eps) were sorted first, in an attempt to make it grab the double rather than ending up getting a single ep and then getting a double-ep that contained that single ep. Not 100% sure if this really needs to be fixed or not. Result: Double eps are no longer preferred unless they are first in task.
- [x] Add a test to explicitly verify we are respecting incoming entry order.

### DIFF
--- a/flexget/plugins/filter/series.py
+++ b/flexget/plugins/filter/series.py
@@ -1,6 +1,7 @@
 from __future__ import unicode_literals, division, absolute_import
 
 import argparse
+import itertools
 import logging
 import re
 import sys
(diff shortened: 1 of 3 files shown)


gazpachoking:
I'm thinking perhaps prioritizing double-eps isn't so important and we can just remove the test. They aren't very common, and they would be even less common to show up alongside a single episode that they contain. Worst case scenario is getting both a single ep then the double ep that also contains it.

flexget/plugins/filter/series.py
```diff
@@ -1961,14 +1958,11 @@ def process_propers(self, config, episode, entries):
         """
 
         pass_filter = []
-        best_propers = []
-        # Since eps is sorted by quality then proper_count we always see the highest proper for a quality first.
-        (last_qual, best_proper) = (None, 0)
+        # First find best available proper for each quality without modifying incoming entry order
+        sorted_entries = sorted(entries, key=lambda e: (e['quality'], e['proper_count']), reverse=True)
+        best_propers = {q: next(e) for q, e in itertools.groupby(sorted_entries, key=lambda e: e['quality'])}
```
liiight: god help who has to debug this one liner 😆

gazpachoking: That's why I split it into a 2 liner. 😉

PR #2313, opened 2019-01-27
Remove discover sorting

gazpachoking:
### Motivation for changes:
Stop overriding search order returned from search plugins with discover plugin.

This should probably be released alongside #2312 so the upgrade actions can be combined. #2311 should be released either before, or with this.

### Detailed changes:
- Discover plugin on longer does any sorting of entries.
- Search plugins that were filling search_sort field were all using torrent_availability as the metric. Change all those search plugins to provide that field directly, so user can sort by that if desired.
- Behavior of 'limit' option in discover plugin is now changed. Entry which urlrewrite_search returns may also change. See TODO below.

### Upgrade action needed:
To maintain same behavior of current release, the following can be added to tasks:
```
sort_by:
  field: torrent_availability
  reverse: true
```

### TODO:
- [ ] Many of the search plugins were sorting entries by torrent_availability before returning them. Should this be expanded to all the search plugins that provide it? That would maintain current behavior with regards to `limit` option in discover plugin, and which entry was chosen by urlrewrite_search.

### DIFF
--- a/flexget/plugins/input/discover.py
+++ b/flexget/plugins/input/discover.py
@@ -118,8 +118,7 @@ class Discover(object):
                         continue
                     log.debug('Discovered %s entries from %s', len(search_results), plugin_name)
                     if config.get('limit'):
-                        search_results = sorted(search_results, reverse=True,
-                                                key=lambda x: x.get('search_sort', ''))[:config['limit']]
+                        search_results = search_results[:config['limit']]
                     for e in search_results:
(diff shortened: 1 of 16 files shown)


gazpachoking:
Going forward, all search plugins should do whatever sorting they deem best to give most relevant entries first. Some of them may need to be updated after this pull to keep discover plugin 'limit' option working optimally.

PR #2315, opened 2019-01-28
Remove utils/search.py

paranoidi:
### Motivation for changes:

Avoid core pollution

### Detailed changes:
- Remove utils/search.py and unused code from the core


flexget/plugins/sites/__init__.py
```diff
@@ -0,0 +1,36 @@
+"""
+Utilities for search plugins, please try to avoid adding new ones.
+"""
```
liiight: adding code to __init__.py? not super recommended

flexget/plugins/sites/limetorrents.py
```diff
@@ -1,18 +1,34 @@
...
 from flexget.utils.requests import RequestException
 from flexget.utils.soup import get_soup
-from flexget.utils.search import torrent_availability, clean_symbols
+from flexget.plugins.sites import torrent_availability
 from flexget.utils.tools import parse_filesize
 
 log = logging.getLogger('limetorrents')
 
+
+def clean_symbols(text):
```
liiight: why is this in this plugin and not the main utils module?

flexget/plugins/sites/__init__.py
paranoidi: still better than plugin code in core

flexget/plugins/sites/limetorrents.py
paranoidi: it's only used in this one plugin, I rather keep it that way

PR #2318, opened 2019-01-30
GuessIt Parser -- Initialize a MatchesDict()

XVicarious:
### Motivation for changes:
I didn't read and the previous pull only fixed the first level of issues.

### Detailed changes:
- Instead of checking if its a special kind of dictionary, initialize as a MatchesDict and it works like magic. Passes the tests too.

### DIFF
--- a/flexget/components/parsing/parsers/parser_guessit.py
+++ b/flexget/components/parsing/parsers/parser_guessit.py
@@ -227,7 +227,7 @@ class ParserGuessit(object):
             guess_result = guessit_api.guessit(native(data), options=guessit_options)
         except GuessitException:
             log.warning('Parsing %s with guessit failed. Most likely a unicode error.', data)
-            guess_result = {}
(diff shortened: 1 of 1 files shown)


flexget/plugins/parsers/parser_guessit.py
```diff
@@ -220,7 +220,7 @@ def parse_series(self, data, **kwargs):
             guess_result = guessit_api.guessit(native(data), options=guessit_options)
         except GuessitException:
             log.warning('Parsing %s with guessit failed. Most likely a unicode error.', data)
-            guess_result = {}
+            guess_result = MatchesDict()
```
gazpachoking: Couldn't we just abort here in this case? Something like:
```suggestion
            return SeriesParseResult(data=data, valid=False)
```

gazpachoking:
I think this is good now, we just need to resolve the conflicts before merging.

Issue #2335, opened 2019-02-09
descargas2020 plugin broken (again)


Aokromes:
### Expected behaviour:
it downloads torrent properly.

### Steps to reproduce:
add this rss http://feeds.feedburner.com/descargas2020new
flexget execute
you will get:
2019-02-09 16:00 INFO     download      series_sigo     Downloading: Single Parents (Proper) - Temporada 1  [HDTV 720p][Cap.109][AC3 5.1 Castellano][]
2019-02-09 16:00 ERROR    entry         series_sigo     Failed Single Parents (Proper) - Temporada 1  [HDTV 720p][Cap.109][AC3 5.1 Castellano][] (Unexpected html content received from https://descargas2020.com/descargar/serie-en-hd/single-parents/temporada-1/capitulo-09/ - maybe a login page?)
2019-02-09 16:00 INFO     task          series_sigo     Rerunning the task in case better resolution can be achieved.

#### Config:

```
tasks:
  series_sigo:
    rss:
      url: http://feeds.feedburner.com/descargas2020new
      link: link
    all_series: yes
    seen: local
    regexp:
      reject:
        - PCDVD
    accept_all: yes
    thetvdb_lookup: yes
    no_entries_ok: yes
    set:
      filename: "{{title|pathscrub}}.torrent"
    download: /home/flexget/torrent/
```
  
#### Log:

2019-02-09 16:00 INFO     download      series_sigo     Downloading: Single Parents (Proper) - Temporada 1  [HDTV 720p][Cap.109][AC3 5.1 Castellano][]
2019-02-09 16:00 ERROR    entry         series_sigo     Failed Single Parents (Proper) - Temporada 1  [HDTV 720p][Cap.109][AC3 5.1 Castellano][] (Unexpected html content received from https://descargas2020.com/descargar/serie-en-hd/single-parents/temporada-1/capitulo-09/ - maybe a login page?)
2019-02-09 16:00 INFO     task          series_sigo     Rerunning the task in case better resolution can be achieved.

```
paste log output here
```
</details>

### Additional information:

- FlexGet version: 2.19.5
- Python version: 2.7.13
- Installation method: i don't remember
- Using daemon (yes/no): no
- OS and version: debian 9 x64
- Link to crash log:

daveraver:
[...]

PR #2337, opened 2019-02-12
Fix NoneType on self.requests. Add support to https url.

malomehi:
### Motivation for changes:
- https urls are not currently supported
- self.requests in `parse_download_page` should never be `None`

### Detailed changes:
- Changed regexp to accept both http and https url
- Update `self.requests` with value from `Session()`. This property when used in `parse_download_page` should never be `None`

### Addressed issues:
- Fixes # .

```
paste_config_here
```
```
paste output here
```
#### To Do:

- [ ] Stuff..

### DIFF
--- a/flexget/components/sites/sites/descargas2020.py
+++ b/flexget/components/sites/sites/descargas2020.py
@@ -32,7 +32,7 @@ class UrlRewriteDescargas2020(object):
     # urlrewriter API
     def url_rewritable(self, task, entry):
         url = entry['url']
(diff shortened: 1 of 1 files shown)


Aokromes:
add closes #2335 :) (works for me)

gazpachoking:
Thanks! I cleaned this up a bit more as well, seems there were some problems when we transitioned this from plugins to components. dd616b51cc645fbe59b2f066947af82e70e2114a

PR #2338, opened 2019-02-12
Update manipulate.py re.search

chaosmaker:
re.search only allows for the first hit to be returned. By adding the option to use re.findall multiple matches can be returned to the user using the same method as re.search.

### Motivation for changes:
re.search only allows a single hit to be returned during extract
### Detailed changes:
- add boolean to select findall, and return matches identifi

### Addressed issues:
- Fixes # .

- Feathub #[71](https://feathub.com/Flexget/Flexget/+71).71 

```
paste_config_here
added boolean for findall: true
```
```
paste output here
```
#### To Do:

- [ ] Stuff..


flexget/plugins/modify/manipulate.py
```diff
@@ -117,12 +119,18 @@ def process(self, entry, jobs):
...
                         log.warning('Cannot extract, field `%s` is not present' % from_field)
                         continue
-                    match = re.search(config['extract'], field_value, re.I | re.U)
-                    if match:
-                        groups = [x for x in match.groups() if x is not None]
-                        log.debug('groups: %s' % groups)
-                        field_value = config.get('separator', ' ').join(groups).strip()
+                    if config.get('findall'):
+                        match = re.findall(config['extract'], field_value)
+                        log.debug('all matches: %s' % match)
```
cvium: Use logger string interpolation
```suggestion
                        log.debug('all matches: %s', match)
```

flexget/plugins/modify/manipulate.py
```diff
@@ -117,12 +119,18 @@ def process(self, entry, jobs):
...
+                        field_value = config.get('separator', ' ').join(match).strip()
                         log.debug('field `%s` after extract: `%s`' % (field, field_value))
+                    else:
+                        match = re.search(config['extract'], field_value, re.I | re.U)
+                        if match:
+                            groups = [x for x in match.groups() if x is not None]
+                            log.debug('groups: %s' % groups)
```
cvium:
```suggestion
                            log.debug('groups: %s', groups)
```

flexget/plugins/modify/manipulate.py
chaosmaker: I guess then it might be worth changing all log.debug in this file to that notation then. As all the others use '%' hence why I used it there.

chaosmaker:
I've altered all instances in manipulate.py with the minor corrections that @cvium provided.

gazpachoking:
I'm good with this, should we put an underscore in the option though? `find_all`?

gazpachoking:
Also, it looks like the behavior of only capturing groups is different depending on whether findall is specified. Shouldn't that stay consistent?

gazpachoking:
In fact, it looks like this new config,
```
findall: yes
extract: anything
```
would be equivalent to this already possible config:
```
extract: "(?:(anything).*)+"
```
Is that true? I agree the syntax on that already possible one is more complicated, just trying to figure out the best/simplest way to allow this feature.

chaosmaker:
I'm not 100% sure but I just had a quick try and my regex that I use to capture links, only finds the first instance when I do this:
(?:href\s*=\s*\"((?:[\w-]+://?|www[.])[^\s()<>]+(?:\([\w\d]+\)|[^[:punct:]\s]|))\".*)+

While findall using this:
href\s*=\s*\"((?:[\w-]+://?|www[.])[^\s()<>]+(?:\([\w\d]+\)|[^[:punct:]\s]|))\"
Does find all. But then maybe my regex is wrong haha.

Personally having the option of specifying it this way is easier but happy to be wrong.

gazpachoking:
Aha, seems like python's `re` module doesn't support repeated capture groups. Well, I'm on board for this then, my 2 questions still remain:
1. Should the behavior stay consistent with regular extract, and only extract capture groups?
2. Should we call it find_all rather than findall?

chaosmaker:
> Aha, seems like python's `re` module doesn't support repeated capture groups. Well, I'm on board for this then, my 2 questions still remain:
> 
> 1. Should the behavior stay consistent with regular extract, and only extract capture groups?
> 2. Should we call it find_all rather than findall?
> 
> (I'm leaning towards 'yes' on both of these questions.)

I'm not 100% sure what you mean by 1. In my understanding findall should only extract capture groups, but if they occur more than once they will all be captured. 
I have no issue naming it find_all if you think that this would make it easier to understand what it does. 
Once pulled the documentation needs to be changed online too.

chaosmaker:
I didn't realise this was never finished. Can we finalise this pull please?

flexget/plugins/modify/manipulate.py
```diff
@@ -120,27 +122,31 @@ def process(self, entry, jobs):
...
                 if config.get('remove'):
                     if field in entry:
                         del entry[field]
                         modified = True
                     continue
-
                 if 'extract' in config:
                     if not field_value:
-                        logger.warning('Cannot extract, field `{}` is not present', from_field)
+                        logger.warning('Cannot extract, field `%s` is not present', from_field)
```
liiight, line 132: we use [loguru](https://github.com/Delgan/loguru) as our logger, so please retain the usage of `{}` for logging

chaosmaker: Apologies I didn't notice! They should all be in the loguru style now.

chaosmaker:
@cvium @liiight @gazpachoking  Any chance to merge this? I think all the requested changes have been applied?

gazpachoking:
@chaosmaker Merged, thanks! Can you update the wiki page to document this option?

PR #2344, opened 2019-02-19
Allow `set` plugin can now set entry fields to types other than str

gazpachoking:
### Motivation for changes:
If you want to manipulate a number field it is not currently possible.

### Detailed changes:
- If set plugin templates evaluate to a type other than string, the type is not cast to a string.

### Addressed issues:
- https://discuss.flexget.com/t/content-size-for-magnet-links-from-xml-feed/4572

```
set:
  some_field: {{some_int_field/2}}
```
#### To Do:

- [x] Should a 'native' parameter be added to the Entry.render method rather than constructing the NativeTemplate in set plugin?

### DIFF
--- a/flexget/plugins/modify/set_field.py
+++ b/flexget/plugins/modify/set_field.py
@@ -1,13 +1,12 @@
 from __future__ import unicode_literals, division, absolute_import
 from builtins import *  # noqa pylint: disable=unused-import, redefined-builtin
-from past.builtins import basestring
 
 import logging
 from functools import partial
 
 from flexget import plugin
 from flexget.event import event
-from flexget.utils.template import RenderError, FlexGetNativeTemplate
+from flexget.utils.template import RenderError
 
 log = logging.getLogger('set')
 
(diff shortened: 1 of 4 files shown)


gazpachoking:
So, I'm generally a fan of this 'native' rendering for the way we are using jinja templates. I think it allows more user manipulation of entry fields with the only cost being quotes sometimes needed if you explicitly need a string. @paranoidi @liiight @cvium Anything I'm overlooking?

flexget/plugins/modify/set_field.py
```diff
@@ -34,7 +33,7 @@ def modify(self, entry, config, errors=True):
         """This can be called from a plugin to add set values to an entry"""
         for field in config:
             # If this doesn't appear to be a jinja template, just set it right away.
-            if not isinstance(config[field], basestring) or '{' not in config[field]:
+            if not isinstance(config[field], str) or '{' not in config[field]:
```
paranoidi: Doesn't this remove unicode support? Why swtich from basestring -> str, we are not yet python3? :)

cvium: It's a redundant import actually. Because of `from builtins import *`, `str` is equivalent to `basestring`.

Source: https://python-future.org/compatible_idioms.html#basestring

gazpachoking: Actually, 'str' is equivalent to 'unicode' on python 2 (because of the `from builtins import *`.) We don't actually allow byte strings in any of these places.

gazpachoking:
@liiight Yeah, good plan, added.

PR #2345, opened 2019-02-27
added support for dd+5.1 quality to guessit parser

manacker:
### Motivation for changes:
The guessit parsing plugin doesn't recognize the "dd+5.1" audio quality, this pull request fixes that.

### Detailed changes:
- when guessit audio_channels == "5.1" and guessit audio_codec == "Dolby Digital Plus" return audio quality "dd+5.1"

### DIFF
--- a/flexget/components/parsing/parsers/parser_guessit.py
+++ b/flexget/components/parsing/parsers/parser_guessit.py
@@ -157,7 +157,9 @@ class ParserGuessit(object):
         # unlike the other components, audio can be a bit iffy with multiple codecs, so we limit it to one
         if 'dts' in audio and any(hd in audio_profile for hd in ['hd', 'master audio']):
             audio = ['dtshd']
-        elif '5.1' in audio_channels and any(dd in audio for dd in ['dolby digital']):
+        elif '5.1' in audio_channels and 'dolby digital plus' in audio:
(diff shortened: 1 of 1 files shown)


flexget/components/parsing/parsers/parser_guessit.py
```diff
@@ -157,6 +157,8 @@ def _quality(self, guessit_result):
         # unlike the other components, audio can be a bit iffy with multiple codecs, so we limit it to one
         if 'dts' in audio and any(hd in audio_profile for hd in ['hd', 'master audio']):
             audio = ['dtshd']
+        elif '5.1' in audio_channels and any(dd in audio for dd in ['dolby digital plus']):
```
gazpachoking: We don't need the 'any' here do we?
```suggestion
        elif '5.1' in audio_channels and 'dolby digital plus' in audio:
```

manacker: To be honest, I just duplicated the following lines, which don't need an 'any' either (anymore). I amended my PR.

gazpachoking: Gotcha. Well, looks good to me now, thanks!

gazpachoking:
Looks good, thanks!

PR #2346, opened 2019-03-03
add support for popular/trending list on trakt

heikobornholdt:
### Motivation for changes:
I need that feature for myself.

### Detailed changes:
Add support for `trending`/`popular` list for `trakt_list` plugin.

- Feathub #[40](https://feathub.com/Flexget/Flexget/+40).

### Config usage:
```yml
trakt_list:
  username: <trakt username>
  list: popular # new values: trending, popular
  type: shows  # allowed values for trending/popular list: movies, shows
```

### DIFF
--- a/flexget/components/trakt/trakt_list.py
+++ b/flexget/components/trakt/trakt_list.py
@@ -17,7 +17,7 @@ from . import db
 
 
 log = logging.getLogger('trakt_list')
-IMMUTABLE_LISTS = []
+IMMUTABLE_LISTS = ['trending', 'popular']
 
 
 def generate_show_title(item):
--- a/flexget/components/trakt/trakt_list.py
+++ b/flexget/components/trakt/trakt_list.py
@@ -182,7 +182,7 @@ class TraktSet(MutableSet):
     @property
     def items(self):
         if self._items is None:
-            if self.config['list'] in ['collection', 'watched'] and self.config['type'] == 'auto':
+            if self.config['list'] in ['collection', 'watched', 'trending', 'popular'] and self.config['type'] == 'auto':
                 raise plugin.PluginError(
                     '`type` cannot be `auto` for %s list.' % self.config['list']
                 )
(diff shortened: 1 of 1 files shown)


gazpachoking:
Looks good to me. Technically we don't need username/account in the config for these, but I think it might complicated the schema more than it's worth to fix that.

gazpachoking:
On second thought, it might not be too bad. I think updating the current 'anyOf' in the config would release the requirement for username for these lists:
```
        'anyOf': [
            {'anyOf': [{'required': ['username']}, {'required': ['account']}]},
            # The 'trending' and 'popular' lists don't require an username
            {'properties': {'list': {'enum': ['trending', 'popular']}}}
        ],
```
I haven't tested though, not sure if there are any other implications.

gazpachoking:
Oh, one more thing. `IMMUTABLE_LISTS` should be updated to have 'trending' and 'popular' in there, so that a good error message is produced if an user tries to edit them.

heikobornholdt:
Thanks for the feedback. I've updated the PR.

gazpachoking:
Excellent, thanks! Can you update the wiki with the new options as well?

snickers2k:
awesome! only thing i'm "missing" is a limiter. How is this solved? How many movies are "trending/popular" ?

snickers2k:
okay.. again, why no limiiter ? Now i have 155k ! 155 069 items in my list .. thanks for that

heikobornholdt:
Sorry to hear that. I agree that we need to limit the number of entries.

I am not sure that we should implement a limiter only for trakt or as global feature for any kind of lists.
@gazpachoking What do you think about that?

gazpachoking:
Yeah, guess a limiter is needed if those lists are so big. Not sure the best way to go. We could make a plugin similar to limit_new which acts earlier and just cuts off the entries at a certain number (limit_new isn't) appropriate for this case since it doesn't act until after filtering, and adds supplemental entries to the backlog.) This wouldn't be optimal for plugins that have to scan through multiple pages for results, as they'd do extra work if they were getting results past those limits. In that case the only good option I see would be to add limit to the plugin itself.

Easiest way to go is probably just add a limit option to trakt_list. Maybe the default should be dynamic, and set to unlimited for regular lists, but limit popular and trending to something sane by default.

gazpachoking:
Limit option has been added now in #2366

snickers2k:
thank you for taking the time!

btw. trakt support was really quick to answer my email back and deleted the watchlist for me. next time i use a custom-watchlist for testing i guess :D

PR #2354, opened 2019-03-24
Cronitor plugin

liiight:
### Motivation for changes:
Wanted an uptime monitor. Cronitor is simple and free
### Detailed changes:
- Added cronitor plugin

```yaml
cronitor: ABC123
```
Or:
```yaml
cronitor:
  monitor_code: ABC123
  on_start: yes
  on_abort: no
  auth_key: FOOBAR
```

### DIFF
--- a/flexget/components/notify/notifiers/cronitor.py
+++ b/flexget/components/notify/notifiers/cronitor.py
@@ -45,8 +45,10 @@ class Cronitor(object):
 
     def __init__(self):
         self.config = None
+        self.task_name = None
 
-    def prepare_config(self, config):
+    @staticmethod
+    def prepare_config(config):
         if isinstance(config, str):
             config = {"monitor_code": config}
         config.setdefault("on_start", True)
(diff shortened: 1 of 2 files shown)


requirements.txt
```diff
@@ -21,7 +21,6 @@ flask-login==0.4.0
 flask-restful==0.3.6
 flask-restplus==0.10.1
 flask==1.0.2
-future==0.16.0
```
cvium: Hmm?

liiight: Slippage

flexget/components/notify/notifiers/cronitor.py
```diff
@@ -0,0 +1,91 @@
+from builtins import *
```
cvium: Missing future? Absolute import etc.

flexget/components/notify/notifiers/cronitor.py
```diff
@@ -0,0 +1,91 @@
...
+                },
+                "required": ["monitor_code"],
+                "additionalProperties": False,
+            },
+            {"type": "string"},
+        ]
+    }
+
+    def __init__(self):
+        self.config = None
```
paranoidi: This is not good practice as it leaves (possibly) behind all kinds of objects after plugin execution has finished.

liiight: it's either this or prepare config and pass it for every task execution phase. any recommendation?

flexget/components/notify/notifiers/cronitor.py
liiight: not sure what you mean

gazpachoking: `from __future__ import unicode_literals...` I think he means.

requirements.txt
```diff
@@ -30,7 +30,6 @@ itsdangerous==0.24        # via flask
 jinja2==2.10
 jsonschema==2.6.0
 markupsafe==1.0           # via jinja2
-path.py==11.5.0
```
gazpachoking: Do we not need path.py anymore? Did this leak over from python 3 only branch?

liiight: yeah...

PR #2367, opened 2019-04-16
[add] types to kitsu input

vxcamiloxv:
### Motivation for changes:
Kitsu input doesn't support the types from the API, so if someone needs to add the series in one list and the movies in another one are pretty hard, this adds the `type` option similar to `my_anime_list` input.

### Detailed changes:
- Adds `type` option to filter the results

### Documentation of Options:
- Types strings[]: 
  - type.ona "ONA"
  - type.ova "OVA"
  - type.TV "TV"
  - type.movie "movie"
  - type.music "music"
  - type.special "special"


gazpachoking:
Looks good to me, thanks!

PR #2368, opened 2019-04-16
Remove '...' from text in sites.utils.normalize_scene()

andrey-yantsen:
### Motivation for changes:
Currently it's not possible to find show `Travel Man: 48 Hours In...` using discovery+rarbg, because of the ending dots. I'm not sure where is better to strip them, so I've decided to do it globally in `NextSeriesEpisodes` plugin.

### Detailed changes:
- if the show name contains `...` — the title without them would be added to `search_strings`

### DIFF
--- a/flexget/components/sites/utils.py
+++ b/flexget/components/sites/utils.py
@@ -29,8 +29,9 @@ def normalize_scene(text):
     #     ABCDEFGHIJKLMNOPQRSTUVWXYZ
     #     abcdefghijklmnopqrstuvwxyz
     #     0123456789-._()
+    text = normalize('NFKD', text).encode('ASCII', 'ignore').decode()
     return re.sub(
-        r'[^a-zA-Z0-9 \-._()]', '', normalize('NFKD', text).encode('ASCII', 'ignore').decode()
+        r'[^a-zA-Z0-9 \-._()]', '', text.replace('...', '')
     )
 
 


flexget/components/series/next_series_episodes.py
```diff
@@ -55,6 +55,8 @@ def search_entry(self, series, season, episode, task, rerun=True):
         paren_match = re.match(r'(.+?)( \(.+\))?$', series.name)
         if paren_match.group(2):
             alts.append(paren_match.group(1))
+        if '...' in series.name and series.name.replace('...') not in alts:
+            alts.append(series.name.replace('...'))
```
gazpachoking: Looks like you are missing the blank argument here in these replace statements. i.e. `...replace('...', '')`

andrey-yantsen: Omg, sorry! Fixed now :)

andrey-yantsen:
Yeah, I'm not sure about the proper location as well. May be it'd be better to place dots removal in rarbg plugin... I mean just remove all the dots before starting the search. What do you thing?

gazpachoking:
Yeah, I was thinking that might be more appropriate as well. This might actually be a good spot https://github.com/Flexget/Flexget/blob/1400d84aee8c4f45b36ce109f870bb3e8c5a836e/flexget/components/sites/utils.py#L22
I believe rarbg plugin is already using that to normalize the search names before doing the search, and other search plugins can use it as well.

andrey-yantsen:
Yeah, sounds like a good idea, thanks! I've moved the replacement there.

gazpachoking:
Seems good to me now, thanks!

PR #2369, opened 2019-04-22
Allow slack buttons/actions

lingfish:
### Motivation for changes:
As per Feathub entry, buttons would be useful.

### Detailed changes:
- Extend schema to allow actions

- Feathub #[75](https://feathub.com/Flexget/Flexget/+75).

```
              attachments:
...
                  callback_id: "wopr_game"
                  actions:
                  - name: game
                    text: Chess
                    type: button
                    value: chess
```

#### To Do:

- [ ] dependencies in the schema don't seem to be working/done right by me; assistance here would be appreciated


gazpachoking:
I think to fix your 'dependencies' in the schema it needs to be at the same level as the `'required': ['fallback'],` line. Right now it's a couple levels too far out.

PR #2372, opened 2019-04-24
[wip] Selenium plugin for advanced website interactions

thawn:
### Motivation for changes:
some sites don't play well with pythons requests. For example, cloudflare implements a ddos protection on some sites that require a real browser with javascript to circumvent. Therefore, this plugin uses selenium to automate a real firefox browser and interact with these sites. For now, the plugin supports cloudflare ddos circumvention, imdb login and general site login.
After successful interaction with the website, the session cookies are copied to the task.requests Session object. Therefore, no changes to other plugins are needed to make this work. Tested successfully with the rlsbb and rmz plugins that stopped working because of cloudflare ddos protection.

### Detailed changes:
The selenium plugin supports three actions:
1. `cloudflare_ddos`: circumvent cloudflare protection. Parameters:
    * 'url`: url to page that is protected by cloudflare
2. `imdb_login`: log in to imdb. So that the rest of the task can be performed as logged in user. Parameters: 
    * `username`: email used for login, 
    * `password`: password used for login.
3. `login`: log in to arbirtrary page. Parameters: 
    * `url`: url to login page, 
    * `username`: user name (or email) used for login,
    * `password`: password used for login,
    * `input_username_id`: id tag of field where username (or email) is to be entered,
    * `input_password_id`: id tag of field where password is to be entered,
    * `login_button_id`: id tag of button that must be clicked in order to log in,
    * `element_after_login_id`: id of an element that appears on the page only after login, used to wait for successful loading of the page after login.
 

### Addressed issues:
- Fixes problems with accessing pages that are protected by cloudflare ddos protection.

- With additional work on imdb_list plugin, this may help circumvent problems with imdb asking for captchas during login.
- Could help fix problems with pages that require javascript during login

**Configuration example for cloudflare DDOS protection:**
```yaml
selenium:
  action: cloudflare_ddos
  parameters:
    url: http://example.com
```

**Configuration example for imdb login**
```yaml
selenium:
  action: imdb_login
  parameters:
    username: me@example.com
    password: YouWon'tGuess
```

**Configuration example for general login**
```yaml
selenium:
  action: login
  parameters:
    url: http://example.com/login
    username: MyName
    password: YouWon'tGuess
    input_username_id: username_input_field_id
    input_password_id: password_input_field_id
    login_button_id: button_id
    element_after_login_id: logout_button_id
```
```
log.error('Error importing selenium: %s' % e)
log.debug('The selenium plugin does not support this action: %s. Allowed values for the "action" config parameter are: %s' % config['action'], self.implemented_actions)
log.debug('Added cookie: %s' % pformat(cookie))
log.verbose('Passed cloudflare DDOS protection for url: "%s". Page title is now: %s', parameters['url'], self.driver.title)
log.warning('Could not circumvent cloudflare protection for url: "%s". Page title is now: %s',parameters['url'], self.driver.title)
log.debug('Cookies now stored in task.requests.cookies: %s' % pformat(task.requests.cookies))
log.warning('Could not verify the presence of an element with the id "%s". Login may have failed.' %parameters['element_after_login_id'])
log.verbose('Logged in at url: "%s". Page title is now: %s', parameters['url'], self.driver.title)
```
#### To Do:

- [ ] Adapt imdb_list to use selenium.
- [ ] In case this plugin is accepted, write a documentation on the wiki (mostly copy an d paste from the class documentation within the code).

### DIFF
--- a/flexget/plugins/operate/selenium.py
+++ b/flexget/plugins/operate/selenium.py
@@ -36,9 +36,19 @@ class Selenium(object):
     ## Configuration
 
     Currently, the following actions are supported:
-    1. `cloudflare_ddos`: circumvent cloudflare protection. Parameters: 'url`: url to page that is protected by cloudflare
-    2. `imdb_login`: log in to imdb. So that the rest of the task can be performed as logged in user. Parameters: `username`: email used for login, `password`: password used for login.
-    3. `login`: log in to arbirtrary page. Parameters: `url`: url to login page, `username`: user name (or email) used for login, `password`: password used for login, `input_username_id`: id tag of field where username (or email) is to be entered, `input_password_id`: id tag of field where password is to be entered, `login_button_id`: id tag of button that must be clicked in order to log in, `element_after_login_id`: id of an element that appears on the page only after login, used to wait for successful loading of the page after login.
+    1. `cloudflare_ddos`: circumvent cloudflare protection. Parameters:
+        * 'url`: url to page that is protected by cloudflare
+    2. `imdb_login`: log in to imdb. So that the rest of the task can be performed as logged in user. Parameters: 
+        * `username`: email used for login, 
+        * `password`: password used for login.
+    3. `login`: log in to arbirtrary page. Parameters: 
+        * `url`: url to login page, 
+        * `username`: user name (or email) used for login,
+        * `password`: password used for login,
+        * `input_username_id`: id tag of field where username (or email) is to be entered,
+        * `input_password_id`: id tag of field where password is to be entered,
+        * `login_button_id`: id tag of button that must be clicked in order to log in,
+        * `element_after_login_id`: id of an element that appears on the page only after login, used to wait for successful loading of the page after login.
 
     **Configuration example for cloudflare DDOS protection:**
     ```yaml
@@ -114,7 +124,7 @@ class Selenium(object):
     '''
     schema = {
         'type': 'object',
-        'properties': {'chromedriver': {'type': 'string'}, 'action': {'type': 'string'}, 'parameters': {'type': 'object'}},
+        'properties': {'action': {'type': 'string'}, 'parameters': {'type': 'object'}},
         'additionalProperties': False,
     }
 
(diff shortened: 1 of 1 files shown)


paranoidi:
This is pretty wild and quite heavy way to go, I would preferably rely on simplier headless V8 engines where possible .. 

As for the implementation, the way this is implemented is not quite how we do plugins. To be more exact, let's think "action: imdb_login" .. what parameters does it take? The schema can not define those as parameters is shared with other actions ...

The better way would be to make a selenium component-plugin that has sub-plugins which implement interface "selenium", check `components/parsing/´ and other components. Then each of those can define their own schema. This would produce configuration like:

```
selenium:
  imdb_login:
    username: foo
    password: bar
```

Roughly in a directory structure:

../components/selenium/selenium.py
../components/selenium/imdb_login.py
../components/selenium/login.py
../components/selenium/cloudflare_ddos

As plugin names are in a global namespace those sub-plugin names would need a bit thinking so that they don't cause problems down the line .. as the selenium plugin does not provide any configuration for subplugins it might be just better to make them individual plugins in there and prefix with "selenium", ie. "selenium_imdb_login, selenium_login, selenium_cloudflare_ddos".

thawn:
@paranoidi Thank you very much for taking the time to analyze my pull request. 

_This is pretty wild and quite heavy way to go, I would preferably rely on simplier headless V8 engines where possible .._

The problem with the cloudflare ddos protection is that they are quite keen on making it difficult to circumvent it in an automated way. I did try simpler engines before but nothing worked. Even selenium with chromium headless does not always work. Therefore, I finally ended up with selenium and headless firefox, which works. Thus this plugin is meant as a last resort for accessing websites that won't let you automate them in any other way.

_As for the implementation, the way this is implemented is not quite how we do plugins._

Thanks a lot for your detailed advice for refactoring the plugin such that its configuration is more consistent with other plugins. I will get to work on that as soon as I find the time.


I have a question with regard to the selenium_imdb_login plugin: I noticed, that IMDB does not require captchas when I am using my regular browser or the selenium plugin. Therefore, I added the imdb_login function to the selenium plugin. This is meant to help the imdb_list plugin obtain updated login cookies. 
In principle the selenium_imdb_login plugin would only have to be run if the imdb_list plugin needs fresh login cookies. What would be the best way to implement this?

crawln45:
For this:

> I have a question with regard to the selenium_imdb_login plugin: I noticed, that IMDB does not require captchas when I am using my regular browser or the selenium plugin. Therefore, I added the imdb_login function to the selenium plugin. This is meant to help the imdb_list plugin obtain updated login cookies.
In principle the selenium_imdb_login plugin would only have to be run if the imdb_list plugin needs fresh login cookies. What would be the best way to implement this?
> 
I would assume we would do this the same way we do with the trakt plugin and assign the username/password not in the config but as a configuration of the plugin. This way we can call this plugin in the imdb_list plugin to allow imdb_list to call the function/w username/pass as needed to validate and store the cookie for authentication. Is how I would see that working the best.

PR #2388, opened 2019-05-09
Do not request more pages than required in trakt_list

andrey-yantsen:
### Motivation for changes:

ATM list `popular` returns 160k items, and we request them all, even if there is a `limit` set in config. See #2386 for details.

### Detailed changes:
- Combined code responsible for pagination and the actual entries processing
- Pagination processing stops as soon as we reach the limit of required entries
- No more extra request to detect whether pagination is required

### Addressed issues:
- Fixes #2386

### Config usage if relevant (new plugin or updated schema):
```
tasks:
    task:
        trakt_list:
            list: popular
            type: movies
            limit: 10
```

#### To Do:

- [x] Set default limit to something
- [ ] Update wiki

Issue #2386, opened 2019-05-08
trakt_list pulls all entries in list before limiting


jawilson:
### Expected behaviour:

When setting `limit` in the `trakt_list` plugin, Flexget should only fetch up to that many items from Trakt.

### Actual behaviour:
The plugin grabs _all_ items (I've seen upwards of 160k in the "popular" list) in the list before filtering all but `limit` items.

### Steps to reproduce:
# Add the following to a task:
```yaml
    discover:
      what:
        - trakt_list:
            list: popular
            type: movies
            limit: 50
```
# Run the task with debug logging and you will see it can take over 5 minutes to pull all the items

#### Config:
https://github.com/jawilson/flexget-config
  
#### Log:
<details>
 <summary>(click to expand)</summary>

```
2019-05-08 14:34 VERBOSE  trakt_list    Auto-Movies     Retrieving `movies` list `popular`
2019-05-08 14:34 DEBUG    utils.requests Auto-Movies     GETing URL https://api.trakt.tv/movies/popular with args () and kwargs {'allow_redirects': True, u'timeout': 30}
2019-05-08 14:34 DEBUG    trakt_list    Auto-Movies     Response is paginated. Number of items: 160121, number of pages: 161.0
2019-05-08 14:34 DEBUG    utils.requests Auto-Movies     Waiting 1.91 seconds until next request to trakt.tv
2019-05-08 14:34 DEBUG    utils.requests Auto-Movies     GETing URL https://api.trakt.tv/movies/popular with args () and kwargs {'allow_redirects': True, 'params': {u'limit': 1000, u'page': 1}, u'timeout': 30}
2019-05-08 14:34 DEBUG    utils.requests Auto-Movies     Waiting 1.95 seconds until next request to trakt.tv
2019-05-08 14:34 DEBUG    utils.requests Auto-Movies     GETing URL https://api.trakt.tv/movies/popular with args () and kwargs {'allow_redirects': True, 'params': {u'limit': 1000, u'page': 2}, u'timeout': 30}
2019-05-08 14:34 DEBUG    utils.requests Auto-Movies     Waiting 1.96 seconds until next request to trakt.tv
2019-05-08 14:34 DEBUG    utils.requests Auto-Movies     GETing URL https://api.trakt.tv/movies/popular with args () and kwargs {'allow_redirects': True, 'params': {u'limit': 1000, u'page': 3}, u'timeout': 30}
[...]

PR #2390, opened 2019-05-10
Limit plugin

gazpachoking:
### Motivation for changes:
We needed a way to limit entries created by trakt_list plugin #2386, #2388. We added a limit option to that plugin. This is an experiment to see if we could use a more generic approach to limit any input plugin, while still allowing the optimization of not having the input plugin request more than is necessary from its source. (i.e. limiting HTTP or database requests)

### Detailed changes:
- Allows input plugins to be generators
- Adds 'limit' plugin, which takes a limit amount, as well as the configuration for another input plugin

```
limit:
  amount: 5
  from:
    rss: http://blah.com
```

#### To Do:

- [x] Verify whether we like this idea
- [x] Convert input plugins which might take a long time to get many entries to generators
- [x] Deprecate limit options from plugins that implemented themselves (T411, Letterboxd)
- [x] Figure out how to best handle `@cached` decorator with generator inputs

### DIFF
--- a/flexget/plugins/input/limit.py
+++ b/flexget/plugins/input/limit.py
@@ -10,7 +10,7 @@ from flexget.event import event
 log = logging.getLogger('limit')
 
 
-class PluginInputs(object):
+class PluginLimit(object):
     """
     Limits the number of entries an input plugin can produce.
     """
@@ -52,4 +52,4 @@ class PluginInputs(object):
 
 @event('plugin.register')
 def register_plugin():
-    plugin.register(PluginInputs, 'limit', api_ver=2)
+    plugin.register(PluginLimit, 'limit', api_ver=2)
--- a/flexget/plugins/input/limit.py
+++ b/flexget/plugins/input/limit.py
@@ -1,0 +1,55 @@
+from __future__ import unicode_literals, division, absolute_import
+from builtins import *  # noqa pylint: disable=unused-import, redefined-builtin
+
+import itertools
+import logging
+
+from flexget import plugin
+from flexget.event import event
+
+log = logging.getLogger('limit')
+
+
+class PluginInputs(object):
+    """
+    Limits the number of entries an input plugin can produce.
+    """
+
+    schema = {
+        'type': 'object',
+        'properties': {
+            'amount': {'type': 'integer', 'minimum': -1},
+            'plugin': {
+                'allOf': [
+                    {'$ref': '/schema/plugins?phase=input'},
+                    {
+                        'maxProperties': 1,
+                        'error_maxProperties': 'Plugin options within limit plugin must be indented 2 more spaces than '
+                        'the first letter of the plugin name.',
+                        'minProperties': 1,
+                    },
+                ]
+            },
+        },
+        'required': ['amount', 'plugin'],
+        'additionalProperties': False,
+    }
+
+    def on_task_input(self, task, config):
+        for input_name, input_config in config['plugin'].items():
+            input = plugin.get_plugin_by_name(input_name)
+            method = input.phase_handlers['input']
+            try:
+                result = method(task, input_config)
+            except plugin.PluginError as e:
+                log.warning('Error during input plugin %s: %s' % (input_name, e))
+                continue
+            # A 0 or -1 limit means don't limit.
+            if config['amount'] < 1:
+                return result
+            return itertools.islice(result, config['amount'])
+
+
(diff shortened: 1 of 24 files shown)


flexget/plugins/input/limit.py
```diff
@@ -0,0 +1,55 @@
...
+class PluginInputs(object):
+    """
+    Limits the number of entries an input plugin can produce.
+    """
+
+    schema = {
+        'type': 'object',
+        'properties': {
+            'amount': {'type': 'integer', 'minimum': -1},
+            'plugin': {
```
paranoidi: Change to "from" to be in line with discover etc

gazpachoking: Oh, good idea, that's much nicer.

gazpachoking:
Looks like we like this, so I fixed up a few more things:
- Fixed the key for the subplugin to be called 'from'
- Switched over any input plugins I saw that were fetching successive pages to be generators
- Converted 'inputs' and 'sequence' plugins to both pass through generators

gazpachoking:
@paranoidi @andrey-yantsen So, now the question is what to do with trakt_list plugin. Do we keep the limit option in there, or remove that again in favor of this? Do we keep the default limit for trakt_list, or maybe have a log message that recommends using this plugin if the number of entries is bigger than a certain amount?

andrey-yantsen:
I suggeest removing the limit from trakt_list, and adding a check to an "upper" level, and if any input plugin except `limit` produces more than, let's say, 1000 items — send a notification to the log. Therefore we'll get the code with less cohesion level, where input plugins just produce items and don't care about amounts.

andrey-yantsen:
BTW, feel free to steal my modified trakt_list from #2388, and convert it to the generator :)

andrey-yantsen:
Shouldn't we deprecate `max_results` from `Letterboxd` and `T411` as well?

gazpachoking:
@liiight limit_new is a bit different, that just limits the amount of accepted entries per run, and saves the rest for later runs

@andrey-yantsen You are right, those should probably be deprecated with this

gazpachoking:
Hmm, things to figure out:
- What the hell changed about trakt_list_add test that it doesn't work anymore.
- How to handle our `@cached` input plugin decorator with generator style inputs. In memory persistence should be really easy, we can just use `itertools.tee` and store one copy of the resultant generator in cache. DB persistence on the other hand....

andrey-yantsen:
> What the hell changed about trakt_list_add test that it doesn't work anymore.

Basically, you've fixed `TraktSet.clear()` method and now tests are making one request less :) previously it was populating items as the first step of the test, when the object is empty. Check out my latest commit to #2388.

gazpachoking:
> > What the hell changed about trakt_list_add test that it doesn't work anymore.
> 
> Basically, you've fixed `TraktSet.clear()` method and now tests are making one request less :) previously it was populating items as the first step of the test, when the object is empty. Check out my latest commit to #2388.

Oh awesome, thanks for the investigation. :)

gazpachoking:
@andrey-yantsen Is t411 a plugin anymore? I don't see it anywhere.

andrey-yantsen:
Probably no - by some reason I’ve searched in wiki, not in github.

cvium:
t411 as a site does not exist anymore

andrey-yantsen:
from_digest plugin has a limit as well: https://github.com/Flexget/Flexget/blob/a4037da07f7d7707650596670c467ab3b7e748c9/flexget/plugins/operate/digest.py#L100

gazpachoking:
Okay, I've modified the input caching to work with these new generator inputs. I think it's probably fine, but some feedback would be appreciated. Basic gist of how it works:
- Entries don't get cached to memory until they are iterated over. The memory cache also stores the generator though, so another task can iterate farther if the first task limited the input.
- Entries don't get cached to database at all, unless every entry is iterated over, since we can't cache the partially exhausted generator to db along with the partial entry list.

In practice, none of this probably matters, it's only protecting against the case of 2 tasks using the same input, but limited to a different quantity of entries (or increasing the limit number between runs while entries are cached to db.)

gazpachoking:
Well, I haven't thought of any better ways to handle `@cached` decorator, so I've become a bit more confident on the solution I implemented. I think we are probably good to go on this PR unless anyone else has something.

gazpachoking:
Going with it.

PR #2394, opened 2019-05-19
[add] from_transmission: Add 'transmission_error_state'

Karlson2k:
### Motivation for changes:
Cleanup of torrents may remove finished, but not moved to final location torrents.

### Detailed changes:
I use dedicated miniserver for downloading new torrents. Miniserver uses local storage for unfinished torrents, but move fully downloaded torrents to main NAS. Sometimes torrent cannot be moved to final location due to various reasons. Currently I cannot distiguish between completed and moved to final destination torrents and finished but failed to move to proper location torrent.
This change add `transmission_error_state` attribute with human-readable status. In particular,  error status `local_error` means that files cannot be created locally or cannot be moved to final location.

### DIFF
--- a/flexget/plugins/clients/transmission.py
+++ b/flexget/plugins/clients/transmission.py
@@ -240,6 +240,13 @@ class PluginTransmissionInput(TransmissionBase):
             entry['transmission_trackers'] = [t['announce'] for t in torrent.trackers]
             entry['transmission_seed_ratio_ok'] = seed_ratio_ok
             entry['transmission_idle_limit_ok'] = idle_limit_ok
+            st_error_to_desc = {
+                0: 'OK',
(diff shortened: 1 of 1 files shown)


gazpachoking:
I think transmission provides this directly, which would be preferable to duplicating the mapping try the 'errorString' property.

Karlson2k:
@gazpachoking I checked only with "OK" error state, "errorString" was empty.
By checking sources of transmission and other clients, errorString have description for some specific error, like "permission denied for /var/local/lib/flexget/incoming", while "error" contains type of error.
For automation proposes. type of error is much more useful than specific error description.

gazpachoking:
Gotcha, looks good to me then.

PR #2395, opened 2019-05-19
[add] from_transmission: Add human-readable 'status_desc'

Karlson2k:
### Motivation for changes:
Use more user-friendly textual torrent statuses instead of numbers.

### Detailed changes:
This PR add `'transmission_status_desc'` for entries from transmission. While maintaining backward compatibility with existing configurations by still using numeric `'transmission_status'`, new field greatly improve readability of flexget configuration.

### DIFF
--- a/flexget/plugins/clients/transmission.py
+++ b/flexget/plugins/clients/transmission.py
@@ -237,6 +237,16 @@ class PluginTransmissionInput(TransmissionBase):
                     log.debug(
                         'error when requesting transmissionrpc attribute %s', attr, exc_info=True
                     )
+            st_status_to_desc = {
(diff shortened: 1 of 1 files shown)


gazpachoking:
Hmm, so right now it's returning an integer? I sorta thought we were already returning the text status. Is there a way to do this without adding the int to string mapping in flexget? Transmissionrpc should be doing that for us https://bitbucket.org/blueluna/transmissionrpc/src/2d0a22304f0c74fb537221c3c8b49a603a9480eb/transmissionrpc/torrent.py#lines-26:34

Karlson2k:
Sorry for buzz, actually status is string already.

PR #2402, opened 2019-06-01
[WIP] Add pre-commit config

gazpachoking:
### Motivation for changes:
Make it easier to standardize our code with isort and black. Have easy instructions for contributors to make sure they remember too.

### Detailed changes:
- Move black and isort config to pyproject.toml
- Add a [pre-commit](https://pre-commit.com/) config

### Usage
```
pip install pre-config
<go to your flexget checkout dir>
pre-config install
<proceed to get automatic isort and black whenever committing>
```
#### To Do:
- [ ] Run these tools on everything, to make future diffs better?
- [ ] Do we switch to standard black string style? (double quotes)
- [ ] Enable CI check for black/isort compliance?


gazpachoking:
After adding the option to not edit string formatting, there wasn't too much change since the last time we ran black, so I went ahead and updated that. isort on the other hand will touch a whole bunch of stuff, do we want to run that on everything?

gazpachoking:
@paranoidi @liiight @cvium Everyone good with this? I guess there isn't really a downside, because you don't _have_ to install pre-commit. The real question would be do we want to convert any more stuff now (isort, or quote styles), and/or start enforcing this more.

gazpachoking:
I added a circleci job that verifies whether we diverge from this auto-formatting... is that something we want? I guess it would be a bit better if it only checked the files changed in the current PR, but I'm not exactly sure how that would be set up. If we fix all files now and start enforcing though that probably doesn't matter.

gazpachoking:
I've made a mess of this. Going to start over with just updating isort and black configs, then go further from there.

PR #2410, opened 2019-06-04
Implemented new jinja tests

Karlson2k:
### Motivation for changes:
Sometimes transmission do not report any error when files are not moved to final location from 'incomplete' directory. Need simple check for presence of files on filesystem when cleaning completed torrents

### Detailed changes:
As follow-up for PR #2394. Added `is fs_file` jinja test. Also added `fs_dir` and `fs_link` tests for additional flexibility. 

example:
``` YAML
  clean-torr:
    from_transmission:
      username: '{? trans.usr ?}'
      password: '{? trans.pwd ?}'
    disable: [seen, seen_info_hash]
    no_entries_ok: yes
    if:
      - transmission_error_state == 'local_error': reject
      - transmission_progress < 100: reject
      - ('/path/transmission/incomplete/' ~ title) is fs_file: reject
      - transmission_date_added + timedelta(minutes=95) < now: accept
    transmission:
      username: '{? trans.usr ?}'
      password: '{? trans.pwd ?}'
      action: remove
```

### DIFF
--- a/flexget/utils/template.py
+++ b/flexget/utils/template.py
@@ -7,6 +7,7 @@ import logging
 import os
 import re
 import locale
+import os.path
 from copy import copy
 from datetime import datetime, date, time
 
@@ -140,6 +141,21 @@ def filter_default(value, default_value='', boolean=True):
 filter_d = filter_default
 
 
+def is_fs_file(pathname):
+    """Test whether item is existing file in filesystem"""
+    return os.path.isfile(pathname)
+
+
+def is_fs_dir(pathname):
+    """Test whether item is existing directory in filesystem"""
+    return os.path.isdir(pathname)
+
+
+def is_fs_link(pathname):
+    """Test whether item is existing link in filesystem"""
+    return os.path.islink(pathname)
+
+
 class FlexGetTemplate(Template):
     """Adds lazy lookup support when rendering templates."""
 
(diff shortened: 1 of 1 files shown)


gazpachoking:
Should we make these real jinja tests, rather than just passing the functions in to the context?

Karlson2k:
> Should we make these real jinja tests, rather than just passing the functions in to the context?

What advantages will we have with junja test?
Will we be able to use constructs like `isfile('/some/path/{{title}}')`?

Could you point me where to look for jinja tests implementations?

gazpachoking:
Yeah, I'm starting to second guess that idea. It would put the syntax like `('/some/path/' ~ title) is file` A slight advantage is that it could prevent name collisions, but it's probably unlikely that anyone has entry fields called `isfile`. It would also let us use the tests anywhere else we allow jinja, but I don't really see many use cases for these outside of if plugin.

I could go either way on this, I'll leave it up to you. (Or anyone else can weigh in if they feel so inclined.) Here are the docs on how it could be done:
http://jinja.pocoo.org/docs/2.10/api/#custom-tests
We'd put those into this file, and probably do something similar to how we add custom filters.
https://github.com/Flexget/Flexget/blob/develop/flexget/utils/template.py

Karlson2k:
Updated with `fsys_` prefix for tests. I think `is link` (without prefix) could be confusing.

Maybe some tests for testsuite could be added later if PR is accepted.

Karlson2k:
Renamed PR title to avoid double entries in changelog.

gazpachoking:
Looks good. I'm thinking 'fs' is maybe a more traditional prefix for filesystem rather than fsys though.

Karlson2k:
@gazpachoking, updated with `fs_` prefix.

gazpachoking:
Awesome, looks good to me. Want to update our [jinja](/Jinja) wiki page with this too?

Karlson2k:
> Awesome, looks good to me. Want to update our [jinja](/Jinja) wiki page with this too?

@gazpachoking, updated, but on another URL: https://flexget.com/Jinja :)

PR #2414, opened 2019-06-12
Add seeder and leecher counts, minor refactoring

ianstalk:
### Motivation for changes:
The primary motivation was a feature request on the forum.

The secondary motivation was that the logic is all jammed into one function without much abstraction. I moved the part that parses fields out into its own method because that's what I was touching.

### Detailed changes:
- Move torrent field parsing into `_populate_entry_from_torrent_attrs`
- Add torrent_seeds and torrent_leeches

### Addressed issues:
-N/A

- https://discuss.flexget.com/t/can-plugin-from-transmission-get-more-information/4732/3

### DIFF
--- a/flexget/plugins/clients/transmission.py
+++ b/flexget/plugins/clients/transmission.py
@@ -214,29 +214,9 @@ class PluginTransmissionInput(TransmissionBase):
             if config['host'] in ('localhost', '127.0.0.1'):
                 entry['location'] = torrent.torrentFile
                 entry['url'] = 'file://' + torrent.torrentFile
-            for attr in [
-                'id',
-                'comment',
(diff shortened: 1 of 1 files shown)


gazpachoking:
Looks good to me. I'm wondering if this works differently on different versions of transmission though.. the current rpc specs say that field should be per tracker, rather than a torrent attribute directly. (i.e. `torrent.trackerStats[0].seedCount` or so) I'm not sure why transmission would do it this way though, it seems like you should be able to get the total number of known seeds/leeches, as there could be ones from different trackers, or from dht, or webseeds.

ianstalk:
I don't think this is going to work, actually. Closing the PR.

PR #2422, opened 2019-07-29
[change] Add pagination headers to history api even if no items are present

kristenmills:
### Motivation for changes:
Making Total Count explicitly 0 makes it easier for me to handle infinite scroll for this plugin from the web-ui. 

### Detailed changes:
- Add pagination headers in empty response from GET history api endpoint


gazpachoking:
Looks good to me.

PR #2434, opened 2019-08-26
[fix] thetvdb component retrieve first aired date

KuroSetsuna29:
- fallback to retrieve the first episode's aired date when series's first aired date not specified

### Motivation for changes:
TheTVDB stopped providing the series's first aired date correctly. This fix is to fallback to getting the aired date from the first episode if it is not available.

### Detailed changes:
- If `first_aired` is not set, query TheTVDB API for the first episode and use it's `first_aired` value


gazpachoking:
Seems like a good solution to me, thanks!

KuroSetsuna29:
Sorry, I found an issue. Sometimes the first episode absoluteNumber is not setup correctly either, really seems like a mess over at TheTVDB.

The fix is to use `?airedSeason=1&airedEpisode=1` instead of `?absoluteNumber=1`

PR #2437, opened 2019-08-27
[fix] thetvdb component retrieve first aired date

KuroSetsuna29:
Fixes issues addressed in PR #2434 

### Motivation for changes:
TheTVDB stopped providing the series's first aired date correctly. This fix is to fallback to getting the aired date from the first episode if it is not available.

### Detailed changes:
- fix error handling to handle gracefully
- use airedSeason and airedEpisode instead of aboluteNumber

### DIFF
--- a/flexget/components/thetvdb/api_tvdb.py
+++ b/flexget/components/thetvdb/api_tvdb.py
@@ -212,11 +212,12 @@ class TVDBSeries(Base):
         self._genres = [TVDBGenre(id=name) for name in series['genre']] if series['genre'] else []
 
         if self.first_aired is None:
+            log.debug('Falling back to getting first episode aired date for series %s', self.name)
             try:
-                episode = TVDBRequest().get('series/%s/episodes/query?absoluteNumber=1' % self.id, language=language)
+                episode = TVDBRequest().get('series/%s/episodes/query?airedSeason=1&airedEpisode=1' % self.id, language=language)
                 self.first_aired = episode[0]['firstAired']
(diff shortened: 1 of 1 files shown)


flexget/components/thetvdb/api_tvdb.py
```diff
@@ -212,11 +212,12 @@ def __init__(self, tvdb_id, language):
         self._genres = [TVDBGenre(id=name) for name in series['genre']] if series['genre'] else []
 
         if self.first_aired is None:
+            log.debug('Falling back to getting first episode aired date for series %s', self.name)
             try:
-                episode = TVDBRequest().get('series/%s/episodes/query?absoluteNumber=1' % self.id, language=language)
+                episode = TVDBRequest().get('series/%s/episodes/query?airedSeason=1&airedEpisode=1' % self.id, language=language)
```
cvium: What if the first aired episode is a special?

KuroSetsuna29: So I just got back and was about to start implementing change to get first episode that aired, however I found an example that shows that TheTVDB uses the first season episode as its official air date.
https://www.thetvdb.com/series/good-witch

So this suggests that we should only use 1x01 air date.

stevezau:
Any update on this? want to remove the `Error while updating from tvdb (Error updating first_aired from tvdb first episode:` error from logs

gazpachoking:
I think we decided this was good, let's go with it.

PR #2446, opened 2019-09-27
Update NPO_watchlists tests

mfonville:
### Motivation for changes:
The NPO website had some changes; updated the tests to reflect this


mfonville:
I don't know why CircleCI failed, it seems as if it has not applied the first of the two commits for running the tests on the updated cassettes?

Locally all the tests do pass :-/

mfonville:
Re-ran the tests and cassettes, now it seems to be ok

mfonville:
Could this (trivial) update to the tests please be merged; my server runs with every NPO-website these tests to check for any regressions, so having working tests greatly helps ;-)

gazpachoking:
Yep, looks good to me.

PR #2449, opened 2019-10-02
Allow to add paused torrents to qbittorrent

esanchezm:
### Motivation for changes:
I switched from Transmission to qBittorrent and I found that I couldn't add torrents in a paused state.

### Detailed changes:
- Added `add_paused` token to qBittorrent configuration that allows torrents to be added in paused state

```YAML

  my_qbittorrent_server:
    qbittorrent:
      host: 127.0.0.1
      port: 8080
      path: /download/
      add_paused: yes
```

### DIFF
--- a/flexget/plugins/clients/qbittorrent.py
+++ b/flexget/plugins/clients/qbittorrent.py
@@ -154,7 +154,7 @@ class OutputQBitTorrent(object):
                 form_data['label'] = label  # qBittorrent v3.3.3-
                 form_data['category'] = label  # qBittorrent v3.3.4+
 
-            add_paused = entry.get('paused', config.get('paused'))
+            add_paused = entry.get('add_paused', config.get('add_paused'))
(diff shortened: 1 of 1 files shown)


flexget/plugins/clients/qbittorrent.py
```diff
@@ -152,6 +154,10 @@ def add_entries(self, task, config):
                 form_data['label'] = label  # qBittorrent v3.3.3-
                 form_data['category'] = label  # qBittorrent v3.3.4+
 
+            add_paused = entry.get('paused', config.get('paused'))
```
gazpachoking: Shouldn't these both be `add_paused` rather than just `paused`?

esanchezm: You're right, thanks!

gazpachoking:
Thanks!

PR #2451, opened 2019-10-04
Switch to GitHub Actions for CI

gazpachoking:
### Motivation for changes:
Try out github actions as a replacement for circleci
Hopefully we don't need to maintain custom docker images this way

#### To Do:

- [ ] Get it working
- [ ] Verify we like it better than circleci
- [ ] Add a release workflow

### DIFF
--- a/flexget/tests/conftest.py
+++ b/flexget/tests/conftest.py
@@ -206,8 +206,7 @@ def pytest_runtest_setup(item):
 @pytest.yield_fixture()
 def filecopy(request):
     out_files = []
-    marker = request.node.get_closest_marker('filecopy')
-    if marker is not None:
+    for marker in request.node.iter_markers('filecopy'):
         copy_list = marker.args[0] if len(marker.args) == 1 else [marker.args]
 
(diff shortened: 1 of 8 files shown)


flexget/tests/conftest.py
```diff
@@ -237,7 +238,7 @@ def filecopy(request):
                 else:
                     f.remove()
             except OSError as e:
-                print("couldn't remove %s: %s" % (f, e))
+                print ("couldn't remove %s: %s" % (f, e))
```
liiight: added space by mistake?

gazpachoking: Hmm, it happened twice in that commit, I remember seeing it in the diff after I ran black, but assumed it was fixing the extra space, not adding one...

gazpachoking: Ahh, I must need to upgrade black on that machine. https://github.com/psf/black/issues/752

gazpachoking:
Some stuff unrelated to actions was caught up in here. Merged that into develop directly and moved actions to new PR #2488

PR #2454, opened 2019-10-12
Improve developer documentation slightly

### LINKED ISSUE #2376: readme: git virtualenv doc: maybe recommend python 3 and venv?
In the README (and in https://flexget.readthedocs.io/en/latest/develop/index.html) the instructions to run flexget from git recommend to use `virtualenv`, but in python3 there is a native `venv`: maybe these docs could be updated to recommend use python3 and `venv` ?

tobbez:
* Minor RST fixup. Read the Docs' parser didn't seem to care while
    pandoc did.
  * Remove some unused hyperlink targets.
  * Recommend venv rather than virtualenv.
  * Some minor language improvements.

Fixes #2376.

### DIFF
--- a/docs/develop/index.rst
+++ b/docs/develop/index.rst
@@ -24,7 +24,7 @@ Core
 * Some smaller misc libraries
 
 HTTPServer
-~~~~~
+~~~~~~~~~~
 
 * Flask
 * Jinja2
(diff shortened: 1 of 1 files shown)


liiight:
Thanks!

PR #2461, opened 2019-10-19
[add] passthepopcorn sort by individual torrents

tarantoj:
Lets user configure grouping search parameter,
as compared to default implicit 'grouping=1'

### Motivation for changes:
By default, passthepopcorn groups search results by movie id, influencing how results are returned using the original `order_by` parameter.

For example, searching for a movie by string, ordering by seeders, will return the first entry in the highest seeded group. So searching for 'get', sorted by seeders, returns 
'Get Out (2017)  x264 / MKV / Blu-ray / 720x302 / Scene ', rather than 
'Get Out (2017) x264 / MKV / Blu-ray / 1080p / With Commentary'

```
passthepopcorn:
  grouping: no (default yes)
```
#### To Do:

- [ ] Add corresponding documentation on https://flexget.com/Searches/passthepopcorn

This is my first pull request for the project, feel free to let me know of any problems.


liiight:
Thanks for this!

PR #2466, opened 2019-10-24
[fix] remove horriblesubs scraper in favour of cfscraper plugin

vxcamiloxv:
### Motivation for changes:
`horriblesubs` fail because it requires the old scraper lib as dependence, so not make sense have tow different libs to do the same, also it should use the `cfscraper` plugin to have that responsibility unify

### Detailed changes:
- Remove custom horriblesubs scraper and use task request that has `cfscraper` instance

### Addressed issues:
- No Yet

### DIFF
--- a/flexget/components/sites/sites/horriblesubs.py
+++ b/flexget/components/sites/sites/horriblesubs.py
@@ -83,25 +83,12 @@ class HorribleSubs(object):
             )
         return entries
 
-    @staticmethod
-    def scraper():
-        try:
-            import cfscrape
-        except ImportError as e:
-            log.debug('Error importing cfscrape: %s', e)
-            raise plugin.DependencyError(
-                'cfscraper', 'cfscrape', 'cfscrape module required. ImportError: %s' % e
-            )
-        else:
-            return cfscrape.create_scraper()
-
     @cached('horriblesubs')
     def on_task_input(self, task, config):
         if not config:
             return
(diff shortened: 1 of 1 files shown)


liiight:
can you show logs that this fix works for you?

vxcamiloxv:
@liiight  sure
screenshot-2019-10-25_06-44-40

gazpachoking:
Am I right in guessing that the user has to manually include the cfscraper plugin in the task along with horriblesubs? If so, maybe it should catch the error we get when we don't have it and print a helpful message about it being required. Or, we could have it enable that plugin automatically.

vxcamiloxv:
@gazpachoking makes sense, so catch the error or enabled it automatically?

gazpachoking:
Not sure. Catching the error would leave room for other plugins to replace cfscraper, but automatically enabling it might make it a bit easier for users, given that it's the only way to make it work right now.

vxcamiloxv:
@gazpachoking actually `horriblesubs` not needs scraper https://horriblesubs.info/api.php?method=getlatest

without cfscraper plugin
screenshot-2019-10-25_08-51-38

gazpachoking:
Oh, excellent. I guess they just removed the cloudflare protection.

gazpachoking:
Thanks! We can update the wiki page to removed the bit about needing cfscrape installed then.

PR #2472, opened 2019-10-27
[fix] anilist: Larger lists no longer limited to the first 50 items

### LINKED ISSUE #2471: AniList plugin only loads the first page
### Expected behaviour:

Using the AniList plugin on my account (`username: Kazakuri`) should create 840 entries with completed status

### Actual behaviour:

It creates 50 entries.

### Steps to reproduce:

1. Use the AniList input plugin with an account that has more than 50 entries of a given status.
2. Observe results.

#### Config:
```yaml
    tasks:
  example:
    anilist:
      username: kazakuri
      status:
        - completed
      release_status:
        - finished
```
  
#### Log:
<details>

```
2019-10-26 23:09 VERBOSE  input_cache   example          Restored entries from cache
2019-10-26 23:09 VERBOSE  details       example          Produced 50 entries.
2019-10-26 23:09 WARNING  task          example          Task does not have any filter plugins to accept entries. You need at least one to accept the entries you  want.
2019-10-26 23:09 VERBOSE  details       example          Summary - Accepted: 0 (Rejected: 0 Undecided: 50 Failed: 0)
2019-10-26 23:09 WARNING  task          example          Task doesn't have any output plugins, you should add (at least) one!
```
</details>

### Additional information:

- FlexGet version: 2.21.32
- Python version: 3.7.5
- Installation method: pip
- Using daemon (yes/no): no
- OS and version: Alpine Linux 3.10

The [query being built](https://github.com/Flexget/Flexget/blob/ea11cdccce95c6513d8be9978bdd9d0baeb88b7b/flexget/plugins/input/anilist.py#L96) wraps the `MediaList` type with the `Page` type, but by design the `Page` type limits the response to 50 entries.

#### Querying Each Page

You can query the number of pages in the same query and use that to iterate over all the available pages in a loop.

#### Using MediaListCollection and Querying Each Chunk

Instead of using `Page` With `MediaList`, you can query with `MediaListCollection` instead to retrieve up to 500 entries per chunk.

BrutuZ:
### Motivation for changes:
Bug report. Previous method was limited to 50 entries, this patch handles the pagination
### Detailed changes:
- Changed query to use an object able to carry more elements (500 instead of 50)
- Handle pagination if needed
- Add optional `title_only` option to schema for slimmer result

### Schema update:
```diff
anilist:
  username: <<string | required>>
  status: <<enum | optional>>
  release_status: <<enum | optional>>
  format: <<enum | optional>>
+ title_only: <<boolean | optional>>
```
### Addressed issues:
- Fixes #2471 .

### To Do:
- [x] Test
- [x] Review
- [ ] Merge
- [ ] Update wiki - Document `title_only`

### DIFF
--- a/flexget/plugins/input/anilist.py
+++ b/flexget/plugins/input/anilist.py
@@ -40,6 +40,7 @@ class AniList(object):
         - <all|tv|tv_short|movie|special|ova|ona>
         - <tv|tv_short|movie|special|ova|ona>
         ...
+      title_only: <boolean>
     """
 
     schema = {
@@ -88,7 +89,6 @@ class AniList(object):
         log.debug('selected_release_status: %s' % selected_release_status)
         log.debug('selected_formats: %s' % selected_formats)
 
-        list_json = []
         req_variables = {'user': config['username']}
         req_chunk = 1
         req_fields = ('status, title{ romaji, english }, synonyms, siteUrl, idMal, format, episodes, '
--- a/flexget/plugins/input/anilist.py
+++ b/flexget/plugins/input/anilist.py
@@ -93,7 +93,7 @@ class AniList(object):
         req_chunk = 1
         req_fields = ('status, title{ romaji, english }, synonyms, siteUrl, idMal, format, episodes, '
                     'trailer{ site, id }, coverImage{ large }, bannerImage, genres, tags{ name }, '
-                    'externalLinks{ site, url }' if not lightweight else 'status, format, title{ romaji }')
+                    'externalLinks{ site, url }' if not lightweight else 'status, format, title{ romaji, english }')
         while req_chunk:
             req_query = ('query ($user: String){collection: MediaListCollection(userName: $user, type: ANIME, '
                         'perChunk: 500, chunk: %s, status_in: [%s]){ hasNextChunk, statuses: lists { list: entries { '
--- a/flexget/plugins/input/anilist.py
+++ b/flexget/plugins/input/anilist.py
@@ -105,10 +105,11 @@ class AniList(object):
                 raise plugin.PluginError('Error reading list - {url}'.format(url=e))
 
             try:
-                list_json.append(list_response.json()['data'])
+                list_response = list_response.json()['data']
+                list_json.append(list_response)
             except ValueError:
                 raise plugin.PluginError('Invalid JSON response')
-            req_chunk = req_chunk + 1 if list_response.json()['data']['collection']['hasNextChunk'] else False
+            req_chunk = req_chunk + 1 if list_response['collection']['hasNextChunk'] else False
 
         log.debug('JSON output: %s' % list_json)
         for list_index in list_json:
--- a/flexget/plugins/input/anilist.py
+++ b/flexget/plugins/input/anilist.py
@@ -40,7 +40,6 @@ class AniList(object):
         - <all|tv|tv_short|movie|special|ova|ona>
         - <tv|tv_short|movie|special|ova|ona>
         ...
-      title_only: <boolean>
     """
 
     schema = {
@@ -58,8 +57,7 @@ class AniList(object):
                     ),
                     'format': one_or_more(
                         {'type': 'string', 'enum': ANIME_FORMAT}, unique_items=True
-                    ),
-                    'title_only': {'type': 'boolean'}
+                    )
                 },
                 'required': ['username'],
                 'additionalProperties': False,
@@ -74,7 +72,6 @@ class AniList(object):
         selected_list_status = config['status'] if 'status' in config else ['current', 'planning']
         selected_release_status = config['release_status'] if 'release_status' in config else ['all']
         selected_formats = config['format'] if 'format' in config else ['all']
-        lightweight = config.get('title_only', False)
 
         if not isinstance(selected_list_status, list):
             selected_list_status = [selected_list_status]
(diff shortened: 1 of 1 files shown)


gazpachoking:
Should we change this to a generator type input (`yield` the entries, rather than making a list and returning them all?) That way it would work properly with the `limit` plugin and doesn't cause extra page lookups if you only want a certain number of pages.

BrutuZ:
Integration with other control plugins such as `limit` is always nice. How would I go about that?

BrutuZ:
Took a look at how `kitsu` handles the `limit` plugin for reference.
Changed this to use the same as suggested and passed a couple manual tests.

gazpachoking:
Yep, just changing to yield the entries is the only thing you need to do for full limit compatibility, looks good.

I'm not so sure we need the 'lightweight' option. It's already fetched all that data, so we don't save on network requests, right? It shouldn't really increase processing time just to save a few more fields to the entry.

BrutuZ:
It requests less data with `titles_only` \ "lightweight" (4 fields opposed to 16).
Using that was also the only way I could successfully complete the test task from #2471 without running out of memory 😳

flexget/plugins/input/anilist.py
```diff
@@ -67,12 +68,12 @@ class AniList(object):
 
     @cached('anilist', persist='2 hours')
     def on_task_input(self, task, config):
-        entries = []
         if isinstance(config, str):
             config = {'username': config}
         selected_list_status = config['status'] if 'status' in config else ['current', 'planning']
         selected_release_status = config['release_status'] if 'release_status' in config else ['all']
         selected_formats = config['format'] if 'format' in config else ['all']
+        lightweight = config['title_only'] if 'title_only' in config else False
```
gazpachoking: This sort of statement is more easily written using `dict.get`
```suggestion
        lightweight = config.get('title_only', False)
```

gazpachoking:
> Using that was also the only way I could successfully complete the test task from #2471 without running out of memory 😳

Hmm, that seems sorta crazy. How much memory do you have? I wonder what could suck up that much memory.

BrutuZ:
8GB RAM. From the little I could gather by the log, seems to be during the caching process
```
2019-10-27 14:47 DEBUG    remember_rej  anilist         Task config has changed since last run, purging remembered entries.
2019-10-27 14:47 DEBUG    input_cache   anilist         cache name: anilist_94f5cc2c84cb2d66899df3e0ddd3de66 (has: )
2019-10-27 14:47 DEBUG    input_cache   anilist         storing entries to cache anilist_94f5cc2c84cb2d66899df3e0ddd3de66
2019-10-27 14:47 DEBUG    anilist       anilist         selected_list_status: [u'completed']
2019-10-27 14:47 DEBUG    anilist       anilist         selected_release_status: [u'finished']
2019-10-27 14:47 DEBUG    anilist       anilist         selected_formats: [u'all']
2019-10-27 14:47 DEBUG    utils.requests anilist         POSTing URL <<Snipped to save space>>
2019-10-27 14:47 DEBUG    anilist       anilist         JSON output: <<Snipped to save space>>
2019-10-27 15:04 DEBUG    input_cache   anilist         Storing cache anilist_f67582754e8f0a24b1e1dc9549da8992 to database.
2019-10-27 15:04 DEBUG    backlog.db    anilist         0 entries purged from backlog                                      
2019-10-27 15:05 CRITICAL task          anilist         BUG: Unhandled error in plugin backlog:                            
2019-10-27 15:05 CRITICAL task_queue                    BUG: Unhandled exception during task queue run loop.               
Exception in thread task_queue:                                                                                            
  File "c:\python27\Lib\threading.py", line 801, in __bootstrap_inner                                                      
    self.run()                                                                                                             
  File "c:\python27\Lib\threading.py", line 754, in run                                                                    
    self.__target(*self.__args, **self.__kwargs)                                                                           
  File "~\flexget\.install\lib\site-packages\flexget\task_queue.py", line 56, in run                         
    self.current_task.manager.crash_report()                                                                               
  File "~\flexget\.install\lib\site-packages\flexget\manager.py", line 1074, in crash_report                 
    outfile.writelines(logger.debug_buffer)                                                                                
  File "~\flexget\.install\lib\codecs.py", line 712, in writelines                                           
    return self.writer.writelines(list)                                                                                    
  File "~\flexget\.install\lib\codecs.py", line 377, in writelines                                           
    self.write(''.join(list))                                                                                              
MemoryError                                                                                                                
                                                                                                                           
2019-10-27 15:05 INFO     ipc.rpyc                      listener closed                                                    
2019-10-27 15:05 DEBUG    util.simple_persistence                 Flushing simple persistence for task None to db.         
2019-10-27 15:05 DEBUG    manager                       Shutting down                                                      
2019-10-27 15:05 INFO     manager                       Removed test database                                              
2019-10-27 15:05 DEBUG    manager                       Removed ~\flexget\.test-config-lock

gazpachoking:
Hmm, it looks like the memory error is from when it tries to write out the crash report. I wonder what the error that caused the original crash was. Perhaps the same thing, but from somewhere else.

BrutuZ:
Given that error is not related to the `anilist` plugin, is the PR good to go?

paranoidi:
That out of memory issue is scaring me ...

BrutuZ:
Interesting, just retested the same code and data - now in Python 3.7 instead of my previous 2.7 - and it not only completed without running out of memory but was also **a lot** faster.
Given how FlexGet3 doesn't support Python 2.7 anymore we should be good.

gazpachoking:
Hmm, something weird happened here. I'm not sure why github is showing unrelated commits and changes in this pull. Looks like conflicts need to be resolved as well.

BrutuZ:
Little Git mistake on my part, should be good now

gazpachoking:
I resolved the conflicts and merged this. Hopefully I've not messed it up. Probably want to give it a test with latest develop branch.

Thanks!

PR #2477, opened 2019-11-04
correct season and episode variable for trakt aired

jorritsmit:
### Motivation for changes:
Flexget gave an error when using the 'aired' context. This was because the wrong variables were set.

### Detailed changes:
- renamed the variables to the correct names


gazpachoking:
Looks good, thanks!

PR #2478, opened 2019-11-05
Add subtask input plugin

gazpachoking:
### Motivation for changes:
Be able to manipulate and filter entries before passing in to plugins like configure_series, or discover.

### Detailed changes:
- Creates a new input plugin which produces the resulting accepted entries from another task run.

```
tasks:
  task_a:
     mock:
      - title: blah
      accept_all: yes
  task_b:
    discover:
      what:
      - from_task: task_a
      from:
      - somewhere
```
#### To Do:

- [x] Do we like the name 'subtask' best? What about 'from_task'? (going with from_task)
- [x] Currently for logging the subtask name is `parenttask>subtask` is that good?
  good enough for me.
- [x] Should the 5 minute caching mechanism be enabled for subtask results?
  Decided against this for now at least.
- [ ] Is there an alternative way to disable the seen plugin, or some other solution to make that not necessary?
- [x] Do we allow/require subtasks to be in a new root level 'subtasks' key? If not, is there a better way to prevent them from being executed by default from cli or daemon scheduling?
  Skipping this for now. Can always be added later.
- [x] Do we pass on the options of the parent task to the subtask? --inject is probably a very bad to pass on, but do we want/need other things passed on?
  Let's not pass any options through right now. If we discover something that should be we can re-think.


liiight:
I totally get the need for this, but this feels very hacky. I think ideally we want that ability to chain tasks to each other and that should be the direction, and not neccesarily introduce new hybrid solutions like this.

gazpachoking:
@liiight How do you mean chain tasks?

gazpachoking:
I sorta like this method. It doesn't really allow anything different than clearing then saving to an entry_list and using it somewhere else, but this way feels like it might be a bit easier. I've considered chaining tasks with something similar to run_task, but that passes entries along, but that doesn't help us in cases where we want to use the entries inside discover, or configure_series

liiight:
> @liiight How do you mean chain tasks?

I've tried putting something down only to get something almost similar to what you did, so NM :)

gazpachoking:
@liiight @paranoidi @cvium Any thoughts on the TODO checklist questions? I like this method, but want to get a little feedback on those points before merging it in.

gazpachoking:
@liiight @paranoidi Is `from_task` (ala from_deluge, from_transmission) a better name than subtask?

cvium:
`from_task` is clearer for me

gazpachoking:
> `from_task` is clearer for me

That's the way I'm leaning as well. I think it makes it more clear it's just an input. Especially if already familiar with from_deluge etc.

gazpachoking:
Wiki update: https://github.com/Flexget/wiki/commit/e336ec46fb5f099eef081e86d30c28835606fd5d

metaMMA:
Does the flexget.com search database need to be updated when adding new plugins? I searched for "from_task" in the search bar on flexget.com "https://flexget.com/_search?q=from_task" but it comes up empty. I guessed the url and it took me to the page, so I know it DOES exist. I think there may have been a handful of other plugins that don't come up in a search, too. But I cannot remember which ones...

gazpachoking:
> Does the flexget.com search database need to be updated when adding new plugins? I searched for "from_task" in the search bar on flexget.com "https://flexget.com/_search?q=from_task" but it comes up empty. I guessed the url and it took me to the page, so I know it DOES exist. I think there may have been a handful of other plugins that don't come up in a search, too. But I cannot remember which ones...

Seems like maybe the search index update was getting bypassed when syncing from the github repo (or something.) I've added a weekly cronjob to re-index, hopefully that keeps things working better.

PR #2481, opened 2019-11-06
Add a plugin to parametrize input plugins

gazpachoking:
### Motivation for changes:
Allows parametrizing input plugins. e.g. getting a list of rss sources from a csv file, then getting the rss items from each of those sources. Parsing an html page for other pages, then parsing those other pages for entries.

```
tasks:
  rss_from_csv:
    parameterize:
      plugin:
        rss: "{{url}}"
      using:
        csv: /my/list_of_rss_feeds
  two_level_html:
    parameterize:
      plugin:
        html: "{{url}}"
      using:
        html: https://toplevelpage.com
```

#### To Do:

- [ ] Decide whether name and configuration layout can be improved.
- [x] Validation doesn't work right in the 'plugin' section of config right now. It should use input plugin validation, but there are issues because e.g. `{{url}}` is not a valid url, and fails validation.


metaMMA:
@gazpachoking 

Can you elaborate on the motivation for the `parameterize` name? 

Maybe I'm not understanding exactly what this plugin can accomplish, though. 

As a basic user, I'm not sure I would know what this plugin does when looking for plugins that had its function. 

I'm more partial to `subtask` or maybe another name.

gazpachoking:
This one is different than subtask, I have a separate PR with that one. This one will run an input plugin multiple times, changing the configuration each time based on values from another set of entries.

flexget/plugins/input/parameterize.py
```diff
@@ -0,0 +1,62 @@
...
+
+
+class Parameterize(object):
+
+    schema = {
+        'type': 'object',
+        'properties': {
+            # TODO: This should really be input plugin validation, but some format validators don't deal well with Jinja
+            'plugin': {'type': 'object'},
+            'using': {'$ref': '/schema/plugins?phase=input'},
```
liiight: maybe name this `from`?

flexget/plugins/input/parameterize.py
```diff
@@ -0,0 +1,62 @@
...
+                log.warning('Error during input plugin %s: %s' % (param_input_name, e))
+                continue
+            for param_entry in param_entries:
+                for subj_input_name, subj_input_config in config['plugin'].items():
+                    subj_input = plugin.get_plugin_by_name(subj_input_name)
+                    method = subj_input.phase_handlers['input']
+                    subj_input_config = _parameterize(subj_input_config, param_entry)
+                    try:
+                        result = method(task, subj_input_config)
+                    except plugin.PluginError as e:
```
liiight: i get why you're catching `PluginError` but I wonder if this could have adverse side effects since the expectation is to abort the task

liiight:
This is pretty cool actually. I wonder if `parameterize` isn't a bit on the technical side. maybe `repeat`, `use_with`, `reuse`, `loop`?

metaMMA:
Yeah, I should have read the OP before commenting. A use-case for this plugin is similar to something I had envisioned for a `subtask` use-case.

The name makes a lot more sense, now that I understand what it does. I think there is definitely a good case for keeping plugin names short and concise. I'll throw a couple names out there anyway: 

- `scrape_for_inputs`
- `inputs_from`
- `discover_from`

Now that I'm looking at the `discover` [manpage](https://flexget.com/Plugins/discover), I wonder if this is similar enough to the current use of `discover` to justify adding these features there?

I haven't looked at the code to see if this is even remotely possible. I'll take a look later today and weigh-in further.

flexget/plugins/input/parameterize.py
gazpachoking: It's there right now because I copied it from somewhere else. 😉 But I'm not sure what the expectation should be actually. It's not guaranteed that if one of the sets of parameters fail that they all will. Would we still want it to try the others?

flexget/plugins/input/parameterize.py
gazpachoking: I'd think renaming the other section would make more sense for 'from', if either of them do. This is the part that generates the parameters, not that does the entry creating. In my head, the config was describing a sentence e.g. 'Parameterize the html plugin, using settings from rss plugin.' Where in this case 'html' would be the 'plugin' and 'rss' would be the 'using' in the config. I agree this all might be a bit technical naming, but I'm not sure what other sorts of namings would make it any simpler yet. It's a sort of technical concept.

gazpachoking:
> Now that I'm looking at the `discover` [manpage](https://flexget.com/Plugins/discover), I wonder if this is similar enough to the current use of `discover` to justify adding these features there?

Yeah, very similar to discover. I implemented this as a search plugin for discover at first, but the release date lookup and repeat limiting discover does doesn't go along with this use case very well, which is why I thought maybe splitting it would be good. The other reason was so that hopefully the names could make it more clear what's happening, though from the current feedback they probably don't yet.

> Yeah, I should have read the OP before commenting. A use-case for this plugin is similar to something I had envisioned for a `subtask` use-case.

Yeah, the reason I thought of both this and subtask plugin was based on your use case. I think your use case would involve a subtask being used in the 'using' section of this plugins config.

> This is pretty cool actually. I wonder if `parameterize` isn't a bit on the technical side. maybe `repeat`, `use_with`, `reuse`, `loop`

I agree it might be a bit technical. I worry those other names don't convey properly that it's changing the settings of the input plugin each time through the loop. Not sure though.

gazpachoking:
Maybe better names for the 2 config keys would be 'input' and 'parameters'? (If the plugin name stays as parameterize.)

BrutuZ:
Tossing another idea in the hat:
```yaml
forward_output:
  from: <<plugin>
  to: <<plugin>>
```

metaMMA:
I was playing around with this plugin today. I got it to use an input (`soup_parse`) that was "using" a different input (`json`). The issue I had to overcome was that the schema of the specified "plugin" (`soup_parse`) isn't really imported into `parameterize` so any defaults that are specified in the input schema aren't evaluated. Once I explicitly assigned all necessary values for my input in the config, it worked as expected.

```
tasks:
  parameterize_test:
    accept_all: yes
    parameterize:
      plugin:
        soup_parse:
          ...
          ...
      using:
        json:
          ...
          ...
    list_add:
      - entry_list: test
```

metaMMA:
I'll see if I can figure out a way to get that functionality back.

It was a nice surprise that `domain_delay` worked as expected. I wasn't sure how the loop would affect that function.

gazpachoking:
Yeah the validation is a problem. It could be implemented the same way as it is for the 'using' param, except then we run into problems with jinja expressions not being a valid url.

gazpachoking:
@metaMMA Give it a try now. I loosened our url validator to allow jinja, and enabled config validation on the 'plugin' part of the config.

metaMMA:
It's working well!

gazpachoking:
Maybe `configure_input` is a less technical name? It's similar to `configure_series` which is a similar principle to what we are doing here.

soloam:
I know that this is very old! But just stepped into it... And I think this would be a great addition to flexget! A advanced plugging, to more advanced users and configurations!

gazpachoking:
I think the naming was the only thing to consider (other than whether it should go in at all) and I think I'm sorta happy with the current names. It's a bit technical, but this plugin is a bit technical, so maybe the name scares off some people that shouldn't be diving this deep. 😛
I'll have to take a peek to see if everything else was good, it's been a long time.

soloam:
I just tested this and It worked great! I was able to retrieve several files with filesystem and use them in the text plugin to retrieve the content! This is so powerfull and so useful that is to good to be passed away.

I don't care a lot about the naming, but configure_input would be better, that's what we are doing, configuring input. When I scroll the flexget plugins list, "configure_input" would catch my attention better to the end goal. But don't let that be reason to hold you back from merging :)

PR #2484, opened 2019-11-07
Added 'encoding' to schema, so that configs will pass schema te…

metaMMA:
### Motivation for changes:
`encoding` is referenced on the webpage regarding this plugin, but was missing from the schema. Using it in a config would cause a schema failure.
### Detailed changes:
- added a single line to the schema to prevent schema test failure.

### Addressed issues:
- schema failure when config should pass schema test


liiight:
thanks!

PR #2485, opened 2019-11-09
Added 'serialize' method.

metaMMA:
### Motivation for changes: 
Serialization was needed to output the the entry list as a `JSON` file.

### Detailed changes: 
- Added a method to the `Entry` class called `serialize`. `entry.serialize()` returns a python dictionary that is ready to be used in file outputs like `json.dump`, or for use in writing `text`, `csv`, `html`, etc. files.  It was added here, so that anywhere an entry exists, it can always be serialized easily.

#### To Do:

- [ ] Discuss if this is wanted/needed
- [ ] Discuss best location for this method

### DIFF
--- a/flexget/entry.py
+++ b/flexget/entry.py
@@ -8,7 +8,7 @@ import functools
 import logging
 from collections import Mapping
 
-from flexget.utils import qualities, json
+from flexget.utils import qualities
 from flexget.plugin import PluginError
 from flexget.utils.lazy_dict import LazyDict, LazyLookup
 from flexget.utils.template import render_from_entry, FlexGetTemplate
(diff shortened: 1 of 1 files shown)


liiight:
I'm not sure this is needed. You can already call `dict(entry)` to get a dict, and I think the custom serialization of non trivial fields should go in our json util (which they already do).

metaMMA:
I think the only issue might be with the `Quality` object. If I use the `datetime_encoding=True` flag, I'm not sure it will convert the `Quality` object to a string.

liiight:
Perhaps that will need an additional encoder then

paranoidi:
In order this to be useful it should be able to serialize & deserialize lazy fields and other object based fields like quality.

metaMMA:
> In order this to be useful it should be able to serialize & deserialize lazy fields and other object based fields like quality.

Agreed. I'm hoping to start work on the `de-serialization` method this week.

paranoidi:
Superseded by #2517

PR #2488, opened 2019-11-11
Switch to GitHub Actions for CI

gazpachoking:
### Motivation for changes:
Second try, first in #2451
Try out github actions as a replacement for circleci
Hopefully we don't need to maintain custom docker images this way

This also move the changelog updating script to running on actions. The wiki now pulls from github repo and pushes changes back, so changes can be made directly on github wiki repo.

#### Caveats:
- It's a couple minutes slower than circleci.
- PR tests are run on the merged result, meaning they won't run while there are merge conflicts.

#### To Do:

- [x] Get it working
- [x] Verify we like it better than circleci
- [x] Add a release workflow


gazpachoking:
@paranoidi @liiight @cvium I'm pretty happy with this, you guys on board? My prime motivation is not needing to maintain docker images.

.github/workflows/main.yml
```diff
@@ -0,0 +1,88 @@
...
+    - run: |
+        echo ${{github.event_name}}
+        echo ${{github.event}}
+  tests:
+    name: Run Tests
+    runs-on: ${{ matrix.operating-system }}
+    strategy:
+      matrix:
+        python-version: [3.6, 3.7]
+        operating-system: [ubuntu-latest] # Get linux working before windows. windows-latest
```
liiight: Is this stable enough to run on windows yet?

release.sh
```diff
@@ -32,13 +38,14 @@ if git log --skip 1 origin/master..origin/develop|grep '^commit '; then
...
   git commit -m "Prepare v`python dev_tools.py version`"
+  git branch -f develop HEAD
 
   # master branch should be at the release we tagged
   git branch -f master ${VERSION}
   # If either of the new branches are not fast forwards, the push will be rejected
   git push origin master develop
   # Make sure our branches push before pushing tag
-  git push --tags
+  #git push --tags
```
liiight: Why not push tags?

gazpachoking: Oops, just forgot to re-enable after testing. Good catch.

.github/workflows/main.yml
gazpachoking: Let's find out.

gazpachoking: Couple issues. One is that the extract plugin tests can't actually tell if rar is installed so don't get skipped if it isn't. The other is in the torrent_match tests, it's using text files to match against, which get an extra byte in size on windows, due to being checked out with `\n` turned into `\n\r` I'm guessing.

PR #2489, opened 2019-11-11
Randomize order of inputs in 'inputs' list.

metaMMA:
### Motivation for changes:
Making a large number of requests to smaller or less popular websites on a regular (scheduled) basis may result in throttling or timeouts. Randomizing the order of the requests (as well as utilizing `domain_delay`) may mitigate this risk.

### Detailed changes:
- Import `shuffle` method from `random` and use it on the list of inputs, before executing the input plugin.

### DIFF
--- a/flexget/plugins/input/inputs.py
+++ b/flexget/plugins/input/inputs.py
@@ -3,6 +3,8 @@ from builtins import *  # noqa pylint: disable=unused-import, redefined-builtin
 
 import logging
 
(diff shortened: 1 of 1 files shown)


gazpachoking:
Don't think I like this. Right now you can prioritize which inputs are checked first, (and thus have the first chance to be accepted,) by listing them out in order.
Seems like a niche case which would only affect the access time slightly.

PR #2490, opened 2019-11-11
Randomize scheduler

metaMMA:
### Motivation for changes:
Making a large number of requests to smaller or less popular websites on a regular (scheduled) basis may result in throttling or timeouts. Randomizing the timing of the requests (as well as utilizing `domain_delay`) may mitigate this risk.

### Detailed changes:
- Added the ability to use `'randint(<minimum integer>, <maximum integer>)'` as a property value in `Advanced Cron Interval` in `scheduler.py`

The following config will run the list of tasks once an hour (from 8AM to 10PM) at a randomly generated minute in the assigned interval (5-35).
```
schedules:
  - tasks: [list, of, tasks]
    schedule:
      minute: 'randint(5,35)'
      hour: 8-22
```
```
TBA
```
#### To Do:

- [ ] Discuss if this can be safely done.
- [ ] Currently requires triggering `update_config` method from `manager.py` - how do we safely trigger this.
- [ ] Should triggering the `update_config` be minimized based on the location of the `randint()` entry? ie. If `randint()` is in the `day` property, trigger the `update_config` method once a month. If it's in the `minute` property, trigger the method once an hour, etc.
- [ ] Update manpage on `flexget.com`

### DIFF
--- a/flexget/components/scheduler/scheduler.py
+++ b/flexget/components/scheduler/scheduler.py
@@ -4,6 +4,7 @@ import hashlib
 import logging
 import os
 import struct
+from random import randint
 from builtins import *  # noqa pylint: disable=unused-import, redefined-builtin
 
 import pytz
--- a/flexget/components/scheduler/scheduler.py
+++ b/flexget/components/scheduler/scheduler.py
@@ -185,7 +185,16 @@ def setup_jobs(manager):
         if 'interval' in job_config:
             trigger, trigger_args = 'interval', job_config['interval']
         else:
-            trigger, trigger_args = 'cron', job_config['schedule']
+            schedule = job_config['schedule']
+            for unit, val in schedule.items():
+                if 'randint' in str(val):
+                    try:
+                        new_val = eval(val)
+                    except (TypeError, NameError, ValueError) as e:
+                        log.warning('The value assigned to %s: %s is not valid. Defaulting to 1: %s' %(unit, val, e))
+                        new_val = 1
+                    schedule[unit] = new_val
+            trigger, trigger_args = 'cron', schedule]
         tasks = job_config['tasks']
         if not isinstance(tasks, list):
             tasks = [tasks]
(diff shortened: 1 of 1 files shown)


gazpachoking:
Looks like apscheduler allows a 'jitter' option (for both cron and regular interval scheduling) to randomize the times slightly. I think a nicer way to implement this would just be to allow that key in our config. https://apscheduler.readthedocs.io/en/latest/modules/triggers/interval.html

metaMMA:
@gazpachoking 
Ah yes, `jitter` provide the exact functionality I'm looking for

gazpachoking:
I allowed that jitter option in our config now. https://github.com/Flexget/Flexget/commit/06953ce70d59a96d799076475e20b8872ad5644c

paranoidi:
@gazpachoking & @metaMMA  .. it should be probably on by default to make any impact?

gazpachoking:
@paranoidi Hmm, not sure I agree about making it default, but I guess a small number wouldn't hurt. What were you thinking?

metaMMA:
I'm not too sure on what a good default would be. I feel like most tasks won't be run more often than once every ten minutes, so a `jitter=45` seems like it would work well. However, if there are people running tasks to scan directories for files every minute or two (for example), I'm not sure what the consequences would be...

gazpachoking:
I'm sorta leaning to leaving it as 0 default, I think it would be surprising for users to set up a schedule that isn't followed exactly. Especially if you are setting up local tasks that run frequently. If we do make it default, I'd vote for something very small, 15 seconds or even less (the variability is 2*jitter, since it can move the time plus or minus the jitter value.)

metaMMA:
I'm leaning that way, too. In general, if you set a schedule, you expect it to be followed on time. 

User input being required to turn jitter on makes sense to me.

gazpachoking:
@paranoidi Feel free to input more, but closing this out for the moment based on current discussion.

PR #2492, opened 2019-11-12
Prevent errors on PirateBay plugin when scanning ads

mfonville:
### Motivation for changes:
Currently the Piratebay plugin gives errors when scanning their site, because of advertisements that are also marked as class="detLink". This change detects if the ULed user was piratebay, which indicates that it is an advertisement.

```
DEBUG    utils.requests Series-Trakt    GETing URL ... with args () and kwargs {'allow_redirects': True, 'timeout': 30}
DEBUG    plugin        Series-Trakt    decorator caught ValueError. handled traceback:
  File "flexget/plugin.py", line 119, in wrapped_func
    return func(*args, **kwargs)
  File "flexget/components/sites/sites/piratebay.py", line 164, in search
    entry['torrent_seeds'] = int(tds[-2].contents[0])
ERROR    discover      Series-Trakt    Error searching with piratebay: invalid literal for int() with base 10: '\n\t\t\t\t'

```

### DIFF
--- a/flexget/components/sites/sites/piratebay.py
+++ b/flexget/components/sites/sites/piratebay.py
@@ -164,7 +164,8 @@ class UrlRewritePirateBay(object):
                     href = self.url + href
                 entry['url'] = href
                 row = link.parent.parent.parent
-                if row.find_all('a', attrs={'class': 'detDesc'})[0].contents[0] == "piratebay ":
+                description = row.find_all('a', attrs={'class': 'detDesc'})
+                if description and description[0].contents[0] == "piratebay ":
                     log.debug('Advertisement entry. Skipping.')
                     continue
                 tds = row.find_all('td')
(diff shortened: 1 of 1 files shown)


flexget/components/sites/sites/piratebay.py
```diff
@@ -163,7 +163,11 @@ def search(self, task, entry, config=None):
                 if href.startswith('/'):  # relative link?
                     href = self.url + href
                 entry['url'] = href
-                tds = link.parent.parent.parent.find_all('td')
+                row = link.parent.parent.parent
+                if row.find_all('a', attrs={'class': 'detDesc'})[0].contents[0] == "piratebay ":
```
paranoidi: if no `class: detDesc` is found this will crash, more robust approach would be preferable

mfonville: Ok, I tried to make it more robust

mfonville:
@paranoidi is this ready for you to merge?

paranoidi:
Sure, thanks for contributions!

PR #2493, opened 2019-11-13
update Dockerfile to use python3 and add webUI v2

X-dark:
This is the updated Dockerfile I use with success for a while. The one in develop uses python2 which is obviously not working anymore.

I am pulling the webUI v2 from the last release on respective GitHub, instead of building it. This could be changed if needed.

I don't build webUI v1 because it is deprecated.

I am using it mounting my config folder in /root/.flexget and with the command `daemon start`.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,13 +1,22 @@
 FROM alpine
 MAINTAINER Rémi Alvergnat <toilal.dev@gmail.com>
 
-RUN apk add --update python py-pip ca-certificates && rm -rf /var/cache/apk/*
+RUN apk add --update python3 ca-certificates nodejs build-base python3-dev libffi-dev openssl-dev unzip && \
+    rm -rf /var/cache/apk/*
 
 ADD . /opt/flexget
 WORKDIR /opt/flexget
 
-RUN pip install paver
-RUN pip install -e .
+RUN pip3 install -e . && \
+    pip3 install transmissionrpc && \
+    pip3 install deluge-client && \
+    pip3 install cloudscraper
+
+WORKDIR /opt/flexget/flexget/ui/v2
+RUN wget https://github.com/Flexget/webui/releases/latest/download/dist.zip && \
+    unzip dist.zip && \
+    rm dist.zip
+WORKDIR /opt/flexget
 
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -1,13 +1,22 @@
...
-RUN apk add --update python py-pip ca-certificates && rm -rf /var/cache/apk/*
+RUN apk add --update python3 ca-certificates nodejs build-base python3-dev libffi-dev openssl-dev unzip && \
+    rm -rf /var/cache/apk/*
 
 ADD . /opt/flexget
 WORKDIR /opt/flexget
 
-RUN pip install paver
-RUN pip install -e .
+RUN pip3 install paver
```
gazpachoking: Is paver needed for something else? We haven't used it for flexget for quite a while.

Dockerfile
```diff
@@ -1,13 +1,22 @@
...
+RUN pip3 install -e .
+RUN pip3 install transmissionrpc
```
gazpachoking: Hmm, should we include other optional deps for common plugins, like deluge-client?

Dockerfile
paranoidi: Paver is not needed for sure. Also you might want to chain these run commands together (&&) as each command creates unnecessary docker layer.

X-dark: Yes, just forgot to remove it. Done.

Dockerfile
X-dark: Yes make sense. Do you have a corresponding dep list?

gazpachoking: We don't. I guess let's just put deluge-client in here and we can add other things later if we deem necessary.

X-dark: OK. Added the dep.

Dockerfile
```diff
@@ -1,13 +1,21 @@
 FROM alpine
 MAINTAINER Rémi Alvergnat <toilal.dev@gmail.com>
```
X-dark: Is this maintainer field still accurate?

gazpachoking: No idea. @Toilal still around?

PR #2501, opened 2019-11-18
Support Python 3.8

gazpachoking:
### Motivation for changes:
Work with python 3.8

### Detailed changes:
- Updated dependencies to versions that work on 3.8
- Fix an issue with modifying a dictionary during iteration
- Start CI testing on 3.8

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -21,7 +21,7 @@ jobs:
     runs-on: ${{ matrix.operating-system }}
     strategy:
       matrix:
-        python-version: [3.6, 3.7]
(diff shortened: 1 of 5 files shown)


metaMMA:
:grin:

grr:
Flexget with Python 3.8 fails for me without the following patch:
https://pastebin.com/4dbATFfp

gazpachoking:
@grr That first change you mention should only cause a SyntaxWarning, not stop execution (but I have fixed it, thanks for finding that.) The other change was already included with the 3.8 pull request. We had a problem with the release right after, so you probably still had an older version.

PR #2505, opened 2019-11-20
[fix] Update TVDB Episode Schema

BrutuZ:
### Motivation for changes:
TVDB changed the key name and type for episodes in their v3 update from `'director': 'string'` to `'directors': ['array']`

[As noted on their documentation](https://api.thetvdb.com/swagger#!/Episodes/get_episodes_id):
> Returns the full information for a given episode id. **Deprecation Warning:** The director key will be deprecated in favor of the new directors key in a future release.

Which causes the following error during episode metadata lookup:
`WARNING  api_tvdb      task     Error while updating from tvdb ('director'), using cached data.`
### Detailed changes:
- Renamed field and convert list to string


gazpachoking:
Hmm, I think they need an update on what 'deprecated' means. 😛
Thanks!

PR #2508, opened 2019-11-23
Add rename config on symlink as move plugin

guillaumelamirand:
### Motivation for changes:
I wanted to be able to override target name when using symlink as it can be done with move plugin.

### Detailed changes:
- Add entry link_to property to set the link target name
- Add config rename in order to override target link name
- Clean target link with pathscrub

```
symlink:
      link_type: hard
      existing: ignore
      to: /Series/{{series_name}}/Season {{series_season|pad(2)}}
      rename: '{{series_name}} - {{series_id}}{{filename | pathext}}'
```
```
2019-11-23 10:40 VERBOSE  symlink       sort_series     hardlink `/download/S.W.A.T.2017.S01E21.FRENCH.720p.WEB.H264-AMB3R/s.w.a.t.2017.s01e21.french.720p.web.h264-amb3r.mkv` to `/Series/S W A T 2017/Season 01/S W A T 2017 - S01E21.mkv`
```

### DIFF
--- a/flexget/plugins/output/symlink.py
+++ b/flexget/plugins/output/symlink.py
@@ -8,7 +8,6 @@ from flexget.utils.template import RenderError, render_from_entry
 
 log = logging.getLogger('symlink')
 
-
 class Symlink:
 
     schema = {
@@ -29,7 +28,6 @@ class Symlink:
         ]
     }
 
-    destination_field = 'link_to'
 
     def prepare_config(self, config):
         if not isinstance(config, dict):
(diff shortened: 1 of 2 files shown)


flexget/plugins/output/symlink.py
```diff
@@ -27,6 +29,8 @@ class Symlink:
         ]
     }
 
+    destination_field = 'link_to'
```
paranoidi: very weird place for a constant in python, and we don't generally have constants for entry fields

please inline this variable

flexget/plugins/output/symlink.py
```diff
@@ -47,14 +51,41 @@ def on_task_output(self, task, config):
                 entry.fail('Does not have location field for symlinking')
                 continue
             lnkfrom = entry['location']
-            name = os.path.basename(lnkfrom)
-            lnkto = os.path.join(config['to'], name)
+            lnkfrom_path, lnkfrom_name = os.path.split(lnkfrom)
```
paranoidi: I know you did not name this variable, but "lnk" over "link" is saving ONE character and makes code harder to read :)

paranoidi:
Hi, we really appreciate new contributors! Hope to see more :)

flexget/plugins/output/symlink.py
guillaumelamirand: I have based my changes following move code, I felft wierd about it but the same thing it does in move code. The constant is not realy necessary here I deleted it.

flexget/plugins/output/symlink.py
guillaumelamirand: Agreed on this!

paranoidi:
Thanks! I updated the wiki

metaMMA:
> Thanks! I updated the wiki
@paranoidi I like the wiki entry. I hope you don't mind, I made a few clarifications. Feel free to revert or edit, as you see fit. There are two parts I'm unclear on.
'Value replacement' - it may be unclear to newer users what this means. I think this pertains to using jinja replacements from 'known fields', right?
for `existing: ignore` - does that replace the existing symlink with the new one?

guillaumelamirand:
> 2\. for `existing: ignore` - does that replace the existing symlink with the new one?

By default symlink with ignore, just doesn't create a new link. It can be a feature to add the availability to overwrite it. I can do another PR for that.

metaMMA:
I'm still new to working on FlexGet, and I don't quite understand how the behavior of this plugin changes when using `existing: ignore` versus when using `existing: fail`

guillaumelamirand:
When using ignore it just does nothing if an existing file is existing,
when using fail the entry is marked has failed.

PR #2514, opened 2019-12-02
Newznab update

metaMMA:
### Motivation for changes:
Searches on unlisted categories cannot currently be executed. This allows a search across all categories
### Detailed changes:
- added a third search option besides `do_search_movies` and `do_search_tvsearch`. Method is called `do_search_all`

same as before, except it can now accept `category: all`

#### To Do:

- [ ] Discuss whether we should work on building out the categories

#### Developement credit to flexget user 'SBan'


flexget/components/sites/sites/newznab.py
```diff
@@ -123,6 +127,11 @@ def do_search_movie(self, arg_entry, task, config=None):
         url = config['url'] + '&imdbid=' + imdb_id
         return self.fill_entries_for_url(url, task)
 
+    def do_search_all(self, arg_entry, task, config=None):
+        log.info('Searching for %s' % arg_entry['title'])
```
liiight: please replace the `%s` with a `,` to let the logger do its own string interpolation

flexget/components/sites/sites/newznab.py
```diff
@@ -123,6 +127,11 @@ def do_search_movie(self, arg_entry, task, config=None):
...
+        url = config['url'] + '&q=%s' % arg_entry['title']
```
liiight: fstrings ftw

metaMMA:
@liiight I simply copied the original code and used it for searching all categories. I made the two updates you suggested. Should I go ahead and make those two changes in the original author's methods, as well?

paranoidi:
@metaMMA do as you see best, that's pretty much how we operate here. Let us know when to merge :)

metaMMA:
@gazpachoking @liiight 
I think this one is ready to be merged.

flexget/components/sites/sites/newznab.py
```diff
@@ -102,14 +106,10 @@ def do_search_tvsearch(self, arg_entry, task, config=None):
...
+            lookup = f"&rid={arg_entry.get('tvrage_id')}"
         else:
-            lookup = '&q=%s' % quote(arg_entry['series_name'])
-        url = (
-            config['url']
-            + lookup
-            + '&season=%s&ep=%s' % (arg_entry['series_season'], arg_entry['series_episode'])
-        )
+            lookup = f"&q={quote(arg_entry['series_name'])}"
+        url = f"{config['url']}{lookup}&season={arg_entry['series_season']}&ep={arg_entry['series_episode']}"
```
liiight: i think that if you use the params keyword, you could avoid using the urlquote as it does so internally via requests:
```python
params = {
    'season': arg_entry['series_season'],
    'ep': arg_entry['series_episode'],
    'q': arg_entry['series_name'],
}
```

metaMMA: I think I formatted everything correctly. Found another place where we needed to make sure 'title' was encoded for URLs, too. Removed import, and changed one line to use f-strings.

metaMMA:
@liiight How's it look, now?

flexget/components/sites/sites/newznab.py
```diff
@@ -102,14 +106,12 @@ def do_search_tvsearch(self, arg_entry, task, config=None):
...
-            lookup = '&q=%s' % quote(arg_entry['series_name'])
-        url = (
-            config['url']
-            + lookup
-            + '&season=%s&ep=%s' % (arg_entry['series_season'], arg_entry['series_episode'])
-        )
+            params = {'q': arg_entry['series_name']}
+        params['season'] = arg_entry['series_season']
+        params['ep'] = arg_entry['series_episode']
+        url = f"{config['url']}{urlencode(params)}"
```
liiight: it's be best if you passed params as its own dict and then you wouldn't have to take care of the econding or concating the url. Let reqeuest handle that by passing it to the `params` keyword

liiight:
@metaMMA just that params change and then we're good to go

metaMMA:
@liiight I don't understand why it's showing this conflict. Otherwise, I think it should be good. 

I've never used requests with parameters, so I followed some tutorials and basically copied/pasted what I thought it should look like. Let me know if there is anything left to amend.

gazpachoking:
I probably would have left the 'params' dict separate from the config, just a style thing, to keep the user specified 'config' separate from our computed params. Not a big enough deal I think it's a blocker though. LGTM

PR #2517, opened 2019-12-04
Entry serialization

gazpachoking:
### Motivation for changes:
Have a lossless format to serialize entries so they can be stored and restored from database.

### Detailed changes:
- Adds a utility that allows python classes to define how to serialize and deserialize themselves

#### To Do:

- [x] Lazy fields
- [x] Convert remaining lazy lookups to the new system
- [x] Add a common method to store entries to DB
- [x] Handle nested objects? e.g. a list of Quality objects
- [x] Store native json serializable stuff without all the overhead?
- [x] Transition appropriate plugins to new db storage
  - [x] Backlog
  - [x] Others? (everything using entry_synonym is transitioned)


gazpachoking:
Should we try to drop the overhead for types that are natively json serializable? We could just check if each item is a dict, and has our special keys, and if so do our special deserialization, if not, keep the native json type.

flexget/tests/test_serialization.py
```diff
@@ -0,0 +1,26 @@
...
+            'datetimefield': datetime.datetime(1999, 9, 9, 9, 9),
+            'qualityfield': qualities.Quality('720p hdtv'),
+            'nestedlist': [qualities.Quality('1080p')],
+            'nesteddict': {'a': datetime.date(1999, 9, 9)},
+        })
+        serialized = entry1.dumps()
+        print(serialized)
+        entry2 = entry.Entry.loads(serialized)
+        # Use the underlying dict, so we compare all fields
+        assert entry1.store == entry2.store
```
liiight: Maybe add a test that `dict(entry1) == dict(entry2)`?

flexget/utils/serialization.py
```diff
@@ -0,0 +1,145 @@
...
+        if Serializable._registry is None:
+            Serializable._registry = {c.serializer_name(): c for c in Serializable.__subclasses__()}
+        return Serializable._registry
+
+    @abstractmethod
+    def _serialize(self):
+        """Return a plain python datatype which is json serializable."""
+        pass
+
+    def serialize(self):
```
liiight: May I be so bold to suggest adding some return types? Especially for public api methods like this

flexget/utils/serialization.py
```diff
@@ -0,0 +1,145 @@
...
+    """
+    if isinstance(value, Serializable):
+        return value.serialize()
+    if isinstance(value, datetime.datetime):
+        return DateTimeSerializer.serialize(value)
+    if isinstance(value, datetime.date):
+        return DateSerializer.serialize(value)
+    elif isinstance(value, (str, int, float, dict, list)):
+        return BuiltinSerializer.serialize(value)
+    else:
```
liiight: else is redundant here

flexget/tests/test_serialization.py
gazpachoking: That's what this is, but it avoids any of the lazy field or other magic Entry normally might do.

flexget/entry.py
```diff
@@ -194,13 +197,13 @@ def __setitem__(self, key, value):
         # url and original_url handling
         if key == 'url':
             if not isinstance(value, (str, LazyLookup)):
-                raise PluginError('Tried to set %r url to %r' % (self.get('title'), value))
+                raise plugin.PluginError('Tried to set %r url to %r' % (self.get('title'), value))
```
liiight: maybe make those fstrings while you're at it

flexget/entry.py
```diff
@@ -301,3 +334,35 @@ def __hash__(self):
...
+        self._func = None
+
+    def __call__(self, func):
+        self._func = func
+        lazy_func_registry[self.name] = self
+        return func
+
+    @property
+    def function(self):
+        if '.' in self._func.__qualname__:
```
liiight: maybe something in the `inspect` module can help with something more deterministic than this, or do you think that is robust enough?

flexget/utils/serialization.py
```diff
@@ -0,0 +1,129 @@
...
+        return [serialize(v) for v in value]
+    if isinstance(value, dict):
+        return {k: serialize(v) for k, v in value.items()}
+    if isinstance(value, datetime.datetime):
+        return DateTimeSerializer.serialize(value)
+    if isinstance(value, datetime.date):
+        return DateSerializer.serialize(value)
+    if isinstance(value, (str, int, float)):
+        return value
+    raise TypeError('%r is not serializable', value)
```
liiight: should be fstring

flexget/entry.py
gazpachoking: I think it's probably fine. I'd say let's wait until we find an example where it causes a problem to change it.

gazpachoking:
Ran in to a road block with the lazy lookup serialization. Set plugin uses lazy lookups, but they are created dynamically, which doesn't work with the system I was creating in this pull. Anyone have a good idea?

metaMMA:
@gazpachoking I think this pull request renders mine obsolete: https://github.com/Flexget/Flexget/pull/2485

Should I go ahead and close mine?

gazpachoking:
Overcame the issue with serializing dynamic lazy fields. Still need to convert remaining lazy lookups to the new system though.

flexget/utils/serialization.py
```diff
@@ -0,0 +1,128 @@
...
+    if isinstance(value, dict):
+        if all(key in value for key in ('serializer', 'version', 'value')):
+            return _registry()[value['serializer']].deserialize(value)
+        return {k: deserialize(v) for k, v in value.items()}
+    if isinstance(value, list):
+        return [deserialize(v) for v in value]
+    return value
+
+
+class Serializable(ABC):
```
paranoidi: Some docstring would be nice, how and when to use this.

flexget/utils/serialization.py
```diff
@@ -0,0 +1,128 @@
...
+    @abstractmethod
+    def _serialize(self):
```
paranoidi: I don't think abstractmethods are generally prefixed with _, which heavily implies internal functionality not to be messed with. More confusingly there is public `serialize` method which is NOT overridden.

Very confusing situation with `serialize` and `_serialize`.

flexget/utils/serialization.py
```diff
@@ -0,0 +1,128 @@
...
+class DateSerializer(Serializable):
+    @classmethod
+    def serializer_name(cls):
+        return 'date'
+
+    def _serialize(self):
+        pass
+
+    @classmethod
+    def serialize(cls, value):
```
paranoidi: Method signature does not match base method. Changing method signature in subclass is bad practice. Type hinting/argument naming would make this more readable as well. I think `value` here refers to a entry field, so why not call the argument `field`. Or even better `datetime: field` if it only serializes that type of fields.

flexget/utils/serialization.py
```diff
@@ -0,0 +1,128 @@
...
+    @classmethod
+    def _deserialize(cls, data, version):
+        return datetime.datetime.strptime(data, ISO8601_FMT)
+
+
+_registry_cache = None
+
+
+def _registry():
+    global _registry_cache
```
paranoidi: If you initialize `_registry_cache` as dict straight away you don't need to use global or make checks and initialize it as adding to dict works fine without global.

flexget/utils/serialization.py
gazpachoking: How are you thinking it should work? The idea was that `serialize` is the public method that can be called on any instances of a serializable object, and that `_serialize` is the private implementation, not called external to the class. I based it on how [marshmallow works](https://github.com/marshmallow-code/marshmallow/blob/dev/src/marshmallow/fields.py#L367)

flexget/utils/serialization.py
gazpachoking: The return type is 'any json serializable type', which I wrote in the docstring, but am not sure how better to document.

flexget/utils/serialization.py
gazpachoking: Yeah, I went a bit back and forth on how to implement this. It doesn't work quite the same as other things, because it's for serializing a bulitin type, which we don't directly subclass like we do for our own custom types. I sort of emulated how the custom types work so that the deserialization didn't need any custom handling.

flexget/utils/serialization.py
gazpachoking: True. I guess I was thinking about telling the difference from when there were no subclasses defined, but that situation doesn't matter at all.

flexget/utils/serialization.py
gazpachoking: K.

flexget/utils/serialization.py
gazpachoking: Hmm. I know this works fine, but it still irks me that PyCharm highlights this. I'm tempted to leave the global in there just to avoid it. 😛 
image

gazpachoking: Hmm. Actually, it does not work fine. Is this a new behavior in later pythons?

gazpachoking: Doh. Just me being silly. I was assigning to that variable rather than mutating it.

gazpachoking: > If you initialize `_registry_cache` as dict straight away you don't need to use global or make checks and initialize it as adding to dict works fine without global.

Still needs to make checks and initialize, otherwise it gets initialized at the module level, which is before all the other subclasses of Serializable have been created.

gazpachoking:
Refactored the system a bit in order to have serializers work the same for built in types, and our own types. The methods on classes should no longer be called directly now (e.g. `Entry().serialize()` ❌ ,) instead, the module level functions should be the only interface (`serialize(Entry())` ✔️ .)

gazpachoking:
Pulling the trigger.

PR #2519, opened 2019-12-06
Archive search clean

soloam:
Clean the searches from colon form more exact searches

With this search movies like "Playmobil: The Movie (2019)" can be searched and returned in the archive!


paranoidi:
And here I thought I am only one using the archive plugin. It is very slow with a lot of entries though data should be purged from there after couple of years .. in case you want to step and improve it :)

Thanks for this!

soloam:
I use it a lot to make discover work in websites that only have rss! :)

Who can I purge it? Is there like a cli command that removes torrents with some years? It would be nice to remove torrents from archive when they give several 404 on download!

Thank you

paranoidi:
@soloam  There's even fired that should remove any old items (`@event('manager.db_cleanup')`), lets say over 2 years. Removing failed entries from archive would not be a bad idea either, causing some other possibly better one being tried next time.

paranoidi:
@soloam another issue with current implementation is that it does not really work that well or quickly. There are now much better ways to do this that I am aware of. SQLite does support full text searching:

https://sqlite.org/fts5.html

https://stackoverflow.com/questions/49893974/how-to-use-sqlalchemy-to-create-a-full-text-search-index-on-sqlite-and-query-it

But this gets pretty advanced use quickly ..

@gazpachoking if we have built in support for persisting entry lists. It should probably support fts, and expiring entries (ie. the db_cleanup) ?

Could we somehow make it extensible so that plugins like archive can use it? By adding new tables like `Tag` etc.

gazpachoking:
@paranoidi I'm no longer quite so sure about the common db entry storage, due to the mutability of entries, but I think we could still use the new serialization technique to improve our current `entry_synonym` tool. https://github.com/Flexget/Flexget/blob/develop/flexget/utils/database.py#L81 This would make it easy to for any plugin to store entries into the database, and improve the current version, because it would allow custom class support, as well as lazy fields.

PR #2524, opened 2019-12-09
[change] `tvdb_lookup` - Always have the main poster as the first entry of `tvdb_posters`

BrutuZ:
### Motivation for changes:
TheTVDB returns posters in a random order and Flexget caps the list to 5 entries. I like to use the main poster (curated to be used by the site) on my notifications and given the aforementioned actions, it's not always on the list, and when it is there is no way to determine which entry t is.
### Detailed changes:
- ~~[unrelated] Changed URL for the posters CDN to the new one post TheTVDB's v3 migration. There are currently 301 redirects in place at the old URL but they did recommend changing it wherever possible.~~
- Request the main poster path from the `/series/{id}` endpoint and start the list with it
- Append up to 5 posters to the list (not 4 just in case the main one is not defined)
- Trim the list down to 5 entries

### Addressed issues:
- Fixes *my OCD*.

### DIFF
--- a/flexget/components/thetvdb/api_tvdb.py
+++ b/flexget/components/thetvdb/api_tvdb.py
@@ -24,7 +24,7 @@ SEARCH_RESULT_EXPIRATION_DAYS = 3
 class TVDBRequest:
     API_KEY = '4D297D8CFDE0E105'
     BASE_URL = 'https://api.thetvdb.com/'
-    BANNER_URL = 'https://artworks.thetvdb.com/banners/'
+    BANNER_URL = 'http://thetvdb.com/banners/'
 
     def __init__(self, username=None, account_id=None, api_key=None):
         self.username = username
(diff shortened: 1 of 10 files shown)


flexget/components/thetvdb/api_tvdb.py
```diff
@@ -257,10 +257,12 @@ def get_posters(self):
         if not self._posters:
             log.debug('Getting top 5 posters for series %s', self.name)
             try:
+                poster_main = TVDBRequest().get('series/%s' % self.id).get('poster')
```
liiight: please use fstring

BrutuZ: Sure thing. Only this new line or should I go ahead and convert all other `%s` replacements on the file for f-strings too?

liiight: All please 😁

BrutuZ: On it 👍 .
Just one more question (I hope): Leave log entries alone to let it handle the parameters internally or convert them to f-strings too?

liiight:
Yes, leave logger to do it's own interpolation

liiight:
Thanks for this!

PR #2525, opened 2019-12-11
Support qBittorrent v4.1+, fixes #2522

pospqsjac:
### Motivation for changes:
Plugin qBittorrent stopped working with qBittorrent v4.2

### Detailed changes:
- Update qBittorrent WebAPIm, [Web API Documentation](https://github.com/qbittorrent/qBittorrent/wiki/Web-API-Documentation)

### Addressed issues:
- Fixes #2522

PR #2527, opened 2019-12-11
Fixes wrong API adresses in #2525

pospqsjac:
### Motivation for changes:
Incorrect address for adding torrents in older API version in qBittorrent.


paranoidi:
Thanks!

PR #2528, opened 2019-12-11
Allow downloading in sequential order in qBittorrent

pospqsjac:


paranoidi:
Looks good otherwise but the option name is too verbose / long.

`sequential_download: yes` would be good.

flexget/plugins/clients/qbittorrent.py
```diff
@@ -47,6 +47,7 @@ class OutputQBitTorrent:
                     'maxdownspeed': {'type': 'integer'},
                     'fail_html': {'type': 'boolean'},
                     'add_paused': {'type': 'boolean'},
+                    'download_in_sequential_order': {'type': 'boolean'},
```
paranoidi: Rename

PR #2529, opened 2019-12-12
Loguru!

gazpachoking:
### Motivation for changes:
Switch to simpler and more modern logging system.

### Detailed changes:
- Support bracket lazy formatting of log messages. % formatting can be left behind completely now
- Split the capture system so logs and stdout are captured separately. Logs can now be captured in their unstructured form instead of in already formatted text only.
- Change our naming convention for logging:
    - All loggers (the things you call .info etc. on) are called `logger` as in the standard for loguru
    - Our logging file has been renamed to `log`

### To Do:

- [x] Convert plugins to using luguru logger instead of stdlib. (Probably make a script or regex to facilitate this.)
  - [x] Do we switch to the loguru standard of calling the logger `logger` instead of `log`? (yes)
- [x] Is it awkward to have `flexget.logger` module named so similarly to the loguru `logger`? (renamed `log`)


kristenmills:
> Do we switch to the loguru standard of calling the logger `logger` instead of `log`.  Is it awkward to have `flexget.logger` module named so similarly to the `loguru `logger`?

I think the loguru standard makes more sense semantically.  I think if the module is renamed to `flexget.logging` or something of the sort, i think it seems like it will feel less awkward. 

```py3
from loguru import logger
from flexget.logging import capture_logs

logger_ctx = logger.bind(name='random')
```

gazpachoking:
@kristenmills I think I agree on both points. But if we are going to call it logger, I think it's fine to have the contextual one named the same. i.e.
```logger = logger.bind(name='blah')```

gazpachoking:
Hmm, depending on pycharm settings, naming our file 'logging.py' can mess things up with it getting confused with the stdlib logging module. With proper pycharm settings it can be avoided, but I wonder if we shouldn't avoid that hazard. Not sure what to call it in that case, as I sorta don't want it to conflict with loguru 'logger' either.

liiight:
Maybe `flexget_logger`?

gazpachoking:
> Maybe `flexget_logger`?

flexget.flexget_logger feel a bit redundant. Now I'm reconsidering naming it just logging, which does require setting up pycharm correctly, but I don't think that's as bad anymore.

paranoidi:
Why not just use name `log`, it's short, simple and does not overlap with logging?

gazpachoking:
> Why not just use name `log`, it's short, simple and does not overlap with logging?

I like it.

PR #2530, opened 2019-12-12
Fix for "UnboundLocalError: local variable 'r' referenced before assi…

planeturban:
…gnment"

Return blank entries list instead of crashing when failing to fetch rss.

### Motivation for changes:
fill_entries_for_url() keeps crashing when nzb.su have intermittent problems. 

### Detailed changes:
````
        try:
            r = task.requests.get(url)
        except RequestException as e:
            log.error("Failed fetching '%s': %s" % (url, e))

        rss = feedparser.parse(r.content)
````  
Since r never gets defined with the try block I've added an else to the block to keep it from crashing. It'll return a blank entry list instead.

### Addressed issues:
- Fixes crash.

N/A
N/A

N/A (It doesn't crash anymore. :) )
#### To Do:
Nothing


planeturban:
I'll merge this in a couple of days, after real-life tests has been run for more than 30 minutes. :)

(If no one minds, that is.)

gazpachoking:
Looks good to me 👍

paranoidi:
PR needs refreshing after logging upgrade.

planeturban:
Done and done. Merging.

PR #2532, opened 2019-12-13
add IRC password config entry

nozn:
### Motivation for changes:

Some servers and IRC bouncers require a password to connect. The option is available in the IRC library, but was not exposed through flexget's config. 

### Detailed changes:
- 

### Addressed issues:
- Fixes # .

```
irc:
  foo-irc:
    server: foo.bar.baz
    password: supersecretpassword
```
```
```
#### To Do:

- [x] it's a trivial change, and it works for me, but it might be useful to write some sort of test to validate that.


paranoidi:
Did you test passwordless servers after the change?

nozn:
Yes, when the password is None, SimpleIrcBot does not attempt to use it.

PR #2541, opened 2019-12-23
Colorize entry action in logs

liiight:
### Motivation for changes:
It'd be nice to have some distinction in logs between accept/reject.

### Detailed changes:
- Colorized entry action in log

#### To Do:

- [x] This currently won't work as it's pending https://github.com/Delgan/loguru/pull/193

### DIFF
--- a/flexget/entry.py
+++ b/flexget/entry.py
@@ -19,12 +19,15 @@ class EntryState(Enum):
 
     @property
     def color(self) -> str:
-        return {self.ACCEPTED: 'g', self.REJECTED: 'r', self.FAILED: 'R', self.UNDECIDED: 'd',}[
-            self
-        ]
+        return {
+            self.ACCEPTED: 'green',
+            self.REJECTED: 'red',
+            self.FAILED: 'RED',
+            self.UNDECIDED: 'dim',
+        }[self]
 
     @property
-    def log_color(self) -> str:
+    def log_markup(self) -> str:
         return f'<{self.color}>{self.value.upper()}</>'
 
 
--- a/flexget/entry.py
+++ b/flexget/entry.py
@@ -19,12 +19,9 @@ class EntryState(Enum):
 
     @property
     def color(self) -> str:
-        return {
-            self.ACCEPTED: 'g',
-            self.REJECTED: 'r',
-            self.FAILED: 'R',
-            self.UNDECIDED: 'd',
-        }[self]
+        return {self.ACCEPTED: 'g', self.REJECTED: 'r', self.FAILED: 'R', self.UNDECIDED: 'd',}[
+            self
+        ]
 
     @property
     def log_color(self) -> str:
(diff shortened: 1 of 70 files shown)


gazpachoking:
I re-implemented the verbose method as discussed in https://github.com/Delgan/loguru/issues/192, so this should work properly now.

liiight:
I'll verify and merge

liiight:
I ran `autoformat` and it formatted a lot of unrelated stuff... I think we should integrate pre-commit soon

flexget/entry.py
```diff
@@ -10,6 +11,23 @@
...
 
+class EntryState(Enum):
+    ACCEPTED = 'accepted'
+    REJECTED = 'rejected'
+    FAILED = 'failed'
+    UNDECIDED = 'undecided'
+
+    @property
+    def color(self) -> str:
+        return {self.ACCEPTED: 'g', self.REJECTED: 'r', self.FAILED: 'R', self.UNDECIDED: 'd',}[
```
gazpachoking: I think we could afford full names here rather than abbreviations, just for clarity's sake. I had to look up what 'd' was.

flexget/entry.py
```diff
@@ -10,6 +11,23 @@
...
+            self
+        ]
+
+    @property
+    def log_color(self) -> str:
+        return f'<{self.color}>{self.value.upper()}</>'
```
gazpachoking: Not sure if it's worth putting this in here. The name is sorta confusing if we do leave it. Maybe `log_markup`?

gazpachoking:
We should investigate having the autoformat command only run on files change

> I ran `autoformat` and it formatted a lot of unrelated stuff... I think we should integrate pre-commit soon

Agreed on pre-commit, although I tried it on my work (windows) machine and it was not easy to get set up here. If I remember correctly, it went smoothly on my home windows machine, I'll add the pre-commit script when I get there and can test it properly. Also, maybe we should modify the autoformat dev tool to only run on files changed in the git working tree by default, and require some extra param to do everything.

flexget/entry.py
liiight: I agree, although it was worth it to see you write that you had to look up the D 😆

flexget/entry.py
liiight: Agreed on the name. I kind of like adding these enhancements to enums, I use it a lot in other projects. Seems more logical in this case than to hold a dict of value to colors somewhere else

liiight:
Yeah, I should have probably skipped yhecautoformats stuff but I saw that it didnt render the changes completely unreadable.
I definitely agree that we should use this and a lot more enums in our codebase. Enums are one of the nicest features that python3 introduced IMO, and we should definitely work them in

gazpachoking:
@liiight I like that simple escaping solution. We can figure out a better way if we start coloring more stuff and need to.

liiight:
I agree. This was a bit trickier since we generically markup messages. I imagine that if we'd want to use this in a more specific use case it won't be an issue

PR #2550, opened 2019-12-30
[change] trakt_calendar: automatically handle more than 31 days

cvium:
### Motivation for changes:
Trakt API only allows up to 31 days.
### Detailed changes:
- Automatically call the API the required number of times to get the full calendar.

### DIFF
--- a/flexget/components/trakt/trakt_calendar.py
+++ b/flexget/components/trakt/trakt_calendar.py
@@ -1,4 +1,5 @@
 import datetime
+import math
 
 from loguru import logger
 from requests import RequestException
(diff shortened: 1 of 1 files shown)


flexget/components/trakt/trakt_calendar.py
```diff
@@ -86,54 +89,65 @@ class TraktCalendar:
...
-        for result in results:
-            e = Entry()
-            e.update_using_map(self.series_map, result['show'])
-            if config['type'] == 'episodes':
-                e.update_using_map(self.episode_map, result['episode'])
 
-            title = e['trakt_series_name']
-            if not config['strip_dates']:
-                title = '{0} ({1})'.format(title, e['trakt_series_year'])
+        for _ in range(number_of_calls):
```
gazpachoking: Probably fine the way you have it, something like this way feels a bit more pythonic to me though.
```
for start_day in range(0, config['days'], 31):
    num_days = min(config['days'] - start_day, 31)
```

gazpachoking:
LGTM, thanks!

PR #2556, opened 2020-01-04
Remove webui v1

kristenmills:
### Motivation for changes:
Web UI v2 has parity so no longer necessary to run them side by side

### Detailed changes:
- Remove webui v1 code
- Remove logic to build webui v1
- Move webui to run at root instead of /v2
- Add logging that the run_v2 is no longer necessary in config

### Log and/or tests output (preferably both):
```
web_server: yes
tasks: []
```
```
2020-01-03 23:13:50 INFO     web_server_daemon                 Running web server at IP 0.0.0.0:5050
2020-01-03 23:13:50 INFO     web_server_daemon                 Initiating API
2020-01-03 23:13:50 INFO     web_server_daemon                 Registering WebUI
2020-01-03 23:13:50 INFO     scheduler                     No schedules defined in config. Defaulting to run all tasks on a 1 hour interval.
2020-01-03 23:13:50 INFO     scheduler                     Starting scheduler
2020-01-03 23:13:50 INFO     web_server                    Web interface available at http://192.168.86.244:5050
```
```
web_server: 
  run_v2: yes
tasks: []
```
```
2020-01-03 23:14:29 INFO     web_server_daemon                 Running web server at IP 0.0.0.0:5050
2020-01-03 23:14:29 INFO     web_server_daemon                 Initiating API
2020-01-03 23:14:29 WARNING  web_server_daemon                 run_v2: true is redundant and can be removed
2020-01-03 23:14:29 INFO     web_server_daemon                 Registering WebUI
2020-01-03 23:14:29 INFO     web_server                    Web interface available at http://192.168.86.244:5050
2020-01-03 23:14:29 INFO     scheduler                     No schedules defined in config. Defaulting to run all tasks on a 1 hour interval.
2020-01-03 23:14:29 INFO     scheduler                     Starting scheduler
```
#### To Do:

- [ ] Update relevant wiki stuff

PR #2557, opened 2020-01-04
Move v2 to root

kristenmills:
### Motivation for changes:
Instead of getting rid of webuv1 right away (#2556), I'd like to move webui_v1 to `/v1/` and v2 to `/`

### Detailed changes:
- Move `/` -> `v1`
- Redirect `/v1/api` -> `/api`
- Move `/v2` -> `/`
```yaml
web_server:
   run_v1: yes
```

### DIFF
--- a/flexget/plugins/daemon/web_server.py
+++ b/flexget/plugins/daemon/web_server.py
@@ -25,7 +25,10 @@ web_config_schema = {
                 'ssl_private_key': {'type': 'string'},
                 'web_ui': {'type': 'boolean'},
                 'base_url': {'type': 'string'},
-                'run_v2': {'type': 'boolean'},
+                'run_v2': {
+                    'type': 'boolean',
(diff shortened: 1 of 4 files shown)


flexget/plugins/daemon/web_server.py
```diff
@@ -100,11 +102,14 @@ def register_web_server(manager):
     # Register WebUI
     if web_server_config.get('web_ui'):
         if web_server_config.get('run_v2'):
-            logger.info('Registering WebUI v2')
-            register_web_ui_v2(web_server_config)
+            logger.warning('Run V2 is deprecated and can be removed from the config')
```
liiight: We have a way to set a field as deprecated I think, not sure it'll work on a root level schema though... Better ask @gazpachoking

paranoidi:
Remember to increase version number for flexget. From run_v1 should probably drop the v1

liiight:
I think that a minor flexget version bump would be in order

PR #2567, opened 2020-01-12
Api schema

gazpachoking:
### Motivation for changes:
Add an api endpoint to get config schemas, for use by the web editor.

### Detailed changes:
- Schemas are available under /api/schema/
- All `$ref`s are rewritten to the actual uri that the referenced schema is available from
- `id` fields are added to each schema with their uri

### DIFF
--- a/flexget/api/core/schema.py
+++ b/flexget/api/core/schema.py
@@ -28,7 +28,6 @@ def rewrite_ref(identifier, base_url):
 def rewrite_refs(schema, base_url):
     """
     Make sure any $refs in the schema point properly back to this endpoint.
-
     """
     if isinstance(schema, dict):
         if '$ref' in schema:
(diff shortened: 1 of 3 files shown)


flexget/api/core/schema.py
```diff
@@ -0,0 +1,66 @@
...
+@api.doc(params={'path': 'Path of schema'})
+@api.response(NotFoundError)
+class SchemaAPI(APIResource):
+    @api.response(200, model=schema_api_list)
+    def get(self, path, session=None):
+        """ Get schema definition """
+        try:
+            schema = resolve_ref(request.full_path)
+        except RefResolutionError:
+            raise NotFoundError('invalid schema path')
```
liiight: I'd add the path to this exception

PR #2570, opened 2020-01-12
Add manual release github actions

kristenmills:
### Motivation for changes:
I added this to webui when we were switching over to github actions so i figured I'd  add it over in main repo as well. Basically this gives the ability to kick off a github deployment (which is super easy to do in slack) as a way to do a manual release if necessary.


gazpachoking:
LGTM!

PR #2573, opened 2020-01-14
Refactored hebits to use bs4 and upgrade bs4 (#2573). Closes #2571

### LINKED ISSUE #2571: Remove request_html dependency
Seems like `request_html` has a bunch of problematic dependencies. Hebits plugin will be refactored to use BS4 instead.

Related #2569

liiight:
### Motivation for changes:
Get rid of requests_html as it's not worth the added and problematic deps

### Detailed changes:
- Removed requests_html from deps
- Refactors hebits to ue BS4
- Upgraded BS4 version due to CSS selector upgrade

### Addressed issues:
- Fixes #2571

### DIFF
--- a/flexget/components/sites/sites/hebits.py
+++ b/flexget/components/sites/sites/hebits.py
@@ -108,6 +108,14 @@ class SearchHeBits:
     profile_link = f"{base_url}my.php"
 
     @staticmethod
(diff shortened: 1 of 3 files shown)


gazpachoking:
Other than that comment, looks good to me!

PR #2576, opened 2020-01-15
[FIX] http -> https in URLs

GreyKat:
### Motivation for changes:
Since yesterday was getting this eror in log
> 2020-01-14 12:55:04 CRITICAL plugin        LostFilm        Received not 200 (OK) status
>2020-01-14 12:55:04 WARNING  task          LostFilm        Aborting task (plugin: lostfilm)

### Detailed changes:
After this modification
```
-        if rss.get('status') != 200:
-            raise PluginError('Received not 200 (OK) status')
+        status = rss.get('status')
+        if status != 200:
+            raise PluginError('Received %s status instead of 200 (OK)' % status)
```
found out that actually http code was 301.
So finally fixing URL  to use https did the trick.

### DIFF
--- a/flexget/components/sites/sites/lostfilm.py
+++ b/flexget/components/sites/sites/lostfilm.py
@@ -28,7 +28,7 @@ quality_map = {
     'HD': ('720p', 'mkv'),
 }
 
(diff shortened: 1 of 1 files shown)


paranoidi:
Thanks

PR #2589, opened 2020-02-01
[fix] Reorder audio qualities based on bitrates

0x022b:
### Motivation for changes:

The order of audio qualities was incorrect.

### Detailed changes:
 
Though sound quality is mostly subjective, audio qualities are now
ordered based on maximum supported bitrate. Even though FLAC, Dolby
TrueHD and DTS-HD MA are all lossless audio codecs the current order
takes into account the fact that FLAC is not standard audio codec for
video containers and that Dolby TrueHD and DTS-HD MA both contain a
core bitstream that is compatible with older players. DTS codecs are
placed above equivalent Dolby codecs because those are usually the
recommended choice between the two.

### Addressed issues:

- Fixes TODOs in code

```
1314 passed, 28 skipped, 4 xfailed, 5 xpassed, 59 warnings in 397.51s (0:06:37)
```

### DIFF
--- a/flexget/utils/qualities.py
+++ b/flexget/utils/qualities.py
@@ -147,19 +147,18 @@ _codecs = [
     QualityComponent('codec', 40, 'h265', '[hx].?265|hevc'),
     QualityComponent('codec', 50, '10bit', '10.?bit|hi10p'),
 ]
-channels = r'(?:(?:[^\w+]?[1-7][\W_]?(?:0|1|ch)))'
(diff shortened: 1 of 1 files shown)


0x022b:
On a second thought I'm going to add unit tests to this pull request to make sure that AC-3/DD 5.1 < DD+ 5.1 < TrueHD and DTS < DTS-HD.

gazpachoking:
I'm good with it. I suppose we should we include a minor version bump and add it to UpgradeActions though, in case people were relying on the old behavior.

PR #2594, opened 2020-02-02
[fix] Series plugin doesn't pick the entry with best quality

0x022b:
### Motivation for changes:

Documentation and code both tell that series plugin will get the best quality in either given timeframe or within series entries but it doesn't actually do that. Series plugin assumes the first entry is the best quality entry.

### Detailed changes:

Order series entries by quality from best to worst. 

```
    @pytest.mark.parametrize('reverse', [False, True])
    def test_input_order_preserved(self, manager, execute_task, reverse):
        """If multiple versions of an episode are acceptable, make sure the first one is accepted."""
        entries = [
            Entry(title='Some Show S01E01 720p proper', url='http://a'),
            Entry(title='Some Show S01E01 1080p', url='http://b'),
        ]
        if reverse:
            entries.reverse()
        task = execute_task('test_input_order_preserved', options={'inject': entries})
>       assert task.accepted[0] == entries[0], 'first entry should have been accepted'
E       AssertionError: first entry should have been accepted
E       assert <Entry(title=...ate=accepted)> == <Entry(title=...te=undecided)>
E         -<Entry(title=Some Show S01E01 1080p,state=accepted)>
E         +<Entry(title=Some Show S01E01 720p proper,state=undecided)>
E         Full diff:
E         - <Entry(title=Some Show S01E01 1080p,state=accepted)>
E         ?                               ^^^         ^ ^^^^
E         + <Entry(title=Some Show S01E01 720p proper,state=undecided)>
E         ?                               ^^  +++++++       ^^^^ ^^

2 failed, 1314 passed, 28 skipped, 4 xfailed, 5 xpassed, 59 warnings in 383.88s (0:06:23)
```
#### To Do:

- [ ] Decide what to do with failing unit tests seen above

### DIFF
--- a/flexget/components/series/series.py
+++ b/flexget/components/series/series.py
@@ -733,8 +733,10 @@ class FilterSeries(FilterSeriesBase):
                     entry.reject('entity has already been downloaded')
                 continue
 
-            best = entries[0]
             logger.debug('continuing w. entities: {}', [e['title'] for e in entries])
+            # sort qualities from best to worst
+            entries.sort(key=lambda e: (e['quality']), reverse=True)
+            best = entries[0]
             logger.debug('best entity is: `{}`', best['title'])
 
             # episode tracking. used only with season and sequence based series
@@ -774,7 +776,7 @@ class FilterSeries(FilterSeriesBase):
                     continue
 
             # Just pick the best ep if we get here
-            reason = reason or 'choosing first acceptable match'
+            reason = reason or 'choosing best acceptable match'
             best.accept(reason)
 
             # need to reject all other episode/season packs for an accepted season during the task,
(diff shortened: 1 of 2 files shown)


0x022b:
The failing unit tests are in conflict with what the documentation says so I'm not sure if the test should be removed or updated to reflect to new behaviour which is inline with the documentation.

gazpachoking:
If we still have documentation that says the best quality is accepted, I think that's what we need to update. We removed the internal sorting from the series plugin so that the user could sort the entries based on other criteria (say, e.g. you want the worst quality within your acceptable quality range, or you want to prioritize releases from certain groups.)

If you are trying to get the old behavior, you just have to add the sort_by plugin in the task.
```yaml
sort_by:
  field: quality
  reverse: yes
```

gazpachoking:
For reference, here is where we removed the quality sorting. https://github.com/Flexget/Flexget/pull/2312

0x022b:
OK good to know. Somehow I was able to miss that the default behaviour was changed. I can go through the documentation and update it where necessary. I'm closing this as the plugin is working as expected.

PR #2599, opened 2020-02-06
[fix] convert to timezone aware datetime for execution time in the api

kristenmills:
### Motivation for changes:
Since we store the the datetimes without timezones, when they are json encoded in the api, it is assumed that they are in UTC (which they aren't). This adds the system local timezone before json encoding,  so that it will properly offset the time

### DIFF
--- a/flexget/components/status/db.py
+++ b/flexget/components/status/db.py
@@ -48,7 +48,7 @@ class StatusTask(Base):
             select([func.max(TaskExecution.start)])
             .where(TaskExecution.task_id == cls.id)
             .correlate(StatusTask.__table__)
-            .label('last_execuftion_time')
+            .label('last_execution_time')
         )
 
     def to_dict(self):
--- a/flexget/components/status/db.py
+++ b/flexget/components/status/db.py
@@ -48,11 +48,15 @@ class StatusTask(Base):
             select([func.max(TaskExecution.start)])
             .where(TaskExecution.task_id == cls.id)
             .correlate(StatusTask.__table__)
-            .label('last_execution_time')
+            .label('last_execuftion_time')
         )
 
     def to_dict(self):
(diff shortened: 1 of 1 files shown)


flexget/components/status/db.py
```diff
@@ -48,11 +48,16 @@ def last_execution_time(cls):
             select([func.max(TaskExecution.start)])
             .where(TaskExecution.task_id == cls.id)
             .correlate(StatusTask.__table__)
-            .label('last_execution_time')
+            .label('last_execuftion_time')
```
cvium: Is this change intentional?

kristenmills: nope thanks

metaMMA:
I just wanted to say thanks for taking this on. I think this will be a useful feature!

flexget/components/status/db.py
```diff
@@ -52,7 +52,12 @@ def last_execution_time(cls):
         )
 
     def to_dict(self):
-        return {'id': self.id, 'name': self.name, 'last_execution_time': self.last_execution_time}
+        return {
+            'id': self.id,
+            'name': self.name,
+            'last_execution_time': self.last_execution_time
+            and self.last_execution_time.astimezone(),
```
cvium: Is this a new syntax though? `last_execution_time` should be a datetime? But won't the `and` make it a bool?

kristenmills: `and` does not coerce the value to a bool. `a and b` is equivalent to if `a` is truthy return `b` else return `a`.

cvium: I do think `self.last_execution_time.astimezone() if self.last_execution_time else None` is easier to read though.

gazpachoking:
LGTM!

PR #2601, opened 2020-02-08
[feature] Update Slack notifier to handle two separate versions, Legacy and Block Kit

signe:
### Motivation for changes:

The current slack notifier uses webhook formatting which is considered legacy/deprecated.
Adding a new notifier allows for an upgrade without breaking changes to the existing plugin.
The two plugins can run concurrently.

### Detailed changes:

Add slack2 notifier plugin.

The Block Kit format is wildly flexible, allowing for blocks of text, images, dividers, and context lines to appear in any order.  Up to 50 blocks are supported.

```
notify:
  entries:
    via:
      - slack2:
          web_hook_url: "https://hooks.slack.com/services/..."
          blocks:
          - section:
              text: '<{{ tvdb_url }}|{{ series_name }} ({{ series_id }})>'
          - section:
              text: '{{ tvdb_ep_overview }}'
              image:
                url: "https://api.slack.com/img/blocks/bkb_template_images/plants.png"
                alt_text: 'plants'
              fields:
                - '*Score*'
                - '*Genres*'
                - '{{ imdb_score }}'
                - "{{ imdb_genres | join(', ') | title }}"
                - '*Rating*'
                - '*Cast*'
                - '{{ imdb_mpaa_rating }}'
                - >
                  {% for key, value in imdb_actors.items() %}{{ value }}{% if not loop.last %}, {% endif %}
                  {% endfor %}
          - image:
              url: "{{ tvdb_banner }}"
              alt_text: "{{ series_name }} Banner"
          - divider: True
          - context:
              - image:
                  url: 'https://image.freepik.com/free-photo/red-drawing-pin_1156-445.jpg'
                  alt_text: 'red pin'
              - text: ':round_pushpin:'
              - text: 'Task: {{ task }}'
```

### Example output
image

### DIFF
--- a/flexget/components/notify/notifiers/slack.py
+++ b/flexget/components/notify/notifiers/slack.py
@@ -15,12 +15,13 @@ logger = logger.bind(name=plugin_name)
 
 class SlackNotifier:
     """
-    Example:
+    Example (version 1):
 
       notify:
         entries:
           via:
             - slack:
+                version: 1
                 web_hook_url: <string>
                 [channel: <string>] (override channel, use "@username" or "#channel")
                 [username: <string>] (override username)
@@ -28,12 +29,72 @@ class SlackNotifier:
                 [icon_url: <string>] (override emoji icon)
                 [attachments: <array>[<object>]] (override attachments)
 
+    Example (version 2):
+
+      notify:
+        entries:
+          via:
+            - slack2:
+                version: 2
+                web_hook_url: "https://hooks.slack.com/services/..."
+                blocks:
+                  - section:
+                      text: '<{{ tvdb_url }}|{{ series_name }} ({{ series_id }})>'
+                  - section:
+                      text: '{{ tvdb_ep_overview }}'
+                      image:
+                        url: "https://api.slack.com/img/blocks/bkb_template_images/plants.png"
+                        alt_text: 'plants'
+                      fields:
+                        - '*Score*'
+                        - '*Genres*'
+                        - '{{ imdb_score }}'
+                        - "{{ imdb_genres | join(', ') | title }}"
+                        - '*Rating*'
+                        - '*Cast*'
+                        - '{{ imdb_mpaa_rating }}'
+                        - >
+                          {% for key, value in imdb_actors.items() %}{{ value }}{% if not loop.last %}, {% endif %}
+                          {% endfor %}
+                  - image:
+                      url: "{{ tvdb_banner }}"
+                      alt_text: "{{ series_name }} Banner"
+                  - context:
+                      - image:
+                          url: 'https://image.freepik.com/free-photo/red-drawing-pin_1156-445.jpg'
+                          alt_text: 'red pin'
+                      - text: ':round_pushpin:'
+                      - text: 'Task: {{ task }}'
     """
 
     schema = {
+        'definitions': {
+            'image_block': {
+                'type': 'object',
+                'properties': {
+                    'url': {'type': 'string', 'format': 'uri'},
+                    'alt_text': {'type': 'string'},
+                },
+                'required': ['url', 'alt_text'],
+                'additionalProperties': False,
+            },
+            'image_block_w_title': {
+                'type': 'object',
+                'properties': {
+                    'url': {'type': 'string', 'format': 'uri'},
+                    'title': {'type': 'string'},
+                    'alt_text': {'type': 'string'},
+                },
+                'required': ['url', 'alt_text'],
+                'additionalProperties': False,
+            }
+
+        },
+
         'type': 'object',
         'properties': {
             'web_hook_url': {'type': 'string', 'format': 'uri'},
+            'version': {'type': 'integer', 'enum': [1, 2], 'default': 1},
             'username': {'type': 'string', 'default': 'Flexget'},
             'icon_url': {'type': 'string', 'format': 'uri'},
             'icon_emoji': {'type': 'string'},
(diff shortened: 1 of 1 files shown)


flexget/components/notify/notifiers/slack2.py
```diff
@@ -0,0 +1,237 @@
...
+                                {'required': ['text']},
+                                {'required': ['fields']}
+                            ],
+                            'additionalProperties': False,
+                        },
+                        'image': {
+                            'type': 'object',
+                            'properties': {
+                                'text': {'type': 'string'},
+                                'image': {
```
liiight: seems like you can reuse some definitions. You can set those to different python variables or use json schema's own `definitions` attribute

signe:
I've updated the code to be a part of the original slack notifier, handling two schema variations, keyed on a version field (which defaults to the current plugin format).

Updated commit message:

```
    [feature] Update Slack notifier to handle two separate versions

    Version 1 (default) supports the existing plugin format using the Legacy webhook parameters
    Version 2 supports the Block Kit webhook format
```

flexget/components/notify/notifiers/slack.py
```diff
@@ -15,25 +15,86 @@
...
+                'required': ['url', 'alt_text'],
+                'additionalProperties': False,
+            }
+
+        },
+
         'type': 'object',
         'properties': {
             'web_hook_url': {'type': 'string', 'format': 'uri'},
+            'version': {'type': 'integer', 'enum': [1, 2], 'default': 1},
```
liiight: interesting. Perhaps a simpler approach would be to declare the two schema independently and then use `oneOf` between them. WDYT?

flexget/components/notify/notifiers/slack.py
```diff
@@ -106,16 +255,104 @@ def notify(self, title, message, config):
...
-            'channel': config.get('channel'),
-            'attachments': config.get('attachments'),
-        }
-        if config.get('icon_emoji'):
-            notification['icon_emoji'] = ':%s:' % config['icon_emoji'].strip(':')
-        if config.get('icon_url'):
-            notification['icon_url'] = config['icon_url']
+
+        import json
+        logger.trace(json.dumps(self.schema))
```
paranoidi: leftover debugs

flexget/components/notify/notifiers/slack.py
paranoidi: Yeah, we don't generally do versioning on configuration. Version can be inferred from using "blocks" key here?

lingfish:
It is a pity this never got merged :(

gazpachoking:
If somebody wants to revive it and address previous comments that would be great!

flexget/components/notify/notifiers/slack.py
vivodi: That's an excellent point and a very valid approach to structuring the schema. Thank you for the suggestion!

The idea of defining two completely independent schemas and then using `oneOf` to select between them is indeed a very clean, modular pattern.

**The advantages of that approach would be:**
*   **Clarity and Modularity:** Each version's schema (`v1_schema`, `v2_schema`) would be a self-contained, complete unit, making it very easy to read and understand the rules for a single version.
*   **Reusability:** We could easily reference a specific version's schema from elsewhere if needed.

**The primary reason I opted for the original structure was to avoid duplication and adhere to the DRY (Don't Repeat Yourself) principle.**

With two separate schemas, common properties like `web_hook_url`, `username`, `channel`, `icon_url`, etc., would have to be defined identically in both places. The main drawback of this duplication is maintainability:
*   If we ever needed to add a new common property or modify an existing one (e.g., add a `proxy_url` field), we would have to remember to update it in both schemas. This could easily lead to inconsistencies or bugs down the line.

The original pattern—defining all possible properties at the top level and then using conditional logic (`anyOf`) to apply specific rules—centralizes the common fields. This ensures that any changes to those shared properties only need to be made in one place.

While it makes the top-level structure slightly more complex to read at first glance, it makes the schema more robust and easier to maintain in the long run.

flexget/components/notify/notifiers/slack.py
```diff
@@ -106,16 +255,104 @@ def notify(self, title, message, config):
...
+
+        # version 1 (default) is the Legacy formatting
+        if config.get('version') == 1:
+            # username, channel, icon_emoji, and icon_url are deprecated and ignored by the Slack API,
+            # therefore they've been removed from the posted data
+            notification = {
+                'text': message,
+                'attachments': config.get('attachments'),
+            }
+
```
vivodi, line 270: @signe In [https://api.slack.com/legacy/custom-integrations/messaging/webhooks](https://api.slack.com/legacy/custom-integrations/messaging/webhooks), it only says “Legacy custom integrations lack newer features and will be deprecated and possibly removed in the future.” It does not say “they are ignored by the Slack API.”

flexget/components/notify/notifiers/slack.py
vivodi: > interesting. Perhaps a simpler approach would be to declare the two schema independently and then use `oneOf` between them. WDYT?

@liiight That's an excellent point and a very valid approach to structuring the schema. Thank you for the suggestion!

The idea of defining two completely independent schemas and then using `oneOf` to select between them is indeed a very clean, modular pattern.

**The advantages of that approach would be:**
*   **Clarity and Modularity:** Each version's schema (`v1_schema`, `v2_schema`) would be a self-contained, complete unit, making it very easy to read and understand the rules for a single version.
*   **Reusability:** We could easily reference a specific version's schema from elsewhere if needed.

**The primary reason I opted for the original structure was to avoid duplication and adhere to the DRY (Don't Repeat Yourself) principle.**

With two separate schemas, common properties like `web_hook_url`, `username`, `channel`, `icon_url`, etc., would have to be defined identically in both places. The main drawback of this duplication is maintainability:
*   If we ever needed to add a new common property or modify an existing one (e.g., add a `proxy_url` field), we would have to remember to update it in both schemas. This could easily lead to inconsistencies or bugs down the line.

The original pattern—defining all possible properties at the top level and then using conditional logic (`anyOf`) to apply specific rules—centralizes the common fields. This ensures that any changes to those shared properties only need to be made in one place.

While it makes the top-level structure slightly more complex to read at first glance, it makes the schema more robust and easier to maintain in the long run.

PR #2602, opened 2020-02-11
Switch to using Enum to track Entry state

gazpachoking:
### Motivation for changes:
While adding some type hints, I realized there was already an `EntryState` enum, and thought maybe we should start actually using it to store entry state.

There was some hacky stuff in digest plugin regarding entry state, I didn't even try to fix that in this pull, just updated it to still work.

### DIFF
--- a/flexget/plugins/operate/digest.py
+++ b/flexget/plugins/operate/digest.py
@@ -7,7 +7,7 @@ from sqlalchemy import Column, DateTime, Integer, Unicode, select
 from flexget import db_schema, plugin
 from flexget.config_schema import one_or_more
 from flexget.db_schema import versioned_base
-from flexget.entry import Entry
+from flexget.entry import Entry, EntryState
 from flexget.event import event
 from flexget.manager import Session
 from flexget.utils import json, serialization
@@ -96,10 +96,11 @@ class OutputDigest:
         config = self.prepare_config(config)
         with Session() as session:
             for entry in task.all_entries:
-                if entry.state not in config['state']:
+                state = str(entry.state)
+                if state not in config['state']:
                     continue
                 entry['digest_task'] = task.name
-                entry['digest_state'] = entry.state
+                entry['digest_state'] = state
                 session.add(DigestEntry(list=config['list'], entry=entry))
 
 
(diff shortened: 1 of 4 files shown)


flexget/plugins/operate/digest.py
```diff
@@ -96,10 +96,11 @@ def on_task_learn(self, task, config):
         config = self.prepare_config(config)
         with Session() as session:
             for entry in task.all_entries:
-                if entry.state not in config['state']:
+                state = str(entry.state)
+                if state not in config['state']:
```
liiight: hmm, maybe a custom json schema format that returns `EntryState`?

gazpachoking: Not sure what you mean?

gazpachoking:
> This seems very nice. I wonder if there could be unforeseen effects to this via config. Maybe people have checks that rely on entry states being strings? does that make sense?

I added a custom `__str__` so that when used in a string context (jinja template, etc.) it should come out the same.

liiight:
> I added a custom __str__ so that when used in a string context (jinja template, etc.) it should come out the same.

Maybe add a test for this?

gazpachoking:
> > I added a custom **str** so that when used in a string context (jinja template, etc.) it should come out the same.
> 
> Maybe add a test for this?

Hmm, actually, it could be an issue inside a jinja expression, like in the `if` plugin. I'll add some backwards compatibility and a test for that.

gazpachoking:
> Hmm, actually, it could be an issue inside a jinja expression, like in the `if` plugin. I'll add some backwards compatibility and a test for that.

Strike that, we can't access entry state from the if plugin (although, perhaps that should be added.) Are there any other places outside of string formatting this could be an issue anyone can think of?

gazpachoking:
I added in `__eq__` and `__hash__` which should make the enum states pretty compatible with the old plain string states, even though I'm not sure there's anywhere in the config where that could matter. I'm thinking we should allow accessing the state from jinja though, probably in a separate pull. The compatibility stuff I just added allows `entry.state == 'accepted'` and `entry.state in ['accepted', 'rejected']` type stuff again.

gazpachoking:
Held off on any explicit tests, since we aren't actually relying on any special behavior of the entry state beyond what normal tests verify (that I know of.) If we allow it in jinja later, we can add some tests to verify that behavior works as expected.

PR #2605, opened 2020-02-13
Dockerfile: multi-stage building

X-dark:
### Motivation for changes:

Current Docker image contains build deps and is unnecessary big.

### Detailed changes:

- Use multi-stage building to separate build from install. The resulting image is ~300MB instead of ~800MB.
- Use official python image (based on alpine) instead of installing python on an alpine image


metaMMA:
I like this. Thanks!

paranoidi:
Yeah, this is way better than what we have now. Thanks a lot!

PR #2607, opened 2020-02-13
Simplify versioned_base implementation

gazpachoking:
### Motivation for changes:
While doing some type hint stuff, I noticed there should be easier ways to handle the versioned_base stuff. There might even be ways to simplify it further, but this is what I came up with at first. Slightly nervous to merge it straight away, so going to let it sit here for a bit. 😝 If anyone notices problems, let me know.

### DIFF
--- a/flexget/config_schema.py
+++ b/flexget/config_schema.py
@@ -2,7 +2,7 @@ import datetime
 import os
 import re
 from collections import defaultdict
-from typing import Any, Callable, Optional, Union
+from typing import Any, Callable, Dict, Optional, Union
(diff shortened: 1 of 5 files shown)


gazpachoking:
I tested creating a new db, running against an existing db, and running a reset-plugin to clear out the db for one plugin then have it recreated. I guess I should try to downgrade to some version before the last db upgrade and see if the upgrade goes okay too, as I'm not certain we have any good db upgrade tests.

gazpachoking:
Oops, didn't mean to commit those other hinting things to this branch. Meh, whatever.

PR #2615, opened 2020-03-08
Increase tries to solve challenge

ExceptionalError:
### Motivation for changes:
Increase tries to solve cloudflare challenge giving a higher probability of solving it correctly. 

### Addressed issues:
- #2614

### DIFF
--- a/flexget/plugins/operate/cfscraper.py
+++ b/flexget/plugins/operate/cfscraper.py
@@ -45,7 +45,7 @@ class CFScraper:
                     ('Upgrade-Insecure-Requests', '1'),
                 ]
             )
-            task.requests = CFScrapeWrapper.create_scraper(task.requests)
+            task.requests = CFScrapeWrapper.create_scraper(task.requests, solveDepth=5)
 
 
 @event('plugin.register')


ExceptionalError:
@paranoidi :
Could you merge this one, too?

BrutuZ:
```
2020-03-09 22:23:41 CRITICAL task          http-ddl        BUG: Unhandled error in plugin cfscraper: __init__() got an unexpected keyword argument 'solveDepth'
TypeError: __init__() got an unexpected keyword argument 'solveDepth'
```

gazpachoking:
@BrutuZ  Looks like cloudscraper 1.2.13+ is required now. We don't have such a good way to automatically update, or notify about these optional requirements though. Not sure the best way to improve that.

BrutuZ:
Figured the same, udpated CloudScraper, restarted the Flexget Daemon, still getting the same error.

ExceptionalError:
Same here - I had problems updating my docker container but thought I was running the latest version. Sorry for the bug.

Also it seems that even with 1.2.26 the error still persists - I have requested the correct parameter name from the cloudscraper maintainer (again :wink: ).

Btw a container running the latest dev branch would be very nice ...

ExceptionalError:
This is an inheritance problem caused by this wrapper class:
```python
class CFScrapeWrapper(Session, cloudscraper.CloudScraper):
    """
    This class allows the FlexGet session to inherit from CloudScraper instead of the requests.Session directly.
    """
```

```python
@classmethod
def create_scraper(cls, sess=None, **kwargs):
    """
    Convenience function for creating a ready-to-go CloudScraper object.
    """
    scraper = cls(**kwargs)
````


call:
```python
task.requests = CFScrapeWrapper.create_scraper(task.requests, solveDepth=5)
```

I am unsure how to fix this - some help might be appreciated.

gazpachoking:
> The `solveDepth` gets passed to [Session](https://github.com/Flexget/Flexget/blob/45fdc8df6211fe6c64e1b0caa0452fd640b5b162/flexget/utils/requests.py#L173) which then raises the error.

But, Session doesn't have a create_scraper method, no?

gazpachoking:
Perhaps the inheritance order can just be reveresed for the wrapper class? Or else, `__init__` can be defined in the wrapper class to call just cfscraper init.

ExceptionalError:
Imho this would be the easiest solution

```python
class CFScrapeWrapper(Session, cloudscraper.CloudScraper):
    def __init__(self, *args, **kwargs):
        cloudscraper.CloudScraper.__init__(self, *args, **kwargs)
        Session.__init__(self, *args, **kwargs)
```
the call to ``CloudScraper`` removes the keys from the dict so ``Session`` doesn't throw an error.

gazpachoking:
We could also add a `**kwargs` to our Session, and pass it to super, to make it play better with multiple inheritance. (Pretty sure `super` in that case would be cfscraper, not requests, but I'm not positive on that.)

https://github.com/Flexget/Flexget/blob/45fdc8df6211fe6c64e1b0caa0452fd640b5b162/flexget/utils/requests.py#L180-L182

ExceptionalError:
Yea - that would probably the cleaner solution

PR #2616, opened 2020-03-08
Rate-Limit only for GET-Requests

ExceptionalError:
Limit only for get requests so cloudflare challenge gets answered timely (#2614)

### Motivation for changes:
Long timeout seems to reset the cloudflare challenge

### Detailed changes:
- Rate limiting only for get-requests so cloudflare gets the answer directly after the challenge


paranoidi:
Thanks for contributing!

PR #2617, opened 2020-03-10
[Anilist] Evaluate if the english title is provided

BrutuZ:
### Motivation for changes:
Found out some entries in Anilist don't have a proper English title (there are a number of valid reasons), which causes a `None` entry to be added to the alternate_name list
### Detailed changes:
- Evaluate if the English title is provided and not already in the alternate_name \ synonym \ a.k.a. list provided before inserting it on first position.


gazpachoking:
LGTM

PR #2618, opened 2020-03-10
Revert solveDepth

ExceptionalError:
#2615

PR #2619, opened 2020-03-10
Exceptional error patch rate limit

ExceptionalError:
Bugfix for 
#2618
#2615

### DIFF
--- a/flexget/utils/requests.py
+++ b/flexget/utils/requests.py
@@ -235,7 +235,8 @@ class Session(requests.Session):
             )
 
         # Run domain limiters for this url
-        limit_domains(url, self.domain_limiters)
+        if method == 'GET':
+            limit_domains(url, self.domain_limiters)
 
         kwargs.setdefault('timeout', self.timeout)
         raise_status = kwargs.pop('raise_status', True)
--- a/flexget/utils/requests.py
+++ b/flexget/utils/requests.py
@@ -177,9 +177,9 @@ class Session(requests.Session):
 
     """
 
-    def __init__(self, timeout: int = 30, max_retries: int = 1) -> None:
+    def __init__(self, timeout: int = 30, max_retries: int = 1, **kwargs) -> None:
         """Set some defaults for our session if not explicitly defined."""
-        super().__init__()
+        super().__init__(**kwargs)
         self.timeout = timeout
         self.stream = True
         self.adapters['http://'].max_retries = max_retries
--- a/flexget/plugins/operate/cfscraper.py
+++ b/flexget/plugins/operate/cfscraper.py
@@ -45,7 +45,7 @@ class CFScraper:
                     ('Upgrade-Insecure-Requests', '1'),
                 ]
             )
-            task.requests = CFScrapeWrapper.create_scraper(task.requests)
+            task.requests = CFScrapeWrapper.create_scraper(task.requests, solveDepth=5)
 
 
 @event('plugin.register')


ExceptionalError:
@paranoidi can you check if I have done it correctly now?

flexget/utils/requests.py
```diff
@@ -235,7 +235,8 @@ def request(self, method: str, url: str, *args, **kwargs) -> requests.Response:
             )
 
         # Run domain limiters for this url
-        limit_domains(url, self.domain_limiters)
+        if method == 'GET':
```
paranoidi: Hmm, this affects all ratelimiting done by everywhere. Some might require post requests to be limited.

ExceptionalError: Yes - but limiting the POST after a GET brakes cloudscraper functionality.
Tbh I don't see much real life usage for it.

paranoidi: But this util is not spesific to cfscrape and is very likely used for example "domain_delay" plugin.

@gazpachoking thoughts?

paranoidi: Merged since this addresses issues ..

gazpachoking: Yeah, I'm hesitant to disable all the limiters for all requests other than GET. I wonder if we can do something specific in cfscrape plugin to only disable it on the pertinent request.

gazpachoking: Idea: add a keyword arg to disable limiters on individual requests here
https://github.com/Flexget/Flexget/blob/e0e5f9b7d8cfbf8a7748091bd8fbc9cbb9f2dc66/flexget/utils/requests.py#L222
Then, override `solve_cf_challenge` in our cfscraper wrapper to just add that keywor arg to `original_kwargs` then call `super`
https://github.com/Anorov/cloudflare-scrape/blob/9692fe7ff3c17b76ddf0f4d50d3dba7d1791c9c6/cfscrape/__init__.py#L150

ExceptionalError:
@paranoidi 
The broken version is still on pypi - could you make a release so the bugfix is available?

flexget/utils/requests.py
ExceptionalError: Why not like this:

```python
class Session(requests.Session):
    LIMITER_ENABLED = True

    ....
    def request(self, method: str, url: str, *args, **kwargs) -> requests.Response:
        ....
        if self.LIMITER_ENABLED:
             limit_domains(url, self.domain_limiters)


class CFScrapeWrapper(Session, cloudscraper.CloudScraper):
    
    def __init__(self, *args, **kwargs):
        super().__init(*args, **kwargs)
        self.LIMITER_ENABLED = False
```

gazpachoking: That would disable all limits to all sites when the cfscraper plugin was used, rather than just the problematic challenges

ExceptionalError: Hm, you're right. How about this:

```python
class CFScrapeWrapper(Session, cloudscraper.CloudScraper):
    
    def __init__(self, *args, **kwargs):
        super().__init(*args, **kwargs)
        self.LIMITER_ENABLED = True

     def solve_cf_challenge(self, *args, **kwargs):
         self.LIMITER_ENABLED = False
         super().solve_cf_challenge(*args, **kwargs)
         self.LIMITER_ENABLED = True
```

gazpachoking: That would have about the same effect as my proposed solution, except in my way we pass the variable in, rather than setting and unsetting some session level variable. Just feels a little cleaner to me to have it as an argument to request method.

ExceptionalError: But wouldn't you need to pass the variable to all function calls in ``CloudScraper`` which could use the ``request`` function?

gazpachoking: This is my proposed solution. (Seems I was looking at the wrong module before, I think pep8 was murdered by the one we are using.) I don't know any sites to test against, so I'm not sure if it's working.

https://github.com/Flexget/Flexget/compare/cfscraper_limiter_fix

ExceptionalError: Seems I can't comment on your code I'll do it here.

I'd rather not pop the 'disable_limiters' because the cloudscraper module does multiple trys to solve the challenge. If you pop it only the first skips the limiters.

How about this:
```python
           def Challenge_Response(self, resp, **kwargs):
                kwargs.setdefault('disable_limiters', True)
                super().Challenge_Response(resp, **kwargs)
                kwargs.pop('disable_limiters')
```

```python
        if 'disable_limiters' not in kwargs:
            limit_domains(url, self.domain_limiters)
```

gazpachoking: It only pops that inside the request method. The next time cloudscraper calls it it will be true again.

gazpachoking: Just checking if 'disable_limiters' is in kwargs is also not good, because if someone explicitly passed `disable_limiters=False` it would still disable them.

ExceptionalError: I guess we're good to go then :+1:

gazpachoking: Did you test out my proposal? So far my solution was only theoretical, I'd feel better once it's confirmed it actually fixes what it's meant to. ;)

gazpachoking: It probably wasn't working, as I forgot to return the result. Probably fixed now though.

gazpachoking: Opened #2620 for testing and further discussion.

PR #2620, opened 2020-03-11
Cfscraper limiter fix

gazpachoking:
### Motivation for changes:
Apparently cloudflare challenges get broken by our domain limiting system. This disables our limiting system when responding to a cloudflare challenge. Makes the solution in #2619 less heavy handed.

Haven't tested this myself, would be good if someone who had the issue can confirm this works.

### DIFF
--- a/flexget/plugins/operate/cfscraper.py
+++ b/flexget/plugins/operate/cfscraper.py
@@ -34,6 +34,10 @@ class CFScraper:
             This class allows the FlexGet session to inherit from CloudScraper instead of the requests.Session directly.
             """
 
+            def Challenge_Response(self, resp, **kwargs):
+                kwargs.setdefault('disable_limiters', True)
+                super().Challenge_Response(resp, **kwargs)
+
         if config is True:
             task.requests.headers = OrderedDict(
                 [
--- a/flexget/plugins/operate/cfscraper.py
+++ b/flexget/plugins/operate/cfscraper.py
@@ -36,7 +36,10 @@ class CFScraper:
 
             def Challenge_Response(self, resp, **kwargs):
                 """Make sure limiters are disabled when doing a cloudflare challenge."""
-                kwargs['disable_limiters'] = True
+                if not self.is_reCaptcha_Challenge(resp):
+                    # If this is a recaptcha challenge, the request gets sent straight to requests, not our subclass,
+                    # so it can't have any extra arguments that requests doesn't expect.
+                    kwargs['disable_limiters'] = True
                 return super().Challenge_Response(resp, **kwargs)
 
         if config is True:
--- a/flexget/plugins/operate/cfscraper.py
+++ b/flexget/plugins/operate/cfscraper.py
@@ -35,8 +35,9 @@ class CFScraper:
             """
 
             def Challenge_Response(self, resp, **kwargs):
-                kwargs.setdefault('disable_limiters', True)
-                super().Challenge_Response(resp, **kwargs)
+                """Make sure limiters are disabled when doing a cloudflare challenge."""
+                kwargs['disable_limiters'] = True
+                return super().Challenge_Response(resp, **kwargs)
 
         if config is True:
             task.requests.headers = OrderedDict(
--- a/flexget/utils/requests.py
+++ b/flexget/utils/requests.py
@@ -75,7 +75,11 @@ class TokenBucketLimiter(DomainLimiter):
     state_cache = {}
 
     def __init__(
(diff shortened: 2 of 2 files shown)


ExceptionalError:
Your change allows the first request in ``Challenge_Response`` to go without rate limiting.
For this change to work you assume:
- there is exactly one request made in ``Challenge_Response``
- this is the only part where requests are made in the CloudScraper library
If one of these two are not true the implementation breaks.

It feels the cleaner solution would be to add the http method to the token limiter.
That way the POST would follow up immediately after a GET.
Also a POST/PUT/DEL could also reset the GET limiter.
That way it'll not be CFS specific and would work as intended without the assumption of the both points.

Anyway - If I figure out how to modify the docker container I'll happily test it

gazpachoking:
>     * there is exactly one request made in `Challenge_Response`

All requests in that function will have the disable_limiters keyword argument, so that shouldn't be true.

>     * this is the only part where requests are made in the CloudScraper library
>       If one of these two are not true the implementation breaks.

It wasn't an assumption, I checked the source ;)

gazpachoking:
Hmm, I did see one potential issue: If it gets a captcha reponse, Challenge_Response calls `super().request`, which would crash because `disable_limiters` is an unknown arugment..

gazpachoking:
> It feels the cleaner solution would be to add the http method to the token limiter.
> That way the POST would follow up immediately after a GET.
> Also a POST/PUT/DEL could also reset the GET limiter.
> That way it'll not be CFS specific and would work as intended without the assumption of the both points.

Not exactly sure what you mean, but the whole point of the limiters is to make sure requests are not directly in a row, no matter the kind of request. This exception is meant to be cloudflare specific, based on your report that the challenge response doesn't work if delayed. (Did you verify that btw? In your original report it said that was a guess.)

ExceptionalError:
> It wasn't an assumption, I checked the source ;)

I thought so, but someone has to check the source every time it changes to ensure this is still true. Otherwise flexget will crash for users which have rate limiting in place but not for users who don't use it.


> Not exactly sure what you mean, but the whole point of the limiters is to make sure requests are not directly in a row, no matter the kind of request. 

I mean instead of one limiter per domain have four limiters. A successful POST/PUT/DEL call will increase the GET limiter, too.
There is still an effective rate limiting in place, but the next non GET request will always follow immediately after a GET request, e.g. the POST will always follow immediately upon a GET request.
That way all challenge response implementations will work without having to create exceptions for each one.

> This exception is meant to be cloudflare specific, based on your report that the challenge response doesn't work if delayed. (Did you verify that btw? In your original report it said that was a guess.)

Yes, with the originally suggested 'GET' workaround everything works fine, without the 'GET' I get crashes.

gazpachoking:
> But if you pop it after the first request it will not be there for the second request

The pop doesn't happen in Challenge_Response, so it will remain for all requests. it gets passed down to our requests method by **kwargs, which means any mutations won't be propagated back.

Is there a reason other than this cloudflare issue you think that other requests should have less rate limiting? If a site has e.g. a limit of 100 requests per hour, it won't care what kind of requests they are, so writing exceptions into the limiters risks running afoul of their rules.

ExceptionalError:
> The pop doesn't happen in Challenge_Response, so it will remain for all requests. it gets passed down to our requests method by **kwargs, which means any mutations won't be propagated back.

But doesn't ``Challenge_Response`` call ``request`` which will then pop the ``disable_limiters``.
I am just confused - sorry.


> Is there a reason other than this cloudflare issue you think that other requests should have less rate limiting? If a site has e.g. a limit of 100 requests per hour, it won't care what kind of requests they are, so writing exceptions into the limiters risks running afoul of their rules.

No

gazpachoking:
Here's a simplified example of what's happening
```python
def request(**kwargs):
    kwargs.pop('something')

def Challenge_Response(**kwargs):
    kwargs['something'] = True
    request(**kwargs)
    # kwargs still has 'something' included here, it doesn't matter what happens in `request`,
    # as the actual dictionary isn't being passed when done via ** keyword arguments
    request(**kwargs)
```

gazpachoking:
> I thought so, but someone has to check the source every time it changes to ensure this is still true. Otherwise flexget will crash for users which have rate limiting in place but not for users who don't use it.

This bit is still true, if cloudscraper changes how it does things our customization could break. I think I'd prefer to see if this happens rather than mess with how the domain limiters work for every request.

ExceptionalError:
> Here's a simplified example of what's happening

Thank you for your example. 
I was under the assumption that **kwargs was the passed dict, but of course it is the keyword args from the function. 
Now it's clear - thanks!

PR #2621, opened 2020-03-12
Added Tray icon

liiight:
### Motivation for changes:
A nice tray icon to be added on daemon mode run.

### Detailed changes:
- Added `TrayIcon` and moved `manager.start()` to its own thread


paranoidi:
Tray icon should be enabled by default, no need to make it unnecessay complicated to use.

flexget/__init__.py
```diff
@@ -1,12 +1,16 @@
...
 import os
 import sys
+from pathlib import Path
 
-# __version__ import need to be first in order to avoid circular import within logger
-from ._version import __version__  # noqa
 from flexget import log
 from flexget.manager import Manager
 
+# __version__ import need to be first in order to avoid circular import within logger
```
gazpachoking: Oops, this got moved to not be first. Going to break things?

gazpachoking:
Not sure if we can do this in a plugin given the requirements for running in main thread and blocking. Might have to be just built in if we add it.

liiight:
Changed it to threads. It's ugly though since there's a dependency between `TrayIcon` and `Manager`. tray needs manager for its shutdown and reload config methods, and manager needs to call tray_icon.stop() when it shutsdown...

@gazpachoking @paranoidi any ideas on how to make this nicer?

flexget/__init__.py
liiight: fixed and added an exclude in pre-commit

liiight:
Seems tests are failing since by running pip-compile on a mac, it compiled so mac specific deps. I guess in this case we need to add the higher level package (pystray) directly to the requirement.txt file and let it resolve on the client machine. 
This will be a problem with future pip-compiles though...

flexget/__init__.py
```diff
@@ -38,7 +47,14 @@ def main(args=None):
                     os.path.join(manager.config_base, manager.options.profile),
                 )
             else:
-                manager.start()
+                m = threading.Thread(target=manager.start, daemon=True)
+                m.start()
+                if tray_icon:
+                    while not manager_loaded:
```
gazpachoking: This will wait forever if we are launching flexget not in daemon mode since the manager.daemon.started event will never fire.

flexget/__init__.py
```diff
@@ -1,12 +1,21 @@
...
 
 # __version__ import need to be first in order to avoid circular import within logger
 from ._version import __version__  # noqa
 from flexget import log
+from flexget.event import event
+from flexget.tray_icon import tray_icon
 from flexget.manager import Manager
 
+logger = logger.bind(name='main')
+manager_loaded = False
```
gazpachoking: This should probably be a `threading.Event`

liiight: thanks, my threading game is weak. Event indeed seems perfect here

flexget/__init__.py
liiight: fixed by setting a timeout on the event.wait()

PR #2632, opened 2020-03-21
Enum events

liiight:
### Motivation for changes:
Enum fits well for events.

### Detailed changes:
- Created `EventType` and changed all explicit events to this type
- All method accept `Union[EventType, str]` since plugin phase handlers are created dynamically so they will remain as str.
- Changed `_events` data structure to be a defaultdict for a slightly nicer syntax

### To do

- [ ] Add tests

### DIFF
--- a/docs/develop/database.rst
+++ b/docs/develop/database.rst
@@ -60,6 +60,6 @@ If the plugin accumulates data into database that should be cleaned at some poin
 
 .. code-block:: python
 
-   @event('manager.db_cleanup')
+   @event(EventType.manager__db_cleanup)
    def db_cleanup(session):
        # cleanup actions here
--- a/docs/develop/list_interface.rst
+++ b/docs/develop/list_interface.rst
@@ -197,7 +197,7 @@ After creating the base class, the plugin class itself need to be created.
             return list(DBEntrySet(config))
 
 
-    @event('plugin.register')
+    @event(EventType.plugin__register)
     def register_plugin():
         plugin.register(EntryList, 'entry_list', api_ver=2, interfaces=['list', 'task'])
 
--- a/docs/develop/tdd.rst
+++ b/docs/develop/tdd.rst
@@ -44,13 +44,13 @@ Within this file we will add our plugin.
 .. testcode::
 
    from flexget import plugin
-   from flexget.event import event
+   from flexget.event import event, EventType
 
 
    class Hello:
        pass
 
-   @event('plugin.register')
+   @event(EventType.plugin__register)
    def register_plugin():
        plugin.register(Hello, 'hello', api_ver=2)
 
@@ -95,7 +95,7 @@ Continue by implementing the test case.
 .. testcode::
 
    from flexget import plugin
-   from flexget.event import event
+   from flexget.event import event, EventType
 
 
    class Hello:
@@ -103,7 +103,7 @@ Continue by implementing the test case.
            for entry in task.entries:
                entry['hello'] = True
 
-   @event('plugin.register')
+   @event(EventType.plugin__register)
    def register_plugin():
        plugin.register(Hello, 'hello', api_ver=2)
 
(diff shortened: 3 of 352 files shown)


paranoidi:
Gaah, do not like. Dunder in variable names :(

liiight:
What if I'll separate to several Enum classes like `ManagerEvent` `OptionsEvent` etc? That could resolve both issues

paranoidi:
Yeah, I was thinking something along those lines but wa too tired.

If those enums are in events.py it does not help much though. What do you think about inner classes? Make inner class in whoever is declaring & firing events.

ie.

Manager.Events.register

liiight:
I was experimenting with that but nested enums must use `.value`:

```python
class EventType(Enum):
    class Manager(Enum):
        register = 'register's

e = EventType.Manager.value.register
```
Not that nice...

WDYT?

paranoidi:
I didn't mean nested event classes, instead something like:

```
class Manager:
    class Events(Enum):
      STARTUP = 'manager.startup'
```

But yeah, even with this one would need to always use `.value` which sucks, unless we (can) tweak event dispatching to do that based on type ..

liiight:
I don't see how that's better than separate classes:
```python
class ManagerEvent(Enum):
    STARTUP = 'manager.startup'
```
Also, I think the best place for this is the events module, since I don't think it'd be nice to add import to more core module from different plugins

paranoidi:
Ease of import is a poor decision mechanism since by that logic everything should be in a single module ;)

I think it like this: event.py is just the framework for dispatching and handling events. Events are created by some entity (manager, plugin framework, plugins, whatever) and hooked by others. Placing Managers or plugins events into a event library module itself spreads the entity & responsibility further. If everything related is together it's easier to understand and manage, and more critically, to add or remove.

https://programmingisterrible.com/post/139222674273/write-code-that-is-easy-to-delete-not-easy-to

liiight:
Ok, so let's just create enum class in the relevant modules then, not need for nested anything

paranoidi:
That works too, and I'm fine with that :)

Nesting the enums using convention would make it alightly clearer who declares what events (eg. multiple plugins in a single module).

@gazpachoking comments?

gazpachoking:
Is our goal to just prevent typos when firing/reacting to events? The enum classes are then a way to 'register' events? Maybe just make an explicit register method then, and can be used like
```
@event.manager.startup
def myfunc(arg):
    pass

fire_event.manager.startup(arg)
```
If we like enums for registering, that could still be done, though at that point a list of strings sounds easier.
```
@register_events('manager')
class ManagerEvents(Enum):
    startup = 'startup'

# or
register_events('manager', ['startup'])
```
I'm also fine with enums as you guys are thinking, there are ways to avoid the `.value` calls too if that's an issue.

liiight:
At this point it's starting to feel like we're trying to rationalize use cases retroactively for this. I'm fine with just dropping it all together as the current implementation seems to work fine for everyone

gazpachoking:
Sorry, didn't mean to put the kibosh on your work. Just wanted to back up to the point of figuring out what we were fixing, and whether this was a good way.

paranoidi:
Nothing sucks more than throwing away coding efforts, but sometimes it needs to happen. Hard to say what is best here. Constants (enums) are nice, but so is the flexibility with event usage. Maybe keep it simple is good rule of thumb to follow here.

liiight:
> Sorry, didn't mean to put the kibosh on your work. Just wanted to back up to the point of figuring out what we were fixing, and whether this was a good way.

Not at all dude, this was a POC and I'm glad you guys gave feedback. I was overly naive with my approach to events and I'm satisfied to know that for sure as much as I would have been with actually merging this, for real :)

> Maybe keep it simple is good rule of thumb to follow here.

Exactly, strings seems to fit best here, even with their downsides as opposed to enums.

I'm happy to close this :)

PR #2638, opened 2020-03-23
Added tarantool plugin

StefftheEmperor2:
I added a plugin to handle input/output to tarantool queues

### DIFF
--- a/flexget/plugins/clients/tarantool_queue.py
+++ b/flexget/plugins/clients/tarantool_queue.py
@@ -5,6 +5,7 @@ from flexget import plugin
 
 logger = logger.bind(name='tarantool_queue')
 
+
 class TarantoolTube:
     def __init__(self, queue, tube_name):
         self._queue = queue
@@ -22,6 +23,7 @@ class TarantoolTube:
     def ack(self, task_id):
         return self._queue.call('queue.tube.{0}:ack'.format(self._tube_name), task_id)
 
+
 class TarantoolQueue:
     def __init__(self, tarantool_module, host, port, user, password):
         self._server = tarantool_module.connect(host, port, user=user, password=password)
@@ -38,6 +40,7 @@ class TarantoolQueue:
     def call(self, function, *args):
         return self._server.call(function, *args)
 
+
 class PluginTarantoolQueueBase:
     schema = {
         'anyOf': [
@@ -52,6 +55,7 @@ class PluginTarantoolQueueBase:
                     'tube': {'type': 'string'},
                     'chunk_size': {'type': 'integer', 'default': 30}
                 },
+                'required': ['host', 'port', 'user', 'password', 'tube'],
                 'additionalProperties': False,
             },
         ]
@@ -59,17 +63,18 @@ class PluginTarantoolQueueBase:
 
     def get_queue(self, tarantool_module, config):
         queue = TarantoolQueue(tarantool_module,
-                                     config.get('host'),
-                                     config.get('port'),
-                                     config.get('user'),
-                                     config.get('password'))
+                               config.get('host'),
+                               config.get('port'),
+                               config.get('user'),
+                               config.get('password'))
 
         return queue
 
-    def get_tube(self, tarantool_module, config):
+    def get_tube_of_queue(self, tarantool_module, config):
         return self.get_queue(tarantool_module, config).get_tube(config.get('tube'))
 
-    def get_tarantool_module(self):
+    @staticmethod
+    def get_tarantool_module():
         try:
             import tarantool
         except:
@@ -80,7 +85,7 @@ class PluginTarantoolQueueBase:
 class PluginTarantoolQueueInput(PluginTarantoolQueueBase):
     def on_task_input(self, task, config):
         tarantool_module = self.get_tarantool_module()
-        tube = self.get_tube(tarantool_module, config)
+        tube = self.get_tube_of_queue(tarantool_module, config)
         tarantool_entries = tube.take(0)
         do_continue = True
         entry_count = 0
@@ -106,12 +111,13 @@ class PluginTarantoolQueueOutput(PluginTarantoolQueueBase):
             return
 
         tarantool_module = self.get_tarantool_module()
-        tube = self.get_tube(tarantool_module, config)
+        tube = self.get_tube_of_queue(tarantool_module, config)
         for entry in task.accepted:
             try:
                 tube.put(Entry.serialize(entry))
             except Exception as e:
                 entry.fail(e)
+
 
 @event('plugin.register')
 def register_plugin():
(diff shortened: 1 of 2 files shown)


flexget/plugins/clients/tarantool_queue.py
```diff
@@ -0,0 +1,145 @@
...
+            {'type': 'boolean'},
+            {
+                'type': 'object',
+                'properties': {
+                    'host': {'type': 'string'},
+                    'port': {'type': 'integer'},
+                    'user': {'type': 'string'},
+                    'password': {'type': 'string'},
+                    'tube': {'type': 'string'},
+                    'enabled': {'type': 'boolean'},
```
paranoidi: we don't do option called enabled, instead

`tarantool: false`

flexget/plugins/clients/tarantool_queue.py
```diff
@@ -0,0 +1,145 @@
...
+        self._tube_name = tube_name
+
+    def put(self, data):
+        return self._queue.call('queue.tube.'+self._tube_name+':put', data)
+
+    def take(self, timeout=None):
+        return self._queue.call('queue.tube.'+self._tube_name+':take', timeout)
+
+    def ack(self, task_id):
+        return self._queue.call('queue.tube.' + self._tube_name + ':ack', task_id)
```
paranoidi: preferably use string formatting over +

flexget/plugins/clients/tarantool_queue.py
```diff
@@ -0,0 +1,145 @@
...
+                    'password': {'type': 'string'},
+                    'tube': {'type': 'string'},
+                    'enabled': {'type': 'boolean'},
+                },
+                'additionalProperties': False,
+            },
+        ]
+    }
+
+    def __init__(self):
```
paranoidi: Plugins should avoid `__init__`. We want them to be stateless. These can often "leak" memory. At minimum config should be kept out as it leads too easily into situation where reload does not apply.

flexget/plugins/clients/tarantool_queue.py
```diff
@@ -0,0 +1,145 @@
...
+
+        for entry in task.accepted:
+            try:
+                self.tube.put(Entry.serialize(entry))
+            except Exception as e:
+                entry.fail(e)
+
+@event('plugin.register')
+def register_plugin():
+    plugin.register(PluginTarantoolQueueInput, 'tarantool_queue_input', api_ver=2)
```
paranoidi: Plugin naming convention, we don't have `output` or `input` in names. Input should be just `from_tarantool` and output just `tarantool`.

flexget/plugins/clients/tarantool_queue.py
```diff
@@ -0,0 +1,145 @@
...
+
+            self._is_prepared = True
+
+class PluginTarantoolQueueInput(PluginTarantoolQueueBase):
+    def on_task_input(self, task, config):
+        if not config.get('enabled', True):
+            return
+
+        self.config = config
+        tarantool_entries = self.tube.take(0)
```
paranoidi: this can be None or list, not ideal type usage. Empty list is preferable and makes following checks easier.

`tarantool_entries is not None and len(tarantool_entries) > 0`

->

`while tarantool_entries`

flexget/plugins/clients/tarantool_queue.py
```diff
@@ -0,0 +1,145 @@
...
+        do_continue = True
+        entry_count = 0
```
paranoidi: Unnecessary variable, same as `len(entries)`

flexget/plugins/clients/tarantool_queue.py
```diff
@@ -0,0 +1,145 @@
...
+        entries = []
+        while tarantool_entries is not None and len(tarantool_entries) > 0 and do_continue:
+            for tarantool_entry in tarantool_entries:
+                entry_id, entry_type, entry_data = tarantool_entry
+                entry = Entry.deserialize(entry_data, None)
+                entries.append(entry)
+                self.tube.ack(entry_id)
+                entry_count += 1
+                if entry_count > 30:
```
paranoidi: Why 30? Should this be configurable?

paranoidi:
Nice to have new contributors!

flexget/plugins/clients/tarantool_queue.py
StefftheEmperor2: Thats not true. there are more entries added in the loop.

StefftheEmperor2:
Im still pretty new to python. Hope you're patient with me ;-)

flexget/plugins/clients/tarantool_queue.py
```diff
@@ -0,0 +1,119 @@
...
+        queue = TarantoolQueue(tarantool_module,
+                                     config.get('host'),
+                                     config.get('port'),
+                                     config.get('user'),
+                                     config.get('password'))
+
+        return queue
+
+    def get_tube(self, tarantool_module, config):
+        return self.get_queue(tarantool_module, config).get_tube(config.get('tube'))
```
metaMMA: I see you're re-using the method name `get_tube` from ` class TarantoolQueue: `. The way this is written makes is look almost recursive. I wonder if it would be better to choose a different name, if you can still have it makes sense,

flexget/plugins/clients/tarantool_queue.py
```diff
@@ -0,0 +1,119 @@
...
+            {'type': 'boolean'},
+            {
+                'type': 'object',
+                'properties': {
+                    'host': {'type': 'string'},
+                    'port': {'type': 'integer'},
+                    'user': {'type': 'string'},
+                    'password': {'type': 'string'},
+                    'tube': {'type': 'string'},
+                    'chunk_size': {'type': 'integer', 'default': 30}
```
metaMMA: You'll want to setup the schema with `required` fields. It looks like, as of now, the config would pass even with missing fields.  More info can be found here (https://flexget.readthedocs.io/en/latest/develop/schemas.html) or message me on slack and I can assist.

metaMMA:
> Im still pretty new to python. Hope you're patient with me ;-)

Was in the same boat recently. I think you'll be getting lots of useful feedback and suggestions. Thanks for contributing!

flexget/plugins/clients/tarantool_queue.py
liiight: make that fstrings ;)

flexget/plugins/clients/tarantool_queue.py
```diff
@@ -0,0 +1,125 @@
...
+
+class PluginTarantoolQueueInput(PluginTarantoolQueueBase):
+    def on_task_input(self, task, config):
+        tarantool_module = self.get_tarantool_module()
+        tube = self.get_tube_of_queue(tarantool_module, config)
+        tarantool_entries = tube.take(0)
+        do_continue = True
+        entry_count = 0
+        entries = []
+        while tarantool_entries and do_continue:
```
liiight, line 93: no need for a control variable, you can just `break`

PR #2639, opened 2020-03-24
Serienjunkies urlrewriter: Added seasonpack fallback. Added hosters

StefftheEmperor2:
Signed-off-by: Stefan Bischoff <stefan.bischoff@9plus.de>

Serienjunkies.org changed their hosters and often, the way they update series: As share-online went down, they had to provide some alternatives. Previously, the uploaders where encouaged to only use share-online, uploaded an cz (cant even remember what it stands for). Now they may use every hoster they want. In the past, every episode got its own link. Now they often generate crypted folders and update them, as new episodes are broadcasted. There is a new fallback to handle that case. I also changed the config of the hoster configuration. This can now be a list. The best (=first) match is used then.

### DIFF
--- a/flexget/components/sites/sites/serienjunkies.py
+++ b/flexget/components/sites/sites/serienjunkies.py
@@ -10,6 +10,7 @@ from flexget.components.sites.urlrewriting import UrlRewritingError
 from flexget.event import event
 from flexget.utils import requests
 from flexget.utils.soup import get_soup
+from flexget.config_schema import one_or_more
 
 logger = logger.bind(name='serienjunkies')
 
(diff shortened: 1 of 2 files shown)


flexget/components/sites/sites/serienjunkies.py
```diff
@@ -107,43 +179,82 @@ def parse_downloads(self, series_url, search_title):
...
+            do_exit = True
+
+        if not do_exit:
             # find episode language
             episode_lang = episode.find_previous('strong', text=re.compile('Sprache')).next_sibling
             if not episode_lang:
-                logger.warning('No language found for: {}', series_url)
-                continue
+                logger.warning('No language found for: {}', episode_lang)
+                do_exit = True
```
ianstalk, line 215: The way you use this flag is kind of awkward and makes your code really hard to follow. Can you refactor it to be cleaner?

flexget/components/sites/sites/serienjunkies.py
```diff
@@ -27,10 +28,80 @@
...
+    'ddl': ['ddl.to'],
+    'filefactory': ['filefactory.com'],
+    'filer': ['filer.net'],
+    'nitroflare': ['nitroflare.com'],
+    'oboom': ['oboom.com'],
+    'rapidgator': ['rapidgator.net'],
+    'turbobit': ['turbobit.net'],
+    'uploaded': ['uploaded.to', 'uploaded.net'],
+    'share-online': ['share-online.to']
+}
```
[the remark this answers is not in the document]
paranoidi: Yeah, these look like various file hosters. We should have a single plugin for each.

StefftheEmperor2:
Can you help me with that. I'm new to python and that is beyond my skillset. The map should not change very often. And it's not a configuration. Could be exported to a file anyway.

flexget/components/sites/sites/serienjunkies.py
```diff
@@ -27,10 +28,80 @@
...
+
+class SerienjunkiesMatch:
+    def __init__(self):
+        self.hoster = None
+        self.quality = float("inf")
+        self.found = False
+
+    @classmethod
+    def factory(cls, href, next_sibling, hoster_config):
+        instance = cls()
```
liiight, line 59: if the first thing you do is init the class then there is no reason to make this a classmethod, you can just move this logic to the init (or keep it as a separate function that init calls)

flexget/components/sites/sites/serienjunkies.py
```diff
@@ -27,10 +28,80 @@
...
+        if hoster == 'so':
+            hoster = 'share-online'
+
+        if hoster in ['uploaded', 'share-online']:
+            if hoster == 'uploaded':
+                url_tag = 'ul'
+            if hoster == 'share-online':
+                url_tag = 'so'
+            pattern = (
+                    r'http:\/\/download\.serienjunkies\.org.*%s_.*\.html' % url_tag
```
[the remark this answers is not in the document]
liiight: also please use fstrings

flexget/components/sites/sites/serienjunkies.py
```diff
@@ -27,10 +28,80 @@
...
+                instance.found = True
+        return instance
+
+    @staticmethod
+    def does_match(href, next_sibling, hoster):
+        match_found = False
+        if hoster == 'ul':
+            hoster = 'uploaded'
+        if hoster == 'so':
+            hoster = 'share-online'
```
liiight, line 87: more pythonic to use dict for this:
```suggestion
		hoster = {'ul': 'uploaded', 'so': 'share-online'}.get(hoster, hoster)
```

flexget/components/sites/sites/serienjunkies.py
```diff
@@ -27,10 +28,80 @@
...
+
+        if hoster in ['uploaded', 'share-online']:
+            if hoster == 'uploaded':
+                url_tag = 'ul'
```
liiight, line 91: so you switch the hoster and then reuse its value to get the original value as a url-tag? that's kinda weird.

flexget/components/sites/sites/serienjunkies.py
```diff
@@ -107,43 +179,82 @@ def parse_downloads(self, series_url, search_title):
             if not episode_title:
                 continue
 
-            # find download container
-            episode = episode_title.parent
-            if not episode:
-                continue
+            urls = self.find_urls(urls, episode_title)
+
+        if len(urls) == 0 and season_pack_fallback:
```
liiight, line 184: more pythonic to do this:
```suggestion
        if not urls and season_pack_fallback:
```

flexget/components/sites/sites/serienjunkies.py
```diff
@@ -107,43 +179,82 @@ def parse_downloads(self, series_url, search_title):
...
+                    urls = self.find_urls(urls, season_title)
 
+        return urls
+
+    def find_urls(self, urls, episode_title):
+        # find download container
+        episode = episode_title.parent
+        do_exit = False
+        if not episode:
+            do_exit = True
```
liiight, line 208: seems like you can just return in many place instead of using this flag

PR #2643, opened 2020-03-27
Add missing deps for Docker

X-dark:
* zlib-dev and jpeg-dev are needed for Pillow building
* packaging is needed at runtime for clients/pyload plugin

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -7,6 +7,8 @@ RUN apk add --no-cache --upgrade \
         build-base \
         libffi-dev \
         openssl-dev \
+        zlib-dev \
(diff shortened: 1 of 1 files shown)


paranoidi:
@gazpachoking  This makes me think we should probably try to introduce a better dependency declaration to plugins so we can get a list of these programmatically. Would possibly open better ways to notify/warn user. Maybe even automatic dependency installation via pip ?

paranoidi:
@X-dark pillow requirements were loosened, is this still required?

X-dark:
Nope. Works now without.

PR #2648, opened 2020-04-08
[add] from_transmission: Add 'transmission_errorString'

fenduru:
### Motivation for changes:
Currently there is no way to differentiate between a tracker error due to temporary downtime, vs permanent deletion of the torrent, making it difficult to automate cleanup tasks.

### Detailed changes:
- Adds [`errorString`](https://github.com/transmission/transmission/blob/master/extras/rpc-spec.txt#L195) to the whitelist of Transmission RPC fields to be populated on entries.


paranoidi:
Meeseeks good changes, thanks for your contribution

PR #2652, opened 2020-04-13
[fix] Removed pillow and pystray as dependencies

### LINKED ISSUE #2651: Dependencies on non pure Python modules
Hi,

Sorry for [x-posting](https://discuss.flexget.com/t/dependencies-on-non-pure-python-modules/5086). This is kind of a development issue so it might fit here better than in the forum. Feel free to just close this and reply on Discourse.

When upgrading Flexget I stumbled upon the fact that now it depends on Pillow, as per:

https://github.com/Flexget/Flexget/pull/2621

My Flexget runs on an embedded aarch64 device so every time you add a dep on a module that links against C libraries (Pillow depends on zlib and libjpeg) I have to cross compile the necessary native extensions, distribute them etc. Either that or manually installing a skimmed `requirements.txt` and finally installing Flexget passing `--no-dependencies` to `pip install`.

Does adding a tray icon justify this hassle for users? No saying that’s wrong of course, just trying to bring to your attention this use case and the potential annoyance for users.

Could perhaps these things be made optional at install/build time by relaxing the hard dependency?

Thanks.

liiight:
### Motivation for changes:
Pillow needs compilation which hurts Flexget's compatibility.

### Detailed changes:
- Removed `pillow` and `pystray` from deps. If someone wants to run the tray icon, they can install them manually.

### Addressed issues:
- Fixes #2651

### DIFF
--- a/flexget/tray_icon.py
+++ b/flexget/tray_icon.py
@@ -17,7 +17,7 @@ try:
 
     _import_success = True
 except Exception as e:
-    logger.debug('Could not import pystray: {}', e)
+    logger.debug('Could not load tray icon: {}', e)
     _import_success = False
 
 
--- a/requirements.in
+++ b/requirements.in
@@ -30,4 +30,3 @@ pyparsing>=2.0.3
 zxcvbn-python
 progressbar==2.5
 more-itertools
-Pillow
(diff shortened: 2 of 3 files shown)


alirz1:
So when i try to install flexget on a synology Nas, using Python 3.x instlaled via entware.
It fails while trying to install pillow. So looks like that dependency is still there?

Im using this article for the isntallation of flexget: 
https://flexget.com/InstallWizard/SynologyNAS/PythonPackage

2sheds:
> It fails while trying to install pillow. So looks like that dependency is still there?
Came here for a similar thing. This change hasn't made it into a new release yet, so if you install from pip (which is on 3.1.51) it will still have the dependency. Our options right now are to wait until 3.1.52 or build from source.

PR #2658, opened 2020-04-21
[change] new search plugin for Filelist with API

putyn:
### Motivation for changes:
Filelist released an API which can be used to search for torrents, this should replace the old search plugin which scraped the site.

### DIFF
--- a/flexget/components/sites/sites/filelist_api.py
+++ b/flexget/components/sites/sites/filelist_api.py
@@ -1,0 +1,155 @@
+from http import HTTPStatus
+from loguru import logger
+from flexget import plugin
+from flexget.config_schema import one_or_more
+from flexget.entry import Entry
+from flexget.event import event
+from flexget.utils.requests import RequestException
+from flexget.utils.requests import Session as RequestSession
+
+logger = logger.bind(name='filelist')
+requests = RequestSession()
+
+
+class SearchFileList:
+    """
+        FileList search plugin w/ API w00t.
+    """
+
+    api_url = 'https://filelist.io/api.php'
+
+    valid_categories = {
+        'Anime': '24',
+        'Audio': '11',
+        'Desene': '15',
+        'Diverse': '18',
+        'Docs': '16',
+        'FLAC': '5',
+        'Filme 3D': '25',
+        'Filme 4K': '6',
+        'Filme 4K Blu-Ray': '26',
+        'Filme Blu-Ray': '20',
+        'Filme DVD': '2',
+        'Filme DVD-RO': '3',
+        'Filme HD': '4',
+        'Filme HD-RO': '19',
+        'Filme SD': '1',
+        'Jocuri Console': '10',
+        'Jocuri PC': '9',
+        'Linux': '17',
+        'Mobile': '22',
+        'Programe': '8',
+        'Seriale 4K': '27',
+        'Seriale HD': '21',
+        'Seriale SD': '23',
+        'Sport': '13',
+        'TV': '14',
+        'Videoclip': '12',
+        'XXX': '7',
+    }
+
+    valid_extras = ['internal', 'moderated', 'freeleech', 'doubleup']
+    
+    schema = {
+        'type': 'object',
+        'properties': {
+            'username': {'type': 'string'},
+            'passkey': {'type': 'string'},
+            'category': one_or_more({'type': 'string', 'enum': list(valid_categories.keys())}),
+            'internal': {'type': 'boolean', 'default': False},
+            'moderated': {'type': 'boolean', 'default': False},
+            'freeleech': {'type': 'boolean', 'default': False},
+            'doubleup': {'type': 'boolean', 'default': False},
+        },
+        'required': ['username', 'passkey'],
+        'additionalProperties': False,
+    }
+
+    def get(self, url, params):
+
+        try:
+            response = requests.get(url, params=params, raise_status=False)
+        except RequestException as e:
+            raise plugin.PluginError(f'FileList request failed badly! {e}')
+
+        if not response.ok:
+            http_status = HTTPStatus(response.status_code)
+            error_message = response.json().get('error', http_status.description)
+
+            raise plugin.PluginError(f'FileList request failed; err code: {http_status}; err msg: `{error_message}`')
+
+        return response
+
+    @plugin.internet(logger)
+    def search(self, task, entry, config):
+        """
+            Search for entries on FileList
+        """
+        entries = []
+
+        # mandatory params
+        params = {'username': config['username'], 'passkey': config['passkey'], 'action': 'search-torrents'}
+
+        # category
+        if config.get('category'):
+            params['category'] = (
+                ','.join(self.valid_categories[cat] for cat in config['category'])
+                if isinstance(config.get('category'), list)
+                else self.valid_categories[config.get('category')]
+            )
+
+        # extras: internal release, moderated torrent, freeleech
+        params.update({extra: 1 for extra in valid_extras if config.get(extra)})
+
+        # set season/episode if series
+        if entry.get('series_episode'):
+            params['episode'] = entry.get('series_episode')
+        if entry.get('series_season'):
+            params['season'] = entry.get('series_season')
+
+        for search_title in entry.get('search_strings', [entry.get('title')]):
+
+            if entry.get('imdb_id'):
+                params['type'] = 'imdb'
+                params['query'] = entry.get('imdb_id')
+                params['name'] = search_title
+            else:
+                params['type'] = 'name'
+                params['query'] = search_title
+
+            # get them results
+            try:
+                response = self.get(self.api_url, params)
+            except RequestException as e:
+                raise plugin.PluginError(f'FileList request failed badly! {e}')
+
+            results = response.json()
+            if not results:
+                logger.verbose('No torrent found on Filelist for `{}`', search_title)
+            else:
+                logger.verbose('{} torrent(s) were found on Filelist for `{}`', len(results), search_title)
+
+            for result in results:
+                entry = Entry()
+
+                entry['title'] = result['name']
+                entry['url'] = result['download_link']
+                entry['imdb'] = result['imdb']
+                # size is returned in bytes but expected in MiB
+                entry['content_size'] = result['size'] / 2 ** 20
+                entry['torrent_snatches'] = result['times_completed']
+                entry['torrent_seeds'] = result['seeders']
+                entry['torrent_leeches'] = result['leechers']
+                entry['torrent_internal'] = bool(result['internal'])
+                entry['torrent_moderated'] = bool(result['moderated'])
+                entry['torrent_freeleech'] = bool(result['freeleech'])
+                entry['torrent_genres'] = [genres.strip() for genres in result['small_description'].split(',')]
+
+                entries.append(entry)
+
+        return entries
+
+
+@event('plugin.register')
+def register_plugin():
+    plugin.register(SearchFileList, 'filelist_api', interfaces=['search'], api_ver=2)
--- a/flexget/components/sites/sites/filelist.py
+++ b/flexget/components/sites/sites/filelist.py
@@ -74,6 +74,7 @@ class SearchFileList:
 
     schema = {
         'type': 'object',
+        'deprecated': 'plugin filelist is deprecated, please consider using plugin filelist_api',
         'properties': {
             'username': {'type': 'string'},
             'password': {'type': 'string'},


flexget/components/sites/sites/filelist_api.py
```diff
@@ -0,0 +1,180 @@
...
+from flexget.entry import Entry
+from flexget.event import event
+from flexget.utils.requests import TimedLimiter, RequestException
+from flexget.utils.requests import Session as RequestSession
+
+logger = logger.bind(name='filelist')
+requests = RequestSession()
+
+
+class SearchFileList(object):
```
liiight, line 13: No need to inherit from object

flexget/components/sites/sites/filelist_api.py
```diff
@@ -0,0 +1,180 @@
...
+    }
+
+    def get(self, url, params):
+
+        try:
+            response = requests.get(url, params=params, raise_status=False)
+        except RequestException as e:
+            raise plugin.PluginError(f'FileList request failed badly! {e}')
+
+        if response.status_code in self.error_codes.keys():
```
liiight, line 82: You can check `if not response.ok`. Also, there's no need to map these statuses, python has that built in:https://docs.python.org/3/library/http.html#http.HTTPStatus

flexget/components/sites/sites/filelist_api.py
```diff
@@ -0,0 +1,180 @@
...
+            # get error message if we have one, else use generic
+            error_message = response.json().get('error', self.error_codes[response.status_code])
+
+            raise plugin.PluginError(
+                f'FileList request failed; err code: {response.status_code}; err msg: `{error_message}`'
+            )
+
+        return response
+
+    def dump_entry(self, entry):
```
liiight, line 92: This seems redundant, you can just log the entry itself

flexget/components/sites/sites/filelist_api.py
```diff
@@ -0,0 +1,180 @@
...
+        entry_items = {key: value for key, value in entry.items()}
+        logger.debug(f'entry_items={entry_items}')
+
+    @plugin.internet(logger)
+    def search(self, task, entry, config):
+        """
+            Search for entries on FileList
+        """
+        entries = list()
```
liiight, line 101: Best practice is to use a literal list
```
entries = []
```

flexget/components/sites/sites/filelist_api.py
```diff
@@ -0,0 +1,180 @@
...
+        # TODO remove later
+        self.dump_entry(entry)
+
+        # mandatory params
+        params = {'username': config['username'], 'passkey': config['passkey'], 'action': 'search-torrents'}
+
+        # category
+        if config.get('category'):
+            params['category'] = (
+                ','.join([self.valid_categories[cat] for cat in config['category']])
```
liiight, line 112: You can pass a generator to join, no need for the list comprehension here

flexget/components/sites/sites/filelist_api.py
```diff
@@ -0,0 +1,180 @@
...
+            try:
+                response = self.get(self.api_url, params)
+                # TODO remove later
+                logger.debug(f'Requesting: {response.url}')
+            except RequestException as e:
+                raise plugin.PluginError(f'FileList request failed badly! {e}')
+
+            results = response.json()
+            if not results:
+                logger.info(f'No torrent found on Filelist for `{search_title}`')
```
liiight, line 151: Dont use fstrings in logger. See how it's used in other loggers. Also, I'm not sure this should be info level, verbose should be sufficient

flexget/components/sites/sites/filelist_api.py
```diff
@@ -0,0 +1,180 @@
...
+                logger.info(f'I have found {len(results)} torrent(s) on Filelist for `{search_title}`')
+
+            for result in results:
+                entry = Entry()
+
+                entry['title'] = result['name']
+                entry['url'] = result['download_link']
+                entry['imdb'] = result['imdb']
+                # size is returned in bytes but expected in MiB
+                entry['content_size'] = result['size'] / 2 ** 20
```
liiight, line 162: We have a util for this

flexget/components/sites/sites/filelist_api.py
putyn: it was _inherited_ from the old plugin 😄

flexget/components/sites/sites/filelist_api.py
putyn: I am mapping the error codes, because those are specific to the implementation of the API. I saw that in most case the API will also return an error message, however not always so the error_codes dict holds some generic message.

flexget/components/sites/sites/filelist_api.py
putyn: yes...also inherited from another plugin, will log the entry 😄

flexget/components/sites/sites/filelist_api.py
putyn: will do

flexget/components/sites/sites/filelist_api.py
putyn: `','.join(self.valid_categories[cat] for cat in config['category'])` you mean like this ? TIL

flexget/components/sites/sites/filelist_api.py
putyn: I will change the level to verbose and I will look into fstrings and logger.

flexget/components/sites/sites/filelist_api.py
putyn: I have only found this https://github.com/Flexget/Flexget/blob/e33eed8f1a508555aeaffa3bf3b24600a8fea420/flexget/utils/tools.py#L388 however that takes a string as input

flexget/components/sites/sites/filelist_api.py
```diff
@@ -0,0 +1,176 @@
...
+        except RequestException as e:
+            raise plugin.PluginError(f'FileList request failed badly! {e}')
+
+        if response.status_code in self.error_codes.keys():
+            # get error message if we have one, else use generic
+            error_message = response.json().get('error', self.error_codes[response.status_code])
+
+            raise plugin.PluginError(
+                f'FileList request failed; err code: {response.status_code}; err msg: `{error_message}`'
+            )
```
liiight, line 88: consider something like this:
```suggestion
        if not response.ok:
        	status = HTTPStatus(response.status_code)
            error_message = response.json().get('error', status.description)

            raise plugin.PluginError(
                f'FileList request failed; err code: {status}; err msg: `{error_message}`'
            )
```
pretty much exactly what you want but without needing to hold that mapping

flexget/components/sites/sites/filelist_api.py
liiight: I guess that fine

flexget/components/sites/sites/filelist_api.py
liiight: I know, not blaming you here, just an observation

flexget/components/sites/sites/filelist_api.py
putyn: should be fixed now, I guess your suggestion is better as it handles all possible status codes, not only the one explicitly mapped.

putyn:
@liiight - is there a way to make the other plugin deprecated and move users to the new one ?

flexget/components/sites/sites/filelist_api.py
```diff
@@ -0,0 +1,166 @@
...
+            # TODO remove later
+            logger.debug('get params: {}', params)
+
+            # get them results
+            try:
+                response = self.get(self.api_url, params)
+                # TODO remove later
+                logger.debug('Requesting: {}', response.url)
+            except RequestException as e:
+                raise plugin.PluginError('FileList request failed badly! {}', e)
```
liiight, line 133: this should be fstring as its not a logger

flexget/components/sites/sites/filelist_api.py
```diff
@@ -0,0 +1,166 @@
...
+                params['query'] = search_title
+
+            # TODO remove later
+            logger.debug('get params: {}', params)
+
+            # get them results
+            try:
+                response = self.get(self.api_url, params)
+                # TODO remove later
+                logger.debug('Requesting: {}', response.url)
```
liiight, line 131: we already debug requests, no need for this

flexget/components/sites/sites/filelist_api.py
```diff
@@ -0,0 +1,166 @@
...
+            except RequestException as e:
+                raise plugin.PluginError('FileList request failed badly! {}', e)
+
+            results = response.json()
+            if not results:
```
liiight, line 136: will this be an empty list if there are no results?

flexget/components/sites/sites/filelist_api.py
```diff
@@ -0,0 +1,166 @@
...
+                logger.verbose('No torrent found on Filelist for `{}`', search_title)
+            else:
+                logger.verbose('I have found {} torrent(s) on Filelist for `{}`', len(results), search_title)
```
liiight, line 139: please phrase this like other log messages

flexget/components/sites/sites/filelist_api.py
```diff
@@ -0,0 +1,166 @@
...
+            params['category'] = (
+                ','.join(self.valid_categories[cat] for cat in config['category'])
+                if isinstance(config.get('category'), list)
+                else self.valid_categories[config.get('category')]
+            )
+
+        # extras: internal release, moderated torrent, freeleech
+        for extra_name, extra_value in self.valid_extras.items():
+            if config.get(extra_name):
+                params[extra_name] = extra_value
```
liiight, line 106: seems a bit redundant to hold these "extra" in an additional dict, espicially since their values is always one. consider this:
```suggestion
        params.update({extra: 1 for extra in ('internal', 'moderated', 'freeleech', 'doubleup') if config.get(extra)})
```

liiight:
@putyn you can use the 'deprecated' keyword in the schema of the old plugin. See how its used in other places

flexget/components/sites/sites/filelist_api.py
putyn: hmm looks like I missed that 👍

flexget/components/sites/sites/filelist_api.py
putyn: yes, no need for this anymore.

flexget/components/sites/sites/filelist_api.py
putyn: as far as i remember if i don't get any result it will be an empty list, hopefully they don't change the behavior 😄

flexget/components/sites/sites/filelist_api.py
putyn: anything in mind ? can't really think of a better phrase

flexget/components/sites/sites/filelist_api.py
putyn: while true the value of the extras are always one, i will still like to keep it this ways since if they introduce more extras with different values i will just need to update the dict.

flexget/components/sites/sites/filelist_api.py
liiight:
```suggestion
                logger.verbose('{} torrent(s) were found for `{}`', len(results), search_title)
```

flexget/components/sites/sites/filelist_api.py
liiight: the `1` is just simple way to pass `True` via params, so this value will probably not change. How about this for a compromise, set the class attribute to hold a tuple of the names and iterate over those

putyn: willing to take a list over a tuple ? 😄 if yes, all required changes are committed

flexget/components/sites/sites/filelist_api.py
```diff
@@ -0,0 +1,157 @@
...
+                entry['torrent_snatches'] = result['times_completed']
+                entry['torrent_seeds'] = result['seeders']
+                entry['torrent_leeches'] = result['leechers']
+                entry['torrent_internal'] = bool(result['internal'])
+                entry['torrent_moderated'] = bool(result['moderated'])
+                entry['torrent_freeleech'] = bool(result['freeleech'])
+                entry['torrent_genres'] = [genres.strip() for genres in result['small_description'].split(',')]
+
+                # TODO remove later
+                logger.debug(entry)
```
liiight, line 149: seems redundant to me to debug the entry

liiight:
Just last comment remaining and we're good to go 😃

putyn:
Done, I also squashed all the commits in one hopefully I did it right as I am still learning to use git 😄 
Regarding the old plugin, should we deprecate it ? If yes new pull request or can we include the change in this one ?

liiight:
You can include the change in this PR

putyn:
old plugin deprecated, I think we are good to go 😄

putyn:
@liiight - do you think we need more change ?

liiight:
Merged. Thanks for your patience and collaboration on this @putyn!

PR #2664, opened 2020-05-02
New episode URL in RSS

gorynychzmey:
### Motivation for changes:
Link to eposide in RSS was changed.
### Detailed changes:
- Changed regexp to parse season and episode


paranoidi:
Thanks!

PR #2675, opened 2020-05-16
Fix deploy

kristenmills:
### Motivation for changes:
Currently the release script isn't working, hopefully this fixes. 

### Detailed explanation:
I think the issue has been that for whatever reason, skipped steps used to not show up in the checks list but they've started show up. And the way the deployment API works, deploys will fail if there are unsuccessful checks on the PR (and with the release skipped check showing up on every commit, that's considered not passing). We're not seeing this problem over on webui because webui release script is a seperate workflow. The reason it wasn't on this repo, was to make sure that we ran the tests before making a release. This made since prior to the change to using the deployment API, but it's not necessary now. So hopefully this fixes. the issue

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,0 +1,57 @@
+name: Build and release to PyPI
+on:
+  deployment: {}
+jobs:
+  release:
+    runs-on: ubuntu-latest
+    env:
+      PYTHON_VERSION: "3.7"
+    strategy:
+      matrix:
+        node: [12.x]
+    steps:
+    - uses: actions/checkout@v1.1.0
+      with:
+        token: ${{ secrets.flexgetbot_pat }}
+    - name: Setup Git User
+      run: |
+        git config user.email ${{ secrets.git_email }}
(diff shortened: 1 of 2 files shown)


.github/workflows/release.yml
```diff
@@ -0,0 +1,56 @@
...
+        token: ${{ secrets.flexgetbot_pat }}
+    - name: Setup Git User
+      run: |
+        git config user.email ${{ secrets.git_email }}
+        git config user.name ${{ secrets.git_user }}
+    - name: Set up Python ${{ env.PYTHON_VERSION }}
+      uses: actions/setup-python@v1
+      with:
+        python-version: ${{ env.PYTHON_VERSION }}
+      # TODO: Uncomment this when supported: https://github.com/actions/cache/issues/63
```
kristenmills, line 24: WOW this was fixed yesterday 🙏 
https://github.com/actions/cache/pull/263
https://github.com/actions/cache/issues/63

kristenmills: i'm going to just wait until the v2 releases before using this functionality but i'll change the comment

PR #2684, opened 2020-05-25
Avoids running crossmatch from if no entries

soloam:
### Motivation for changes:

Avoids running the from input plugin if no entries to check. This is very useful if, for example, the from is using the "from_task" plugin. This will avoid running the task in the "from_task" if no entries exists to be matched.

### Detailed changes:
- Check the len of the entries iterator before running the from in crossmatch

### DIFF
--- a/flexget/plugins/filter/crossmatch.py
+++ b/flexget/plugins/filter/crossmatch.py
@@ -43,6 +43,10 @@ class CrossMatch:
         action = config['action']
         all_fields = config['all_fields']
 
+        if len(task.entries) == 0:
+          logger.trace('Stopping crossmatch filter because of no entries to check')
+          return
+        
         match_entries = aggregate_inputs(task, config['from'])
 
         # perform action on intersecting entries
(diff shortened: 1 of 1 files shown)


flexget/plugins/filter/crossmatch.py
```diff
@@ -43,6 +43,10 @@ def on_task_filter(self, task, config):
         action = config['action']
         all_fields = config['all_fields']
 
+        if len(task.entries) == 0:
```
gazpachoking, line 46: Seems a reasonable optimization. The more pythonic way to put this would be:
```suggestion
        if not task.entries:
```

soloam: Yes that is a better way! Do I need to change the commit? Or is it done?

Thank you

gazpachoking: I merged it in, thanks!

gazpachoking:
Thanks!

PR #2685, opened 2020-05-25
Fail entries instead of throwing a PluginError when Jinja templating fails

ianstalk:
### Motivation for changes:
Flexget currently aborts the entire task when JINJA2 templating failes (e.g. when the `all_series` plugin fails to parse an entry). This is super annoying, since it skips processing any additional entries after the failures.

### Detailed changes:
- Swap out raising a `PluginError` for using `entry.fail()`


paranoidi:
How about copy and other file operations? All should behave same.

ianstalk:
> How about copy and other file operations? All should behave same.

They all use the same code, despite the misleading file name.

ianstalk:
I renamed the file because it's definitely confusing.

liiight:
LGTM

Issue #2681, opened 2020-05-20
Plugin `flexget.components.imdb.api` failed to import dependencies


ctag:
### Expected behaviour:

Arch Linux flexget AUR package. Python3.8 and kernel 5.4.41.
I'm attempting to start flexget with `flexget daemon start`. It appears to start, but spews errors.

### Steps to reproduce:
- Step 1: Installed flexget package and created a minimal config.yml
- Step 2: Start flexget with `flexget daemon start`

#### Config:
```yaml
tasks:
  test task:
    rss: https://www.archlinux.org/feeds/releases/
    accept_all: yes
    download: /mnt/linux/
```
  
#### Log:
<details>
 <summary>(click to expand)</summary>

```
[deluge@bns-anarch ~]$ /usr/bin/flexget daemon start
2020-05-20 10:27:27 CRITICAL plugin                        Plugin `flexget.plugins.daemon.web_server` failed to import dependencies
Traceback (most recent call last):

  File "/usr/bin/flexget", line 11, in <module>
    load_entry_point('FlexGet==3.1.30', 'console_scripts', 'flexget')()
    -> <function load_entry_point at 0x7f6a38231b80>
  File "/usr/lib/python3.8/site-packages/flexget/__init__.py", line 41, in main
    manager.start()
    |       -> <function Manager.start at 0x7f6a36514a60>
    -> <flexget.manager.Manager object at 0x7f6a3650f7f0>
  File "/usr/lib/python3.8/site-packages/flexget/manager.py", line 366, in start
    self.initialize()
    |    -> <function Manager.initialize at 0x7f6a36514820>
    -> <flexget.manager.Manager object at 0x7f6a3650f7f0>
  File "/usr/lib/python3.8/site-packages/flexget/manager.py", line 206, in initialize
    plugin.load_plugins(
    |      -> <function load_plugins at 0x7f6a3688f820>
    -> <module 'flexget.plugin' from '/usr/lib/python3.8/site-packages/flexget/plugin.py'>
  File "/usr/lib/python3.8/site-packages/flexget/plugin.py", line 555, in load_plugins
    _load_plugins_from_dirs(extra_plugins)
    |                       -> ['/srv/deluge/.config/flexget/plugins', '/usr/lib/python3.8/site-packages/flexget/plugins']
    -> <function _load_plugins_from_dirs at 0x7f6a3688f670>
[...]

PR #2689, opened 2020-05-31
Switch from flask-restplus to flask-restx

ianstalk:
### Motivation for changes:
Flask-RESTPlus is abandoned, and is no longer compatible with the latest version of  werkzeug, on which it depends. The community has forked the project, and created Flask-RESTX, which seems to be a drop-in replacement.

### Detailed changes:
- change `flask-restplus==0.10.1` out for `flask-restx==0.2.0` (the latest stable release of flask-restx).
- change all references to `flask_restplus` to `flask_restx`

### Addressed issues:
- Fixes https://github.com/Flexget/Flexget/issues/2681 .

#### To Do:

- The API test `test_crash_logs_without_crash_log` fails, but the same seems to happen with `develop`, so I think that test was already broken

### DIFF
--- a/flexget/api/app.py
+++ b/flexget/api/app.py
@@ -7,8 +7,8 @@ from functools import partial, wraps
 from flask import Flask, jsonify, make_response, request
 from flask_compress import Compress
 from flask_cors import CORS
-from flask_restplus import Api as RestPlusAPI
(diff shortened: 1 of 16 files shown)


liiight:
LGTM

PR #2693, opened 2020-06-03
Expand scope of disable plugin

gazpachoking:
### Motivation for changes:
This is a revival of #2295. Part of that PR I already merged in because it made sense and the old way was causing bugs. e8a474654672caf6c667a5e40c3dde5c8000f5fc Right now, it's hard or impossible to disable certain plugins like estimators or urlrewriters which don't get configured directly in a task. disable plugin also does not currently work for plugins that are run by other plugins. e.g. inside sequence, or configure_series, or limit

### Detailed changes:

- Adds get_plugin_by_name method to `Task`s. This throws a ValueError if the plugin is disabled. (is that the best error to throw?)
- Adds get_plugins method to `Task`s. This just skips any plugins disabled on that task.
Any plugins should use these methods over the functions provided in `flexget.plugin` when executing anything in a task context.

TODO:
- [ ] Decide if this is good.
- [ ] Update disable_urlrewriters plugin in favor of this method.
- [ ] Add some more tests
- [ ] Make it work for parsers?
- [ ] Make it work for the movie_metainfo plugins?

### DIFF
--- a/flexget/components/estimate_release/estimate_release.py
+++ b/flexget/components/estimate_release/estimate_release.py
@@ -21,9 +21,13 @@ class EstimateRelease:
         """
 
         logger.debug(entry['title'])
-        estimators = [
-            e.instance.estimate for e in plugin.get_plugins(interface='estimate_release')
-        ]
+        # If the entry is in a task context, use the task interface to get estimator plugins, so that we don't get ones
+        # that are disabled on that task.
+        if entry.task:
+            get_plugins = entry.task.get_plugins
+        else:
+            get_plugins = plugin.get_plugins
+        estimators = [e.instance.estimate for e in get_plugins(interface='estimate_release')]
         for estimator in sorted(
             estimators, key=lambda e: getattr(e, 'priority', plugin.PRIORITY_DEFAULT), reverse=True
         ):
--- a/flexget/task.py
+++ b/flexget/task.py
@@ -6,7 +6,7 @@ import random
 import string
 import threading
 from functools import total_ordering, wraps
-from typing import Union, Iterable, List
+from typing import Union, Iterable, List, Optional
 
 from loguru import logger
 from sqlalchemy import Column, Integer, String, Unicode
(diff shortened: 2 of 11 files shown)


flexget/task.py
```diff
@@ -428,8 +434,27 @@ def find_entry(self, category='entries', **values):
                 return entry
         return None
 
-    def plugins(self, phase=None):
-        """Get currently enabled plugins.
+    def get_plugin_by_name(self, name: str) -> PluginInfo:
+        if name in self.disabled_plugins:
+            raise ValueError(f'`{name}` plugin is disabled for this task.')
```
liiight, line 439: I think a dedicated `DisabledPluginError` would be more explicit and obvious in this case

flexget/task.py
```diff
@@ -428,8 +434,27 @@ def find_entry(self, category='entries', **values):
...
+        return get_plugin_by_name(name)
+
+    def get_plugins(
+        self,
+        phase: Optional[str] = None,
+        interface: Optional[str] = None,
+        category: Optional[str] = None,
```
liiight, line 446: maybe it's an opportunity to move plugin categories into an enum?

flexget/components/estimate_release/estimate_release.py
```diff
@@ -21,9 +21,13 @@ def estimate(self, entry):
         """
 
         logger.debug(entry['title'])
-        estimators = [
-            e.instance.estimate for e in plugin.get_plugins(interface='estimate_release')
-        ]
+        # If the entry is in a task context, use the task interface to get estimator plugins, so that we don't get ones
+        # that are disabled on that task.
```
liiight, line 25: not sure i follow on how an entry can NOT be in a task context

gazpachoking: I'm not sure if it really can either. But I remember weirdness when accessing entry.task at some point before, so this was sorta protecting against that. I'm also not really a fan of entry.task, I'm wondering if we shouldn't just switch to passing task as a separate parameter.

flexget/task.py
gazpachoking: Yeah, that might be better.

flexget/components/estimate_release/estimate_release.py
liiight: makes sense. It'll also create a single place for this logic which would be better

PR #2697, opened 2020-06-08
[add] Added free parameter to iptorrents search plugin

bobcatthebaker:
### Motivation for changes:
To be able to search IPTorrents for freeleech-only results.
### Detailed changes:
- Added a `free` parameter to the IPTorrents search plugin schema that will include `&free=on` in the URL if set to `True`.

### Addressed issues:
- N/A

- N/A

```
    discover:
      what:
        - next_series_episodes: yes
      from:
        - iptorrents:
            rss_key: ***
            uid: ***
            password: ***
            free: True <---
```
```
> bin/flexget execute --tasks SEARCH_TV --discover-now --no-cache
2020-06-08 15:57:28 VERBOSE  task_queue                    There are 1 tasks to execute. Shutdown will commence when they have completed.
2020-06-08 15:57:28 VERBOSE  series        SEARCH_TV       Series `RuPaul's Drag Race` is already configured in series plugin
2020-06-08 15:57:28 VERBOSE  series        SEARCH_TV       Series `Better Call Saul` is already configured in series plugin
2020-06-08 15:57:28 VERBOSE  discover      SEARCH_TV       Discovering 2 titles ...
2020-06-08 15:57:28 INFO     discover      SEARCH_TV       Ignoring interval because of --discover-now
2020-06-08 15:57:28 VERBOSE  discover      SEARCH_TV       Searching for `RuPaul's Drag Race S01E01` with plugin `iptorrents` (1 of 2)
2020-06-08 15:57:30 VERBOSE  discover      SEARCH_TV       Searching for `Better Call Saul S01E01` with plugin `iptorrents` (2 of 2)
2020-06-08 15:57:31 VERBOSE  details       SEARCH_TV       Produced 42 entries.
2020-06-08 15:57:31 VERBOSE  task          SEARCH_TV       ACCEPTED: `RuPauls Drag Race S01E01 1080p WEBRip x264...` by series plugin because choosing first acceptable match
2020-06-08 15:57:31 VERBOSE  task          SEARCH_TV       ACCEPTED: `Better Call Saul S01E01 Uno 1080p WEB-DL DD5 1 H 264...` by series plugin because choosing first acceptable match
2020-06-08 15:57:31 VERBOSE  limit_new     SEARCH_TV       Allowed RuPauls Drag Race S01E01 1080p WEBRip x264... (https://iptorrents.com/...)
2020-06-08 15:57:31 VERBOSE  limit_new     SEARCH_TV       Allowed Better Call Saul S01E01 Uno 1080p WEB-DL DD5 1 H 264... (https://iptorrents.com/...)
2020-06-08 15:57:31 VERBOSE  details       SEARCH_TV       Summary - Accepted: 2 (Rejected: 0 Undecided: 40 Failed: 0)
```
#### To Do:

- [ ] Pending maintainer recommendations

### DIFF
--- a/flexget/components/sites/sites/iptorrents.py
+++ b/flexget/components/sites/sites/iptorrents.py
@@ -86,6 +86,7 @@ class UrlRewriteIPTorrents:
             'category': one_or_more(
                 {'oneOf': [{'type': 'integer'}, {'type': 'string', 'enum': list(CATEGORIES)}]}
             ),
+            'free': {'type': 'boolean', 'default': False},
         },
         'required': ['rss_key', 'uid', 'password'],
         'additionalProperties': False,
@@ -135,9 +136,14 @@ class UrlRewriteIPTorrents:
             query = normalize_unicode(search_string)
             query = quote_plus(query.encode('utf8'))
 
-            url = "{base_url}/t?{filter}&q={query}&qf=".format(
-                base_url=BASE_URL, filter=filter_url, query=query
-            )
+            if config.get('free'):
+                url = "{base_url}/t?{filter}&free=on&q={query}&qf=".format(
+                    base_url=BASE_URL, filter=filter_url, query=query
+                )
+            else:
+                url = "{base_url}/t?{filter}&q={query}&qf=".format(
+                    base_url=BASE_URL, filter=filter_url, query=query
+                )
             logger.debug('searching with url: {}', url)
             req = requests.get(
                 url, cookies={'uid': str(config['uid']), 'pass': config['password']}
--- a/flexget/components/sites/sites/iptorrents.py
+++ b/flexget/components/sites/sites/iptorrents.py
@@ -55,6 +55,7 @@ CATEGORIES = {
 }
 
 BASE_URL = 'https://iptorrents.com'
+SEARCH_URL = 'https://iptorrents.com/t?'
 
 
 class UrlRewriteIPTorrents:
@@ -107,7 +108,7 @@ class UrlRewriteIPTorrents:
             logger.error("Didn't actually get a URL...")
         else:
             logger.debug('Got the URL: {}', entry['url'])
-        if entry['url'].startswith(BASE_URL + '/t?'):
+        if entry['url'].startswith(SEARCH_URL):
             # use search
             results = self.search(task, entry)
             if not results:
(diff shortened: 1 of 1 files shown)


bobcatthebaker:
IPTorrents functionality is great, but there is one major hindrance that I've experienced with it beyond the feature I'm implementing in this PR.

IPTorrents searches by the IMDB ID when you use the `Movies` or `TV shows` tabs. For TV show episodes, you can append a season number and episode number in the format `S01E01` alongside the IMDB ID and it will still work as expected.

Currently Flexget appears to use the title of the show or movie which does not always return results, especially in the case of TV shows. I've tried inspecting the requests and responses to/from IPTorrents and I can't figure out where the IMDB ID succeeds in matching listings because it's not in the title and not in the tags.

Anyway, I'd like to implement searching by IMDB ID as a feature alongside this. Is there a list of plugins/components that would need to support injecting the IMDB ID into the `search_strings` entry field? I know one of them would be `next_series_episodes`.

flexget/components/sites/sites/iptorrents.py
```diff
@@ -135,9 +136,14 @@ def search(self, task, entry, config=None):
             query = normalize_unicode(search_string)
             query = quote_plus(query.encode('utf8'))
 
-            url = "{base_url}/t?{filter}&q={query}&qf=".format(
-                base_url=BASE_URL, filter=filter_url, query=query
-            )
+            if config.get('free'):
```
liiight, line 139: i know it's not your code but we might as well improve it. All query params should be passed inside a params dict and not added to the url string like that.
Also, please use fstrings instead of `.format()`. 

If it's possible to change these formats to fstring in other places in the plugin that would be swell

bobcatthebaker: I believe I've addressed these improvements, please let me know if there's anything further that I can do.

bobcatthebaker:
I left another comment earlier about another feature request, if that could be addressed I'd like to get started on that in my spare time.

bobcatthebaker:
I did some various testing to be sure before I gave the all-clear, and I discovered that the plugin was ultimately searching with URLs like `https://www.iptorrents.com/t?q=Leave%2BNo%2BTrace%25282018%2529&qf=` rather than `https://www.iptorrents.com/t?q=Leave+No+Trace+%282018%29&qf=#torrents` due to the usage of the of the `quote_plus` method from `urllib.parse`.

I also discovered that the freeleech-only search depends on where you place the `free=on` parameter in the URL. It must come directly after the `/t?` so it cannot depend on the params dict. I've added a new URL specifically for freeleech-only search that works with the rest of the params that may be passed.

liiight:
@bobcatthebaker thank you for this!

PR #2699, opened 2020-06-17
[fix] manual: actually allow running in from_task context

delroth:
### Motivation for changes:

`from_task` + `manual` is documented to work, and was clearly designed to work, but doesn't actually work because the boolean logic checking the `allow_manual` option is broken.

cc @gazpachoking who implemented `from_task`

### Detailed changes:
- Fixed the broken boolean logic
- Modified the from_task test to actually exercise this code path.

(Note: also re-formatted with Black.)

### DIFF
--- a/flexget/plugins/operate/manual.py
+++ b/flexget/plugins/operate/manual.py
@@ -18,10 +18,8 @@ class ManualTask:
             return
         # If --task hasn't been specified disable this plugin
         if (
-            not task.options.tasks
-            or task.name not in task.options.tasks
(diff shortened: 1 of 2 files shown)


delroth:
Ah, nevermind, seems like the semantics of allow_manual is actually not what I understood it to be. So this fix doesn't really make sense. I'll move this to an issue report, tl;dr: right now a subtask with `manual: yes` isn't getting run unless it's explicitly listed on the command line.

delroth:
#2700

PR #2706, opened 2020-06-24
WIP: SFTP unit tests

ianstalk:
Putting up an early WIP PR to get feedback. I couldn't find a PyTest mock SFTP fixture so I'm trying to just do it with pure mocks. I'm having trouble figuring out a way to simulate callback behavior; any suggestions?

### DIFF
--- a/flexget/components/ftp/sftp_client.py
+++ b/flexget/components/ftp/sftp_client.py
@@ -26,324 +26,6 @@ NodeHandler = Callable[[str], None]
 logger = logger.bind(name='sftp_client')
 
 
-class SftpClient:
-    def __init__(
-        self,
-        host: str,
-        port: int,
-        username: str,
-        password: Optional[str] = None,
-        private_key: Optional[str] = None,
-        private_key_pass: Optional[str] = None,
-        connection_tries: int = 3,
-    ):
-
-        if not pysftp:
-            raise plugin.DependencyError(
-                issued_by='sftp_client',
-                missing='pysftp',
-                message='sftp client requires the pysftp Python module.',
-            )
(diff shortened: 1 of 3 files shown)


flexget/components/ftp/sftp_client.py
```diff
@@ -26,38 +26,230 @@
 logger = logger.bind(name='sftp_client')
 
 
-class SftpClient:
+class SftpError(Exception):
```
ianstalk, line 29: Sorry these changes are so huge. There are two reasons:

1) I made an `SftpClientBuilder` class and moved a bunch of logic out of `SftpClient` both for separation of concerns, and to allow for better dependency injection.
2) Objects declared below a given class can't be used as a type annotation in the constructor, which forced me to move SftpClient to the bottom.

liiight:
This is way too big to fit in my brain. The only feedback I can give is please run black :)

flexget/components/ftp/sftp_client.py
```diff
@@ -26,38 +26,230 @@
...
+
+    def get_null_handler(self) -> NodeHandler:
+        """
+        Builds a noop node handler suitable for use with pysftp.Connection.walktree
+        """
+        return partial(Handlers.null_node_handler)
+
+
+class Handlers:
+    @classmethod
```
paranoidi, line 106: Complicated use of classmethods, they are generally alternative constructors.

ianstalk: I basically want to use them like static methods, but with access to the other methods in the class. Would it be preferable to fully qualify them instead?

PR #2710, opened 2020-07-22
[fix] `anilist`: Exception if field is not set on the site

BrutuZ:
### Motivation for changes:
Discovered that site entries could have empty fields, usually ones added from very early announcements or teasers. That would result in exceptions due to parsing errors since some fields were taken for granted.
### Detailed changes:
- Replaced most direct references to dict keys with `dict().get()`
- Convert to string before performing case operations to avoid `NoneType` exceptions
- Changed `al_links` output:
  - From Dicts inside a List: `[{'site': '*site_name*', 'url': '*site_url*'}, ... ]`
  - To single Dict: `{'*site_name*': '*site_url*', ... }`
- Alphabetize `Entry()` fields to improve code readability

### DIFF
--- a/flexget/plugins/input/anilist.py
+++ b/flexget/plugins/input/anilist.py
@@ -122,56 +122,65 @@ class AniList(object):
             try:
                 list_response = list_response.json()['data']
                 logger.debug('JSON output: {}', list_response)
-                for list_status in list_response['collection']['statuses']:
+                for list_status in list_response.get('collection', {}).get('statuses'):
                     if (
                         selected_list_name
-                        and list_status['name'].lower() not in selected_list_name
+                        and str(list_status.get('name')).lower() not in selected_list_name
                     ):
                         continue
                     for anime in list_status['list']:
-                        anime = anime['anime']
+                        anime = anime.get('anime')
                         has_selected_release_status = (
-                            anime['status'].lower() in selected_release_status
+                            str(anime.get('status')).lower() in selected_release_status
                             or 'all' in selected_release_status
                         )
                         has_selected_type = (
-                            anime['format'].lower() in selected_formats
+                            str(anime.get('format')).lower() in selected_formats
                             or 'all' in selected_formats
                         )
                         if has_selected_type and has_selected_release_status:
                             entry = Entry()
-                            entry['title'] = anime['title']['romaji']
-                            entry['al_title'] = anime['title']
-                            entry['al_format'] = anime['format']
-                            entry['al_release_status'] = anime['status'].capitalize()
-                            entry['al_list'] = list_status['name']
-                            entry['al_list_status'] = (
-                                list_status['status'].capitalize()
-                                if list_status.get('status')
-                                else ''
-                            )
-                            entry['alternate_name'] = anime.get('synonyms', [])
-                            if (
-                                anime['title'].get('english')
-                                and anime['title'].get('english') != anime['title']['romaji']
-                                and anime['title'].get('english') not in entry['alternate_name']
-                            ):
-                                entry['alternate_name'].insert(0, anime['title']['english'])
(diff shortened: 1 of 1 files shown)


flexget/plugins/input/anilist.py
```diff
@@ -122,54 +122,65 @@ def on_task_input(self, task, config):
             try:
                 list_response = list_response.json()['data']
                 logger.debug('JSON output: {}', list_response)
-                for list_status in list_response['collection']['statuses']:
+                for list_status in list_response.get('collection', {}).get('statuses'):
```
liiight, line 125: while you're going overly defensive, maybe add a default to `.get('statuses')`

flexget/plugins/input/anilist.py
```diff
@@ -122,54 +122,65 @@ def on_task_input(self, task, config):
...
                     if (
                         selected_list_name
-                        and list_status['name'].lower() not in selected_list_name
+                        and str(list_status.get('name')).lower() not in selected_list_name
```
ianstalk, line 128: Are all of these fields really all that likely to contain something other than a string?

BrutuZ: They can be empty, in which case trying to perform a case operation like `lower()` results in a `NoneType` exception

flexget/plugins/input/anilist.py
BrutuZ: If it's not defined \ empty, won't the loop just be gracefully skipped as intended?

liiight: no, if it doesn't exist the value will be `None` and it'll crash when you try to iterate over it.
try this:
```suggestion
                for list_status in list_response.get('collection', {}).get('statuses', []):
```

BrutuZ: It is highly unlikely that field would be empty since the status lists are a default structure and AFAIK can't be deleted by the user, but one can never be too safe. Implemented 👍

flexget/plugins/input/anilist.py
ianstalk: The string `"None"` doesn't seem like a good pick-- it looks too much like a real value (see: https://www.wired.com/story/null-license-plate-landed-one-hacker-ticket-hell/). Empty string would be better, or you can check for the presence of a value.

flexget/plugins/input/anilist.py
```diff
@@ -122,54 +122,65 @@ def on_task_input(self, task, config):
...
-                        and list_status['name'].lower() not in selected_list_name
+                        and list_status.get('name', 'None').lower() not in selected_list_name
                     ):
                         continue
                     for anime in list_status['list']:
-                        anime = anime['anime']
+                        anime = anime.get('anime')
                         has_selected_release_status = (
-                            anime['status'].lower() in selected_release_status
+                            anime.get('status', 'None').lower() in selected_release_status
```
ianstalk, line 134: My comment was on an outdated line-- I don't like `'None'` because it could be a valid value. If you absolutely need a value, I'd use an empty string.

BrutuZ:
Empty strings caused an exception when performing case operations, the reason I had the string conversions in place.



Sent from my Samsung Galaxy smartphone.

ianstalk:
Then check if it has a value before manipulating it.

BrutuZ:
Testing for empty value now

BrutuZ:
Bump.

I'm stuck at this 3.1.68 branch since a couple of entries in my list have empty values that aren't handled without this, hence crashing the task.

BrutuZ:
Testing discord bridge

BrutuZ:
Paging @FlexGet

gazpachoking:
Seems good to me, thanks!

PR #2720, opened 2020-08-23
SFTP plugin changes

Winter:
### Motivation for changes:
The motivation to add filename support was so that when I'm downloading content from my server I'm able to automatically name things according to my naming convention. Rather than using the source filename. I've been doing this for nearly two years but only just decided to make my changes onto the actual branch as @ianstalk had cleaned up the spaghetti that it was beforehand. 

### Detailed changes:
- Added a new parameter to the schema, 'filename' 
- Implemented the filename field across sftp.py and sftp_client.py 

### Addressed issues:
- Fixes an issue that I was experiencing where the 'to' field wasn't rendering the jinja2 template.

```
        'type': 'object',
        'properties': {
            'to': {'type': 'string', 'format': 'path'},
            'filename': {'type': 'string', 'default': ''},
            'recursive': {'type': 'boolean', 'default': True},
            'delete_origin': {'type': 'boolean', 'default': False},
            'socket_timeout_sec': {'type': 'integer', 'default': DEFAULT_SOCKET_TIMEOUT_SEC},
            'connection_tries': {'type': 'integer', 'default': DEFAULT_CONNECT_TRIES},
        },
        'required': ['to'],
```
```
Not really much to test as it's a minor change and the output is the same
```
#### To Do:
- Nothing, apart from making sure that path's are properly sanitized

### DIFF
--- a/.github/workflows/changelog.yml
+++ b/.github/workflows/changelog.yml
@@ -13,12 +13,13 @@ jobs:
       PYTHON_VERSION: 3.7
     steps:
     - uses: actions/checkout@v2
+      with:
+        fetch-depth: 0
     - uses: actions/checkout@v2
       with:
         repository: Flexget/wiki
         path: wiki
         token: ${{ secrets.flexgetbot_pat }}
-        fetch-depth: 0
         ref: master
     - name: Setup Git User
       run: |


ianstalk:
The change looks good, but I'm not 100% sure of the utility of it. You could always just download and rename as separate tasks.

Winter:
> 
> 
> The change looks good, but I'm not 100% sure of the utility of it. You could always just download and rename as separate tasks.

Completely true. Which was my original solution, but it just felt unnecessary to have to have a separate task, dedicated renaming things when I it can be done at the time of download. Makes the config a lot more concise, as well as it's something that you would probably 'expect' to be able to do. 

Alternatively, we could just change the 'to' field to accept an absolute file path. Albeit, that does have it's own caveats with requiring more checks in place but doesn't add anything new to the schema; if that is the issue.

liiight (approved):
LGTM. Change looks good an use case seems common, so I don't see a reason against this.

PR #2728, opened 2020-09-12
add ettvcentral.com domain to ettv rewriter

flakyton:
### Motivation for changes:
ETTV has moved

### Detailed changes:
- add ettvcentral.com to the list of ETTV domains


gazpachoking:
Thanks!

Issue #2725, opened 2020-09-07
webUI cost too much time on version query


xmt328:
### Expected behaviour:

query lastest version on background

### Actual behaviour:

visit http://*/api/server/version in api/core/server.py

it will query https://pypi.org/pypi/FlexGet/json for  lastest version info

it cost 20+ seconds on loading page

p.s. maybe it's fast for most country, but it's hard to connect to pypi in China

### Steps to reproduce:
- Step 1: open webUI on any browser, it will loading 20s+

#### Config:
```yaml
web_server:
  bind: 0.0.0.0
  port: 8080
  web_ui: yes
```

### Additional information:

- FlexGet version: 3.1.70
- Python version: 3.7.3
- Installation method: pip
- Using daemon (yes/no): yes
- OS and version: debian 10.2

liiight:
Good point. @kristenmills maybe we can make the option to compare to latest version clickable?

kristenmills:
We currently use the version endpoint to check whether you are already logged in which is why the first request to version blocks page load. Probably should make just make a dedicated endpoint that's cheaper for that purpose.

kristenmills:
This should be fixed in the next version of flexget.

PR #2731, opened 2020-09-20
[add] /user/token to check if you are logged in

kristenmills:
### Motivation for changes:
* Currently we use the version endpoint to determine whether or not you are logged in on webui. This initial call will block rendering of everything until resolved. The version endpoint has to call out to PyPy which is not great in places that it takes a while to resolve. So instead i made a dedicated endpoint to use for this purpose that doesn't really do anything at all other than check auth. 

### Addressed issues:
- #2725 (not fixed as i still have to make webui changes)

### DIFF
--- a/dev-requirements.txt
+++ b/dev-requirements.txt
@@ -1,6 +1,6 @@
 sphinx==1.6.3
 click==6.7
-pylint
+pylint>=2.6.0 #pylint < 2.6 doesn't work with isort5
 coverage==4.4.1
 flake8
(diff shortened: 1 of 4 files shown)


liiight:
I'm on board with the motivation, but adding an endpoint that basically does nothing feels kinda non idiomatic. Maybe change it to get the current user token? That feels more natural, accomplishes the same goal and also actually adds missing functionality.

kristenmills:
Sure i'm fine with doing that.

kristenmills:
@liiight made a `GET user/token` endpoint to use instead

liiight (approved):
LGTM

PR #2738, opened 2020-09-26
Torznab: Add timeout parameter

eSoares:
### Motivation for changes:
This is needed for under some environments with many sources on the torznab, the answer will take more than 30 seconds.
Let the user configure this to adapt according to their needs.

### Detailed changes:
- Added a new timeout parameter to the torznab input plugin.

```
timeout : 120s
```

### DIFF
--- a/flexget/plugins/input/torznab.py
+++ b/flexget/plugins/input/torznab.py
@@ -165,7 +165,7 @@ class Torznab:
         logger.info('Fetching URL: {}', url)
 
         try:
-            response = task.requests.get(url, timeout=self.timeout))
+            response = task.requests.get(url, timeout=self.timeout)
         except RequestException as e:
             raise PluginError("Failed fetching '{}': {}".format(url, e))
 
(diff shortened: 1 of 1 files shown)


eSoares:
As far as I know, this still applies.

gazpachoking:
Do we need to make this an option? Can we just bump it up to a longer value? If it's not inside a loop I don't see a problem with that.

eSoares:
> Do we need to make this an option? Can we just bump it up to a longer value? If it's not inside a loop I don't see a problem with that.

I'm happy to bump to a bigger value, if the answer arrives faster the current users should have no change in using the plugin.
I find useful to be configurable because depending on the number of sources in torznab may need more or less time, creating the need for something more dynamic. 

Maybe bump for 120 seconds in requests made in this plugin?

eSoares:
Specially with the recent issues handling Couldflare anti-bots, having a web browser running (see https://github.com/Jackett/Jackett/issues/9029 ) to handle that, causes a higher round-trip time for get answers, depending on which trackers are used and the machine that is running the software.
I still think this issue exist and should be added the option.

gazpachoking:
Sorry for the delay. Someone want to update the wiki page?

PR #2740, opened 2020-09-28
Log filter plugin

gazpachoking:
### Motivation for changes:

Converts [log_filter](https://github.com/tarzasai/.flexget/blob/master/plugins/log_filter.py) plugin by @tarzasai to work with the new loguru logging system

### Detailed changes:
- Adds `add_filter` and `remove_filter` to the flexget logging system, which allows changing filters after the handlers have been created.

#### To Do:

- [ ] Consider how useful this is/ whether there are better solutions.
- [ ] Consider whether people will use this then want support based on an incomplete log. (maybe exclude the filter from crash report files somehow?)
- [ ] consider whether this should be at root rather than task level
- [ ] Consider config format. Maybe something a bit more flexible like:
```yaml
log_filter:
- level: debug
  logger: backlog
- message: blahblah
  task: mytask  # This would be if it gets configured at root level
```

### DIFF
--- a/flexget/log.py
+++ b/flexget/log.py
@@ -9,7 +9,7 @@ import sys
 import threading
 import uuid
 import warnings
-from typing import Iterator
+from typing import Callable, Iterator
 
 import loguru
 from loguru import logger
@@ -96,6 +96,22 @@ _startup_buffer_id = None
 _logging_started = False
 # Stores the last 100 debug messages
 debug_buffer = collections.deque(maxlen=100)
+_log_filters = []  # Stores filter functions
+
+
+def _log_filterer(record):
+    """This is the function we add to our loguru handlers. It will dynamically use all filters we add later."""
+    return all(f(record) for f in _log_filters)
+
+
+def add_filter(func: Callable[['loguru.Record'], bool]):
+    """Adds a filter function to the log handlers."""
+    _log_filters.append(func)
+
+
+def remove_filter(func: Callable[['loguru.Record'], bool]):
+    """Removes a filter function from the log handlers."""
+    _log_filters.remove(func)
 
 
 def initialize(unit_test: bool = False) -> None:
@@ -168,6 +184,7 @@ def start(
             retention=int(os.environ.get(ENV_MAXCOUNT, 9)),
             encoding='utf-8',
             format=LOG_FORMAT,
+            filter=_log_filterer,
         )
 
     # without --cron we log to console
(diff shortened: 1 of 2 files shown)


ksurl:
this would be useful for ignoring errors from one of my feeds that gets 502 http error sometimes when it's "empty"

BrutuZ:
I used to have the following rules on the old plugin:
```yaml
    log_filter:
      parser_internal:
        - appears to be an episode pack
        - looks like series
      content_size:
        - rejecting
      inputs:
        - HTTP error 502
      trakt_list:
        - 0 movie(s), 0 show(s), 0 episode(s), 0 season(s)
```

frankyw:
It would be great to get this working as-is, and then perhaps iterate on it. I personally think it makes sense at the plugin level, as implemented. Would love to see this working again and included as a usable plugin.

frankyw:
Any updates on this, or any chance of getting this going soon? My log is still a mess of irrelevant messages, and it would be great to deal with this problem. Thanks!

gazpachoking:
Meh, I'm sorta ambivalent on the new config format.  I allowed it at the base level, such that it could be configured for all tasks at once, as well as changed it to allow filtering on task name, level, plugin name, and message. If someone has feelings one way or the other I'm open to anything. This is the current state of the config:
```yaml
log_filter:
# Just filter by the message
- message: blah blah
# Filter a specific message from a specific plugin
- message: some other thing
  plugin: but only for this plugin
# Filter a specific message for a given task
- message: aoeuaoeu
  task: only on this task
# Really specific filter
- message: tetete
  level: info
  task: something
  plugin: series
```

BrutuZ:
The plugin at root level instead of under the task or ´global´ template has the added benefit or issue, depending on situation of cleaning up the task's `--dump-config`.
I personally liked it

soloam:
I like the fact that you could also specify the task and the plugin! I would only prefere if the task and plugin would be one of, string or list. That would allow to add multiple tasks or plugins and not duplicate the entire message and level info! Like 

```yaml

log_filter:
# Just filter by the message
- message: blah blah
# Filter a specific message from a specific plugin
- message: some other thing
  plugin: but only for this plugin
# Really specific filter
- message: tetete
  level: info
  task: something
  plugin: series
# Really Really specific filter
- message: aoeuaoeu
  task: 
    - only on this task
    - or this task
  plugin: 
    - this
    - or that
```

That was the approach that I took on the hooks, and I can use it in any way!

gazpachoking:
Oh man. I was considering going back to the old way because it seemed a bit simpler. Adding another layer of lists inside this config is sorta the opposite of what I was considering. 😝

gazpachoking:
At least for tasks, to avoid repetition it can still be configured at the task level, and repeated across multiple tasks using a template.

BrutuZ:
I just assumed it took lists with wildcards, `schedules` style ._.

gazpachoking:
As for allowing a list of plugins, it seems unlikely that 2 plugins are going to have the same log message needing filtered?

soloam:
No? If more than one task got the same message? What I was suggesting was string or list! If we need one we only set a task/plugin... If more that one, we set list... 

But I might be over thinking this...

gazpachoking:
I think for setting tasks it can be included in the specific task. In fact now I'm thinking making it available at the root level might not have been a good idea at all. The only reason that might be useful is to filter out messages that aren't from a specific task (like from the manager or webui or schedules.)

PR #2713, opened 2020-08-02
switched piratebay to new API due to breaking change from 3rd party

trananhtuan:
### Motivation for changes:
- Current piratebay search plugin has been broken since 3rd party website redesign. Static HTML parsing no longer works.

### Detailed changes:
- Use piratebay JSON API (https://apibay.org) instead.
- Maintain support for old config and mirrors (such as TOR address http://piratebayztemzmv.onion/).

### Addressed issues:
- Static HTML parsing no longer works due to 3rd party change.

### Config usage if relevant (new plugin or updated schema):
```
tasks:
  test-piratebay:
    seen: no
    discover:
      what:
        - rss: https://www.rogerebert.com/feed
      from:
        - piratebay:
            category: highres movies
            sort_by: seeds
            url: https://apibay.org/
      release_estimations: ignore
    accept_all: yes
```

```
C:\Users\tuant\PycharmProjects\Flexget\venv\Scripts\python.exe C:/Users/tuant/PycharmProjects/Flexget/flexget_vanilla.py --loglevel=debug -c test-piratebay.yml execute --discover-now --now
2020-08-02 10:21:05 DEBUG    manager                       Figuring out config load paths
2020-08-02 10:21:05 DEBUG    manager                       Found config: C:\Users\tuant\PycharmProjects\Flexget\test-piratebay.yml
2020-08-02 10:21:05 DEBUG    manager                       Config file C:\Users\tuant\PycharmProjects\Flexget\test-piratebay.yml selected
2020-08-02 10:21:05 DEBUG    manager                       sys.defaultencoding: utf-8
2020-08-02 10:21:05 DEBUG    manager                       sys.getfilesystemencoding: utf-8
2020-08-02 10:21:05 DEBUG    manager                       flexget detected io encoding: utf-8
2020-08-02 10:21:05 DEBUG    manager                       os.path.supports_unicode_filenames: True
2020-08-02 10:21:05 DEBUG    plugin                        Trying to load plugins from: ['C:\\Users\\tuant\\PycharmProjects\\Flexget\\plugins', 'C:\\Users\\tuant\\PycharmProjects\\Flexget\\flexget\\plugins']
2020-08-02 10:21:06 DEBUG    plugin                        Plugin `memusage` requires plugin `ext lib `guppy`` to load.
[...]

PR #2745, opened 2020-10-05
fix url rewriting tests for pirate bay

kristenmills:
### Motivation for changes:
Tests were broken when #2713  was merged. This fixes them

### DIFF
--- a/flexget/components/sites/sites/piratebay.py
+++ b/flexget/components/sites/sites/piratebay.py
@@ -103,7 +103,7 @@ class UrlRewritePirateBay:
         url = entry['url']
         if url.endswith('.torrent'):
             return False
-        return bool(self.url_match.match(url))
+        return bool(self.url_match.match(url) or self.url_search.match(url))
 
     # urlrewriter API
     def url_rewrite(self, task, entry):
--- a/flexget/components/sites/sites/piratebay.py
+++ b/flexget/components/sites/sites/piratebay.py
@@ -103,7 +103,7 @@ class UrlRewritePirateBay:
         url = entry['url']
         if url.endswith('.torrent'):
             return False
-        return bool(self.url_match.match(url) or self.url_search.match(url))
+        return any((self.url_match.match(url), self.url_search.match(url)))
 
     # urlrewriter API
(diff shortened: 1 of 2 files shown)


flexget/components/sites/sites/piratebay.py
```diff
@@ -103,7 +103,7 @@ def url_rewritable(self, task, entry):
         url = entry['url']
         if url.endswith('.torrent'):
             return False
-        return bool(self.url_match.match(url))
+        return bool(self.url_match.match(url) or self.url_search.match(url))
```
liiight, line 106: I'm not sure what the logic here is supposed to do

kristenmills: The old `url_match` regex used to match both the search url and the regular urls...now it does not.

kristenmills: Unclear if that behavior is necessary but we have a test that makes sure it works so i figured we'd expect it to. 
https://github.com/Flexget/Flexget/blob/c7a6ab37c27b11afec6e71503efe2398d8d748cd/flexget/tests/test_urlrewriting.py#L47-L52

liiight: ok, how about something like this then:
```suggestion
        return any((self.url_match.match(url), self.url_search.match(url)))
```

liiight (approved):

flexget/components/sites/sites/piratebay.py
gazpachoking: `any` and `or` both feel fine to me. I'd ditch the `bool()` cast if we stick with `or` though.

gazpachoking (approved):
LGTM. Thanks for bailing me out again @kristenmills. 😝 I kept meaning to fix this and kept putting it off.

flexget/components/sites/sites/piratebay.py
kristenmills: I think the bool cast is still necessary if we use `or` since otherwise it will return the match object. But i can just change it to any if that seems cleaner.

PR #2751, opened 2020-10-15
Add type annotations (top-level files)

vgerak:
### Motivation for changes:
Add/enrich type annotations
### Detailed changes:
- Add type annotation to `update-changelog.py` and files directly under `flexget` module
- Add `mypy_cache/` and `dmypy.json` to gitignore

```
# before
❯ dmypy run -- --follow-imports=skip flexget/*.py  | tail -1
Found 174 errors in 14 files (checked 16 source files)
# after
❯ dmypy run -- --follow-imports=skip flexget/*.py  | tail -1
Found 83 errors in 14 files (checked 16 source files)
```
I can keep adding commits to this PR
### Addressed issues:
- No such issue exists, we can add one if you wish

#### To Do:
- [ ] Add type annotations for the rest of the modules

### DIFF
--- a/flexget/entry.py
+++ b/flexget/entry.py
@@ -2,7 +2,7 @@ import functools
 import types
 import warnings
 from enum import Enum
-from typing import Callable, Iterable, Mapping, Optional, Sequence, Union
+from typing import Callable, Iterable, Mapping, Optional, Sequence, Union, Dict
 
 from loguru import logger
 
@@ -23,10 +23,10 @@ class EntryState(Enum):
     @property
     def color(self) -> str:
         return {
-            self.ACCEPTED: 'green',
-            self.REJECTED: 'red',
-            self.FAILED: 'RED',
-            self.UNDECIDED: 'dim',
+            EntryState.ACCEPTED: 'green',
+            EntryState.REJECTED: 'red',
+            EntryState.FAILED: 'RED',
+            EntryState.UNDECIDED: 'dim',
         }[self]
 
     @property
@@ -34,22 +34,22 @@ class EntryState(Enum):
         return f'<{self.color}>{self.value.upper()}</>'
 
     # __str__, __eq__, and __hash__ are make these states almost interchangeable with the old plain string states
-    def __str__(self):
+    def __str__(self) -> str:
         return self.value
 
-    def __eq__(self, other):
+    def __eq__(self, other) -> bool:
         if isinstance(other, str):
             return self.value == other
         return super().__eq__(other)
 
-    def __hash__(self):
+    def __hash__(self) -> int:
         return hash(self.value)
 
 
 class EntryUnicodeError(Exception):
     """This exception is thrown when trying to set non-unicode compatible field value to entry."""
 
-    def __init__(self, key, value):
+    def __init__(self, key: str, value) -> None:
         self.key = key
         self.value = value
 
@@ -87,7 +87,12 @@ class Entry(LazyDict, Serializer):
         # Make sure constructor does not escape our __setitem__ enforcement
         self.update(*args, **kwargs)
 
-    def trace(self, message: str, operation: Optional[str] = None, plugin: Optional[str] = None):
+    def trace(
+        self,
+        message: Optional[str],
+        operation: str = None,
+        plugin: str = None,
+    ) -> None:
         """
         Adds trace message to the entry which should contain useful information about why
         plugin did not operate on entry. Accept and Reject messages are added to trace automatically.
@@ -102,7 +107,7 @@ class Entry(LazyDict, Serializer):
         if item not in self.traces:
             self.traces.append(item)
 
-    def run_hooks(self, action: str, **kwargs):
+    def run_hooks(self, action: str, **kwargs) -> None:
         """
         Run hooks that have been registered for given ``action``.
 
@@ -112,7 +117,7 @@ class Entry(LazyDict, Serializer):
         for func in self._hooks[action]:
             func(self, **kwargs)
 
-    def add_hook(self, action: str, func: Callable, **kwargs):
+    def add_hook(self, action: str, func: Callable, **kwargs) -> None:
         """
         Add a hook for ``action`` to this entry.
 
@@ -126,7 +131,7 @@ class Entry(LazyDict, Serializer):
         except KeyError:
             raise ValueError('`%s` is not a valid entry action' % action)
 
-    def on_accept(self, func: Callable, **kwargs):
+    def on_accept(self, func: Callable, **kwargs) -> None:
         """
         Register a function to be called when this entry is accepted.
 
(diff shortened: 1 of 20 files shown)


vgerak:
Only annotated the vars that I could deduce with some confidence, so it's not complete, but should be correct.

3 thoughts moving forward:
1. I was thinking to open different PRs for each module, rather than one (pretty big) PR that would keep being updated. This way it would be easier to review/comment and get integrated into the codebase. I leave the decision to the maintainers.
2. Should I convert the logging strings to f-strings instead of `str.format()`? Since lazy logger evaluation (`%`) is not used, should be safe to do so.
3. I get many warnings for `logger.verbose()` calls, I see that loguru.logger does not have that attribute. I assume it's correct (since it would be crashing everywhere otherwise), but I don't see how it's working. Please point me to the file that enables this behaviour, so I can take a look (curiosity).

ianstalk (commented):
Awesome!

flexget/entry.py
```diff
@@ -87,7 +87,12 @@ def __init__(self, *args, **kwargs):
         # Make sure constructor does not escape our __setitem__ enforcement
         self.update(*args, **kwargs)
 
-    def trace(self, message: str, operation: Optional[str] = None, plugin: Optional[str] = None):
+    def trace(
+        self,
+        message: Optional[str],
```
gazpachoking, line 92: Looks like the optional bit got moved to message here instead of the other 2 arguments.

flexget/entry.py
```diff
@@ -171,12 +176,12 @@ def accept(self, reason: Optional[str] = None, **kwargs):
             # Run entry on_accept hooks
             self.run_hooks('accept', reason=reason, **kwargs)
 
-    def reject(self, reason: Optional[str] = None, **kwargs):
+    def reject(self, reason: str = None, **kwargs) -> None:
```
gazpachoking, line 179: Is there a reason you took out the Optional on these?

flexget/manager.py
```diff
@@ -254,9 +255,9 @@ def has_lock(self) -> bool:
 
     def execute(
         self,
-        options: Optional[Union[dict, argparse.Namespace]] = None,
```
gazpachoking, line 257: More `Optional`s I'm not sure should be removed in this file.

gazpachoking (commented):
I only scanned through the first half so far, but looks good to me. I mostly had questions on your changes in the use of `Optional`. Figured I'd just ask if there is a reason you changed them before scanning the rest.

gazpachoking:
> 3 thoughts moving forward:
> 
>     1. I was thinking to open different PRs for each module, rather than one (pretty big) PR that would keep being updated. This way it would be easier to review/comment and get integrated into the codebase. I leave the decision to the maintainers.

Agreed, this would be nice, much easier to check over and merge in stuff that's good in smaller chunks.

>     2. Should I convert the logging strings to f-strings instead of `str.format()`? Since lazy logger evaluation (`%`) is not used, should be safe to do so.

We are using loguru, which does do lazy evaluation using the `format` formatting. I'm all for f strings in most locations, and I don't know how much that lazy evaluation saves us, but might as well leave them in that format for lazy evaluation.

>     3. I get many warnings for `logger.verbose()` calls, I see that loguru.logger does not have that attribute. I assume it's correct (since it would be crashing everywhere otherwise), but I don't see how it's working. Please point me to the file that enables this behaviour, so I can take a look (curiosity).

Loguru is a bit weird with how you add levels, and since we don't (can't) subclass the logger class a lot of tools give warnings on this. 😕 [Here](https://github.com/Flexget/Flexget/blob/afbaa6f3df641336160c69a6dff2e19dce0cc6e5/flexget/log.py#L115-L117) is where it's defined. Here are the [loguru docs](https://loguru.readthedocs.io/en/stable/resources/recipes.html?highlight=custom%20level#using-logging-function-based-on-custom-added-levels) on the matter. If you can figure out a nicer way that PyCharm/mypy won't complain that would be great.

Thanks for all your work on this!

flexget/entry.py
vgerak: At first I made the parameters' annotation follow the `Optional[<type>] = None` approach, but then I noticed that you were not using the [`no_implicit_optional`](https://mypy.readthedocs.io/en/stable/config_file.html#confval-no_implicit_optional) config option (it's `False` by default). When it's not enabled the following 2 annotations for arguments are equivalent:
1. `var: Optional[<type>] = None`
2. `var: <type> = None`

The latter seems more readable, so while converting my initial annotations to the 2nd form I changed some more (like the one above) for consistency. `message` was annotated as `Optional[str]`, as it's not initialised with `None`, but can receive `None` values (e.g. `accept()`, on line 170)

gazpachoking: Aha. I thought there must be some reason. I haven't used mypy much, and didn't know about this. In that case I'm good with it.

vgerak:
> We are using loguru, which does do lazy evaluation using the format formatting. I'm all for f strings in most locations, and I don't know how much that lazy evaluation saves us, but might as well leave them in that format for lazy evaluation.

I couldn't find anything in the [loguru docs](https://loguru.readthedocs.io/en/stable/search.html?q=lazy&check_keywords=yes&area=default) regarding this. I know that stdlib logging is "lazy" (and safer) when using the "%s" formatting, as indicated by the relevant [pylint warnings](http://pylint.pycqa.org/en/latest/technical_reference/features.html#logging-checker-messages) when trying to use `.format()` or f-strings.

I couldn't find anything indicating that loguru treats `.format()` lazily, only when [marking a single entry](https://loguru.readthedocs.io/en/stable/overview.html?highlight=lazy#lazy-evaluation-of-expensive-functions) that should be lazily evaluated if sink is at the appropriate level. I'm not against following the current implementation, just wondering if it is indeed lazy.

gazpachoking:
> 
> 
> > We are using loguru, which does do lazy evaluation using the format formatting. I'm all for f strings in most locations, and I don't know how much that lazy evaluation saves us, but might as well leave them in that format for lazy evaluation.
> 
> I couldn't find anything in the [loguru docs](https://loguru.readthedocs.io/en/stable/search.html?q=lazy&check_keywords=yes&area=default) regarding this. I know that stdlib logging is "lazy" (and safer) when using the "%s" formatting, as indicated by the relevant [pylint warnings](http://pylint.pycqa.org/en/latest/technical_reference/features.html#logging-checker-messages) when trying to use `.format()` or f-strings.
> 
> I couldn't find anything indicating that loguru treats `.format()` lazily, only when [marking a single entry](https://loguru.readthedocs.io/en/stable/overview.html?highlight=lazy#lazy-evaluation-of-expensive-functions) that should be lazily evaluated if sink is at the appropriate level. I'm not against following the current implementation, just wondering if it is indeed lazy.

Hmm. I thought I had read this, but now I'm not so sure

gazpachoking:
Thinking about the lazy log stuff further:
1. We are always logging at debug level in the background, so that we can create the crash logs when needed. This means only trace level stuff would avoid the evaluation.
2. I don't know how much time it actually saves anyway.
3. As you pointed out, loguru doesn't even advertise the `format` call is lazy. (it might be, I haven't dug into the code to find out)

Given that, I'm thinking fstrings are fine. I don't think we should necessarily go replace every logging call right away, but I don't think there is a problem with it. If we actually have expensive logging stuff we want to avoid always calling, maybe we put it on the `trace` level, with the explicit loguru way of making it lazy. @liiight @paranoidi Any thoughts?

PR #2754, opened 2020-10-17
[change] API - Add type annotations to api module

vgerak:
### Motivation for changes:
Continuation of #2751, `api` module

### Detailed changes:
- Added type annotations to api module
- Changed "rel" links generation in `pagination_headers()` (equivalent results, obviously). To me, it's more readable compared to the template parameter replacement method, but, as it's just my opinion, I can revert it if you wish
- [`TypedDict`](https://docs.python.org/3/library/typing.html#typing.TypedDict) usage is wrapped inside `if TYPE_CHECKING` blocks, as it's available on versions >= 3.8.
- Adding `Session` annotation for API methods introduced "new" mypy warnings like the following (previously ignored, as inferred type was `Any` cause of missing annotations):
```Item "None" of "Optional[Session]" has no attribute "query"```
Should all go away when typing decorators that mess with the arguments is implemented. Relevant issues: python/typing/issues/193 , python/mypy#3157
- Changed some `try: ... except Exception: pass` to use [`contextlib.suppress`](https://docs.python.org/3/library/contextlib.html#contextlib.suppress) instead

```
============== 1317 passed, 28 skipped, 4 xfailed, 5 xpassed, 59 warnings in 188.89s (0:03:08) ===============
```

### DIFF
--- a/flexget/api/app.py
+++ b/flexget/api/app.py
@@ -2,15 +2,20 @@ import json
 import os
 import re
 from collections import deque
+from contextlib import suppress
 from functools import partial, wraps
+from typing import Callable, List, Dict, Union, Mapping, Tuple, TYPE_CHECKING
 
-from flask import Flask, jsonify, make_response, request
+from jsonschema.exceptions import ValidationError as SchemaValidationError
+from flask import Flask, Response, jsonify, make_response, request, Request
 from flask_compress import Compress
 from flask_cors import CORS
 from flask_restx import Api as RestxAPI
-from flask_restx import Resource
+from flask_restx import Resource, Model
+from flask_restx.reqparse import RequestParser
 from jsonschema import RefResolutionError
 from loguru import logger
+from sqlalchemy.orm import Session
 from werkzeug.http import generate_etag
 
 from flexget import manager
@@ -24,6 +29,18 @@ __version__ = '1.8.0'
 
 logger = logger.bind(name='api')
 
+if TYPE_CHECKING:
+    from typing import TypedDict
+    _TypeDict = TypedDict('_TypeDict', {'type': str})
+
+    class MessageDict(_TypeDict):
+        properties: Dict[str, _TypeDict]
+        required: List[str]
+
+    PaginationHeaders = TypedDict(
+        'PaginationHeaders', {'Link': str, 'Total-Count': int, 'Count': int}
+    )
+
 
 class APIClient:
     """
@@ -32,13 +49,13 @@ class APIClient:
     It skips http, and is only usable from within the running flexget process.
     """
 
-    def __init__(self):
+    def __init__(self) -> None:
         self.app = api_app.test_client()
 
-    def __getattr__(self, item):
+    def __getattr__(self, item: str) -> 'APIEndpoint':
         return APIEndpoint('/api/' + item, self.get_endpoint)
 
-    def get_endpoint(self, url, data=None, method=None):
+    def get_endpoint(self, url: str, data=None, method: str = None):
         if method is None:
             method = 'POST' if data is not None else 'GET'
         auth_header = dict(Authorization='Token %s' % api_key())
@@ -53,7 +70,7 @@ class APIClient:
 
 
 class APIEndpoint:
-    def __init__(self, endpoint, caller):
+    def __init__(self, endpoint: str, caller: Callable) -> None:
         self.endpoint = endpoint
         self.caller = caller
 
@@ -62,11 +79,11 @@ class APIEndpoint:
 
     __getitem__ = __getattr__
 
-    def __call__(self, data=None, method=None):
+    def __call__(self, data=None, method: str = None):
         return self.caller(self.endpoint, data=data, method=method)
 
 
-def api_version(f):
+def api_version(f: Callable[..., Response]):
     """ Add the 'API-Version' header to all responses """
 
     @wraps(f)
@@ -83,7 +100,7 @@ class APIResource(Resource):
 
     method_decorators = [with_session, api_version]
 
-    def __init__(self, api, *args, **kwargs):
+    def __init__(self, api: RestxAPI, *args, **kwargs) -> None:
         self.manager = manager.manager
         super().__init__(api, *args, **kwargs)
 
@@ -95,7 +112,12 @@ class API(RestxAPI):
       - methods to auto document and handle :class:`ApiError` responses
     """
 
-    def validate(self, model, schema_override=None, description=None):
+    def validate(
+        self,
+        model: Model,
+        schema_override: Dict[str, List[Dict[str, str]]] = None,
+        description=None
+    ):
         """
         When a method is decorated with this, json data submitted to the endpoint will be validated with the given
         `model`. This also auto-documents the expected model, as well as the possible :class:`ValidationError` response.
@@ -121,13 +143,13 @@ class API(RestxAPI):
 
         return decorator
 
-    def response(self, code_or_apierror, description='Success', model=None, **kwargs):
+    def response(self, code_or_apierror, description: str = 'Success', model=None, **kwargs):
         """
         Extends :meth:`flask_restx.Api.response` to allow passing an :class:`ApiError` class instead of
         response code. If an `ApiError` is used, the response code, and expected response model, is automatically
         documented.
         """
-        try:
+        with suppress(TypeError):  # If first argument isn't a class this happens
             if issubclass(code_or_apierror, APIError):
                 description = code_or_apierror.description or description
                 return self.doc(
@@ -139,12 +161,15 @@ class API(RestxAPI):
                     },
                     **kwargs,
                 )
-        except TypeError:
-            # If first argument isn't a class this happens
-            pass
         return self.doc(responses={code_or_apierror: (description, model)}, **kwargs)
 
-    def pagination_parser(self, parser=None, sort_choices=None, default=None, add_sort=None):
+    def pagination_parser(
+        self,
+        parser: RequestParser = None,
+        sort_choices: List[str] = None,
+        default: str = None,
+        add_sort: bool = None
+    ) -> RequestParser:
         """
         Return a standardized pagination parser, to be used for any endpoint that has pagination.
 
@@ -191,7 +216,7 @@ api = API(
     format_checker=format_checker,
 )
 
-base_message = {
+base_message: "MessageDict" = {
     'type': 'object',
     'properties': {
         'status_code': {'type': 'integer'},
@@ -210,7 +235,7 @@ class APIError(Exception):
     status = 'Error'
     response_model = base_message_schema
 
-    def __init__(self, message=None, payload=None):
+    def __init__(self, message: str = None, payload=None) -> None:
         self.message = message
         self.payload = payload
 
@@ -300,14 +325,18 @@ class ValidationError(APIError):
         'parent',
     )
 
-    def __init__(self, validation_errors, message='validation error'):
+    def __init__(
+        self,
+        validation_errors: List[SchemaValidationError],
+        message: str = 'validation error'
+    ) -> None:
         payload = {
             'validation_errors': [self._verror_to_dict(error) for error in validation_errors]
         }
         super().__init__(message, payload=payload)
 
-    def _verror_to_dict(self, error):
-        error_dict = {}
+    def _verror_to_dict(self, error: SchemaValidationError) -> Mapping[str, Union[str, list]]:
+        error_dict: Dict[str, Union[str, list]] = {}
         for attr in self.verror_attrs:
             if isinstance(getattr(error, attr), deque):
                 error_dict[attr] = list(getattr(error, attr))
@@ -319,7 +348,7 @@ class ValidationError(APIError):
 empty_response = api.schema_model('empty', {'type': 'object'})
 
 
-def success_response(message, status_code=200, status='success'):
+def success_response(message: str, status_code: int = 200, status: str = 'success') -> Response:
     rsp_dict = {'message': message, 'status_code': status_code, 'status': status}
     rsp = jsonify(rsp_dict)
     rsp.status_code = status_code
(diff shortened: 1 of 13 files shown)

### CODE REVIEW

File: flexget/api/app.py [hunk 1]
```diff
@@ -24,6 +29,18 @@
 
 logger = logger.bind(name='api')
 
+if TYPE_CHECKING:
+    from typing import TypedDict
```
liiight, line 33: wont this crash before py3.8?

flexget/api/app.py
```diff
@@ -95,7 +112,12 @@ class API(RestxAPI):
       - methods to auto document and handle :class:`ApiError` responses
     """
 
-    def validate(self, model, schema_override=None, description=None):
+    def validate(
+        self,
+        model: Model,
+        schema_override: Dict[str, List[Dict[str, str]]] = None,
+        description=None
+    ):
```
liiight, line 120: should this return `None`?

flexget/api/app.py
```diff
@@ -121,13 +143,13 @@ def wrapper(*args, **kwargs):
...
 
-    def response(self, code_or_apierror, description='Success', model=None, **kwargs):
+    def response(self, code_or_apierror, description: str = 'Success', model=None, **kwargs):
         """
         Extends :meth:`flask_restx.Api.response` to allow passing an :class:`ApiError` class instead of
         response code. If an `ApiError` is used, the response code, and expected response model, is automatically
         documented.
         """
-        try:
+        with suppress(TypeError):  # If first argument isn't a class this happens
```
liiight, line 152: 👍

flexget/api/app.py
```diff
@@ -199,7 +224,7 @@ def pagination_parser(self, parser=None, sort_choices=None, default=None, add_so
         'status': {'type': 'string'},
     },
     'required': ['status_code', 'message', 'status'],
-}
+}  # type: MessageDict
```
liiight, line 227: was this comment left on purpose?

flexget/api/core/authentication.py
```diff
@@ -18,42 +21,37 @@
 
 @login_manager.request_loader
 @with_session
-def load_user_from_request(request, session=None):
+def load_user_from_request(request: Request, session: Session = None) -> Optional[User]:
     auth_value = request.headers.get('Authorization')
 
     if not auth_value:
-        return
+        return None
```
liiight, line 28: redundant usage of `None`

flexget/api/core/authentication.py
```diff
@@ -18,42 +21,37 @@
...
+            credentials = base64.b64decode(auth_value.replace('Basic ', '', 1)).decode()
             username, password = credentials.split(':')
             user = session.query(User).filter(User.name == username).first()
             if user and user.password and check_password_hash(user.password, password):
                 return user
-            else:
-                return None
-        except (TypeError, ValueError):
-            pass
+    return None
```
liiight, line 44: redundant return

flexget/api/core/authentication.py
```diff
@@ -18,42 +21,37 @@
...
 
 
 @login_manager.user_loader
 @with_session
-def load_user(username, session=None):
+def load_user(username: str, session: Session = None) -> Optional[User]:
```
liiight, line 49: the `User` isn't optional here iirc, if there are no matches, calling `.first()` will raise an exception

flexget/api/core/authentication.py
```diff
@@ -123,7 +122,7 @@ def post(self, session=None):
 @auth_api.route('/logout/')
 class LogoutAPI(APIResource):
     @api.response(200, 'Logout successful', model=base_message_schema)
-    def post(self, session=None):
+    def post(self, session: None = None) -> Response:
```
liiight, line 125: type shouldn't be `None`, it's `Session`

flexget/api/core/cached.py
```diff
@@ -22,15 +23,15 @@ class CachedResource(APIResource):
     @api.response(BadRequest)
     @api.response(APIError)
     @api.doc(parser=cached_parser)
-    def get(self, session=None):
+    def get(self, session: None = None) -> Response:
```
liiight, line 26: type shouldn't be None, it's Session

flexget/api/core/database.py
```diff
@@ -44,19 +46,19 @@ def post(self, session=None):
...
-                raise BadRequest('The plugin {} has no stored schema to reset'.format(plugin_name))
+                raise BadRequest(f'The plugin {plugin_name} has no stored schema to reset')
         return success_response(msg)
 
 
 @db_api.route('/plugins/')
 class DBCleanup(APIResource):
     @api.response(200, model=plugins_schema)
-    def get(self, session=None):
+    def get(self, session: None = None) -> Response:
```
liiight, line 62: type shouldn't be None, it's Session

flexget/api/core/plugins.py
```diff
@@ -81,7 +97,7 @@ class PluginsAPI(APIResource):
     @api.response(BadRequest)
     @api.response(NotFoundError)
     @api.doc(parser=plugins_parser)
-    def get(self, session=None):
+    def get(self, session: None = None) -> Response:
```
liiight, line 100: type shouldn't be None, it's Session. last one i'll mention :)

flexget/api/core/server.py
```diff
@@ -130,19 +133,15 @@ class ServerReloadAPI(APIResource):
...
                 self.manager.load_config(output_to_console=False)
             except YAMLError as e:
-                if (
-                    hasattr(e, 'problem')
-                    and hasattr(e, 'context_mark')
-                    and hasattr(e, 'problem_mark')
-                ):
-                    error = {}
+                if isinstance(e, MarkedYAMLError):
+                    error: Dict[str, int] = {}
```
liiight, line 144: seems kind of overkill to me to specify such an internal variable type, but ok

flexget/api/core/server.py
```diff
@@ -8,6 +8,8 @@
 import traceback
 from pathlib import Path
 from time import sleep
+from typing import Dict, IO, Generator
+from typing import Optional as OptionalType
```
liiight, line 12: why alias it?

flexget/manager.py
```diff
@@ -921,20 +922,20 @@ def acquire_lock(self, event: bool = True) -> Iterator:
...
-                    f.write('%s: %s\n' % (key, ipc_info[key]))
+                    f.write(f'{key}: {ipc_info[key]}\n')
 
     def release_lock(self) -> None:
         try:
             os.remove(self.lockfile)
         except OSError as e:
             if e.errno != errno.ENOENT:
                 raise
-            logger.debug('Lockfile {} not found', self.lockfile)
```
liiight, line 935: logger should actually retain the current syntax and not use fstrings. This is a loguru (and stdlib logger as well) feature which allows the logger to decide whether or not to do the string interpolation, based on the configured log level output of the handler.

liiight (changes requested):
Holy crap dude, you put in a TON of work on this, thank you!
I wonder if we should add type testing to the CI/pre-commit. 
@gazpachoking @paranoidi WDYT?

flexget/api/app.py
vgerak: That's why this and all other usages of TypedDict are in `if TYPE_CHECKING` blocks.
These parts will only be considered by mypy and never run by the python interpreter, therefore not crashing.

flexget/api/app.py
vgerak: It returns a decorator, not None.

Type annotations on decorators are still tricky (they're working on it iirc), and I only added annotations on things I was fairly confident I was correct, as to minimise the mental effort needed from the reviewers, therefore this was left (intentionally) untyped.

flexget/api/app.py
vgerak: `MessageDict` is a defined within the `if TYPE_CHECKING` block, and therefore not "available" during execution (as that part would not have been run). Therefore, when using annotations based on TypeDicts, I'm using this syntax instead of `base_message: MessageDict = { ... }`, to avoid errors in python <3.8

flexget/api/core/authentication.py
vgerak: If a function returns a value at one path, mypy expects:
- return values to be specified on all other returns and
- all possible paths to return a value

Simplified example:

```python
from typing import Optional

def foo(n: int) -> Optional[int]:
    if n > 10:
        return n
    elif n < 0:
        return
```
```
main.py:3: error: Missing return statement
main.py:7: error: Return value expected
Found 2 errors in 1 file (checked 1 source file)
```

flexget/api/core/authentication.py
vgerak: expected by mypy, please refer to reply above

flexget/api/app.py
gazpachoking: Couldn't it be `base_message: "MessageDict" = { ... }`? That seems nicer to me than the comment version.

flexget/api/core/authentication.py
vgerak: According to the docs, [`Query.first()`](https://docs.sqlalchemy.org/en/13/orm/query.html#sqlalchemy.orm.query.Query.first) behaves as annotated:

>Return the first result of this Query or None if the result doesn’t contain any row.

Commonly confused with [`Query.one()`](https://docs.sqlalchemy.org/en/13/orm/query.html#sqlalchemy.orm.query.Query.one), which does raise an exception:
> Return exactly one result or raise an exception.

I don't know whether `one()` should really be used here, my goal was to annotate with as minimal functional changes as possible (hopefully none)

flexget/api/core/authentication.py
vgerak: `session` gets injected by the `@with_session` decorator, which is not applied to this method (and all others annotated as `None`). Therefore it's really `None` in this case.

Even if I'd managed to correctly annotate the `with_session()` method, these instances would have to be `Optional[Session]` due to the `None` initialisation

flexget/api/core/cached.py
vgerak: `session` gets injected by the `@with_session` decorator, which is not applied to this method.

flexget/api/core/database.py
vgerak: `session` gets injected by the `@with_session` decorator, which is not applied to this method.

flexget/api/core/server.py
vgerak: `Optional` is imported from pyparsing on [L.24](https://github.com/Flexget/Flexget/blob/d18633038cb9d75661305b4146c0aaee6d92edde/flexget/api/core/server.py#L24)

flexget/api/core/server.py
vgerak: Dict (and other collections) annotations are sometimes required, when mypy cannot infer the type of the dict based on its usage. This is such a hint, to stop it from complaining,

flexget/api/app.py
vgerak: @gazpachoking as per my comment above, using that annotation type for this instance would crash, as it's defined within the `TYPE_CHECKING` block (which is necessary because I'm using TypedDict and that only exists in Python 3.8+). 

Minified example:
```python
❯ bat foo.py
───────┬──────────────────────────────────────────────
       │ File: foo.py
───────┼──────────────────────────────────────────────
   1   │ from typing import TYPE_CHECKING
   2   │
   3   │ if TYPE_CHECKING:
   4   │     from typing import TypedDict
   5   │     TD = TypedDict('TD', {'foo': int})
   6   │
   7   │ d1 = {'foo': 3}  # type: TD
   8   │
   9   │ d2: TD = {'foo': 4}
───────┴──────────────────────────────────────────────
❯ mypy foo.py
Success: no issues found in 1 source file
❯ python foo.py
  File "foo.py", line 9, in <module>
    d2: TD = {'foo': 4}
NameError: name 'TD' is not defined
```

flexget/manager.py
vgerak: There was a short discussion on [another PR](https://github.com/Flexget/Flexget/pull/2751#issuecomment-710082594) I made regarding the lazy evaluation of loguru and using f-strings in logs. The conclusion of this discussion was this (by @gazpachoking) :
> Given that, I'm thinking fstrings are fine. I don't think we should necessarily go replace every logging call right away, but I don't think there is a problem with it. 

I don't mind reverting it if that decision no longer stands

flexget/api/app.py
gazpachoking: Notice the quotes in my version. That makes the type annotation a string, not a runtime reference to the type.

vgerak: Sorry, I missed that. Yeah, we could go with that version since you prefer it, updated in branch

flexget/api/app.py
liiight: Ah cool, didn't know that.

flexget/api/core/authentication.py
liiight: huh. makes sense typing wise I guess, but this kinda conflict with the pythonic approach. Not sure how I feel about this one. I guess if we want to run mypy then there's no other choice, right?

flexget/api/core/database.py
liiight: `with_session` is implicitly called for all `APIResource` method via the `method_decorators` attribute: https://github.com/Flexget/Flexget/blob/5e3b8156247887df8fa74fd26205bb6c2055c186/flexget/api/app.py#L84

flexget/manager.py
liiight: Oh shit, you're right: https://loguru.readthedocs.io/en/stable/resources/migration.html#replacing-isenabledfor-method

Honestly, I kinda dislike that. @gazpachoking having the ability to lazy log on explicitly and by sending a callable, kinda sucks honestly...

gazpachoking: @liiight I'm not sure it really matters. We weren't using `isEnabledFor` anywhere anyway. (That I know about.) I'm fairly confident loguru does the string formatting lazily, but it isn't an advertised feature. I also don't think that really matters too much to us, given that string formatting is not super expensive, and we have debug logging enabled all the time anyway (for crash log.)

flexget/api/core/database.py
vgerak: You are correct (just verified by looking into the flask_restx `Resource` implementation). All such annotations have been changed to use `sqlalchemy.orm.Session`, adding the import wherever required.

vgerak:
branch rebased to latest develop, requested changes performed

liiight (approved):

liiight:
Absolutely amazing work on this @vgerak !!!

mrahtz:
@vgerak I noticed you referenced [python/typing#193](https://github.com/python/typing/issues/193) on variadic generics in this thread. Heads up that we've been working on a draft of a PEP for this in [PEP 646](https://www.python.org/dev/peps/pep-0646/). If this is something you still care about, take a read and let us know any feedback in [this thread](https://mail.python.org/archives/list/typing-sig@python.org/thread/WFV5K2LK3LFQSO63X2KUOCK3VVLAQ374/) in typing-sig. Thanks!

PR #2757, opened 2020-10-18
Add type annotations to utils module

vgerak:
### Motivation for changes:
Continuation of #2751, utils module

### Detailed changes:
- Added type annotations to api module
- Changed some docstrings to better conform to [PEP257](https://www.python.org/dev/peps/pep-0257/#multi-line-docstrings)
- Renamed some variables that shadowed built-ins (checked usages to make sure nothing broke)
- `bitorrent.py`: Changed `decode_item()` to use the iterator from `tokenize()` rather than a partial `next()` callable (more readable IMHO)
 - `requests.py`: Removed `**kwargs` from call of super (`requests.Session`), as it expects nothing
- `template.py`: Changed `locale.format()` to `locale.format_str`, as the format is deprecated
- `tools.py`: Refactored `convert_bytes` implementation to a more concise and readable one (IMHO)
- `simple_persistence.py`: Skipped, as deprecated

Note: in `tools.py`, `ReList` class, the `flags` kwarg is stored, but never used. It's okay since no usages of the class provide it, but perhaps the following line should be changed (if flags is a single value, or a similar approach if we expect a list)
```diff
- item = re.compile(item, re.IGNORECASE | re.UNICODE)
+ item = re.compile(item, re.IGNORECASE | re.UNICODE | self.flags)
```

```
============== 1317 passed, 28 skipped, 4 xfailed, 5 xpassed, 59 warnings in 186.14s (0:03:06) ===============

❯ dmypy run -- --ignore-missing-imports --follow-imports=skip flexget/utils | tail -1
# before
Found 52 errors in 9 files (checked 22 source files)
# after
Found 7 errors in 4 files (checked 22 source files)
```

### DIFF
--- a/flexget/utils/serialization.py
+++ b/flexget/utils/serialization.py
@@ -9,8 +9,8 @@ ISO8601_FMT = '%Y-%m-%dT%H:%M:%SZ'
 
 
 def serialize(value: Any) -> Any:
-    """
-    Convert an object to JSON serializable format.
+    """Convert an object to JSON serializable format.
+
     :param value: Object to serialize.
     :return: JSON serializable representation of this object.
     """
@@ -31,8 +31,8 @@ def serialize(value: Any) -> Any:
 
 
 def deserialize(value: Any) -> Any:
-    """
-    Restore an object stored with this serialization system to its original format.
+    """Restore an object stored with this serialization system to its original format.
+
     :param value: Serialized representation of the object.
     :return: Deserialized object.
     """
@@ -48,9 +48,7 @@ def deserialize(value: Any) -> Any:
 
 
 def dumps(value: Any) -> str:
-    """
-    Dump an object to text using the serialization system.
-    """
+    """Dump an object to text using the serialization system."""
     serialized = serialize(value)
     try:
         return json.dumps(serialized)
@@ -59,9 +57,7 @@ def dumps(value: Any) -> str:
 
 
 def loads(value: str) -> Any:
-    """
-    Restore an object from JSON text created by `dumps`
-    """
+    """Restore an object from JSON text created by `dumps`"""
     return deserialize(json.loads(value))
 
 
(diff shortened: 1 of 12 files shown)


flexget/utils/serialization.py
```diff
@@ -140,11 +135,11 @@ def serializer_handles(cls, value):
         return isinstance(value, set)
 
     @classmethod
-    def serialize(cls, value: set):
+    def serialize(cls, value: Set[T]) -> List[T]:
```
gazpachoking, line 138: Not sure it matters, but I don't think this is strictly true. This function will take a set of T, but then return a list of the serialized form of T. Probably easier to just loosen these definitions and skip the `T`

vgerak: You are correct - generics were removed from annotations.

gazpachoking:
> Note: in `tools.py`, `ReList` class, the `flags` kwarg is stored, but never used. It's okay since no usages of the class provide it, but perhaps the following line should be changed (if flags is a single value, or a similar approach if we expect a list)
> 
> ```diff
> - item = re.compile(item, re.IGNORECASE | re.UNICODE)
> + item = re.compile(item, re.IGNORECASE | re.UNICODE | self.flags)
> ```

Yep, that was probably the intention. Clearly isn't being used now, but I think your fix is the correct one if you want to slip that in.
I like all your other refactorings as well, we should be good to go after this. Thanks!

vgerak:
>Yep, that was probably the intention. Clearly isn't being used now, but I think your fix is the correct one if you want to slip that in.

Do we want to always apply the `re.IGNORECASE | re.UNICODE` flags, or just keep them as default (like now) and if provided during instantiation use the provided ones, therefore changing the line to the below?
```diff
- item = re.compile(item, re.IGNORECASE | re.UNICODE)
+ item = re.compile(item, self.flags)
```

gazpachoking:
> Do we want to always apply the `re.IGNORECASE | re.UNICODE` flags and union with self flags (as per my original suggestion), or just keep them as default (as it is now) and if a new value is provided during instantiation use it, therefore changing the line to the below?

The UNICODE flag can come out entirely, as it's deprecated and default for python 3. Making ignorecase default and overridable sounds good though.

vgerak:
[Pushed changes](https://github.com/Flexget/Flexget/compare/6e1638f6a777e8d1bc79a73d7e03e5f40b393b17..6f6787d7bee2260ef19b19734e602fd76a47209f) - also changed an import to fix the below:
```
DeprecationWarning: Using or importing the ABCs from 'collections' instead of from 'collections.abc' is deprecated since Python 3.3, and in 3.9 it will stop working
```

gazpachoking:
Excellent. Looks good to me, thanks!

PR #2758, opened 2020-10-19
Handle both old and new case for database_uri

gaqzi:
I found this case when installing the latest version of flexget. 
It defaults to an empty string instead of, I presume, the previous default of `None`.

I'm not sure if it's worthwhile keeping the `None` as a case but figured it was the 
"safer" option just in case. Also feels like `if not self.database_uri` should be just enough as well?

### Motivation for changes:

Failing to start flexget from a fresh install from the latest version on Pypi

### DIFF
--- a/flexget/manager.py
+++ b/flexget/manager.py
@@ -792,7 +792,7 @@ class Manager:
             logger.critical('Failed to check SQLAlchemy version, you may need to upgrade it')
 
         # SQLAlchemy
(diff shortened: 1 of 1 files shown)


gazpachoking:
I feel like `if not self.database_uri` is probably nicer.

gazpachoking:
Going to pull this as is, to make sure we make release tomorrow. Thanks!

PR #2767, opened 2020-10-26
Update dockerfile to Python 3.9

X-dark:
### Motivation for changes:

There is a line with Python version in Dockerfile for deploying the web UI. `python:3-alpine` is now Python 3.9, this updates it accordingly.


gazpachoking:
Thanks!

Issue #2769, opened 2020-11-03
IPTorrents search page change causes discovery failure


singron:
The html changed on the search results page. In particular, some of the css classes that were used to locate metadata were removed. I have a change that fixes it. Below is the boilerplate for filing issues.

### Expected behaviour:
Discovery works if you use iptorrents.

### Actual behaviour:
Discovery fails if you use iptorrents. See the snippet from the logs/crash_report.

### Steps to reproduce:
- Step 1: ...

#### Config:
I have a long autogenerated config, and if you don't have an IPT account, then it's probably not useful for you anyway since you need to fill in the tokens. This might be enough? Idk I didn't test it. I'm just going to send a PR after I file this.
```json
{
  "schedules": false,
  "tasks": {
    "search_series": {
      "discover": {
        "from": [
          {
            "iptorrents": {}
          }
        ],
        "interval": "1 hour",
        "what": [
          {
            "next_series_episodes": {
              "from_start": true
            }
          }
        ]
      }
    }
  }
}

```
  
#### Log:
<details>
 <summary>(click to expand)</summary>
```
2020-11-02 23:59:15 CRITICAL task          search_series   BUG: Unhandled error in plugin discover: 'NoneType' object has no attribute 'text'
```

```
  File "/nix/store/hmy4nb9f0hzbzfcbq67pahzjv2hjv6bm-FlexGet-3.1.71/lib/python3.8/site-packages/flexget/components/sites/sites/iptorrents.py", line 169, in search
    entry['title'] = torrent.find('a', href=re.compile('details')).text
    │                │       │              │  └ <function compile at 0x7fe7a4376670>
    │                │       │              └ <module 're' from '/nix/store/346skv0d24rqnf4npknbp9h5bs14j8zy-python3-3.8.6/lib/python3.8/re.py'>
    │                │       └ <function Tag.find at 0x7fe7a203f700>
    │                └ <tr><td class="i p73"><a href="?4"><img alt="Xvid" src="https://cloudstatic.net/ipt/i/i2/TV-XviD.png"/></a><td class="al"><a ...
    └ <unprintable Entry object>
           
AttributeError: 'NoneType' object has no attribute 'text'
```
</details>

### Additional information:
If you are really interested, I'll fill in all the fields, but I don't think they are relevant to the issue and I just want to send a PR.
- FlexGet version:
- Python version:
- Installation method:
- Using daemon (yes/no): 
- OS and version:
- Link to crash log:

PR #2770, opened 2020-11-03
[fix] iptorrents: parse new search page redesign

singron:
A lot of the css classes were removed on the IPT site, which the
previous code relied on. Instead, this parses the table headers.

The html.parser doesn't parse tables correctly so I switched it to
html5lib. Specifically, `<td><td><td>` would create a nested
structure instead of siblings.

### Addressed issues:
- Fixes https://github.com/Flexget/Flexget/issues/2769 .

I don't think there is a test for this? The log looks like the normal log when everything is working. You might have to pass `--discover-now`.

### DIFF
--- a/flexget/components/sites/sites/iptorrents.py
+++ b/flexget/components/sites/sites/iptorrents.py
@@ -157,24 +157,35 @@ class UrlRewriteIPTorrents:
             if '/u/' + str(config['uid']) not in req.text:
                 raise plugin.PluginError("Invalid cookies (user not logged in)...")
 
-            soup = get_soup(req.content, parser="html.parser")
+            soup = get_soup(req.content, parser='html5lib')
             torrents = soup.find('table', {'id': 'torrents'})
+            seeders_idx = None
+            leechers_idx = None
+            for idx, header in enumerate(torrents.thead.find('tr').findAll('th', recursive=False)):
+                header_text = header.text
+                if 'seeders' in header_text:
+                    seeders_idx = idx
+                if 'leechers' in header_text:
+                    leechers_idx = idx
+
+
 
             results = torrents.findAll('tr')
-            for torrent in results:
+            for torrent in torrents.tbody.findAll('tr', recursive=False):
                 if torrent.th and 'ac' in torrent.th.get('class'):
                     # Header column
                     continue
-                if torrent.find('td', {'colspan': '99'}):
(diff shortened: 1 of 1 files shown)


ianstalk (approved):
LGTM

gazpachoking:
Thanks!

Viper442:
Hi.  I am impatient and so I wanted to test this hotfix while waiting for the release.  I copied your iptorrents.py and replaced the current one in my directory.  I'm still receiving the same error, but it just points to a different snippet of code now.  Log file contains a lot of personal information, but the last bit of the error log is quoted below.  Hope this helps.

> File "/path/to/flexget/lib/python3.7/site-packages/flexget/components/sites/sites/iptorrents.py", line 174, in search
>     if torrent.th and 'ac' in torrent.th.get('class'):
>        │                      └ <tr><td class="i p73"><a href="?99">TV/x265</a><td class="al">...
>        └ <tr><td class="i p73"><a href="?99">TV/x265</a><td class="al">...
> 
> AttributeError: 'NoneType' object has no attribute 'text'
> Traceback (most recent call last):
>   File "/path/to/flexget/lib/python3.7/site-packages/flexget/task.py", line 547, in __run_plugin
>     result = method(*args, **kwargs)
>   File "/path/to/flexget/lib/python3.7/site-packages/flexget/event.py", line 20, in __call__
>     return self.func(*args, **kwargs)
>   File "/path/to/flexget/lib/python3.7/site-packages/flexget/plugins/input/discover.py", line 294, in on_task_input
>     return self.execute_searches(config, entries, task)
>   File "/path/to/flexget/lib/python3.7/site-packages/flexget/plugins/input/discover.py", line 134, in execute_searches
>     search_results = search.search(task=task, entry=entry, config=plugin_config)
>   File "/path/to/flexget/lib/python3.7/site-packages/flexget/plugin.py", line 126, in wrapped_func
>     return func(*args, **kwargs)
>   File "/path/to/flexget/lib/python3.7/site-packages/flexget/components/sites/sites/iptorrents.py", line 174, in search
>     continue
> AttributeError: 'NoneType' object has no attribute 'text'

ianstalk:
Are you sure it's using the correct plugin? The exception looks to align with the old code. Also I tested this on my machine and it seems to work.

Viper442:
Regardless of what I posted above, the update pushed today fixed it, so please disregard.  Thank you!

PR #2773, opened 2020-11-04
[add] transmission - Added possibility to bypass queue in transmission

spacemanspiff2007:
### Motivation for changes:

Currently it's not possible to bypass the transmission queue

### Detailed changes:

Added option to start torrents immediately

```
transmission action gets additional key 'start now'
```

### DIFF
--- a/flexget/plugins/clients/transmission.py
+++ b/flexget/plugins/clients/transmission.py
@@ -234,6 +234,9 @@ class PluginTransmissionInput(TransmissionBase):
                     logger.opt(exception=True).debug(
                         'error when requesting transmissionrpc attribute {}', attr
                     )
+            # Availability in percent
+            entry['transmission_availability'] = torrent.desiredAvailable / torrent.leftUntilDone if torrent.leftUntilDone else 0
+            
             entry['transmission_trackers'] = [t['announce'] for t in torrent.trackers]
             entry['transmission_seed_ratio_ok'] = seed_ratio_ok
             entry['transmission_idle_limit_ok'] = idle_limit_ok
--- a/flexget/plugins/clients/transmission.py
+++ b/flexget/plugins/clients/transmission.py
@@ -235,7 +235,7 @@ class PluginTransmissionInput(TransmissionBase):
                         'error when requesting transmissionrpc attribute {}', attr
                     )
             # Availability in percent
-            entry['transmission_availability'] = torrent.desiredAvailable / torrent.leftUntilDone if torrent.leftUntilDone else 0
+            entry['transmission_availability'] = (torrent.desiredAvailable / torrent.leftUntilDone) if torrent.leftUntilDone else 0
             
             entry['transmission_trackers'] = [t['announce'] for t in torrent.trackers]
             entry['transmission_seed_ratio_ok'] = seed_ratio_ok
--- a/flexget/plugins/clients/transmission.py
+++ b/flexget/plugins/clients/transmission.py
@@ -291,7 +291,7 @@ class PluginTransmission(TransmissionBase):
                     'password': {'type': 'string'},
                     'action': {
                         'type': 'string',
-                        'enum': ['add', 'remove', 'purge', 'pause', 'resume', 'start bypass_queue'],
+                        'enum': ['add', 'remove', 'purge', 'pause', 'resume', 'bypass_queue'],
                     },
                     'path': {'type': 'string'},
                     'max_up_speed': {'type': 'number'},
@@ -613,6 +613,8 @@ class PluginTransmission(TransmissionBase):
                 if list(options['change'].keys()):
                     self.client.change_torrent(torrent_info.id, 30, **options['change'])
 
+                start_torrent = partial(self.client.start_torrent, [torrent_info.id])
+
                 if config['action'] == 'add':
                     # if add_paused was defined and set to False start the torrent;
                     # prevents downloading data before we set what files we want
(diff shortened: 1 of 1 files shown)


flexget/plugins/clients/transmission.py
```diff
@@ -286,7 +286,7 @@ class PluginTransmission(TransmissionBase):
                     'password': {'type': 'string'},
                     'action': {
                         'type': 'string',
-                        'enum': ['add', 'remove', 'purge', 'pause', 'resume'],
+                        'enum': ['add', 'remove', 'purge', 'pause', 'resume', 'start now'],
```
liiight, line 289: i'd name this `bypass_queue` to keep it consistent with the original terminology

flexget/plugins/clients/transmission.py
```diff
@@ -631,7 +631,10 @@ def on_task_output(self, task, config):
                 elif config['action'] == 'resume':
                     self.client.start_torrent([torrent_info.id])
                     logger.info('resumed {} in transmission', torrent_info.name)
-
+                elif config['action'] == 'start now':
+                    self.client.start_torrent([torrent_info.id], bypass_queue=True)
```
liiight, line 635: you should use `partial` object here to avoid code duplication

liiight (changes requested):
Some syntactical changes, but if someone with transmission could actually test this, that would be great

flexget/plugins/clients/transmission.py
```diff
@@ -286,7 +286,7 @@ class PluginTransmission(TransmissionBase):
                     'password': {'type': 'string'},
                     'action': {
                         'type': 'string',
-                        'enum': ['add', 'remove', 'purge', 'pause', 'resume'],
+                        'enum': ['add', 'remove', 'purge', 'pause', 'resume', 'start bypass_queue'],
```
spacemanspiff2007, line 289: Was that what you meant?

flexget/plugins/clients/transmission.py
spacemanspiff2007: I am unsure what you mean, could you please elaborate?

ianstalk: He means partial application using `functools.partial` (https://docs.python.org/3/library/functools.html)

flexget/plugins/clients/transmission.py
ianstalk: I think more like this:
`'enum': ['add', 'remove', 'purge', 'pause', 'resume', 'bypass_queue']`

flexget/plugins/clients/transmission.py
spacemanspiff2007: Yes, I got that. But I am unaware how it would prevent code duplication since this is just a one liner.
I'll happily change it but I don't know what to.

flexget/plugins/clients/transmission.py
spacemanspiff2007: transmission uses ``Resume Now`` in the context menu, that's why I came up with ``start now``.
On second thought maybe we should use ``resume now`` so it's consistent with the gui?
My apprehension is that with ``bypass_queue`` it's not really clear what the action does

ianstalk:
Honestly, I don't see why this needs to be a standalone action. Doesn't it make more sense to add it as a bool to config, and then pass that value to `start_torrent`? Is the idea to force an already added torrent through?

spacemanspiff2007:
> Is the idea to force an already added torrent through?

Yes - after adding it is in the queue. I want some to start immediately, but not all.
So I have a second job with more logic that just "forces" some.

flexget/plugins/clients/transmission.py
liiight: I think that bypass_queue is clear enough and also the official term that the application uses, so its best to use it

flexget/plugins/clients/transmission.py
liiight: removed some stuff for brevity:
```python
start_torrent = partial(self.client.start_torrent, [torrent_info.id])
...
elif config['action'] == 'resume':
    start_torrent()
elif config['action'] == 'start now':
    start_torrent(bypass_queue=True)
```

BrutuZ: Wouldn't it be more intuitive to have `bypass_queue` as an optional parameter and pass it with resume?
```diff
                 elif config['action'] == 'resume':
-                    self.client.start_torrent([torrent_info.id])
+                   self.client.start_torrent([torrent_info.id],  bypass_queue=config.get('bypass_queue', False))
                     logger.info('resumed {} in transmission', torrent_info.name)
-                elif config['action'] == 'start now':
-                    self.client.start_torrent([torrent_info.id], bypass_queue=True)
-                    logger.info('started {} in transmission', torrent_info.name)
```

If adding yet another parameter sounds bad, could possibly use the already existing `queue_position` to act as bypass when set to `-1`

flexget/plugins/clients/transmission.py
spacemanspiff2007: It's just that only from ``bypass_queue`` it's imho not clear what the action is doing.
Maybe ``resume (bypass_queue)``?

BrutuZ: That's why plugins are documented 👍

liiight (approved):

liiight:
Thanks for this! Care to update the wiki?

kiracus:
a bug is introduced from this PR https://github.com/Flexget/Flexget/issues/2777

spacemanspiff2007:
I am unsure how to update the wiki - sorry.

ksurl:
Just go to the wiki in a browser and use the sign in with github and then edit. It uses the browser. No coding or git local needed

PR #2774, opened 2020-11-04
Pending list filter option

stevezau:
### Motivation for changes:
Add ability to include items which are pending approval to preform tasks such as

1.clean-ups if older then x days.
2. For shows/movies released on netflix the rating/score on imdb can be empty meaning some end up in pending approval. A few days later people have voted, if it is a high voted movie then having the ability to filter on "pending approval" items will allow you to auto approve if the rating changes.

```
          list_get_pending:
            pending_list:
              list_name: "test_list"
              include: "pending"

          list_get_approved:
            pending_list:
              list_name: "test_list"
              include: "approved"

          list_get_all:
            pending_list:
              list_name: "test_list"
              include: "all"

```

### DIFF
--- a/flexget/components/managed_lists/lists/pending_list/pending_list.py
+++ b/flexget/components/managed_lists/lists/pending_list/pending_list.py
@@ -26,9 +26,13 @@ class PendingListSet(MutableSet):
     def __init__(self, config):
         if not isinstance(config, dict):
             config = {'list_name': config}
-        config.setdefault('include_pending', False)
         self.list_name = config.get('list_name')
-        self.only_approved = False if config.get('include_pending') else True
+
+        self.filter_approved = True
+        if config.get('include') == "pending":
+            self.filter_approved = False
+        if config.get('include') == "all":
+            self.filter_approved = None
 
         with Session() as session:
             if not self._db_list(session):
@@ -48,18 +52,18 @@ class PendingListSet(MutableSet):
                 )
             )
         )
-        if approved:
-            query = query.filter(db.PendingListEntry.approved == True)
+
+        if approved is not None:
+            query = query.filter(db.PendingListEntry.approved == approved)
+
         return query.first()
 
     def __iter__(self):
         with Session() as session:
-            for e in (
-                self._db_list(session)
-                .entries.filter(db.PendingListEntry.approved == self.only_approved)
-                .order_by(db.PendingListEntry.added.desc())
-                .all()
-            ):
+            query = self._db_list(session).entries.order_by(db.PendingListEntry.added.desc())
+            if self.filter_approved is not None:
+                query = query.filter(db.PendingListEntry.approved == self.filter_approved)
+            for e in query.all():
                 logger.debug('returning {}', e.entry)
                 yield e.entry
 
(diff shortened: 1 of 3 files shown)


flexget/components/managed_lists/lists/pending_list/pending_list.py
```diff
@@ -19,15 +19,20 @@ class PendingListSet(MutableSet):
...
             .first()
         )
 
     def __init__(self, config):
-        self.config = config
+        if not isinstance(config, dict):
+            config = {'list_name': config}
+        config.setdefault('include_pending', False)
+        self.list_name = config.get('list_name')
+        self.only_approved = False if config.get('include_pending') else True
```
liiight, line 31:
```suggestion
        self.only_approved = bool(config.get('include_pending'))
```

flexget/components/managed_lists/lists/pending_list/pending_list.py
```diff
@@ -102,12 +107,25 @@ def online(self):
 
     def get(self, entry):
         with Session() as session:
-            match = self._entry_query(session=session, entry=entry, approved=True)
+            match = self._entry_query(session=session, entry=entry, approved=self.only_approved)
```
liiight, line 110: i'm not sure that this will actually include those who are pending, rather than switch select between pending and non pending. If you'd like to select just approved or everything, i'd consider adding the `approved` filter only when needed.

Or, better yet, via enum. Filter by `approved`, `not_approved` or both.

liiight (changes requested):
consider using `enum.Enum` here to distinguish between the 3 scenarios for filtering.

stevezau:
@liiight changed to enum.. Is that what you were thinking?

flexget/components/managed_lists/lists/pending_list/pending_list.py
```diff
@@ -19,15 +19,24 @@ class PendingListSet(MutableSet):
...
             .first()
         )
 
     def __init__(self, config):
-        self.config = config
+        if not isinstance(config, dict):
+            config = {'list_name': config}
+        self.list_name = config.get('list_name')
+
+        self.filter_approved = True
```
liiight, line 31: I meant an actual `Enum`, something like this:
```python
from enum import Enum

class PendingListInclude(Enum):
    pending = False
	approved = True
	all = None
```
Then use it like this:
```python
self.include = PendingListFilter(config['include'])
...
query = self._db_list(session).entries.order_by(db.PendingListEntry.added.desc())
if self.include is not PendingListInclude.all:
    query = query.filter(db.PendingListEntry.approved == self.include.value)
```

flexget/tests/test_pending_list.py
```diff
@@ -67,3 +83,24 @@ def test_list_match(self, execute_task):
...
+        task = execute_task('pending_list_add')
+        assert len(task.entries) == 2
+
+        with Session() as session:
+            entry = session.query(PendingListList).first().entries.first()
+            entry.approved = True
+
+        with Session() as session:
+            entries = session.query(PendingListList).first().entries.all()
+            print(entries)
```
liiight, line 97: forgot to remove

flexget/components/managed_lists/lists/pending_list/pending_list.py
stevezau: @liiight I could do that but  it is  only used in two places.. feels nore complicated then it needs to be?

liiight: Feels nicer to me, but ok. Python enums are great

liiight (approved):

PR #2776, opened 2020-11-07
changed transmission_availability

spacemanspiff2007:
Should now show the same as the gui


flexget/plugins/clients/transmission.py
```diff
@@ -234,8 +234,11 @@ def on_task_input(self, task, config):
                     logger.opt(exception=True).debug(
                         'error when requesting transmissionrpc attribute {}', attr
                     )
+
+            bytes_completed = sum(map(lambda x: x['bytesCompleted'] if x['wanted'] else 0, t.fileStats))
```
spacemanspiff2007, line 238: I am unsure about the ```if x['wanted'] else 0``` part

gazpachoking:
Sorry about the lack of feedback on this. If you've tested, and this is consistent with the UI and working for you, I'm okay with it. Mind updating it to resolve the conflicts?

spacemanspiff2007:
I've switched to a stand alone python script to do these calculations but I've updated the PR with what I am currently using.
However this is untested.

PR #2780, opened 2020-11-09
[change] Sonarr - Handle when the title or id is not found in the entry object

stevezau:
Handle when the title or id is not found in the entry object which means a match is not found.. However the lookup of shows in sonarr returns a show that already exists.


stevezau:
@liiight pretty sure non-realted tests are failing. Got a second to take a look at this?

Issue #2727, opened 2020-09-09
Migrate to new transmission-rpc


gazpachoking:
Looks like the module we use for transmission connection (transmissionrpc) is no longer maintained. Seems there is a fork available now which is under active development ([transmission-rpc](https://github.com/Trim21/transmission-rpc).) I think all that needs to be done to switch out is update the import from `transmissionrpc` to `transmission_rpc`, as well as update documents on how to install. (`pip install transmission-rpc` instead of `pip install transmissionrpc`) Would be nice if a transmission user tested this out and sent a pull request if everything is working.

ksurl:
there is also transmissionrpc-ng which seems to be a direct fork?

gazpachoking:
That looks good as well. Not sure what differences have already diverged, or which will stay more supported. Maybe update our docs for transmission-ng right now since that doesn't require any code changes. Since the import seems to be the only change for the other lib, we could also potentially support both, by trying to fall back to the other import if the first one fails. Not sure if that's advisable though, it would make supporting harder, especially as differences between the code bases emerge.

ianstalk:
From a quick glance, `transmission-rpc` looks to be much more active than `transmissionrpc-ng`. The latter hasn't been touched in 6 months, and has a nearly year-old issue for updating to the latest API version.

ksurl:
transmission-rpc is missing the HTTPErrorHandler function. the transmissionrpc-ng module does have it. I forked the main flexget repo and changed the import lines. installed to a new venv and it fails to run a transmission task with error:

```
2020-11-06 18:13:22 CRITICAL transmission  test             transmission_rpc module version 3.2.1 or higher required.
2020-11-06 18:13:22 WARNING  task          test             Aborting task (plugin: transmission)
```

when running with transmissionrpc-ng, no need to replace import lines since the module name stayed the same. however it is getting an error still, failing to import the module

```
2020-11-06 19:13:16 CRITICAL transmission  tms             Transmissionrpc module version 0.11 or higher required.
2020-11-06 19:13:16 WARNING  task          tms             Aborting task (plugin: transmission)
```

when manually importing the module in a python interpretor, it gives an error about pycurl missing so it seems the author added pycurl as a dependency.

[...]

PR #2791, opened 2020-12-25
Replace obsoleted and unmaintaned transmissionrpc

rsekman:
### Motivation for changes:

[`transmissionrpc`](https://pypi.org/project/transmissionrpc/) is unmaintained. Swap to [`transmission-rpc`](https://pypi.org/project/transmission-rpc/) as suggested in #2727.

### Detailed changes:

Drop-in replacement of imports.
The newer `transmission-rpc` does not provide `HTTPHandlerError`. Instead it raises exceptions from `requests.exceptions`. Error handling on connecting to transmission has to be changed accordingly.

### Addressed issues:
-  #2727  .

### Possible improvements

Error handling could be more fine-grained, but `requests.exceptions` are very detailed as they are.

### DIFF
--- a/flexget/plugins/clients/transmission.py
+++ b/flexget/plugins/clients/transmission.py
@@ -19,8 +19,9 @@ from flexget.utils.template import RenderError
 from flexget.utils.tools import parse_timedelta
 
 try:
-    import transmissionrpc
-    from transmissionrpc import HTTPHandlerError, TransmissionError
+    import transmission_rpc as transmissionrpc
+    from transmission_rpc import TransmissionError
+    import requests.exceptions
 except ImportError:
     # If transmissionrpc is not found, errors will be shown later
     pass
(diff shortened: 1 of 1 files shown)


gazpachoking:
Excellent, looks good to me. Thanks!

ksurl:
From_transmission is not working. See error [here](https://github.com/Flexget/Flexget/issues/2794).

mfonville:
The pip dependencies have not been updated to reflect this change. Users miss the `transmission-rpc` library and get an error:
`CRITICAL transmission  xxx  transmission-rpc module version 0.11 or higher required.`

BrutuZ:
`transmission-rpc` is an optional dependency like many others. Flexget warns it's required when you call a plugin that needs it, the rest is up to you

mfonville:
True that the dependency is optional. So I was wrong to mention it as a pip-dependency.
But users get now when upgrading their Flexget no explicit notice that this dependency has changed. And when executing Flexget and receiving that critical error, I have doubts whether they notice the extra dash in `transmission-rpc` and understand that it is different from the package they have installed in the past.

Maybe an explicit error if `transmissionrpc` has been found but no `transmission-rpc` could help them.

ksurl:
I've added it to the [Upgrade Actions page](https://flexget.com/UpgradeActions)

Karlson2k:
@gazpachoking, could you please look into related issue https://github.com/Flexget/Flexget/issues/2794?

PR #2792, opened 2020-12-26
[change] Pyload: Support for Group Links

kepi:
### Motivation for changes:

RSS plugin allows to use `group_links` option to generate one entry with multiple
URLs, but Pyload wasn't using it.

I was fighting with this for months before I checked the source code and have been using pyloads parsing feature instead. It never cross my mind that it wouldn't be implemented at all.

### Detailed changes:
- Simply load `urls` from `entry` when present.

### DIFF
--- a/flexget/plugins/clients/pyload.py
+++ b/flexget/plugins/clients/pyload.py
@@ -193,7 +193,7 @@ class PluginPyLoad:
 
             parsed = result.json()
 
-            urls = []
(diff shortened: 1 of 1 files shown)


ProteinPig:
same situation, and finally use inputs or sequence plugin to handle multi sources.

gazpachoking:
I don't know anything about pyload, so I'm just going to trust you guys that this does the right thing. 😃

PR #2793, opened 2020-12-27
[change] next_series_episode: Opt to act based on downloaded status

kepi:
### Motivation for changes:

There are cases where it is beneficial to have some plugin which emits all episodes of series or downloaded only. And `next_series_episode` does already almost everything needed.

For me, the main use case is to be able to forget downloads for episodes which were in fact not downloaded. I.e. passed to pyload but pyload failed to download them in the end.

But while working on this I realized that there are more use cases to this so I added also `only` value which might get handy.

I'm not sure if option name is ok and clear. Let me know if there is anything I can do to improve.

### Detailed changes:

New option `downloaded` for plugin `next_series_episode` were value is one of:

- no: (default) Same behavior as before. Don't include already downloaded episodes.
- yes: Include all episodes no matter the downloaded state.
- only: Include only already downloaded episodes.

```
    next_series_episodes:
      backfill: yes
      from_start: yes
      downloaded: only
```

### DIFF
--- a/flexget/components/series/next_series_episodes.py
+++ b/flexget/components/series/next_series_episodes.py
@@ -29,6 +29,11 @@ class NextSeriesEpisodes:
                     'from_start': {'type': 'boolean', 'default': False},
                     'backfill': {'type': 'boolean', 'default': False},
                     'only_same_season': {'type': 'boolean', 'default': False},
+                    'downloaded': {
+                        'type': ['boolean', 'string'],
+                        'enum': [True, False, 'only'],
+                        'default': False,
+                    },
                 },
                 'additionalProperties': False,
             },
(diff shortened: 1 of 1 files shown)


kepi:
Is there any chance to have this merged? Can I help somehow?

kepi:
bump

perriernical:
any updates?

gazpachoking:
I was pretty confused about the actual behavior here before. So when 'downloaded' is True, it emits every episode from the current season? When False it's the previous behavior of only undownloaded eps from this season, when 'only' it's all downloaded eps from current season? How does it interact with backfill?

paranoidi:
Very confusing option, "downloaded: yes | no | only"

aren't yes and only same thing?

PR #2795, opened 2020-12-29
[fix] torznab: Fix None value in description

kepi:
Torznab can receive None value in child.text which leads to generating entry
with None value in description.

Unfortunately `entry.get('description', '')` construct which is used almost
everywhere can't return default value when 'description' key exists in entry but
its value is None.

I believe that better option would be to use different method to handle
defaults and don't count on input plugins to provide correct values, but it
would take major effort to fix everywhere.

### DIFF
--- a/flexget/plugins/input/torznab.py
+++ b/flexget/plugins/input/torznab.py
@@ -190,7 +190,7 @@ class Torznab:
                 if child.tag in ['{http://torznab.com/schemas/2015/feed}attr', 'enclosure']:
                     continue
                 else:
-                    if child.tag in ['description', 'title']:
+                    if child.tag in ['description', 'title'] and child.text:
                         entry[child.tag] = child.text
             entries.append(entry)
         return entries


flexget/plugins/input/torznab.py
```diff
@@ -190,7 +190,7 @@ def create_entries_from_query(self, url, task):
                 if child.tag in ['{http://torznab.com/schemas/2015/feed}attr', 'enclosure']:
                     continue
                 else:
-                    if child.tag in ['description', 'title']:
+                    if child.tag in ['description', 'title'] and child.text != None:
```
gazpachoking, line 193:
```suggestion
                    if child.tag in ['description', 'title'] and child.text:
```
I think it's a bit nicer like this. If we really want to check explicitly for `None` it should be `is not None` rather than `!= None`

kepi: Totally agree with `is not None`. I just realized it on my own and pushed before seen your comment :) 

But you are right that in this context, there is no need to check explicitly for `None`. Fixed.

gazpachoking:
Looks good, thanks!

PR #2800, opened 2021-01-05
[fix] Transmission: Make host config option backwards compatible

DarnisA:
### Motivation for changes:

The new `transmission-rpc` library doesn't use the host parameter for the protocol and path as the old `transmissionrpc` library did, these have to be given as separate parameters.

This PR is to make the host config option backwards compatible by parsing it with urlparse and passing them on. I'm also stripping out `rpc` from the path as the new library adds it on as seen [here](https://github.com/Trim21/transmission-rpc/blob/master/transmission_rpc/client.py#L109).

### Detailed changes:
- Parse the host config value
- If protocol/scheme isn't given, use `http` otherwise urlparse doesn't do well
- Strip out `rpc` from the path if given
- Pass protocol and path to the library

### DIFF
--- a/flexget/plugins/clients/transmission.py
+++ b/flexget/plugins/clients/transmission.py
@@ -55,12 +55,24 @@ class TransmissionBase:
 
     def create_rpc_client(self, config):
         user, password = config.get('username'), config.get('password')
+        urlo = urlparse(config['host'])
+
+        if urlo.scheme == '':
+            urlo = urlparse('http://' + config['host'])
+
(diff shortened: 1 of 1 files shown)


ksurl:
not sure if the fail is legit since trakt has been having issues in the past week.

DarnisA:
Yeah, it looks to be a trakt issue. Not related to this PR.

gazpachoking:
Yep. Had a bit of an issue with the online tests which I've since fixed. This looks good to me. Thanks!

PR #2803, opened 2021-01-08
Dockerfile transmission-rpc

X-dark:
### Motivation for changes:

Pull correct transmission dependency as a followup to #2791 

### Detailed changes:
- Replaces transmissionrpc by transmission-rpc on pip commands
- Explicitly specify the pulling registry of the base image while we are at it


gazpachoking:
LGTM, thanks!

PR #2856, opened 2021-02-14
Update deps for 2021(mostly)

jimman2003:
### Motivation for changes:
Update deps for 2021, starting with latest guessit without breaking the tests..
### Detailed changes:
Update guessit
### Direct
 - guessit
### Indirect
 - pytz
 - python-dateutils
 - (and more)

```
Need help with test output.. Too long to paste.. Probably a log level error..
```
### TO-DO:
- [x] Deal with test errors


jimman2003:
Any help is appreciated!!

jimman2003:
Found why more tests were failing.. The transition to 3.3...So I just upgraded to 3.2 guessit..

gazpachoking:
Looks good. Not sure if there is a bug in new guessit, or maybe something just changed that doesn't play nice with how we expect things to work. Can be sorted out later though.
Is the one per line comments in requirements.txt just a result of new pip-compile version? If so, let's update the `pip-tools` requirement in `dev-requirements.txt` so that the next person running it doesn't end up switching it back to the single line format.

jimman2003:
Yeah,I was running the latest version.. while we are at it let's update black as well

liiight (approved):

gazpachoking:
Thanks!

PR #2869, opened 2021-03-01
Tvdb NoneType Name Error

soloam:
### Motivation for changes:
Some errors my occur (I detected this because I had this error) if no name is detected in tvdb. The method _update_search_strings will try to make lower() and this will make the system crash

### Detailed changes:
- Add if series and series.name to vaidate is existes before updating the search strings

```
2021-02-27 03:48:21 ERROR    lazy_lookup   AutoExtract>t_AutoExtract_GetFiles Unhandled error in lazy lookup plugin: 'NoneType' object has no attribute 'lower'
  File "/home/downloader/flexget/lib/python3.7/site-packages/flexget/utils/lazy_dict.py", line 41, in __getitem__
    callee.func(self.store, *(callee.args or []), **(callee.kwargs or {}))
  File "/home/downloader/flexget/lib/python3.7/site-packages/flexget/components/thetvdb/thetvdb_lookup.py", line 122, in lazy_series_lookup
    return self.series_lookup(entry, language, self.series_map)
  File "/home/downloader/flexget/lib/python3.7/site-packages/flexget/utils/database.py", line 30, in wrapper
    return func(*args, **kwargs)
  File "/home/downloader/flexget/lib/python3.7/site-packages/flexget/components/thetvdb/thetvdb_lookup.py", line 111, in series_lookup
    session=session,
  File "/home/downloader/flexget/lib/python3.7/site-packages/flexget/utils/database.py", line 27, in wrapper
    return func(*args, **kwargs)
  File "/home/downloader/flexget/lib/python3.7/site-packages/flexget/components/thetvdb/api_tvdb.py", line 577, in lookup_series
    _update_search_strings(series, session, search=name)
  File "/home/downloader/flexget/lib/python3.7/site-packages/flexget/components/thetvdb/api_tvdb.py", line 520, in _update_search_strings
    add = [series.name.lower()] + aliases + searches
AttributeError: 'NoneType' object has no attribute 'lower'

```

### DIFF
--- a/flexget/components/thetvdb/api_tvdb.py
+++ b/flexget/components/thetvdb/api_tvdb.py
@@ -574,7 +574,8 @@ def lookup_series(name=None, tvdb_id=None, only_cached=False, session=None, lang
             try:
                 updated_series = TVDBSeries(series.id, language)
                 series = session.merge(updated_series)
-                _update_search_strings(series, session, search=name)
+                if series and series.name:
+                    _update_search_strings(series, session, search=name)
             except LookupError as e:
                 logger.warning(
                     'Error while updating from tvdb ({}), using cached data.', e.args[0]
--- a/flexget/components/thetvdb/api_tvdb.py
+++ b/flexget/components/thetvdb/api_tvdb.py
@@ -205,6 +205,11 @@ class TVDBSeries(Base):
         self._banner = series['banner']
         self._genres = [TVDBGenre(id=name) for name in series['genre']] if series['genre'] else []
 
+        if not self.name:
+            raise LookupError(
+                f'Not possible to get name to series with id {self.id} in language \'{self.language}\''
+            )
+
         if self.first_aired is None:
             logger.debug(
                 'Falling back to getting first episode aired date for series {}', self.name


liiight:
I'm not sure this is the correct approach to handle this. We need to figure out why `series.name` was `None` in the first place, that's the root cause.

soloam:
I agree! But since this can happen, doing tolower without curling eventual none, can give errors! I'll try to get the reason why the name is not filled

soloam:
Ok, I found the root cause! The series that I'm Looking for that triggered me this error https://thetvdb.com/series/esperanca does not return a name in "en" only in "pt" when flexget looks for the series in "en" (the default) it does not return a name, then lower will return a error! From what I see this should not crash, it should handle the error!

_update_search_strings will try to add the name in lowercase. We could, or do my check, or do somting like

 
```python
    name = [series.name] if series.name else []
    add = name + aliases + searches
```
in
https://github.com/Flexget/Flexget/blob/bd823d6ad4cd75147b115554247fe70f74232e41/flexget/components/thetvdb/api_tvdb.py#L516-L529

soloam:
Ok @liiight , I found the root cause! The series that I'm Looking for that triggered me this error https://thetvdb.com/series/esperanca does not return a name in "en" only in "pt" when flexget looks for the series in "en" (the default) it does not return a name, then lower will return a error! From what I see this should not crash, it should handle the error!

_update_search_strings will try to add the name in lowercase. We could, or do my check, or do somting like

 
```python
    name = [series.name] if series.name else []
    add = name + aliases + searches
```
in
https://github.com/Flexget/Flexget/blob/bd823d6ad4cd75147b115554247fe70f74232e41/flexget/components/thetvdb/api_tvdb.py#L516-L529

soloam:
Probably the second is more secure

liiight:
If the search doesn't return a series name, that should be the end of the lookup, no need for it to proceed or to protect the other functions IMO.

soloam:
@liiight I'll wait for your input and change if needed

paranoidi:
IMO, this is not the way, we should be able to trust that series parser has a name that is a string. I'd recommend modifying series parser so that it throws an exception if name is tried to be set to anything but string. That way you will found out where it is being set incorrectly instead of trying to plug holes everywhere.

soloam:
@paranoidi This bug is contained in the TVDB Api! The problem is that it retrieves the show and there is no name in it, then tries to create the search strings setting the name to lower case! That makes the crash! The series parser never takes part on this!

I can make the check the second it retrieves the show from TVDB, and there raise the error to stop.

soloam:
He is returning the show from the ID, and to that shows, the name is not filled!

paranoidi:
@soloam ah, I see, sorry for causing confusion :)

soloam:
No problem @paranoidi . I'm still tracking this! I've actually implemented this changes in my flexget and it's working ok!

gazpachoking:
Do you have a tvdb id example that comes back with a `None` series name? I'm thinking we'd raise a LookupError hereabouts https://github.com/Flexget/Flexget/blob/db02ee1e2778fdec5929d4ea26ddcc43b4a245cb/flexget/components/thetvdb/api_tvdb.py#L189 if there was no seriesname, but I'd be curious what sort of result we are getting from tvdb before we actually implement any fix.

soloam:
I had this one that was giving me this error in the first place

https://www.thetvdb.com/series/esperanca

But from what I can see now it haves the name in english!

The main problem, from what I've seen is that the serie didn't had the name in english, and when we tried to create the search strings it would crash because it was doing the lowercase of none!

soloam:
> Do you have a tvdb id example that comes back with a `None` series name? I'm thinking we'd raise a LookupError hereabouts https://github.com/Flexget/Flexget/blob/db02ee1e2778fdec5929d4ea26ddcc43b4a245cb/flexget/components/thetvdb/api_tvdb.py#L189 if there was no seriesname, but I'd be curious what sort of result we are getting from tvdb before we actually implement any fix.

I had this one that was giving me this error in the first place

https://www.thetvdb.com/series/esperanca

But from what I can see now it haves the name in english!

The main problem, from what I've seen is that the serie didn't had the name in english, and when we tried to create the search strings it would crash because it was doing the lowercase of none!

gazpachoking:
Ahh. I guess maybe if we request a language that the series doesn't have a translation for that's when it ends up happening then? I'm guessing we should just consider it a failed lookup in this case, but I'm far from an expert on the best ways to handle translation stuff.

soloam:
> Ahh. I guess maybe if we request a language that the series doesn't have a translation for that's when it ends up happening then? I'm guessing we should just consider it a failed lookup in this case, but I'm far from an expert on the best ways to handle translation stuff.

English is the defaul request language to the API in flexget! This only happens because I already had pulled the thetvdv_id from other plugin, so we went to the API directly with the Id, and when retrieving the name it was empty (None), resulting in the problems described!

I think this should be handled, to avoid crashes like this one, probably in a different way, or this way plus something else! Probably if the name is empty raise the Lookup error.

gazpachoking:
I'm thinking raising a lookuperror in this function if name is None is probably a nicer way.
https://github.com/Flexget/Flexget/blob/db02ee1e2778fdec5929d4ea26ddcc43b4a245cb/flexget/components/thetvdb/api_tvdb.py#L177

soloam:
> I'm thinking raising a lookuperror in this function if name is None is probably a nicer way.
> https://github.com/Flexget/Flexget/blob/db02ee1e2778fdec5929d4ea26ddcc43b4a245cb/flexget/components/thetvdb/api_tvdb.py#L177

I'll look into it

soloam:
Done @gazpachoking

PR #2874, opened 2021-03-04
Formatted plugins with new black

jimman2003:
### Motivation for changes:
Better readability on docstrings
### Detailed changes:
- Just ran black on plugins folder
### Addressed issues:
- Inconsistant docstrings in plugins


paranoidi:
Thanks!

PR #2885, opened 2021-03-22
Accept highest allowed quality entry from timeframe

### LINKED ISSUE #2448: timeframe doesn't work fully on tasks only started by IRC plugin

0x022b:
### Motivation for changes:

If multiple entries are found within timeframe that meet the target criteria, the first entry that satisfied the criteria is chosen instead of the one with the highest quality.

### Detailed changes:

Sort timeframe entries by quality so that the entry with highest allowed target quality is accepted instead of the first seen entry.

### Addressed issues:

- Should also fix #2448

### DIFF
--- a/flexget/components/series/series.py
+++ b/flexget/components/series/series.py
@@ -861,7 +861,7 @@ class FilterSeries(FilterSeriesBase):
 
     def process_timeframe_target(self, config, entries, downloaded=None):
         """
-        Accepts first episode matching the quality configured for the series.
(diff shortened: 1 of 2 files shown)


gazpachoking:
This was the old behavior. we changed it so that people could prefer based on things other than best quality. The old behavior can still be achieved using sort_by. https://flexget.com/UpgradeActions#series

gazpachoking:
Perhaps we need to update more wiki pages pointing out the use of the sort_by plugin in these sorts of situations.

PR #2891, opened 2021-03-26
Fix remux quality handling

kmikko:
### Motivation for changes:
Remuxes are incorrectly parsed as `bluray` quality.
Also quality score of `remux` should be higher than `bluray` as remuxes merely contain the original, untouched video and audio streams multiplexed into single file container where as `bluray` releases are considered encodes from the original source.

I'm uncertain whether DVD remuxes are considered a thing or not so they're not part of this PR.

### Detailed changes:
- If `guessit` detects entry source either as `bluray` or `ultra hd bluray` and `remux` is found on `other` [properties](https://doc.guessit.io/properties/) mark source as `remux`
- `remux` should have higher quality than `bluray`
- As a bonus detect `AVC` as `h264`

### DIFF
--- a/flexget/components/parsing/parsers/parser_guessit.py
+++ b/flexget/components/parsing/parsers/parser_guessit.py
@@ -138,6 +138,9 @@ class ParserGuessit:
         if 'region 5' in other or 'region c' in other:
             source.append('r5')
 
+        if 'remux' in other and any(bd in source for bd in ['bluray', 'ultra hd bluray']):
+            source = ['remux']
+
         return source
 
     def _quality(self, guessit_result):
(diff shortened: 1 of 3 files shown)


cvium:
I disagree with the quality change. How often do you have the original source remuxed to mkv? It would be insanely big, not to mention inefficient.

cvium:
We don't make the distinction between a bluray rip and a bluray re-encode. As such I'm not sure a remux is automatically better.

kmikko:
Is there a reason for that? I'm wondering what the `remux` quality listed [here](https://flexget.com/Plugins/quality#qualities) is then used for?

I'm running into problems where I'd like to either download entries or upgrade existing ones of only the best possible quality (bluray remux). Currently this seems to be impossible with the commonly used file naming conventions as Bluray remuxes are identified as "just" `bluray`.

BrutuZ:
Wouldn't [reorder_quality](https://flexget.com/Plugins/reorder_quality) be a viable alternative?

kmikko:
That could solve the quality ordering but there's still the problem of distincting bluray encode from remux.

BrutuZ:
Ideally remux wouldn't be handled in quality then, it should receive the Proper\Repack treatment.

PR #2898, opened 2021-04-05
Update some dev-tools to new versions

jimman2003:
### Motivation for changes:
Get pre-commit hooks working again
### Detailed changes:
- Changes to pre-commit config and dev-requirement to allow for newer versions

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,9 +1,9 @@
 repos:
 - repo: https://github.com/pycqa/isort
-  rev: 5.5.1
+  rev: 5.8.0
   hooks:
   - id: isort
 - repo: https://github.com/psf/black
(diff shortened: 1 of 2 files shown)


gazpachoking:
Maybe we should split out a dev-requirements.in which has looser requirements. Having things pinned is nice because the test and release scripts use these versions, and it's not so nice for them to be able to break when some of our dev tools upgrade.

jimman2003:
Good Idea...Starting the above recommendation

gazpachoking:
Cool, looks good. We could probably do [this](https://github.com/jazzband/pip-tools#workflow-for-layered-requirements) with our `dev-requirements.in` as well, to avoid having conflicts between deps shared between dev and regular requirements.

PR #2952, opened 2021-05-01
add transmission-rpc version range

Trim21:
### Motivation for changes:
 avoid upstream break changes

### Detailed changes:
install rpc library with `transmission-rpc>=3.0.0,<4.0.0`


gazpachoking:
I wish we had a better way to specify these optional deps. We can go with this for now given we don't have anything better atm though. Thanks!

PR #2966, opened 2021-05-06
[fix] emby: strip year in search

soloam:
### Motivation for changes:

In emby all search should be preformed without the year

### Detailed changes:
- Strip year from the searches

### DIFF
--- a/flexget/components/emby/api_emby.py
+++ b/flexget/components/emby/api_emby.py
@@ -117,17 +117,6 @@ class EmbyApiBase(ABC):
 
                 break
 
-    @staticmethod
-    def strip_year(name: str) -> str:
-        if not name or not isinstance(name, str):
-            return name
-
-        new_name, new_year = split_title_year(name)
-        if not new_name or new_name == "":
-            return name
-
-        return new_name
-
 
 class EmbyAuth(EmbyApiBase):
     """
@@ -1251,7 +1240,7 @@ class EmbyApiMedia(EmbyApiBase):
         EmbyApi.set_provideres_search_arg(args, **kwargs)
         EmbyApi.set_common_search_arg(args)
 
-        args['SearchTerm'] = EmbyApi.strip_year(kwargs.get('title'))
+        args['SearchTerm'] = split_title_year(kwargs.get('title'))[0]
 
         logger.debug('Search media with: {}', args)
         medias = EmbyApi.resquest_emby(EMBY_ENDPOINT_SEARCH, auth, 'GET', **args)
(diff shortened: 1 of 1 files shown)


flexget/components/emby/api_emby.py
```diff
@@ -117,6 +117,19 @@ def update_using_map(target, field_map: dict, source_item, **kwargs):
...
                 break
 
+    @staticmethod
+    def strip_year(name: str) -> str:
+        new_name = name
+
+        if not isinstance(name, str):
+            return name
+
+        new_name = re.sub(r'^(.*) \(\d{4}\)(.*)', r'\1\2', new_name)
```
BrutuZ, line 127: Wouldn't it be cleaner to search and strip the year disregarding the wildcards before and after?
`new_name = re.sub(r' ?\(\d{4}\)', '', new_name)`

gazpachoking:
We have a utility to split out years, does it work for this purpose, or is yours more specific to this case?
https://github.com/Flexget/Flexget/blob/97ca6d2219c43f723be67bfb7ef2021dbe3c1f72/flexget/utils/tools.py#L357-L379

soloam:
That works for me... I'll change it... Tks

flexget/components/emby/api_emby.py
```diff
@@ -1240,7 +1251,7 @@ def search(**kwargs):
         EmbyApi.set_provideres_search_arg(args, **kwargs)
         EmbyApi.set_common_search_arg(args)
 
-        args['SearchTerm'] = kwargs.get('title')
+        args['SearchTerm'] = EmbyApi.strip_year(kwargs.get('title'))
```
gazpachoking, line 1254: I don't think you need a helper function to call the helper function. The existing split_title_year already does everything your helper does.
```suggestion
        args['SearchTerm'], _ = split_title_year(kwargs.get('title')
```
or
```suggestion
        args['SearchTerm'] = split_title_year(kwargs.get('title')[0]
```
Now that I've written that, I sorta want to change our helper to return a named tuple, so we could use `.title` rather than `[0]`

gazpachoking: I made that tool return a namedtuple now, so you can call `.title` on the result

soloam: I pushed a new change, to remove the declaration of the now unused method

soloam: I added the use of the named tuple

gazpachoking:
Looks good, thanks!

PR #2969, opened 2021-05-09
Add move_data action to Transmission component

ianstalk:
### Motivation for changes:
I would like to download files to an "incomplete" directory before moving them to a different path to be acted on by my file system tasks so that it doesn't attempt to operate on incomplete data (e.g. multipart RAR archives).

### Detailed changes:
- Added move_data action to the transmission plugin

```
  transmission-move-movies:
    priority: 203
    disable: [seen, seen_info_hash]
    from_transmission:
      <<: *transmission
    if:
      - transmission_status != 'seeding': reject
      - transmission_downloadDir == '/var/lib/transmission-daemon/downloads/incomplete/movies': accept
    transmission: 
      enabled: yes
      host: '{? transmission.host ?}'
      port: 9091
      username: '{? transmission.username  ?}'
      password: '{? transmission.password  ?}'
      action: move_data
      path: '/var/lib/transmission-daemon/downloads/movies'
```
```
2021-05-09 13:47:22 VERBOSE  details       transmission-move-movies Produced 10 entries.
2021-05-09 13:47:22 VERBOSE  task          transmission-move-movies ACCEPTED: `XXX` by if plugin because matched requirement: transmission_downloadDir == '/var/lib/transmission-daemon/downloads/movies'
2021-05-09 13:47:22 VERBOSE  details       transmission-move-movies Summary - Accepted: 1 (Rejected: 0 Undecided: 9 Failed: 0)
2021-05-09 13:47:22 VERBOSE  transmission  transmission-move-movies Moving XXX to "/var/lib/transmission-daemon/downloads/movies"
2021-05-09 13:47:22 INFO     transmission  transmission-move-movies set data location for XXX to /var/lib/transmission-daemon/downloads/movies in transmission
```
#### To Do:

- Wiki pages will need to be updated if/when this gets merged.

### DIFF
--- a/flexget/plugins/clients/transmission.py
+++ b/flexget/plugins/clients/transmission.py
@@ -643,8 +643,11 @@ class PluginTransmission(TransmissionBase):
                     start_torrent(bypass_queue=True)
                     logger.info('resumed (bypass queue) {} in transmission', torrent_info.name)
                 elif config['action'] == 'move_data':
-                    self.client.move_torrent_data([torrent_info.id], config['path'])
-                    logger.info('set data location for {} to {} in transmission', torrent_info.name, config['path'])
+                    try:
+                        self.client.move_torrent_data([torrent_info.id], self._render_path(entry, config['path']))
+                        logger.info('set data location for {} to {} in transmission', torrent_info.name, config['path'])
+                    except RenderError as e:
+                        logger.error('Error setting data location for {}: {}', entry['title'], e)
 
             except TransmissionError as e:
                 logger.opt(exception=True).debug('TransmissionError')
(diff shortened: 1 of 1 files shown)


gazpachoking:
Is there a reason we need to add the `move_data` config key?  Shouldn't it just always move if you send an already loaded torrent back to transmission but with a new path set? (That's how the deluge plugin works already, you can use from_deluge and deluge in the same task, and edit any existing options.)

soloam:
Sorry I don't understand.. transmission does this negatively! I use transmission and I do this all the time. Could I be getting something wrong? Basically what you need is to have files during download in one path... and when it's done have it moved to it's final path!

https://github.com/transmission/transmission/wiki/Editing-Configuration-Files

> incomplete-dir: String (default = default locations) Directory to keep files in until torrent is complete.
incomplete-dir-enabled: Boolean (default = false) When enabled, new torrents will download the files to incomplete-dir. When complete, the files will be moved to download-dir.

ianstalk:
@soloam Transmission does this natively, but not on a per torrent basis. I try to download most of my stuff to the SSD in my machine rather than keeping the larger traditional HDDs spun up more than necessary; but sometimes I download something too large to fit on the SSD, and even just the temporary space use is too much.

PR #2976, opened 2021-05-13
[fix] emby: login and list matching

soloam:
### Motivation for changes:

Fixes login and list matching

### Detailed changes:
- The login was not renewing when token changed, fixed
- The list was not matching correctly episodes in some cases, added a new matching

### DIFF
--- a/flexget/components/emby/api_emby.py
+++ b/flexget/components/emby/api_emby.py
@@ -2,6 +2,7 @@ from abc import ABC, abstractmethod, abstractstaticmethod
 from collections.abc import MutableSet
 from datetime import datetime
 import re
+import copy
 from urllib.parse import urlencode
 
 import functools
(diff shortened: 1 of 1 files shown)


soloam:
This can be merged to solve the problems!

Question:
I have created a couple of tests to this, and generated the "cassettes" but when I checked I noticed this:

```
    headers:
      Accept-Encoding:
      - gzip, deflate
      Connection:
      - keep-alive
      User-Agent:
      - FlexGet/3.1.122.dev (www.flexget.com)
      X-Emby-Authorization:
      - Emby UserId=fa55a425b8b842b38789ad380d2002d3, Client=Flexget, Device=Flexget
        Plugin, DeviceId=flexget_plugin, Version=3.1.122.dev
      X-Emby-Token:
      - 6ee98907f4304b3995b641a024f879a4
      accept:
      - application/json

```

This is the header of the request, where I send the version of flexget to appear in the dashboard of emby. Will this break the tests when flexget changes version? The request will take the new flexget version and will not match this header. Does vcr matches this header? Or only matched the url?

gazpachoking:
> This is the header of the request, where I send the version of flexget to appear in the dashboard of emby. Will this break the tests when flexget changes version? The request will take the new flexget version and will not match this header. Does vcr matches this header? Or only matched the url?

Good question. I think VCRs default request matching is smart enough that it allows this. (We haven't run into a problem before, and I believe it's default that we set user agent to flexget version.) It is customizable if it's a problem, but let's not worry about it unless we prove it is.

gazpachoking:
Didn't look to closely, just trusting you on this one. 😉  Thanks!

PR #2979, opened 2021-05-14
[add] INPUT piratebay_list

trananhtuan:
### Motivation for changes:
Enable discovery of fresh and trending torrents from ThePirateBay using its API

```
piratebay_list:
  url: https://apibay.org
  category: HD - Movies
  list: top
  rank: all

OR:

piratebay_list:
  category: HD - Movies
```
#### To Do:

- Add docs.


soloam:
The first thing that occurs to me, even before looking at the code, is that we should not name it *_list.... This will make confusion with the managed lists... This looks like a from.

Tks

PR #2982, opened 2021-05-14
[add] config: allows to use includes in configs

soloam:
### Motivation for changes:

Add the ability to split config into include files.

The existing include is a simple plugin that works in simple tasks... With to massive level of information of config files, maintain  can be messy, so the ability to split the config into config files is welcome. 

I tried to make't as versatile as possible, and to merge perfectly into the existing config, this is my solution to a very requested functionality.

### Detailed changes:

1st: This is a very versatile and anyone can chose the level of complexity that is desired to maintain
2nd: All includes are recursive, meaning if you include a folder, it will dig that folder, this is required and making it not recursive will imply planning a new tag.

This are the current include options:

 - !include <path>: can include a file or a path, if the include is a path, the file tree will be the properties, more clarification about this next
 - !include_list <path>: can include a file or path and will return a merged list. This include must not be used in folders with sub-folders, if you read a file, all the content of that file will be returned as a merged list, if it's a folder, all the content from all the files will be returned as a merged list

Let's look at some examples to help understand the concept:

First Let's see the current file tree, this example was generated to show the full potencial of the includes, with most of the possible scenarios:

```
.
├── config
│   ├── extras
│   │   └── transmission.yaml
│   ├── lists
│   │   ├── movies_src
│   │   │   ├── list_rss1.yaml
│   │   │   └── list_rss2.yaml
│   │   └── movies_that_i_need.yaml
│   ├── tasks
│   │   ├── download_movies.yaml
│   │   └── download_series
│   │       ├── regexp
│   │       │   └── reject_excluding.yaml
│   │       └── template.yaml
│   └── templates
│       ├── download.yaml
│       ├── movies.yaml
│       └── tv
│           └── configure_series
│               ├── from
│               │   └── trakt_list.yaml
│               └── settings.yaml
├── config.yml
├── db-config.sqlite
└── secrets.yaml
```

**Now let's look at the main files**

`secrets.yaml`
```yaml
trakt:
  user: my_user
  list: my_list
```

`config.yaml`
```yaml
variables: secrets.yaml
templates: !include config/templates
tasks: !include config/tasks
```

**Explain:** The config file declares the `variables` file, the `templates` and the `tasks`, both the `templates` and the `tasks` are includes, using the "!include", meaning that will dig the paths for config, using the file tree as the dict structure. The last file set's its content.

**Lets look ate the tasks folder**

```
config/tasks/
├── download_movies.yaml
└── download_series
    ├── regexp
    │   └── reject_excluding.yaml
    └── template.yaml
```

**Explain:** Two tasks are created, the "download_movies" and the "download_series".

**Let's look at the download_movies.yaml file**

This is the simplest approach, the file is in first level of the tree.. is content is returned to the main property (in this case the tasks).

`download_movies.yaml`
```yaml
template:
  - movies
  - download
inputs: !include_list config/lists/movies_src
```

**Explain:** The download_movies task is created and the contend from the file are the plugins used. But we have a new include, this is a different type "include_list". This returns the full content of a file as a list, if more that a file is present, all the contents ar merged.

**Let's look at the config/lists/movies_src folder**

```
config/lists/movies_src/
├── list_rss1.yaml
└── list_rss2.yaml
```

`list_rss1.yaml`
```yaml
- rss:
    url: "http://tracker1.dummy"
- rss:
    url: "http://tracker2.dummy"
- rss:
    url: "http://tracker3.dummy"
```

`list_rss2.yaml`
```yaml
rss:
  url: "http://tracker4.dummy"
```

**Explain:** Since "include_list" was used, all the contents are merged into one list, this will be the result of the merge:

```
  - rss:
      all_entries: true
      ascii: false
      escape: false
      group_links: false
      silent: false
      url: http://tracker1.dummy
  - rss:
      all_entries: true
      ascii: false
      escape: false
      group_links: false
      silent: false
      url: http://tracker2.dummy
  - rss:
      all_entries: true
      ascii: false
      escape: false
      group_links: false
      silent: false
      url: http://tracker3.dummy
  - rss:
      all_entries: true
      ascii: false
      escape: false
      group_links: false
      silent: false
      url: http://tracker4.dummy
```

And this the final result of the task:

```
inputs:
  - rss:
      all_entries: true
      ascii: false
      escape: false
      group_links: false
      silent: false
      url: http://tracker1.dummy
  - rss:
      all_entries: true
      ascii: false
      escape: false
      group_links: false
      silent: false
      url: http://tracker2.dummy
  - rss:
      all_entries: true
      ascii: false
      escape: false
      group_links: false
      silent: false
      url: http://tracker3.dummy
  - rss:
      all_entries: true
      ascii: false
      escape: false
      group_links: false
      silent: false
      url: http://tracker4.dummy
template:
  - movies
  - download
```

**Now let's look ate a more complex example, in the download_series**

This allows to declare the tasks and plugins in files, this use case is very extreme and basically is result of the tree dinging process that the code does. Since each folder is a entry in a dict, it's possible to get something like this, **if desired**:

```
config/tasks/download_series/
├── regexp
│   └── reject_excluding.yaml
└── template.yaml
```

`reject_excluding.yaml`
```yaml
- "trash": { from: url }
```

`template.yaml`
```yaml
- tv
- download
```

The result of this will be the task:

```yaml
regexp:
  reject_excluding:
    - trash:
        from: url
template:
  - tv
  - download
```

I think that with this it's possible to see the returning config from the rest of the files, including the templates. 

File List:
```
.
├── config
│   ├── extras
│   │   └── transmission.yaml
│   ├── lists
│   │   ├── movies_src
│   │   │   ├── list_rss1.yaml
│   │   │   └── list_rss2.yaml
│   │   └── movies_that_i_need.yaml
│   ├── tasks
│   │   ├── download_movies.yaml
│   │   └── download_series
│   │       ├── regexp
│   │       │   └── reject_excluding.yaml
│   │       └── template.yaml
│   └── templates
│       ├── download.yaml
│       ├── movies.yaml
│       └── tv
│           └── configure_series
│               ├── from
│               │   └── trakt_list.yaml
│               └── settings.yaml
├── config.yml
├── db-config.sqlite
└── secrets.yaml
```

Files:

```
./config/lists/movies_src/list_rss1.yaml
- rss:
    url: "http://tracker1.dummy"
- rss:
    url: "http://tracker2.dummy"
- rss:
    url: "http://tracker3.dummy"

./config/lists/movies_src/list_rss2.yaml
rss:
  url: "http://tracker4.dummy"

./config/lists/movies_that_i_need.yaml
- movie_list: movies
- movie_list: movies2

./config/templates/movies.yaml
list_match:
  from: !include config/lists/movies_that_i_need.yaml

./config/templates/tv/configure_series/from/trakt_list.yaml
account: "{? trakt.user ?}"
type: shows
list: "{? trakt.list ?}"

./config/templates/tv/configure_series/settings.yaml
quality: "720p"
identified_by: ep
season_packs: no

./config/templates/download.yaml
transmission: !include config/extras/transmission.yaml

./config/extras/transmission.yaml
host: "localhost"

./config/tasks/download_movies.yaml
template:
  - movies
  - download
inputs: !include_list config/lists/movies_src

./config/tasks/download_series/regexp/reject_excluding.yaml
- "trash": { from: url }

./config/tasks/download_series/template.yaml
- tv
- download
```

config.yaml
```
variables: secrets.yaml
templates: !include config/templates
tasks: !include config/tasks
```

secrets.yaml
```
trakt:
  user: my_user
  list: my_list
```

flexget execute --dump-config
```
--- config from task: download_movies
inputs:
  - rss:
      all_entries: true
      ascii: false
      escape: false
      group_links: false
      silent: false
      url: http://tracker1.dummy
  - rss:
      all_entries: true
      ascii: false
      escape: false
      group_links: false
      silent: false
      url: http://tracker2.dummy
  - rss:
      all_entries: true
      ascii: false
      escape: false
      group_links: false
      silent: false
      url: http://tracker3.dummy
  - rss:
      all_entries: true
      ascii: false
      escape: false
      group_links: false
      silent: false
      url: http://tracker4.dummy
list_match:
  action: accept
  from:
    - movie_list: movies
    - movie_list: movies2
  remove_on_match: true
  single_match: true
template:
  - movies
  - download
transmission:
  host: localhost

---

--- config from task: download_series
configure_series:
  from:
    trakt_list:
      account: my_user
      list: my_list
      strip_dates: false
      type: shows
      username: me
  settings:
    identified_by: ep
    quality: 720p
    season_packs: false
regexp:
  reject_excluding:
    - trash:
        from: url
template:
  - tv
  - download
transmission:
  host: localhost

---
```
#### To Do:

- [ ] Write wiki

### DIFF
--- a/flexget/manager.py
+++ b/flexget/manager.py
@@ -23,6 +23,7 @@ from typing import (  # noqa
     Tuple,
     Type,
     Union,
+    Callable,
 )
 
 import sqlalchemy  # noqa
@@ -64,6 +65,44 @@ logger = logger.bind(name='manager')
 
 manager: Optional['Manager'] = None
 DB_CLEANUP_INTERVAL = timedelta(days=7)
+
+
+class YamlIncludesLoaderMapper:
+    """
+    Class to map Includes
+    """
+
+    _hadler = None
+    _caller = None
+
+    @staticmethod
+    def construct(tag: str, hadler: Callable, caller: 'Manager'):
+        """Factory
+
+        Args:
+            tag (str): Tag to map
+            hadler (Callable): Tag decoder hadler
+            caller (Manager): Manager caller
+        """
+        mapper = YamlIncludesLoaderMapper(tag, hadler, caller)
+
+    def __init__(self, tag: str, hadler: Callable, caller: 'Manager'):
+        """Constructor
+
+        Args:
+            tag (str): Tag to map
+            hadler (Callable): Tag decoder hadler
+            caller (Manager): Manager caller
+        """
+        self._hadler = hadler
+        self._caller = caller
+
+        yaml.SafeLoader.add_constructor(tag, self._costructor)
+
+    def _costructor(self, other, node):
+        """Internal handler"""
+        data = self._hadler(other, node, self._caller)
+        return data
 
 
 class Manager:
@@ -544,6 +583,147 @@ class Manager:
             # Override the default string handling function
             # to always return unicode objects
             return self.construct_scalar(node)
+
+        def includes_merge(src, dst, key, file):
+            """Merges src into dst"""
+            if isinstance(src, dict) and isinstance(dst[key], dict):
+                dst[key].update(src)  # Merge dicts
+            elif isinstance(src, dict):
+                # Only merge dicts with dicts
+                message = f'Not possible to merge `{file}` ({type(src).__name__}) to config ({type(dst[key]).__name__}).'
+                logger.critical(message)
+                raise ValueError(message)
+            elif isinstance(src, list):
+                if isinstance(dst[key], list):
+                    # Merge lists
+                    dst[key] = dst[key] + src
+                elif isinstance(dst[key], dict) and not dst[key]:
+                    # Add list to dict if dict is empty
+                    dst[key] = src
+                else:
+                    # Can't merge lists if not list, or dict not empty
+                    message = f'Not possible to merge `{file}` ({type(src).__name__}) to config ({type(dst[key]).__name__}).'
+                    logger.critical(message)
+                    raise ValueError(message)
+            else:
+                # Merge the rest of the types
+                dst[key] = src
+
+        def construct_yaml_includes(self, node, caller):
+            """
+            Handles Yaml Includes
+            """
+
+            ALLOWED_INCLUDES = ['yml', 'yaml']
+
+            # Tag of the inclide
+            tag = node.tag[1:]
+
+            # Full path to the include
+            include_path = os.path.join(caller.config_base, node.value)
+
+            # Get the complete icluded tree
+            if os.path.isdir(include_path):
+                flist = list_files(include_path)
+            elif os.path.isfile(include_path):
+                flist = [include_path]
+            else:
+                logger.critical(f'Include `{include_path}` is a invalid path')
+                raise ValueError(f'Include `{include_path}` is a invalid path')
+
+            flist.sort()
+
+            include_config = {}
+            for file in flist:
+
+                # Returns file information
+                file_path = os.path.relpath(file, include_path)
+                file_info = os.path.splitext(file)
+                if len(file_info) != 2:
+                    continue
+
+                file_ext = file_info[1][1:]
+                if file_ext not in ALLOWED_INCLUDES:
+                    continue
+
+                file_name = os.path.split(file_info[0])[-1]
+
+                # Gets the complete dict tree from file tree
+                config_path = file_path.split(os.path.sep)
+                config_path.pop()
+
+                # Reads included file
+                with open(file, 'r', encoding='utf-8') as f:
+                    try:
+                        raw_include_config = f.read()
+                    except UnicodeDecodeError:
+                        logger.critical('Config file must be UTF-8 encoded.')
+                        raise ValueError('Config file is not UTF-8 encoded')
+
+                # Build config path based on file path
+                include_config_path = include_config
+                for path in config_path:
+                    path = path.replace(os.path.sep, '')
+                    if not path:
+                        continue
+                    if not path in include_config_path:
+                        include_config_path[path] = {}
+
+                    include_config_path = include_config_path[path]
+
+                if not file_name in include_config_path:
+                    include_config_path[file_name] = {}
+
+                # Generate Include config
+                compiled_include = yaml.safe_load(raw_include_config) or {}
+
+                # Merge includes
+                includes_merge(compiled_include, include_config_path, file_name, file)
+
+            if tag == 'include_list':
+                # If its a merge of lists, convert output to list
+                if isinstance(include_config, dict):
+                    new_include_config = []
+                    for key in include_config:
+                        if isinstance(include_config[key], dict):
+                            new_include_config.append(include_config[key])
+                        elif isinstance(include_config[key], list):
+                            new_include_config += include_config[key]
+                        else:
+                            new_include_config.append(include_config[key])
+                    include_config = new_include_config
+                elif not isinstance(include_config, list):
+                    include_config = [include_config]
+
+            if os.path.isfile(include_path):
+                # If we requested a file, we return only that config
+                include_config = include_config[list(include_config.keys())[0]]
+
+            return include_config
+
+        def list_files(dirname):
+            flist = os.listdir(dirname)
+            return_flist = []
+
+            for file in flist:
+                fpath = os.path.join(dirname, file)
+
+                if os.path.isdir(fpath):
+                    return_flist = return_flist + list_files(fpath)
+                elif os.path.isfile(fpath):
+                    return_flist.append(fpath)
+                else:
+                    continue
+
+            return return_flist
+
+        INCLUDES_HANLERS = {
+            '!include': construct_yaml_includes,
+            '!include_list': construct_yaml_includes,
+        }
+
+        for tag in INCLUDES_HANLERS:
+            YamlIncludesLoaderMapper.construct(tag, INCLUDES_HANLERS[tag], self)
 
         yaml.Loader.add_constructor('tag:yaml.org,2002:str', construct_yaml_str)
         yaml.SafeLoader.add_constructor('tag:yaml.org,2002:str', construct_yaml_str)
--- a/flexget/manager.py
+++ b/flexget/manager.py
@@ -590,7 +590,7 @@ class Manager:
                 dst[key].update(src)  # Merge dicts
             elif isinstance(src, dict):
                 # Only merge dicts with dicts
-                message = f'Not possible to merge `{file}` ({type(src).__name__}) to config ({type(dst[key]).__name__}).'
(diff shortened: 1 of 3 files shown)


soloam:
I changed a small part on how the includes would work... to allow to be more intuitive... Also added the tests to pytest. So this is the solution:

        include - includes a file or dir (search subdirs). Includes the files contents
        include_list - includes a file or dir (search subdirs). Includes the files contents in list
        include_named - includes a file or dir (search subdirs). Includes the file contents, property is the file name
        include_named_list - includes a file or dir (search subdirs). Includes the file contents, property is the file name in list

### include
Is a direct include (if its a dir it search that dir and subdirs for all yaml/yml files) and dumps the exact content into the config.
**Use Case example:** For example to create tasks where the content of the file will have the complete declaration of the task, including its name. Can point to a dir and create/declare all the tasks in that dir.

### include_list
Is a direct include (if its a dir it search that dir and subdirs for all yaml/yml files) and dumps the exact content into the config, but converts into a list. So the output is equal to the "Include" but in list.
**Use Case example:** For example to output lists of inputs, where the file will contain the plugin name and the declaration of that plugin. Can point to a dir and create/declare all the inputs in that dir.

### include_named
Equal to the "include", but the file name will be the top property. If a dir, same logic is applied to all the files.
**Use Case example:** For example to create tasks where the content of the file will have the declaration of the task and the file name will be the task name. Can point to a dir and create/declare all the tasks in that dir.

### include_named_list
Equal to the "include_list", but the file name will be the top property. If a dir, same logic is applied to all the files.
**Use Case example:** For example to output lists of inputs, where the file will contain the plugin the eclaration of that plugin and the file name will be the plugin name. Can point to a dir and create/declare all the inputs in that dir.

gazpachoking:
A few things. I still think we should reduce the scope and simplify this. I think regular `!include` for one file, and `!include_named` for a directory of tasks are probably the things that are most useful.
I think it should probably be split out into its own file. Not sure if it should go at the root of the source, or in utils dir.
Here is something more on the lines of what I was thinking:
https://github.com/Flexget/Flexget/blob/new_include/flexget/yaml_config.py
This was just a quick mockup, so I haven't tested it, or updated your test cases.

soloam:
Updated the new interface... Used your as base and added what in my mind makes more sense.

!include - should always include a file... if it's a folder should scrap all files (in sub-folders) and return them merged
!include_names - same logic as include, but the name of the file is the top property
!include_list - returns always a merged list

Removed the include_named_list and include_tree... They are disposable

This is what I see for a system like this... I think we can find common ground and settle in a solution that we are comfortable and that in the end it's usable and useful to all

soloam:
My thoughts on this:

- If I had to scale the importance of the tags I would say: include > include_list > include_names.
- Recursive include of dir is very useful, if the idea is to split the config and organize, having files splitted in sub folders is useful... For example a master folder with all the tasks, but inside we can split in movie_tasks, series_tasks etc... but in the end the include would include all in the tasks.
- include_list is very useful to include a merged list of inputs, for example... and with that having a folder with all the includes to be used as inputs

gazpachoking:
I'm pretty dubious on the whole merging thing still. As well as the plain `!include` working on directories.

soloam:
If I need to include a butch of files... I would prefer to have all in one place a and get them all at once merged, and if I need to add or remove I add or create files... 

This goes for the include... 

The include_list without the merging it's useless... If without subfolder scraping a and merging is the way... Include_list is useless...

gazpachoking:
I guess my issue is that this seems like a very niche use case, (I really can't imagine it being helpful splitting things up into smaller chunks than a task per file,) and even splitting it up that far doesn't seem like it brings a lot of benefit to me. Given that, every bit of complication (both in the code added, what the code is doing, and different things possible in the config,) makes me more hesitant to merge it. There is also the issue @ksurl mentioned, that the web config editor wouldn't support any of this at all without quite a bit of work on that end.

soloam:
Yes the editor would not allow to change the includes at the moment! But that should not be a show stopper! That can be stated in the wiki, that if you go the include route, that ATM webui editor would not work!

I think that in the end this would be beneficial, I don't see how making recursive would be more complex for the user, it's optional, not mandatory to use! But in the end it's your call, and a section is always better than nothing, so we can go with whatever makes you more comfortable! You have a lot more years of experience in this that me, and if your instincts are buzzing caution...

soloam:
Just to add, for me the most beneficial points of this would come from adding configs like inputs and series config, dynamic, with only the creation of one file! The tasks section on files it's good, but not a top off the line addition!

Then again, something is better than nothing!

I'll support your final call, even if you prefer not to merge nothing at all! Like I said, you have a lot more experience than me one this!

soloam:
@gazpachoking if we go further we need to understand the reason pytest is not rendering the includes!

gazpachoking:
> @gazpachoking if we go further we need to understand the reason pytest is not rendering the includes!

I'd probably just use yaml_config.load to test, is that what you are trying?

soloam:
I created a couple of task's and include file's and executed the tasks... Then I asserted what I was expecting the config to be!

The file is in the PR, but I commented the code to avoid errors

gazpachoking:
Wouldn't be terrible to have an integration test like you describe to test it more completely, but I think most of it could be tested with unit tests, where we skip the rest of the FlexGet framework. Something like this mockup:
```python
f1 = """
tasks: !include f2.yml
"""

f2 = """
a:
  mock:
  - title: a
"""

combined = {
    "tasks": {
        "a": [{"title": "a"}]
    }
}

with open("f1.yml") as f:
    f.write(f1)
    
with open("f2.yml") as f:
    f.write(f2)
    
assert yaml_config.load("f1.yml") == combined
```

soloam:
> Wouldn't be terrible to have an integration test like you describe to test it more completely, but I think most of it could be tested with unit tests, where we skip the rest of the FlexGet framework. Something like this mockup:
> 
> ```python
> f1 = """
> tasks: !include f2.yml
> """
> 
> f2 = """
> a:
>   mock:
>   - title: a
> """
> 
> combined = {
>     "tasks": {
>         "a": [{"title": "a"}]
>     }
> }
> 
> with open("f1.yml") as f:
>     f.write(f1)
>     
> with open("f2.yml") as f:
>     f.write(f2)
>     
> assert yaml_config.load("f1.yml") == combined
> ```

The result should be:
```yaml
combined = {
    "tasks": {
        "a": { "mock": [{"title": "a"}] }
    }
}
```

But yes... test would test the generator. Just would like to understand the reason the reason task.execute worked on my first PR, and not here!

I think first we need to close the end tags and functions to merge... Then we can generate the tests for them

soloam:
Keep this relevant

paranoidi:
This is too granular / complex to my taste. However splitting config into separate more manageable files is something I'd like to see. Ie. each task & template into own file. Not splitting task into multiple files.

soloam:
The biggest advantage would be to allow users to decide the best way to organize it's config! Similar to Home Assistant! Some people would prefer a layout, and other's would prefer other.

PR #3009, opened 2021-05-24
[fix] emby: Research if first search fails

soloam:
### Motivation for changes:

Make more searches with different fields if first search fails! This will help to make a better matching of media with the media in emby

### Detailed changes:
- Add research with movie and serie title if failed by provider (for example imdb_id)


gazpachoking:
Didn't look too closely, but trusting you as the expert on this plugin. 😄

PR #3011, opened 2021-05-25
[add] jinja asciify: Siplify text with jinja

soloam:
### Motivation for changes:

Allow to simplify text with jinja

### Detailed changes:
- Added a new jinja filter
- Added test

```yaml
        tasks:
          asscify_me:
            mock:
              - {"title":"[My](Tìtlê)-ìs a' me^ss", "url":"mock://local" }
                
            accept_all: yes
                
            set:
              title: "{{title|asciify|replace('is','is not')}}"
```
```yaml
title: "My Title is not a mess"
```

### DIFF
--- a/flexget/utils/template.py
+++ b/flexget/utils/template.py
@@ -2,6 +2,7 @@ import locale
 import os
 import os.path
 import re
+from unicodedata import normalize
 from contextlib import suppress
 from copy import copy
 from datetime import date, datetime, time
@@ -134,6 +135,23 @@ def filter_to_date(date_time_val):
 def filter_default(value, default_value: str = '', boolean: bool = True) -> str:
     """Override the built-in Jinja default filter to set the `boolean` param to True by default"""
     return jinja2.filters.do_default(value, default_value, boolean)
+
+
+def filter_asciify(text: str) -> str:
+    """ Siplify text """
+
+    if not isinstance(text, str):
+        return text
+
+    # Replace accented chars by their 'normal' couterparts
+    result = normalize('NFKD', text)
+    # Symbols that should be converted to white space
+    result = re.sub(r'[ \(\)\-_\[\]\.]+', ' ', result)
+    # Leftovers
+    result = re.sub(r"[^a-zA-Z0-9 ]", "", result)
+    # Replace multiple white spaces with one white space
+    result = ' '.join(result.split())
+    return result
 
 
 filter_d = filter_default
--- a/flexget/utils/template.py
+++ b/flexget/utils/template.py
@@ -2,11 +2,11 @@ import locale
 import os
 import os.path
 import re
-from unicodedata import normalize
 from contextlib import suppress
 from copy import copy
 from datetime import date, datetime, time
-from typing import TYPE_CHECKING, Any, List, Mapping, Optional, Union, AnyStr, Type, cast
+from typing import TYPE_CHECKING, Any, AnyStr, List, Mapping, Optional, Type, Union, cast
+from unicodedata import normalize
 
 import jinja2.filters
 from dateutil import parser as dateutil_parse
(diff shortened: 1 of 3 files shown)


flexget/utils/template.py
```diff
@@ -136,6 +137,26 @@ def filter_default(value, default_value: str = '', boolean: bool = True) -> str:
...
+    # Replace accented chars by their 'normal' couterparts
+    result = normalize('NFKD', text)
+
+    # Symbols that should keeped as white space
+    result = re.sub(r'[ \(\)\-_\[\]\.]+', ' ', result)
+    # Leftovers
+    result = re.sub(r"[^a-zA-Z0-9 ]", "", result)
+
+    # Remove Multiple White Spaces to One
+    result = ' '.join(result.split())
```
ianstalk, line 155: This will replace all whitespace with a single space. Probably the desired result, but wanted to make sure it was intentional.

soloam: That's correct! And also remove all leading and trailing white spaces!

This should be done because the replaces can leave multiple white spaces if more that one special char is together

flexget/utils/template.py
```diff
@@ -136,6 +137,26 @@ def filter_default(value, default_value: str = '', boolean: bool = True) -> str:
...
+def filter_asciify(text: str) -> str:
+    """ Siplify text """
+
+    if not isinstance(text, str):
+        return text
+
+    # Replace accented chars by their 'normal' couterparts
+    result = normalize('NFKD', text)
+
+    # Symbols that should keeped as white space
```
BrutuZ, line 149: kept*

Just in case there's another commit to this 😝

soloam:
Sorry about the multiple commits, I tried to merge them without success. You can merge with "merge squash"!

gazpachoking:
I think I'm good with this. Open to input if people think the name should be different. The only thing I'm sorta torn on is how much cleanup it should be doing (just accents? accents and whitespace? accents whitespace and symbols?) open to opinions from anyone who would use this, or something like it. I'm a bit tempted to keep it doing less, because you can always compose more jinja filters with it, but if people would always be using them together then it does make sense to have an all inclusive filter.

soloam:
We could accept a input on the filter to determine the level of cleaning. Like define a defaul and with an input do something else (less or more)

BrutuZ:
I second the addition of optional parameter(s) to fine-tune what to strip or replace.
Not loving the _fy_ in the name tho, I'd leave as just `ascii` or `to_ascii` if really need to be clear it's a conversion outside the documentation, or perhaps something different like `safe_string`

soloam:
I also kinda like the safe_string name! 

When all is settled I make the changes

Something like

```yaml

text: text|safe_string
text2: text|safe_string(true)
```

The first just replaces accent, the second also symbol

soloam:
Any options on this?

gazpachoking:
I think I like `ascii` as the name, and it doesn't strip any symbols (i.e. just tries to smartly convert to ascii.) I think maybe a separate filter to strip the symbols might be better than adding an option? Then you could just `some_field|ascii|strip_symbols` or something to do the whole thing.

BrutuZ:
`strip_symbols` would be just an alias for `re_replace('[^\w\d\s]','')` then?

soloam:
I like the idea of the `some_field|ascii|strip_symbols` I'll make the change

gazpachoking:
> `strip_symbols` would be just an alias for `re_replace('[^\w\d\s]','')` then?

Yeah. Not sure if it's really needed, just slightly friendlier to regex newbies. I mostly just liked the idea of the `ascii` filter being more focused, since you can always compose more filters on top.

gazpachoking:
This still waiting on the above mentioned changes?

paranoidi:
Seems pretty good to me!

soloam:
New config:

```yaml
        tasks:
          asscify_me:
            mock:
              - {"title":"[My](Tìtlê)-ìs a' me^ss", "url":"mock://local" }
                
            accept_all: yes
                
            set:
              title1: "{{title|asciify|replace('is','is still')}}"
              title2: "{{title|strip_symbols|replace('ìs','ìs still')}}"
              title3: "{{title|asciify|strip_symbols|replace('is','is not')}}"
```

soloam:
@gazpachoking things are changed. I'm getting a merge error, tried to sync my repo and commit again but keeps showing!

soloam:
Damm I can't solve this... I'll close this and push a new version based on current DEV

PR #3015, opened 2021-05-26
[add] jinja split year: Jinja split name and year

soloam:
### Motivation for changes:

A clean way to split the year from the title with jinja, that is equal to the logic applied in flexget

### Detailed changes:
- add jinja filter strip_year
- add jinja filter get_year
- add test cases to Jinja filters (can accommodate more filters in the futer, to avoid multiple files per filter)

```yaml
        tasks:
          stripyear:
            mock:
              - {"title":"The Matrix (1999)", "url":"mock://local1" }
              - {"title":"The Matrix", "url":"mock://local2" }
              - {"title":"The Matrix 1999", "url":"mock://local3" }
              - {"title":"2000", "url":"mock://local3" }
              - {"title":"2000 (2020)", "url":"mock://local4" }
              - {"title":"2000 2020", "url":"mock://local5" }
                
            accept_all: yes
                
            set:
              name: "{{title|strip_year}}"
              year: "{{title|get_year}}"
```
```yaml
name: The Matrix
year: 1999

name: The Matrix
year: None

name: The Matrix
year: 1999

name: 2000
year: None

name: 2000
year: 2020

name: 2000
year: 2020
```

### DIFF
--- a/flexget/tests/test_jinja_filters.py
+++ b/flexget/tests/test_jinja_filters.py
@@ -1,0 +1,43 @@
+import pytest
+
+
+class TestJinjaFilters:
+    config = """
+        tasks:
+          stripyear:
+            mock:
+              - {"title":"The Matrix (1999)", "url":"mock://local1" }
+              - {"title":"The Matrix", "url":"mock://local2" }
+              - {"title":"The Matrix 1999", "url":"mock://local3" }
+              - {"title":"2000", "url":"mock://local3" }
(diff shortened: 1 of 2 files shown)


gazpachoking:
This seems like a good idea. The only thing I might do differently is structure the test so that it just renders the jinja templates directly instead of running the whole task. Not really a big deal either way with that though.

soloam:
Do you want me to change the test? Or can we go like this?

PR #3020, opened 2021-05-29
[add] changelog to release

ksurl:
### Motivation for changes:

add commit details to changelog on github releases (based on this [action](https://github.com/Bullrich/generate-release-changelog))

### Detailed changes:
- Adds a step after the release script to generate a changelog. This uses `git tag` to get the previous version and 2 commits back from HEAD (ignore dev and current release commits) to get the relevant commits.
- Uses generated output from changelog step to fill body field in release action

### Needs work:
- testing
- maybe some nicer formatting.


gazpachoking:
Hmm. Would be nice if it pulled it from our semi-curated wiki page with changelog. Maybe this is better than nothing though, as that's a bit more complicated.

ksurl:
I tried reusing the existing python script but timing is an issue. I tried pulling a local copy after it updates to changelog.md file and then running the script to save the changes to a new file since the release action supports pointing to an md file for the body via bodyFile field. Maybe need to pull a before and after. Also the script needs to be modified to only save the diff. It was simpler to use tags as delimiters and commits for body directly. Maybe I'll find a new action to base it off of that's more polished on the formatting and robust in detecting versions.

gazpachoking:
> What if instead it just inserts a link to the Changelog on the wiki at the tagged release (usually the first one but if going to a previous version it should link to that header)

That's a pretty good idea, and should be fairly easy to implement, as the html ref ids are just based on the version number and the date, so it shouldn't need to do any fancy wiki access to generate the links.

ksurl:
Alright let me see what I can whip up

ksurl:
Looks like it is loading the contents of ChangeLog.md with javascript so there aren't any anchors to link to. I can however link to the headers in the file on the GitHub repo at https://github.com/Flexget/wiki/blob/master/ChangeLog.md

gazpachoking:
The anchors are like this https://flexget.com/ChangeLog#31128-2021-05-25 It's the version number without dots, then the release date

gazpachoking:
Let's go with it, and fix/improve it later. 😄 Thanks!

ksurl:
Hmm, the nightly job ran and created a new release. But there is nothing in the release body. Is it because it has to merge it first, and it will show up on the next release?

PR #3026, opened 2021-06-01
[fix] emby: allow emby connect login

soloam:
### Motivation for changes:

Emby allows to login using a public proxy kwon as emby connect, this PR makes it possible to login to a server via the Emby Connect login

Also added the interface to use discover in from_emby

### Detailed changes:
- Add emby connect validation and connect endpoints
- Add interface with discover search in from_emby


gazpachoking:
Thanks!

PR #3030, opened 2021-06-03
[change] aria2: allow entry setting set

soloam:
### Motivation for changes:

Allow the use of path and filename from the entry to set into aria2

### Detailed changes:
- added the mapping of entry 'path' to aria2 options
- added the mapping of entry 'filename' to aria2 optioss
- added the option to retrieve the file extension from a url and append to the filename

```yaml
  test_aria2:
    set:
      path: "/home/downloader/test"
      filename: "{{movie_name}}_new"

    aria2:
      server: test
      secret: my_secrep
      path: /home/downloader/
```


flexget/plugins/clients/aria2.py
```diff
@@ -111,16 +115,48 @@ def add_entry(self, aria2, entry, config):
...
+        filename = entry.get('content_filename', config.get('filename', None))
+        if filename:
+            if config.get('add_extension', False):
+                logger.debug('Getting filename from `{}`', entry['url'])
+                url_info = None
+                try:
+                    ctx = ssl.create_default_context()
+                    ctx.check_hostname = False
+                    ctx.verify_mode = ssl.CERT_NONE
+                    url_info = urlopen(entry['url'], context=ctx)
```
gazpachoking, line 133: Hmm. Using urlopen here, not requests? Is this just to get the headers? A `requests.head` should do this without fetching the full content as well. If we use `task.requests` the user could manually turn off ssl certification with the `verify_ssl` plugin rather than leaving it off permanently here.

flexget/plugins/clients/aria2.py
```diff
@@ -111,16 +115,48 @@ def add_entry(self, aria2, entry, config):
...
+                except Exception as e:
+                    logger.warning('Not possible to retrive file info from `{}`', entry['url'])
+                    entry.fail('Not possible to retrive file info from `{}`', entry['url'])
+                    return
+
+                if url_info and url_info and 'content-disposition' in url_info.headers:
```
gazpachoking, line 139: `url_info and url_info`? 😝

soloam: ups!

flexget/plugins/clients/aria2.py
soloam: The problem, is that head returns 404 in several sites, I tested with a lot, and keep getting 404 in some of them! I can only make it work 100% with urlopen, requests.head sometimes fails, even if I make requests with postman with head (from what I read head is buggy in some servers). Is it too bad to check the request with "task.requests.verify" and if False add the "context=ctx"?

gazpachoking: Maybe if the HEAD request is broken, just do a normal requests.get (or task.requests.get)? I think using requests is preferable to urlopen no matter what.

soloam: But if I do a request.get I'll download the complete movie/show! No? the idea is just to get the name of the file to be able to make replaces and keep the extension! This is useful to rename url's when no information is provided regarding file name or extension on the url (very common)

gazpachoking: You can use this method to make sure not to read the body of the response with requests. https://docs.python-requests.org/en/master/user/advanced/#body-content-workflow

gazpachoking:
I think this is still waiting on those couple changes I commented earlier on?

stranger-danger-zamu:
@soloam I'd use aria2 over transmission so I'm willing to get this working sooner than later. Do you have examples of sites which are returning 404 on HEAD requests? Everything I've tested works so far, but I might just have a smaller pool of types sites I'm visiting. 

@gazpachoking I have a working version based on `requests.head` though I would want to get more tests in from soloam's 404'ing head requests. I can easily implement the body-content-workflow fix if needed, but I've yet to encounter the 404 error.

soloam:
@stranger-danger-zamu I started using aria2 more now and was planning on closing this issue to wrap up all the changes (at the moment I'm using a custom plugin that is the version on my repo, but changed the name to aria2x).

I'll review this PR and improve it so that it can be closed.

The 404 errors that I sometimes get are very sporadic, and strange, because sometimes I get the error, and if I call it gain it works. I'm going to change the code to use "get" to see if it improves! And probably the plugin should fail the entry if it's not able to return the file header (and it's required). So that the entry could be retried in a latter run!

soloam:
I did the changes and the PR is ready to approval! All is working using the task get, that should work with plugins like the ignore certificates!

soloam:
I added the ability to set the extension by entry property.

The config could be:

Get Extension from the original file and add it to the new name
```yaml
    aria2:
      server: "<< server >>"
      secret: "<< secret >>"
      path: "<< download folder>>"
      add_extension: yes
      filename: my_name
```

Force the extension to mkv
```yaml
    aria2:
      server: "<< server >>"
      secret: "<< secret >>"
      path: "<< download folder>>"
      add_extension: mkv
      filename: my_name
```

Force the extension to mkv with set
```yaml

    set:
      content_extension: "mkv"

    aria2:
      server: "<< server >>"
      secret: "<< secret >>"
      path: "<< download folder>>"
      filename: my_name
```

stranger-danger-zamu:
I would definitely change that name from `add_extension`/`set_extension` to something more explicit or spin that type of functionality out to something else (eg. a proper MIME type post processing step).

soloam:
Please explain better, or talk with me in discord!

I can change the name of the property, but I don't understand what you mean by change the functionality! 

I would avoid adding more functionality to this PR. It will only delay further the merge.

stranger-danger-zamu:
Ah, I meant moving the extension manipulation out of this. I would only have this download entry that it gets. Something else (like a different component/plugin/etc) should deal with the extension manipulation. Separation of responsibilities and concerns.

soloam:
Pulling the extension from the file is a must have. Most of the link's to download are coded, and provide little to no information regarding the file name or file type! The download plugin should be able to rename the file, and to do that it needs to pull the extension so that it's able to rename it and keep it's extension!

soloam:
@gazpachoking any comments on this?

flexget/plugins/clients/aria2.py
```diff
@@ -105,22 +114,60 @@ def on_task_output(self, task, config):
...
+        filename = entry.get('content_filename', config.get('filename', None))
+        add_extension = entry.get('content_extension', config.get('add_extension', False))
+
+        if filename:
+            if add_extension:
+                if isinstance(add_extension, bool):
+                    logger.debug('Getting filename from `{}`', entry['url'])
+                    content_disposition = None
+                    try:
+                        response = task.requests.get(entry['url'], headers=None, stream=True)
```
gazpachoking, line 138: If a request is opened with the stream parameter, it has to be closed manually. See https://docs.python-requests.org/en/latest/user/advanced/#body-content-workflow

flexget/plugins/clients/aria2.py
```diff
@@ -105,22 +114,60 @@ def on_task_output(self, task, config):
...
+                        content_disposition = response.headers['content-disposition']
```
gazpachoking, line 139: Like this I mean. Also, is 'content-disposition' guaranteed to be in the headers? Maybe a .get is better.
```suggestion
                        with task.requests.get(entry['url'], headers=None, stream=True) as response:
                            content_disposition = response.headers.get('content-disposition')
```

soloam: done @gazpachoking

gazpachoking:
LGTM, thanks!

PR #3032, opened 2021-06-04
[add] yaml: add yaml list

soloam:
### Motivation for changes:

Allow getting entries from a yaml file and save entries to a yaml file

### Detailed changes:
- Created yaml_list

**Config options**
```yaml
    
   # This
    yaml_list:
      fields: # Fields to save - optional
        - data1
        - data2
        - data3
      path: "/path/to/myyaml.yml"
   
   # Or This
    yaml_list: "/path/to/myyaml.yml"
```

### Example
```yaml
            yaml_list_create:
            disable: seen
            mock:
              - {'title':'My Entry 1','url':'mock://myentry1', 'data':'myentry1'}
              - {'title':'My Entry 2','url':'mock://myentry2', 'data':'myentry2'}
              - {'title':'My Entry 3','url':'mock://myentry3', 'data':'myentry3'}

            accept_all: yes
            list_add:
              - yaml_list: '/path/to/my/yaml_list1.yaml'

```


soloam:
Some extra info

- Added a component because I have planed also a possible yaml_list, a managed list in yaml
- This could be very useful to use with the configure_series, allowing to create series from a yaml and also save new info from that series, like alternate naming
- More advanced, but would fit like a glove with the new [parameterize](https://flexget.com/Plugins/Parameterize) plugin, allowing to create inputs for the plugin, from a yaml file without reloading the config

BrutuZ:
Like an `entry_list` on external files then?
Would the `yaml_list` replace those or be functionally different?

soloam:
I'm working on a yaml_list that should probably replace this!

flexget/components/managed_lists/lists/yaml_list.py
```diff
@@ -0,0 +1,292 @@
...
+        if not item:
+            return None
+
+        item = self._get_item_fields(item)
+        output_item = {}
+        output_item['title'] = title
+        output_item.update(item)
+
+        if 'url' not in output_item:
+            output_item['url'] = f'mock://{simplify(title).replace(" ","_")}'
```
gazpachoking, line 124: Maybe just use `urllib.parse.quote` here?

soloam: Done

flexget/components/managed_lists/lists/yaml_list.py
```diff
@@ -0,0 +1,293 @@
...
+                data[item] = json.loads(data[item])
+            except TypeError:
+                del data[item]
+
+    data.pop('_backlog_snapshot', None)
+
+    return data
+
+
+def simplify(text: str) -> str:
```
paranoidi, line 60: Unused? And why should serialization do some type of normalization to data too?

paranoidi:
I'm curious, what is the use case for this?

gazpachoking:
I was thinking this could be nicer if we could adapt our [json serializer](https://github.com/Flexget/Flexget/blob/develop/flexget/utils/serialization.py) that handles our custom types to also work with yaml. That way entries could round trip and retain fields like quality. It could probably work as is, but it might be nicer if it used yaml tags.

PR #3072, opened 2021-07-10
[add] Flood plugin

Winter:
### Motivation for changes:
Decided to move from ruTorrent to Flood and there was no existing way to add torrents via flexget to Flood. (Which is just a pretty front-end for rTorrent).

### Detailed changes:
- Created a simple plugin 

```yaml

flood:
  url: '...'
  username: '...'
  password: '...'
  directory: '...'
  tags:
    - ...
```

#### To Do:
- [x] Support saving jwt token to database
- [ ] Support for performing CRUD operations on Flood
- [ ] Support for downloading files through Flood

### DIFF
--- a/flexget/plugins/clients/flood.py
+++ b/flexget/plugins/clients/flood.py
@@ -1,130 +1,123 @@
-import os
 from loguru import logger
 from requests import Session
 from requests.exceptions import RequestException
 
 from flexget import plugin
 from flexget.event import event
-from flexget.utils.template import RenderError
-from os import error
 from requests import Session
 from loguru import logger
-import json
-import base64
+
+from flexget.entry import Entry
+from flexget.task import Task
 
 logger = logger.bind(name='flood')
+
+class FloodClient:
+    def __init__(self):
+        self.session = Session()
+        self.connected = False
+
+    def _request(self, method: str, path: str, **kwargs):
+        try:
+            return self.session.request(method, self.url + path, **kwargs)
+        except RequestException as e:
+            raise plugin.PluginError(f'Flood Request Exception: {e}')
+    
+    def authenticate(self, config):
+        self.url = config['url'].strip('/')
+
+        response = self._request('post', '/api/auth/authenticate', data={
+            'username': config['username'],
+            'password': config['password']
+        })
+
+        if response.status_code == 200:
+            self.connected = True
+            logger.debug('Successfully logged into Flood')
+        else:
+            self.connected = False
+            raise plugin.PluginError('Incorrect username or password')
+
+        return self
+    
+    def add_torrent_urls(self, **kwargs):
+        if not self.connected:
+            raise plugin.PluginError('Flood is not connected.')
+
+        # Add's a default 'start': True to the request
+        if not kwargs.get('start'):
+            kwargs['start'] = True
+        
+        response = self._request('post', '/api/torrents/add-urls', json=kwargs)
+
+        if response.status_code == 200:
+            logger.debug('Successfully added torrent(s).')
+        else:
+            # There's no sanity to the codes returned by Flood. 
+            # Both return "code": -32602 and both return status 500.
+            if response.text.find('the input is not a valid.') > 0:
+                raise plugin.PluginError('Not a valid torrent.')
+            elif response.text.find('Info hash already used by another torrent.') > 0:
+                logger.debug('Torrent has already been added to Flood.')
+            else:
+                raise plugin.PluginError(f'Failed to add torrent to Flood. Error {response.status_code}.')
+    
+    def add_torrent_files(self,  **kwargs):
+        return False
 
 class OutputFlood:
     schema = {
        'type': 'object',
         'properties': {
-            'host': {'type': 'string'},
-            'port': {'type': 'integer'},
+            'url': {'type': 'string'},
             'username': {'type': 'string'},
             'password': {'type': 'string'},
-            'directory': {'type': 'string'},
+            'path': {'type': 'string'},
             'tags': {'type': 'array'},
         },
         'additionalProperties': False,
     }
 
-    def __init__(self):
-        self.session = Session()
-        self.connected = False
+    def add_entry(self, config: dict, task: Task, entry: Entry):
+        kwargs = {}
 
-    def _request(self, method, url, **kwargs):
-        try:
-            return self.session.request(method, url, **kwargs)
-        except RequestException as e:
-            raise plugin.PluginError(f'Flood Request Exception: {e}')
-    
-    def authenticate(self, config):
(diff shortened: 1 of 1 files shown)


flexget/plugins/clients/flood.py
```diff
@@ -0,0 +1,131 @@
...
+        else:
+            raise plugin.PluginError('Failed to add torrent to Flood. Error {}'.format(response.status_code))
+
+    #@plugin.priority(120)
+    #def on_task_download(self, task, config):
+
+    @plugin.priority(135)
+    def on_task_output(self, task, config):
+        if task.accepted and self.authenticate(config):
+            self.add_torrent_urls(config, [entry['url'] for entry in task.accepted])
```
paranoidi, line 127: It would be better to pass entries, one might want to use some fields from it in the future like "tag".

flexget/plugins/clients/flood.py
```diff
@@ -0,0 +1,131 @@
...
+            'password': {'type': 'string'},
+            'directory': {'type': 'string'},
+            'tags': {'type': 'array'},
+        },
+        'additionalProperties': False,
+    }
+
+    def __init__(self):
+        self.session = Session()
+        self.connected = False
```
paranoidi, line 33: It would be best not to have anything in the plugin init. This will cause creating unnecessary Sesson() objects even if flood plugin is never used. (Not that it's a huge issue, but still).

Global variables aren't good either. Can session expire between task executions, how does connected flag work in that case?

paranoidi (changes requested):
Somewhat messy. Keep up with the good work though :+1:

paranoidi:
Configuring http endpoints via separate "host" and "port" is unnecessarily inconvenient. Just use "url: https://somehost:1234"

Winter:
@paranoidi I made the requests changes. I ended up just going and making a FloodClient implementation that way it can just re-authenticate every time it adds a torrent. Which, probably could add a timeout before initializing a new instance of FloodClient and re-authenticating as the JWT token doesn't expire for 7 days.

flexget/plugins/clients/flood.py
```diff
@@ -0,0 +1,124 @@
...
+        else:
+            self.flood.add_torrent_urls(**kwargs)
+
+    @plugin.priority(135)
+    def on_task_output(self, task: Task, config: dict):
+        # If we don't have any accepted entries, then let's just stop (possibly redundant?)
+        if not task.accepted:
+            return
+
+        # We're re-authenticating ourselves every time we output
```
cvium, line 115: That seems excessive to me.

> On successful authentication via /authenticate API endpoint, Flood will send a cookie with token to user.

https://github.com/jesec/flood/blob/master/server/util/authUtil.ts#L8

The JWT authentication cookie has a 1 week expiration.

Winter: I did mention that above, I’ll probably do something about that tonight.

flexget/plugins/clients/flood.py
```diff
@@ -0,0 +1,124 @@
...
+            else:
+                kwargs['tags'] = [ entry.render(tags) ]
+
+        if task.manager.options.test:
+            logger.info('Flood Test Mode')
+            logger.info('Would add torrent to Flood with:')
+            logger.info('\tPath: {}', path)
+            logger.info('\tTags: {}', tags)
+        else:
+            self.flood.add_torrent_urls(**kwargs)
```
paranoidi, line 107: Using `**kwargs` in python is pretty annoying. I'd go as far as to claim that user written code should not use it at all. Only frameworks.

Please consider refactoring this so that the argument content is clearly defined and typed. Type hinting is pretty good way to document signatures too.

`**kwargs` is as ambiguous as it can be

flexget/plugins/clients/flood.py
```diff
@@ -0,0 +1,124 @@
...
+            # Both return "code": -32602 and both return status 500.
+            if response.text.find('the input is not a valid.') > 0:
+                raise plugin.PluginError('Not a valid torrent.')
+            elif response.text.find('Info hash already used by another torrent.') > 0:
+                logger.debug('Torrent has already been added to Flood.')
+            else:
+                raise plugin.PluginError(f'Failed to add torrent to Flood. Error {response.status_code}.')
+    
+    def add_torrent_files(self,  **kwargs):
+        return False
```
paranoidi, line 66: Seems pointless method

flexget/plugins/clients/flood.py
```diff
@@ -0,0 +1,124 @@
...
+            self.flood.add_torrent_urls(**kwargs)
+
+    @plugin.priority(135)
+    def on_task_output(self, task: Task, config: dict):
+        # If we don't have any accepted entries, then let's just stop (possibly redundant?)
+        if not task.accepted:
+            return
+
+        # We're re-authenticating ourselves every time we output
+        self.flood = FloodClient().authenticate(config)
```
paranoidi, line 116: It's not a good idea to assign new members outside `__init__`, meaning. You should pass the instance as an argument instead of sidechaining it via self.

The correct way would be to declare this in `__init__` as None, but like I said. We are not fan of those in plugins, so passing it is preferable.

paranoidi (changes requested):
Getting better :+1:

flexget/plugins/clients/flood.py
```diff
@@ -0,0 +1,124 @@
...
+from loguru import logger
+
+from flexget.entry import Entry
+from flexget.task import Task
+
+logger = logger.bind(name='flood')
+
+class FloodClient:
+    def __init__(self):
+        self.session = Session()
```
Trim21, line 17: better to set trust_env to false to avoid Session using proxy environment variables

Winter: Sorry for the late reply, but I can't find `trust_env` being used anywhere else in Flexget. Unless there's a particular reason for it that I'm not aware of?

Winter:
@paranoidi sorry for going MIA for a couple of months, but I've made the requested changes. Apart from storing the JWT is there any other changes?

flexget/plugins/clients/flood.py
crawln45: Personally I would not bother with trust_env.
You can read more about its usage [requests.trust_env
](https://docs.python-requests.org/en/latest/api/#requests.Session.trust_env)
It's a requests options for session to disable proxy usage. I don't see this being much of an issue, I also don't remember any other plugin using this setting during configuration. If such a setting is warranted then it should be applied to the Session wrapper currently configured in [Flexget Request](https://github.com/Flexget/Flexget/blob/develop/flexget/utils/requests.py). Usually this is the request wrapper we use in flexget as there is a domain limiting controls, cookie injections, ect.

PR #3085, opened 2021-07-19
[change] `series_forget`: Episode identifier is now optional. Optional `forget` parameter (originally would just remove)

BrutuZ:
### Motivation for changes:
Was looking for a way to remove a show from `series` plugin from within a task (akin to CLI command `flexget  series <remove | forget> <show_name> [episode]` and bumped into this undocumented plugin.
### Detailed changes:
- Brought its functionality more in line with the `flexget series <remove | forget>` CLI command:
  - Episode is now optional
  - Despite the plugin name, originally it would just remove the show, added option to forget.

```diff
  series_forget: yes
+ series_forget:
+   forget: yes
```
#### To Do:

- [ ] Document this (under series, Lists or on its own?)
- [x] Rename to `series_remove` to avoid confusion with the `forget` parameter? Shouldn't break many existing configs since it is not documented (yet).

### DIFF
--- a/flexget/components/series/series_forget.py
+++ b/flexget/components/series/series_forget.py
@@ -9,24 +9,47 @@ logger = logger.bind(name='series_forget')
 
 
 class OutputSeriesRemove:
-    schema = {'type': 'boolean'}
+    schema = {
+        'oneOf': [
+            {'type': 'boolean'},
+            {'type': 'object', 'properties': {'forget': {'type': 'boolean'}}},
+        ],
+    }
 
     def on_task_output(self, task, config):
         if not config:
             return
+        if isinstance(config, bool):
+            config = {'forget': False}
+        forget = config['forget']
+
         for entry in task.accepted:
-            if 'series_name' in entry and 'series_id' in entry:
-                try:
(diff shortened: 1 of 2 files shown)


BrutuZ:
Post-Dependabot bump

gazpachoking:
Looks good. I think I like the idea of renaming it series_remove, with the option for forget. That would complicate the boolean form of the config though..

BrutuZ:
Just noticed the plugin even registered itself as `series_remove` already on the file's last line. Will rename it then.
I don't see how it will complicate the simplest form of the config though

BrutuZ:
Mistyped the new name on the commit message, hope it's not a problem 😅

gazpachoking:
> I don't see how it will complicate the simplest form of the config though

Oh, I was thinking the bool form became the forget param. I see that's not the case though, so that doesn't complicate anything. Lemme give it another once over and I think we are good to go.

gazpachoking:
LGTM, thanks! Adding a wiki page would be great 😄

BrutuZ:
Documented on https://flexget.com/Plugins/series_remove
Should there also be an entry on https://flexget.com/UpgradeActions due to the renaming?

PR #3118, opened 2021-08-07
[add] torznab: More complete RSS XML entry data

rpatterson:
### Motivation for changes:
I needed access to the `size` child of the RSS `<item>` element like the `rss`
input plugin provides.

### Detailed changes:
While trying to address that I noticed what seemed
like an unnecessarily limited and hard-coded approach to extracting the
`<item>` elements children, so this change implements a more generalized
approach.

I would like to use a more generalized approach for the `<torznab:attr>`
children as well.  The current implementation is, however, sufficient for now
and ripping that out would be more invasive.

In fact, I'd love to see FlexGet use something like
[`xmltodict3`](https://github.com/dart-neitro/xmltodict3) wherever data is
being extracted from XML and inserted into entries so that using RSS or other
XML in FlexGet is consistent throughout.

### DIFF
--- a/flexget/plugins/input/torznab.py
+++ b/flexget/plugins/input/torznab.py
@@ -1,6 +1,7 @@
 from urllib.parse import urlencode
 from xml.etree import ElementTree
 
+import dateutil.parser
 from loguru import logger
 
 from flexget import plugin
@@ -18,6 +19,9 @@ class Torznab:
 
     Handles searching for tv shows and movies, with fallback to simple query strings if these are not available.
     """
+
+    ITEM_TAG_TYPES = dict(category=int, size=int, grabs=int)
+    ITEM_LIST_TAGS = {'category'}
 
     @property
     def schema(self):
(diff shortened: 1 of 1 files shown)


rpatterson:
I noticed [this same CI error](https://github.com/Flexget/Flexget/runs/3270932270) when trying to set up local development:

```
ERROR: Could not find a version that satisfies the requirement dataclasses==0.8 (from versions: 0.1, 0.2, 0.3, 0.4, 0.5, 0.6)
ERROR: No matching distribution found for dataclasses==0.8
```

That package isn't needed nor available for Python 3.7+, so I don't think those failures are related to this PR and I'm sue it's a known issue.

rpatterson:
> This PR is stale because it has been open 150 days with no activity. Remove stale label or comment or this will be closed in 60 days.

I'm still waiting for a response to [my question about the CI failures](https://github.com/Flexget/Flexget/pull/3118#issuecomment-894709450).

gazpachoking:
> I'm still waiting for a response to [my question about the CI failures](https://github.com/Flexget/Flexget/pull/3118#issuecomment-894709450).

Hmm, sorry about that. I'm not sure what would be going wrong there. Can you try rebasing or merging in current develop and seeing if it happens again? Seems like this is too old for me to see the test results.

rpatterson:
> Hmm, sorry about that. I'm not sure what would be going wrong there. Can you try rebasing or merging in current develop and seeing if it happens again? Seems like this is too old for me to see the test results.

Thanks for the followup!  I'm not sure when I'll be to get around to that, but I'll let you know when/if I do.

PR #3123, opened 2021-08-09
[add] transmission: Session data metainfo plugin

rpatterson:
### Motivation for changes:

Add Transmission session data, such as global settings, to each entry.

Useful to filter based on transmssion settings.  For example, don't add
new entries to transmission when a download limit is enabled.

### Detailed changes:

Not much to say here, it's pretty simple.  I copied the `PluginTransmissionInput` class and modified it to simply add the transmission session settings to each entry.  Working for me locally with the example from the docstring.

```
transmission_session:
  username: "myusername"
  password: "mypassword"
if:
  # Don't add items if downloading is paused
  - "transmission_speed_limit_down_enabled and transmission_speed_limit_down == 0":
      "reject"
  - "transmission_alt_speed_enabled and transmission_alt_speed_down == 0":
	    "reject"
```

### DIFF
--- a/flexget/plugins/clients/transmission.py
+++ b/flexget/plugins/clients/transmission.py
@@ -887,8 +887,82 @@ class PluginTransmissionClean(TransmissionBase):
             self.client.remove_torrent(remove_ids, config.get('delete_files'))
 
 
+class PluginTransmissionSession(TransmissionBase):
+    """
+    Add Transmission session data, such as global settings, to each entry.
+
+    Useful to filter based on transmssion settings.  For example, don't add
+    new entries to transmission when a download limit is enabled.
+
+    Example::
+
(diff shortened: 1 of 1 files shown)


paranoidi:
Interesting idea, this is just metadata plugin and should be named as such. I see check_subs breaking the naming convention. Not that the convention is grest as is ...

rpatterson:
> this is just metadata plugin and should be named as such

I don't see any clear example of how I should name it, @paranoidi:

    $ git --no-pager grep -n -e 'etadata' -- './flexget/plugins/**.py'

Can you clarify?

> I see check_subs breaking the naming convention

I don't see `check_subs` in my changes.  Can you clarify?

Thanks!

PR #3126, opened 2021-08-15
[fix] set Manipulate to ignore rejected entries

trananhtuan:
### Motivation for changes:
Currently `manipulate` processes rejected entries, which is problematic. I don't understand the reason for this in the first place.

For example, I want to use `manipulate` to replace `title` with `imdb_name`.  Some entries don't have valid imdb metadata, which causes `manipulate` to crash when trying to set `title` to `None`.  This could be solved by using `required_field: imdb_name` and setting `plugin_priority` for `manipulate` to lower than `required_field` but it doesn't work because `manupulate` will process rejected entries also. This causes the task to crash.

### Detailed changes:
set `manipulate` to ignore rejected entries


paranoidi:
I think this makes more sense.

PR #3144, opened 2021-09-05
[fix] update ettv download link selector

flakyton:
### Detailed changes:
- updated ettv download link selector


paranoidi:
Thanks!

PR #3179, opened 2021-10-10
fix docker build

trim21:
### Motivation for changes:

`python:3-alpine` become `3.10` after python3.10 is released.

But `greenlet==1.0.0` is broken in python3.10, so current dockerfile is not buildable anymore

### Detailed changes:
- set base docker image to `python:3.9-alpine`


Trim21:
Can we have a GitHub workflow to test docker image building?

PR #3180, opened 2021-10-11
ci: build and push docker image to github docker registry

Trim21:
### Motivation for changes:

flexget current has a `Dockerfile`, but no docker image available.

new workflow will release a docker image as `ghcr.io/flexget/flexget` on [GitHub Container Registry](https://docs.github.com/en/packages/working-with-a-github-packages-registry/working-with-the-container-registry)

- for `develop` branch, docker image will be tagged as `ghcr.io/flexget/flexget:develop`
- for a tag `v1.2.3`, docker image will be tagged as `1.2.3`, `1.2`, `1` and  `latest`

On pull-request, a docker image will be built but won't be pushed to registry.

### Detailed changes:
- add a ci to build and push docker image

### DIFF
--- a/.github/workflows/docker-image.yml
+++ b/.github/workflows/docker-image.yml
@@ -1,0 +1,52 @@
+name: Docker Image CI
+
+on:
+  push:
+    branches:
+      - develop
+    tags:
+      - v*.*.*
+  pull_request:
+    branches:
+      - "*"
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+
+      - name: Build Base Docker Image
+        run: docker build --tag flexget:tmp .
+
+      - name: Login to GitHub Container Registry
+        uses: docker/login-action@v1
+        if: github.event_name == 'push'
+        with:
+          registry: ghcr.io
+          username: ${{ github.actor }}
+          password: ${{ github.token }}
+
+      - uses: actions/github-script@v5
+        if: github.event_name == 'push'
+        name: Docker Push
+        with:
+          debug: true
+          script: |
+            const ref = context.ref
+            const versions = []
+            if (ref.startsWith('refs/tags/v')){
+              const tag = ref.substring(11)
+              const [major, minor, _] = tag.split('.')
+              versions.push('latest', tag, `${major}.${minor}`, major)
+            } else if (ref.startsWith('refs/heads/')) {
+              versions.push(ref.substring(11))
+            } else {
+              console.log(ref)
+            }
+            console.log(versions)
+            for (const version of versions) {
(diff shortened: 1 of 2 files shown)


Trim21:
ping

paranoidi:
This looks good to me!

paranoidi:
Our dockerfile is pretty basic though, should we fix that at the same time?

Trim21:
> Our dockerfile is pretty basic though, should we fix that at the same time?

current Dockerfile looks good to me.

gazpachoking:
Should this run on pushes to the master branch instead of develop? That branch only gets pushed to on releases

Trim21:
> Should this run on pushes to the master branch instead of develop? That branch only gets pushed to on releases

this will also run on every tag, there's no need to run on master branch.

gazpachoking:
> this will also run on every tag, there's no need to run on master branch.

I meant master branch instead of develop branch, not in addition to.

Trim21:
> > this will also run on every tag, there's no need to run on master branch.
> 
> I meant master branch instead of develop branch, not in addition to.

develop branch is only tagged as `ghcr.io/flexget/flexget:develop`

gazpachoking:
> develop branch is only tagged as `ghcr.io/flexget/flexget:develop`

I'm not a docker user myself, not sure what that means? My thoughts were just that it should update the docker image only on the daily stable releases to the master branch, not every push to develop branch.

Trim21:
> > develop branch is only tagged as `ghcr.io/flexget/flexget:develop`
> 
> I'm not a docker user myself, not sure what that means? My thoughts were just that it should update the docker image only on the daily stable releases to the master branch, not every push to develop branch.

if a user try to run docker image `ghcr.io/flexget/flexget`, he will get `ghcr.io/flexget/flexget:latest`, which is build from the latest tag.

Only if a user want to run a developing docker image, he can pull `ghcr.io/flexget/flexget:develop` and run a image build from develop branch. otherwise he is using the latest stable release. This branch is released as non-stable image.

gazpachoking:
Gotcha. That seems reasonable to me, we can always adjust if needed later. Given I'm not super up on docker myself, I hope you can pop back in if there are any docker issues in the future!
Thanks!

BrutuZ:
I personally don't really see much point in that image variant on a project with nightly builds on the master branch. The latest and development images will be the same most of the time (within 24h)

gazpachoking:
@Trim21 Hmm, seems there is an issue https://github.com/Flexget/Flexget/runs/4174172448?check_suite_focus=true

I'm guessing it's the
```
const dst = `ghcr.io/${context.actor.toLowerCase()}/flexget:${version}`
```
Probably should just be hardcoded to `flexget` rather than `context.actor`?

Trim21:
> Probably should just be hardcoded to `flexget` rather than `context.actor`?

yes, you are right

Issue #2431, opened 2019-08-19
`series list` crashes depending on terminal size


fongd:
`flexget series list` crashes if the terminal window is narrow enough that at least two columns get removed from output.

### Expected behaviour:

Issuing the command should provide a table of all tracked series and corresponding information for each of them. Columns should be selectively removed from output if the terminal window is too narrow to accommodate all of them.

### Actual behaviour:

If the terminal window is wide enough such that `flexget series list` does not have to remove any columns or only one column has to be removed, the command works as expected:

<pre>
root@bf66f703e948:/config# flexget series list
There is a FlexGet process already running for this config, sending execution there.
Least important column `Downloaded` removed due terminal size

┌─────────────────────────────────────────┬────────┬──────────────────┬──────────┬───────────────┐
│ Name                                    │ Begin  │ Last Encountered │ Age      │ Identified By │
├─────────────────────────────────────────┼────────┼──────────────────┼──────────┼───────────────┤
│ Atlanta                                 │ -      │ S01E10           │ 1020d    │ ep            │
…
</pre>

However, if the terminal window is narrow enough such that two columns need to be removed, Flexget crashes:

<pre>
root@bf66f703e948:/config# flexget series list
There is a FlexGet process already running for this config, sending execution there.
Least important column `Downloaded` removed due terminal size
Least important column `Age` removed due terminal size
Traceback (most recent call last):
  File "/usr/bin/flexget", line 10, in <module>
    sys.exit(main())
  File "/usr/lib/python3.7/site-packages/flexget/__init__.py", line 47, in main
    manager.start()
  File "/usr/lib/python3.7/site-packages/flexget/manager.py", line 381, in start
    client.handle_cli(self.args)
  File "/usr/lib/python3.7/site-packages/rpyc/core/netref.py", line 199, in __call__
    return syncreq(_self, consts.HANDLE_CALL, args, kwargs)
  File "/usr/lib/python3.7/site-packages/rpyc/core/netref.py", line 75, in syncreq
    return conn.sync_request(handler, proxy, *args)
  File "/usr/lib/python3.7/site-packages/rpyc/core/protocol.py", line 471, in sync_request
    return self.async_request(handler, *args, timeout=timeout).value
  File "/usr/lib/python3.7/site-packages/rpyc/core/async_.py", line 97, in value
    raise self._obj
TypeError: '<' not supported between instances of 'NoneType' and 'int'

========= Remote Traceback (1) =========
Traceback (most recent call last):
[...]

PR #3183, opened 2021-10-13
Remove unmaintained terminal libraries in favor of rich

gazpachoking:
### Motivation for changes:
`terminaltables` and `colorclass` libraries are no longer maintained, and at least colorclass isn't compatible with python 3.10

### Detailed changes:
- strips out usage of those two libraries, refactor our table and color support to work using [`rich`](https://github.com/willmcgugan/rich)
- Reworked how some of the console capturing and IPC transport works to be compatible with colors/tables
- Removed the support for dropping of table columns dynamically based on width of terminal. New library seems to more smartly handle column size and wrapping/truncating values.
- Also loses the 'auto' color ability, which tried to make sure colors were visible in all terminals. I couldn't find a good way to replace/replicate that, I think we just need to be smart about where/what colors we use. If there are reports of particular issues we can think of some solutions.
- New tables require strings (or some other thing directly renderable by rich) for all cell values. We have to explicitly cast integers or other types now.

### Addressed issues:
- Maybe addresses #2431 not sure if that was still a problem

#### To Do:

- [ ] Make sure all the plugins using the tables are updated
- [ ] Check console capturing works from API
- [x] something seems to have broken the log coloring with this 😬

### DIFF
--- a/flexget/terminal.py
+++ b/flexget/terminal.py
@@ -111,10 +111,6 @@ class TerminalTable(rich.table.Table):
                 yield rich.segment.Segment.line()
 
 
-class TerminalTableError(Exception):
-    """A CLI table error"""
-
-
 table_parser = ArgumentParser(add_help=False)
 table_parser.add_argument(
     '--table-type',


liiight:
Cool. I'll jump on some of the plugins if you want

gazpachoking:
I'm a little torn on whether to have the plugins just use rich's table class more raw rather than keep the wrapper now that it doesn't do much. It could just be a helper function that takes the cli table options and generates some kwargs for rich table, if even that. @liiight Opinions?

liiight:
well, if we could keep the wrapper as is and switch the library underneath it, that'll be ideal i imagine, being far less time consuming and will easily allow us extendability in the future.

I am not well familiarized (yet) with Rich's API so I'm not sure how much of a redundant boilerplate that will be.

PR #3203, opened 2021-10-20
jsonschema-4.0 removed jsonschema.compat

thermitegod:
### Motivation for changes:
jsonschema starting with 4.0 removed jsonschema.compat

### Detailed changes:
replace str_types with str
replace int_types with int


gazpachoking:
Excellent, thanks!

PR #3212, opened 2021-10-30
[fix] Correct from_piratebay torrent ranks

trananhtuan:
### Motivation for changes:
- Bugfix.
### Detailed changes:
- 

### Addressed issues:
- Fixes the incorrect torrent ranks which causes from_pirateby to crash when it encounters an unknown rank (such as `moderator` or `helper`).


gazpachoking:
Thanks!

PR #3226, opened 2021-11-10
Use name instead of original_name

soloam:
### Motivation for changes:

Use name instead of original_name in thetv_db api

### Detailed changes:
- Changed original_name by name


soloam:
Sorry, git commit messed my commit

soloam:
So all is ok, bot doing

PR #3228, opened 2021-11-10
Use TheTv DB to gt series name

soloam:
### Motivation for changes:

Some shows have different names in emby, and emby api does not handle ok the translation, so we use the tv db api to try to retrieve the shows thetvdb_id and use it to query emby

### Detailed changes:
- Added the thetvdb_api search series to try to get the show

#### To Do:

- [ ] Should only be merged after merging #3226


soloam:
Close to open a new one

PR #3237, opened 2021-11-16
[change] yaml_list: sort yaml, key and encoding

soloam:
### Motivation for changes:

- Sort yaml output: To make it more readble, the output is sorted by keys, but keeps title and key fields on the top
- yaml encoding: Allow to inform the file encoding, making it usable in other languages

### Detailed changes:
- Sort yaml output: sort the fields and add title and key to the top, add sort_keys FALSE to the yaml_dump
- yaml encoding: add encoding to the config, default utf-8 and add that encoding to the yaml_dump

### DIFF
--- a/flexget/components/managed_lists/lists/yaml_list.py
+++ b/flexget/components/managed_lists/lists/yaml_list.py
@@ -18,12 +18,12 @@ logger = logger.bind(name=PLUGIN_NAME)
 
 
 class YamlManagedList(MutableSet):
-    def __init__(self, path: str, fields: list, key: str, update: bool, encoding: str):
+    def __init__(self, path: str, fields: list, update: bool, encoding: str):
         self.filename = path
         self.fields = fields
         self.ecoding = encoding
         self.update = update
-        self.key = key
+        self.key = 'title'
         self.entries = []
         try:
             content = open(self.filename)
@@ -183,7 +183,6 @@ class YamlList:
                     'path': {'type': 'string'},
                     'fields': {'type': 'array', 'items': {'type': 'string'}},
                     'encoding': {'type': 'string', 'default': 'utf-8'},
-                    'key': {'type': 'string', 'default': 'title'},
                     'update': {'type': 'boolean', 'default': False},
                 },
                 'required': ['path'],
@@ -197,7 +196,6 @@ class YamlList:
             config = {'path': config}
         config.setdefault('fields', [])
         config.setdefault('encoding', 'utf-8')
-        config.setdefault('key', 'title')
         config.setdefault('update', False)
         return config
 
--- a/flexget/components/managed_lists/lists/yaml_list.py
+++ b/flexget/components/managed_lists/lists/yaml_list.py
@@ -59,7 +59,9 @@ class YamlManagedList(MutableSet):
             dict: Item with limited keys
         """
 
+
         # Title should allways be the first item
+        required_fields = []
         if self.key != 'title':
             required_fields = [self.key]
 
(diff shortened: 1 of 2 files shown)


gazpachoking:
How exactly might the `key` option be used?

soloam:
The key is what would decide if a entry is new or not!

In the entry_list a very useful functionality is the update of fields with a new entry, this was not the case here! A duplicate entries would be discarded. Based on feedback I noticed that this would be a useful thing to have, so I added the update. While doing that I noticed that managing the key would be useful, to, for example, if you have a list of series, they would be useful to make unique based on the id, and not the title.

gazpachoking:
> In the entry_list a very useful functionality is the update of fields with a new entry, this was not the case here! A duplicate entries would be discarded. Based on feedback I noticed that this would be a useful thing to have, so I added the update. While doing that I noticed that managing the key would be useful, to, for example, if you have a list of series, they would be useful to make unique based on the id, and not the title.

Hmm, making this more similarly to entry_list I think is good, so doing an update rather than discarding sounds good to me. The second bit I'm not so sure of. The rest of flexget is considering entries unique based on their title and url, I'm not so sure it should change here.

soloam:
The main reason is that this list will be managed by user's directly outside of flexget, that's the main use case! And when they do that, having a title that's not very user friendly only to be matched, it's not very intuitive! So the ideia was to keep the title as useful as possible, so that the user knows what he's missing with! 

But for me the main changes are the file encoding (that is making this unusable) and the update! 

So if you what me to remove the key to be pushed, I'll do it!

gazpachoking:
I'm not sure I'm against the 'key' thing, but separating out more bugfixy things from new features definitely makes merging faster/easier. Encoding in utf-8 (with optional override), and updating existing items are definitely 👍 from me.

soloam:
I kinda think it makes sense to let the user decide (title is default) what the key is! Probably not in all the list, but since this is managed outside flexget, it kinda makes sense to let the user have that control if he needs it!

The example of the serie for me is the best! I don't care the name of the serie or if it's translation or not! I would map it by tmdb Id or trakt id. The title would be what ever I like it to be

gazpachoking:
Like I said, I'm not necessarily against it, but putting a bunch of separate changes into one PR makes it much tougher to review and get the more critical changes merged in a timely manner. I still have to think about the implications of how the 'key' feature would work, and I'm not convinced on the naming of that option either. I'd like that discussion to be in a separate PR than the other changes though. Generally, the more separate smaller PRs that can be made the easier it is to decide. (I know it can be tough when you want to do multiple changes to the same plugin though, due to merging issues if you work on the separate branches at the same time.)

soloam:
Ok, @gazpachoking I'll split it!

paranoidi:
Key is also quite ambiguous term, we have talked making some sort of unique content identifier probably close to a decade by now :D

gazpachoking:
> I updated the PR to remove the key from the config... If you ask me, this should be added to all the lists, probably in the interface

Possibly, but we have other parts of the code that already expect, or require titles/urls to be unique. (Most notably seen plugin, which is enabled by default.) The seen plugin can also be configured by an user to enforce the uniqueness of other fields, like tmdb_id.

soloam:
I see! We can think about this in a latter stage! For now this PR solves some issues that are damaging the usability of the plugin, that's the main priority to solve

gazpachoking:
Okay, I rewrote your fixes/improvements a bit here https://github.com/Flexget/Flexget/compare/fcd16ff7b82e...f7acbc8ea650
I hope you don't feel bad that I took over your pull request, everything is based on your work. I just ended up implementing it a bit different while I was testing the behavior.
Thanks for this!

PR #3256, opened 2021-11-26
#3250 [fix] capture sequence type episodes with 4 digits

### LINKED ISSUE #3250: Series plugin unable to parse episode sequence numbers over 3 digits in length
### Expected behaviour:
Flexget series plugin should be able to parse an episode number with over 3 digits in length

### Actual behaviour:
The following error occurs and the episode is not recognised as being a part of the series:
> 2021-11-23 00:04:51 INFO     parser_internal nyaa.si - judas Title `[Judas] One Piece - 1000 [1080p][HEVC x265 10bit][Multi-Subs] (Weekly)` looks like series `One Piece` but cannot find any series numbering.

### Steps to reproduce:
- Have config which references a series entry that has over 999 episodes from RSS
- Run flexget (I personally run in daemon mode, but probably not relevant)
- Episode not recognised

#### Config:
[flexget_config.txt](https://github.com/Flexget/Flexget/files/7585172/flexget_config.txt)

#### Log:
<details>

```
2021-11-23 00:04:51 INFO     parser_internal nyaa.si - judas Title `[Judas] One Piece - 1000 [1080p][HEVC x265 10bit][Multi-Subs] (Weekly)` looks like series `One Piece` but cannot find any series numbering.
```
Let me know if you want the entire log.  None of the log seems interesting.
</details>

### Additional information:

- FlexGet version:
```
3.1.153
You are on the latest release.
```
- Python version: `Python 2.7.18`
- Installation method:  Of flexget?  Via `pip`
- Using daemon (yes/no): Yes
- OS and version:
```
Linux wilko-server 5.4.0-89-generic #100-Ubuntu SMP Fri Sep 24 14:50:10 UTC 2021 x86_64 x86_64 x86_64 GNU/Linux
```
```
No LSB modules are available.
Distributor ID: Ubuntu
Description:    Ubuntu 20.04.3 LTS
Release:        20.04
Codename:       focal
```
- Link to crash log: N/A

Is the problem line here: 
https://github.com/Flexget/Flexget/blob/develop/flexget/components/series/next_series_episodes.py#L46

astrotee:
### Motivation for changes:
+999 sequence type episodes can't be captured (One Piece just reached its 1000 ep, but I don't think it's the only one with more than 999 episodes)
### Detailed changes:
- made the maximum number of digits 4 instead of 5 (don't think there's +9999 episodes series, but I could be wrong!)

### Addressed issues:
- Fixes #3250 

#### To Do:

- check the regex of other types too

### DIFF
--- a/flexget/utils/parsers/series.py
+++ b/flexget/utils/parsers/series.py
@@ -87,7 +87,7 @@ class SeriesParser(TitleParser):
         [
             TitleParser.re_not_in_word(regexp)
             for regexp in [
-                r'(\d{1,3})(?:v(?P<version>\d))?',
+                r'(\d{1,4})(?:v(?P<version>\d))?',
                 r'(?:pt|part)\s?(\d+|%s)' % roman_numeral_re,
             ]
         ]


BrutuZ:
I'm guessing the 3 digits were to prevent possible conflicts with year being identified as episode number?
Unrelated note: Detective Conan passed 1000 episodes a while ago :P

astrotee:
> I'm guessing the 3 digits were to prevent possible conflicts with year being identified as episode number?

Could be, would have to consider that too?
Maybe they could be captured and removed like quality codes and other conflicting/unwanted parts?

BrutuZ:
I'm just guessing that's the reason. Might be worth it to include a test for it (if one doesn't exist already)

gazpachoking:
> I'm guessing the 3 digits were to prevent possible conflicts with year being identified as episode number?

Yeah, I think this was our original reason, years and `1080` can trip it up. I guess it's a good sign that the test suite passed on this PR, maybe those fears are unwarranted.

PR #3263, opened 2021-12-03
use native pip caching

ksurl:
### Detailed changes:
- updated main and release workflows to use native pip caching


gazpachoking:
Thanks!

PR #3271, opened 2021-12-06
[add] qbittorrent: Pass tag(s) when adding torrents to client

whiteout12:
### Motivation for changes:
Qbittorrent has a tag feature which is not yet supported by Flexget's qbittorrent-plugin.
### Detailed changes:
- Added Tags to be passed by Qbittorrent-plugin to Qbittorrent. 
- Renders entry specific tags as well as static tags from config.

Optional config if tags are to be used.
```
templates:
  qbt:
    qbittorrent:
      tags:
        - Flexget
tasks
  Testtask:
    template:
      - qbt
    set:
      tags:
        - '{{quality.resolution}}'
        - Test
```
```
2021-12-06 21:30:05 INFO     qbittorrent   test-qbt        Would add torrent to qBittorrent with:
2021-12-06 21:30:05 INFO     qbittorrent   test-qbt        File: None
2021-12-06 21:30:05 INFO     qbittorrent   test-qbt        Label: None
2021-12-06 21:30:05 INFO     qbittorrent   test-qbt        Tags: Flexget,1080p
2021-12-06 21:30:05 INFO     qbittorrent   test-qbt        Paused: true
```

### DIFF
--- a/flexget/plugins/clients/qbittorrent.py
+++ b/flexget/plugins/clients/qbittorrent.py
@@ -7,7 +7,7 @@ from requests.exceptions import RequestException
 
 from flexget import plugin
 from flexget.event import event
-from flexget.utils.template import RenderError, render
+from flexget.utils.template import RenderError
 
 logger = logger.bind(name='qbittorrent')
 
@@ -228,7 +228,6 @@ class OutputQBitTorrent:
             if label:
                 form_data['label'] = label  # qBittorrent v3.3.3-
                 form_data['category'] = label  # qBittorrent v3.3.4+
-
             try:
                 tags = entry.render(",".join([
                                     ",".join(config.get('tags', '')),
@@ -324,3 +323,4 @@ class OutputQBitTorrent:
 @event('plugin.register')
 def register_plugin():
     plugin.register(OutputQBitTorrent, 'qbittorrent', api_ver=2)
+    
--- a/flexget/plugins/clients/qbittorrent.py
+++ b/flexget/plugins/clients/qbittorrent.py
@@ -45,7 +45,7 @@ class OutputQBitTorrent:
                     'verify_cert': {'type': 'boolean'},
                     'path': {'type': 'string'},
                     'label': {'type': 'string'},
-                    'tags': {'type': 'array'},
+                    'tags': {'type': 'array', 'items': {'type': 'string'}},
                     'maxupspeed': {'type': 'integer'},
                     'maxdownspeed': {'type': 'integer'},
                     'fail_html': {'type': 'boolean'},
--- a/flexget/plugins/clients/qbittorrent.py
+++ b/flexget/plugins/clients/qbittorrent.py
@@ -7,7 +7,7 @@ from requests.exceptions import RequestException
 
 from flexget import plugin
 from flexget.event import event
-from flexget.utils.template import RenderError
+from flexget.utils.template import RenderError, render
 
 logger = logger.bind(name='qbittorrent')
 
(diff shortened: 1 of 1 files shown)


flexget/plugins/clients/qbittorrent.py
```diff
@@ -44,6 +45,7 @@ class OutputQBitTorrent:
                     'verify_cert': {'type': 'boolean'},
                     'path': {'type': 'string'},
                     'label': {'type': 'string'},
+                    'tags': {'type': 'array'},
```
gazpachoking, line 48: Should probably specify that the tags themselves have to be strings here.
```suggestion
                    'tags': {'type': 'array', 'items': {'type': 'string'}},
```

flexget/plugins/clients/qbittorrent.py
```diff
@@ -205,6 +207,7 @@ def prepare_config(config):
         config.setdefault('use_ssl', False)
         config.setdefault('verify_cert', True)
         config.setdefault('label', '')
+        config.setdefault('tags', '')
```
gazpachoking, line 210: tags will be a list if the user enters it, so it's probably nicer to keep it the same type in the default case.
```suggestion
        config.setdefault('tags', [])
```

flexget/plugins/clients/qbittorrent.py
```diff
@@ -225,6 +228,19 @@ def add_entries(self, task, config):
...
+                        [",".join(config.get('tags', '')), ",".join(entry.get('tags', ''))]
+                    ).strip(",")
+                )
+                if tags:
+                    form_data['tags'] = tags
+            except RenderError as e:
+                logger.error('Error rendering tags for {}: {}', entry['title'], e)
+                form_data['tags'] = ",".join(
+                    [",".join(config.get('tags', '')), ",".join(entry.get('tags', ''))]
+                ).strip(",")
```
gazpachoking, line 243: With most other settings, the field from the entry will override what was provided in the config, rather than augment it. I think we should keep this the same for consistency (unless you have a compelling reason otherwise.) Also, it's probably nicer to put an if clause around this whole thing, so that we don't try to do any rendering in the case where no tags were specified.

i.e. something like this to start off:
```python
tags = entry.get('tags', config.get('tags', []))
if tags:
    ...
```

gazpachoking (changes requested):
Looks pretty good to me! I left a few comments in there.

flexget/plugins/clients/qbittorrent.py
whiteout12: The reason I didn't want to override a config tag with an entry tag is simply because there can be multiple tags (as opposed to label och savepath that only have one). This gives the user the possibility to tag all torrents with something very general for every torrent from the config (eg "Flexget") but then also add specific tags from the entry.
Does it make sense?

BrutuZ: Unrelated note: Would it be cleaner to merge both tag lists before converting to string?
```python
form_data['tags'] = ",".join(config.get('tags', [])+entry.get('tags', []))
```

gazpachoking: > The reason I didn't want to override a config tag with an entry tag is simply because there can be multiple tags (as opposed to label och savepath that only have one). This gives the user the possibility to tag all torrents with something very general for every torrent from the config (eg "Flexget") but then also add specific tags from the entry.
Does it make sense?

Yeah, I'm okay with that.

I think we should still make the master list of tags, check if it's not empty, then start doing all the rendering and formatting.

gazpachoking: Similar to @BrutuZ  idea. Something like:
```python
tags = entry.get('tags', []) + config.get('tags', [])
if tags:
    ...
```

whiteout12: Ok thanks for input. Im then going to change to this:

```
            tags = entry.get('tags', []) + config.get('tags', [])
            if tags:
                try:
                    form_data['tags'] = entry.render(",".join(tags))
```
I Agree looks better so thanks the suggestion @BrutuZ

whiteout12:
I have made the changes from last review. Is there anything else I need to fix?

gazpachoking (approved):

gazpachoking:
Looks good to me, thanks!

PR #3273, opened 2021-12-07
[add] `anilist`: Include more metadata (`al_date_start`, `al_date_end`, `mal_id`)

BrutuZ:
### Motivation for changes:
Needed the airing dates (start and end) for additional filtering in my config.
Could also use other databases' IDs - AniDB in particular - to optimize Plex matching by automatically generating `anidb2.id` files for ASS\HAMA upon entry acceptance)

### Detailed changes:
- Additional fields on existing GraphQL request
- New request to API that crossmatches different Databases
- Renamed `al_idMal` field for more sensible `mal_id`

```diff
- {{al_idMal}}
+ {{mal_id}}
+ {{al_date_start}} {{al_date_end}}
```

### DIFF
--- a/flexget/plugins/input/anilist.py
+++ b/flexget/plugins/input/anilist.py
@@ -156,10 +156,10 @@ class AniList(object):
                                 raise plugin.PluginWarning(f'Couldn\'t fetch additional IDs - {e}')
 
                             entry = Entry()
-                            entry['al_id'] = anime.get('id') or ids.get('anilist')
+                            entry['al_id'] = anime.get('id', ids.get('anilist'))
                             entry['anidb_id'] = ids.get('anidb')
                             entry['kitsu_id'] = ids.get('kitsu')
-                            entry['mal_id'] = anime.get('idMal') or ids.get('myanimelist')
+                            entry['mal_id'] = anime.get('idMal', ids.get('myanimelist'))
                             entry['al_banner'] = anime.get('bannerImage')
                             entry['al_cover'] = anime.get('coverImage', {}).get('large')
                             entry['al_date_end'] = (
--- a/flexget/plugins/input/anilist.py
+++ b/flexget/plugins/input/anilist.py
@@ -118,112 +118,104 @@ class AniList(object):
                     'https://graphql.anilist.co',
                     json={'query': req_query, 'variables': req_variables},
                 )
-            except RequestException as e:
-                raise plugin.PluginError('Error reading list - {url}'.format(url=e))
-
-            try:
                 list_response = list_response.json()['data']
-                logger.debug('JSON output: {}', list_response)
-                for list_status in list_response.get('collection', {}).get('statuses', []):
-                    if selected_list_name and (
-                        list_status.get('name')
-                        and list_status.get('name').lower() not in selected_list_name
-                    ):
-                        continue
-                    for anime in list_status['list']:
-                        anime = anime.get('anime')
-                        has_selected_release_status = (
-                            (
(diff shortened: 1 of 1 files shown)


flexget/plugins/input/anilist.py
```diff
@@ -142,14 +144,45 @@ def on_task_input(self, task, config):
...
+                                ids = task.requests.post(
+                                    'https://relations.yuna.moe/api/ids',
+                                    json={'anilist': anime.get('id')},
+                                ).json()
+                            except RequestException as e:
+                                ids = {}
+                                raise plugin.PluginWarning(f'Couldn\'t fetch additional IDs - {e}')
+
                             entry = Entry()
+                            entry['al_id'] = anime.get('id') or ids.get('anilist')
```
paranoidi, line 159: Generally it's better to do
```
anime.get('id', ids.get('anilist'))
```

because using or depends how return value of "id" evaluates in boolean context, not if the field is present or not

paranoidi:
Looks good to me

flexget/plugins/input/anilist.py
BrutuZ: Well, I did get scolded once in another project for nesting and chaining `get()` so I "learned" to favor the `or` syntax 😅

TBF this fallback isn't even necessary, `id` is a required field (possibly the only one) as it's the index. This `or` ends up more like a documentation that the field is available from that second API due to the short-circuit, so I could just comment it instead.
`idMAL` on the other hand might not be present, but if it is the value is also an integer

paranoidi: Hahaa, yeah. I am always a bit torn on this subject. But I think the nesting is correct way as using or does not work as expected if value is 0 or "" for example.

flexget/plugins/input/anilist.py
```diff
@@ -142,14 +144,45 @@ def on_task_input(self, task, config):
...
                         ) or 'all' in selected_formats
+
                         if has_selected_type and has_selected_release_status:
+                            try:
+                                ids = task.requests.post(
+                                    'https://relations.yuna.moe/api/ids',
+                                    json={'anilist': anime.get('id')},
+                                ).json()
+                            except RequestException as e:
+                                ids = {}
```
paranoidi, line 155: Why set `ids` to empty dict here? It's not used ... ?

Also, exception chaining is generally a good idea.

(Besides this PR, holy batman that's a wide exception handing for ValueError)

paranoidi:
Feel free to merge this in any time, I think you have the rights :)

paranoidi (approved):

flexget/plugins/input/anilist.py
BrutuZ: Not completely unexpected, both `0` and `""` would be invalid after all xD

flexget/plugins/input/anilist.py
BrutuZ: If the request isn't successful and `ids` isn't set then the `ids.get()` calls below will return exceptions.

paranoidi: But you're raising PluginWarning immediately after .. which stops the whole plugin execution.

BrutuZ:
> holy batman that's a wide exception handing for ValueError

That never happened 🤐

BrutuZ:
> Feel free to merge this in any time, I think you have the rights :)

I do not. Usually @gazpachoking pushes the button that does the thing xD

Side-note: GitHub's diff view doesn't do any favors when lines only have indentation changes 👀

PR #3286, opened 2021-12-14
[fix] parsing: nested parsing select

soloam:
### Motivation for changes:

When parsing plugin is used nested (from_task for example), or for example in a global template, the selection is cleared in on_task_exit, returning to the internal by default 

### Detailed changes:
- added a nested control to the selection, so that it's only removes the running task config.

``` yaml
templates:
  global:
    parsing:
      movie: guessit
      series: guessit

tasks:
  task_input:
    disable:
      - remember_rejected
      - seen
      - retry_failed
      - seen_info_hash

    accept_all: yes

  get_guess_it:
    disable:
      - remember_rejected
      - seen
      - retry_failed
      - seen_info_hash

    mock:
      - { "title": "My.Series.S01E01.iNTERNAL.720p.x265-ZMNT" }

    configure_series:
      from:
        mock:
          - { "title": "My Series (2021)" }
      settings:
        parse_only: yes

```

### DIFF
--- a/flexget/components/parsing/plugin_parsing.py
+++ b/flexget/components/parsing/plugin_parsing.py
@@ -38,8 +38,6 @@ def init_parsers(manager):
 class PluginParsing:
     """Provides parsing framework"""
 
-    _selected = False
-
     @property
     def schema(self):
         # Create a schema allowing only our registered parsers to be used under the key of each parser type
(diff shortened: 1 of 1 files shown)


flexget/components/parsing/plugin_parsing.py
```diff
@@ -38,6 +38,8 @@ def init_parsers(manager):
 class PluginParsing:
     """Provides parsing framework"""
 
+    _selected = False
```
gazpachoking, line 41: What does this do? Why not just check if `selected_parsers` is empty, rather than keeping track separately? Otherwise it looks good to me.

soloam: I do this check because to avoid pop a entry on a task that didn't set it! Otherwise a inner task that does not set a config would delete the config from the upper one

gazpachoking: Ahh, understood. I'm not sure making it a class property helps though does it? If another task runs before the exit handler of the current task, this will still be True there, even though that task didn't set it, no?

gazpachoking: Maybe something like `task._custom_parser = True`, then do `getattr(task, '_custom_parser', False)` to check.

gazpachoking:
LGTM, thanks!

PR #3313, opened 2022-01-15
Fix new NPO Login for NPO Watchlist

mfonville:
NPO Start now uses a dedicated login service id.npo.nl for all Dutch broadcasting associations
Refreshed the tests

### Motivation for changes:
Fixes current broken NPO Watchlist plugin


gazpachoking:
Thanks!

PR #3329, opened 2022-01-26
Npo login fix

mfonville:
### Motivation for changes:

NPO added ?st=registered to identifying URLs, which we want to ignore.
Updated the logic to drop any query parameters in the identifier URL and fresh run of the tests


mfonville:
Could you please merge this fix? thanks!

gazpachoking:
Looks good, thanks!

PR #3336, opened 2022-02-02
[fix] decompress: check module exists

soloam:
### Motivation for changes:

Decompress plugin was not checking if the rarfile was installed, ignoring and keeping all entries accepted. This would accept all entries and the user would not be notified.

### Detailed changes:
- Check in task start if rarfile is installed
- Created a wrap method to do the check

#### To Do:

- [ ] Breaking change?


gazpachoking:
LGTM! Thanks!

PR #3343, opened 2022-02-03
[add] seen cli: add task options

soloam:
### Motivation for changes:

Adds new option to CLI seen to allow to search, forget, add specifying a task. This is very useful to maintain the entries via cli.

### Detailed changes:
- Added the --tasks option to search, this will search only on that specific tasks.
- search value input was changed and implicit adding of '%' was removed, to add a '%' to the like it should be added a '*', this was done to make the output of search uniform to the actions on forger. Allowing to test a command before removing the entry
- Added the --tasks option to forget, this will forget only on that specific tasks.
- Allow to add '*' to match any char in forget, the forgotten entries should be the same that would output with the same parameters in search.
- Added the --task option to add. This will add a local entry to that specific task

```yaml
tasks:
  seen1:
    mock:
      - { "title": "Title1", "url": "mock://local1" }
      - { "title": "Title2", "url": "mock://local2" }
      - { "title": "Title3", "url": "mock://local3" }
      - { "title": "Title4", "url": "mock://local4" }
    accept_all: yes

  seen2:
    mock:
      - { "title": "Title5", "url": "mock://local5" }
      - { "title": "Title6", "url": "mock://local6" }
      - { "title": "Title7", "url": "mock://local7" }
      - { "title": "Title8", "url": "mock://local8" }
    accept_all: yes
```
```sh
flexget seen add "TitleX" --task seen1
flexget seen add "TitleX" --task seen2
flexget seen search "Title*" --task seen1

flexget seen search "Title*"
```

```sh
flexget execute
flexget seen search "Title*"
```
| Field | Value            |
|-------|------------------|
| Title | Title1           |
| URL   | mock://local1    |
| Task  | seen1            |
| Added | 2022-02-03 10:52 |
|-------|------------------|
| Title | Title2           |
| URL   | mock://local2    |
| Task  | seen1            |
| Added | 2022-02-03 10:52 |
|-------|------------------|
| Title | Title3           |
| URL   | mock://local3    |
| Task  | seen1            |
| Added | 2022-02-03 10:52 |
|-------|------------------|
| Title | Title4           |
| URL   | mock://local4    |
| Task  | seen1            |
| Added | 2022-02-03 10:52 |
|-------|------------------|
| Title | Title5           |
| URL   | mock://local5    |
| Task  | seen2            |
| Added | 2022-02-03 10:52 |
|-------|------------------|
| Title | Title6           |
| URL   | mock://local6    |
| Task  | seen2            |
| Added | 2022-02-03 10:52 |
|-------|------------------|
| Title | Title7           |
| URL   | mock://local7    |
| Task  | seen2            |
| Added | 2022-02-03 10:52 |
|-------|------------------|
| Title | Title8           |
| URL   | mock://local8    |
| Task  | seen2            |
| Added | 2022-02-03 10:52 |

```sh
flexget seen add "TitleX" --task seen1
flexget seen add "TitleX" --task seen2
flexget seen search "TitleX"
```

| Field   | Value            |
|---------|------------------|
| Title   | TitleX           |
| CLI_ADD | TitleX           |
| Task    | seen1            |
| Local   | Yes              |
| Added   | 2022-02-03 10:55 |
|---------|------------------|
| Title   | TitleX           |
| CLI_ADD | TitleX           |
| Task    | seen2            |
| Local   | Yes              |
| Added   | 2022-02-03 10:55 |

```sh
flexget seen forget "TitleX" --tasks seen1
flexget seen search "TitleX"
```

| Field   | Value            |
|---------|------------------|
| Title   | TitleX           |
| CLI_ADD | TitleX           |
| Task    | seen2            |
| Local   | Yes              |
| Added   | 2022-02-03 11:38 |

#### To Do:

- [ ] Update wiki if merged

### DIFF
--- a/flexget/components/seen/cli.py
+++ b/flexget/components/seen/cli.py
@@ -1,6 +1,6 @@
 from flexget import options, plugin
 from flexget.event import event
-from flexget.manager import Manager, TaskNamingError
+from flexget.manager import Manager
 from flexget.terminal import TerminalTable, console, table_parser
 from flexget.utils.database import with_session
 
@@ -35,7 +35,7 @@ def seen_forget(manager: Manager, options):
         for task in options.tasks:
             try:
                 tasks.extend(m for m in manager.matching_tasks(task) if m not in tasks)
-            except TaskNamingError as e:
+            except ValueError as e:
                 console(e)
                 continue
 
@@ -44,7 +44,7 @@ def seen_forget(manager: Manager, options):
         forget_name = forget_name.replace("%", "\\%").replace("_", "\\_")
         forget_name = forget_name.replace("*", "%").replace("?", "_")
 
-    count, fcount = db.forget(forget_name, tasks=tasks, test=options.test_forget)
+    count, fcount = db.forget(forget_name, tasks=tasks, test=options.test)
     console(f'Removed {count} titles ({fcount} fields)')
     manager.config_changed()
 
@@ -98,7 +98,7 @@ def seen_search(manager: Manager, options, session=None):
         for task in options.tasks:
             try:
                 tasks.extend(m for m in manager.matching_tasks(task) if m not in tasks)
-            except TaskNamingError as e:
+            except ValueError as e:
                 console(e)
                 continue
 
@@ -135,10 +135,6 @@ def register_parser_arguments():
         metavar='TASK',
         help='forget only in specified task(s), optionally using glob patterns ("tv-*"). '
         'matching is case-insensitive',
-    )
-
-    forget_parser.add_argument(
-        '--test-forget', action='store_true', default=False, help='test forget query'
     )
 
     forget_parser.add_argument(
--- a/flexget/components/seen/cli.py
+++ b/flexget/components/seen/cli.py
@@ -16,9 +16,9 @@ def do_cli(manager, options):
     if options.seen_action == 'forget':
         seen_forget(manager, options)
     elif options.seen_action == 'add':
-        seen_add(options)
+        seen_add(manager, options)
     elif options.seen_action == 'search':
-        seen_search(options)
+        seen_search(manager, options)
 
 
 def seen_forget(manager, options):
@@ -28,12 +28,20 @@ def seen_forget(manager, options):
         if imdb_id:
             forget_name = imdb_id
 
-    count, fcount = db.forget(forget_name)
-    console('Removed %s titles (%s fields)' % (count, fcount))
+    tasks = manager.tasks_extractor(options.tasks)
+
+    # If tasks are specified it should use pattern matching as search
+    if tasks:
+        forget_name = forget_name.replace("*", "%").replace("?", "_")
+
+    count, fcount = db.forget(forget_name, tasks=tasks)
+    console(f'Removed {count} titles ({fcount} fields)')
     manager.config_changed()
 
 
-def seen_add(options):
+def seen_add(manager, options):
+    DEFAULT_TASK = 'cli_add'
+
     seen_name = options.add_value
     if is_imdb_url(seen_name):
         console('IMDB url detected, try to parse ID')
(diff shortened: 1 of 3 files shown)


crawln45:
I get the want to update this to be have a more regex look. But as this is not a regex evaluation I think it would be best to keep the sql functionality of "%". But that's my own thoughts. As this does not require a regex pattern but is the SQL wildcard expression.

soloam:
I convert to "%" before running the SQL, I think that "*" is more intuitive to the end user! I didn't think that adding complexity with regex would be good idea! If this gets merged and regex is something that would be useful, we can add extra tag to activate the regex mode! Keep the PR smaller and grow as needed! Lesson learned! 🤣

crawln45:
I didn't want users to confuse the usage of the regex "*" as that can lead to attempting to use regex in this field. While the % requires reading its usage.

soloam:
The idea was that user's already use '*' in things like the tasks, we can use "tv-*" for example! Nothing that a update to the wiki wouldn't solve

crawln45:
Yeah the difference is that the config uses yaml wildcard which is why the * wildcard usage is there. The cli uses the sql wildcard through the search function which is sql based and uses the appropriate character to match the usage. I'm not saying one way or the other. They both do the job, but they are different in usage. Sql search can use "%" and "_" for wildcards. While the yaml as far as I am aware of.

flexget/manager.py
```diff
@@ -1052,6 +1037,23 @@ def _shutdown(self) -> None:
         global manager
         manager = None
 
+    def tasks_extractor(self, tasks):
```
gazpachoking, line 1040: Not sure about the name... maybe `matching_tasks`? Also, let's add some type hints here.

Other questions I have, which I haven't decided my full stance:
- Should it raise an exception rather than logging one?
- Should it act on a single string rather than a list?

I'm leaning towards yes on both of those, and let the caller be responsible for looping and logging.

flexget/components/seen/cli.py
```diff
@@ -57,8 +79,11 @@ def seen_search(options, session=None):
         else:
             console("Could not parse IMDB ID")
     else:
-        search_term = '%' + options.search_term + '%'
-    seen_entries = db.search(value=search_term, status=None, session=session)
+        search_term = options.search_term.replace("*", "%").replace("?", "_")
```
gazpachoking, line 82: If we are switching the wildcard, should it automatically escape % and _ from the search so they end up searching for the literal characters?

gazpachoking:
I don't think we use the sql wildcards anywhere else, I'm good with removing them in favor of regular filename matching rules.

soloam:
I did the reviews @gazpachoking . Also added a new flag to forget to test (--test-forget) this will output some information regarding what would be forgotten and the reason:

```sh
2022-02-04 13:21:33 INFO     seen.db                       Testing: would forget entry with title `Title5` of task `seen2`
2022-02-04 13:21:33 INFO     seen.db                       Testing: would forget entry with title `Title6` of task `seen2`
2022-02-04 13:21:33 INFO     seen.db                       Testing: would forget entry with title `Title7` of task `seen2`
2022-02-04 13:21:33 INFO     seen.db                       Testing: would forget entry with title `Title8` of task `seen2`
2022-02-04 13:21:33 INFO     seen.db                       Testing: would forget entry with title `TitleX` of task `seen2`
2022-02-04 13:21:33 INFO     seen.db                       Testing: would forget entry `Title5` of task `seen2` based on field `title` with value `Title5`
2022-02-04 13:21:33 INFO     seen.db                       Testing: would forget entry `Title6` of task `seen2` based on field `title` with value `Title6`
2022-02-04 13:21:33 INFO     seen.db                       Testing: would forget entry `Title7` of task `seen2` based on field `title` with value `Title7`
2022-02-04 13:21:33 INFO     seen.db                       Testing: would forget entry `Title8` of task `seen2` based on field `title` with value `Title8`
2022-02-04 13:21:33 INFO     seen.db                       Testing: would forget entry `TitleX` of task `seen2` based on field `cli_add` with value `TitleX`
```

soloam:
> 

The only reason that I didn't used the regular test, was to avoid cloning the database to something that would be useful to see what would be deleted, also feedback some more info on way it would be removed (title or field)

soloam:
But if it's something that would be better with test I can remove the tag

flexget/components/seen/db.py
```diff
@@ -40,6 +42,8 @@
 logger = logger.bind(name='seen.db')
 Base = db_schema.versioned_base('seen', 4)
 
+ESPCAPE_QUERY = '\\'
```
gazpachoking, line 45:
```suggestion
ESCAPE_QUERY = '\\'
```

flexget/manager.py
```diff
@@ -1052,6 +1050,19 @@ def _shutdown(self) -> None:
...
+    def matching_tasks(self, task: str) -> 'list[str]':
+        """Create list of tasks to run, preserving order"""
+
+        # Consider * the same as not specifying tasks at all (makes sure manual plugin still works)
+        if not task or task == '*':
+            return []
+
+        task_names = [t for t in self.tasks if fnmatch.fnmatchcase(str(t).lower(), task.lower())]
+        if not task_names:
+            raise TaskNamingError(f'`{task}` does not match any tasks')
```
gazpachoking, line 1062: I think this could just be a ValueError

gazpachoking:
> The only reason that I didn't used the regular test, was to avoid cloning the database to something that would be useful to see what would be deleted, also feedback some more info on way it would be removed (title or field)

I'm inclined to just stick with our current option for consistency unless there is a problem. You should be able to check if that option is defined and alter the output though.

soloam:
> > The only reason that I didn't used the regular test, was to avoid cloning the database to something that would be useful to see what would be deleted, also feedback some more info on way it would be removed (title or field)
> 
> I'm inclined to just stick with our current option for consistency unless there is a problem. You should be able to check if that option is defined and alter the output though.

Done!

gazpachoking:
LGTM, thanks!

PR #3375, opened 2022-03-03
Lostfilm fix

Karlson2k:
### Motivation for changes:
When plugin is aborted by exception, then next scheduled run of plugin is not clean.
As the lostfilm site could be unavailable for one or another reason, it makes sense to just log error and return `None` instead of raising exception.

### Detailed changes:
Replaced exception with error in log.
Also updated the list of site mirrors.


gazpachoking:
Seems good to me, thanks!

Karlson2k:
Thanks!

PR #3377, opened 2022-03-06
fix cli templates error

Spider84:
  File "/usr/local/bin/flexget", line 8, in <module>
    sys.exit(main())
  File "/usr/local/lib/python3.9/dist-packages/flexget/__init__.py", line 44, in main
    manager.start()
  File "/usr/local/lib/python3.9/dist-packages/flexget/manager.py", line 383, in start
    self.handle_cli()
  File "/usr/local/lib/python3.9/dist-packages/flexget/manager.py", line 412, in handle_cli
    options.cli_command_callback(self, command_options)
  File "/usr/local/lib/python3.9/dist-packages/flexget/plugins/cli/templates.py", line 31, in list_file_templates
    table.add_row(name, plugin, template.filename, '')
  File "/usr/local/lib/python3.9/dist-packages/rich/table.py", line 423, in add_row
    raise errors.NotRenderableError(
rich.errors.NotRenderableError: unable to render list; a string or other renderable object is required

### DIFF
--- a/flexget/plugins/cli/templates.py
+++ b/flexget/plugins/cli/templates.py
@@ -19,8 +19,11 @@ def list_file_templates(manager, options):
         else:
             plugin = '-'
         name = template_name.replace('.template', '').split('/')
-        if len(name) == 2:
-            name = name[1]
+        if type(name) == list:
+            if len(name) > 0:
+                name = name[-1]
+            else:
+                name = '<ERR>'
         with open(template.filename) as contents:
             table_data.append([name, plugin, template.filename, contents.read()])
 


flexget/plugins/cli/templates.py
```diff
@@ -19,8 +19,11 @@ def list_file_templates(manager, options):
         else:
             plugin = '-'
         name = template_name.replace('.template', '').split('/')
```
gazpachoking, line 21: I think we can just go with this change, and skip all the rest. The result of a `str.split` is always a list, so no need to check that, and there will always be at least one part, because all the templates will have a name.
```suggestion
        name = template_name.replace('.template', '').split('/')[-1]
```

flexget/plugins/cli/templates.py
```diff
@@ -19,8 +19,11 @@ def list_file_templates(manager, options):
...
-        if len(name) == 2:
-            name = name[1]
+        if type(name) == list:
```
paranoidi, line 22: Using `instanceof` to check type is preferable

gazpachoking:
Thanks for the report and PR! Went with a slightly simpler fix.

PR #3392, opened 2022-03-23
[feature] sftp_list - add dirs_only option to exclude files from processing

aidan-:
### Motivation for changes:
In situations where releases are structured in their own directory, processing the individual files (ie a directory of rar's) is unnecessary.  Adding this check to skip files (`dirs_only`) can speed up this list process considerably.

### Detailed changes:
- Add `dirs_only` to exclude files from being turned into potential entries.
- This is implemented very similarly to the existing `files_only` option.
- A very rough benchmark gives a 30% speed improvement in using `sftp_list` with `dirs_only: True` with large directories.

```
sftp_list:
  host: example.com
  username: Username
  private_key: /Users/username/.ssh/id_rsa
  recursive: True
  dirs_only: True
  dirs: 
    - '/path/to/list/'
    - '/another/path/'
```
```
n/a
```

### DIFF
--- a/flexget/components/ftp/sftp.py
+++ b/flexget/components/ftp/sftp.py
@@ -43,6 +43,7 @@ class SftpList:
     get_size:             Indicates whetern to calculate the size of the remote file/directory.
                           WARNING: This can be very slow when computing the size of directories!
     files_only:           Indicates wheter to omit diredtories from the results.
+    dirs_only:            Indicates whether to omit files from the results.
     dirs:                 List of directories to download.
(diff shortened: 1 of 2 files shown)


gazpachoking:
Seems good to me. Do we need to worry about the case when an user accidentally uses both dirs_only, and files_only? Maybe just log a warning, or an error?

gazpachoking:
> 👍 I've opted to log a warning, WDYT?

Seems good to me, thanks!

PR #3393, opened 2022-03-27
[feature] movie_list - add strip_non_alphanum flag to remove non alphanum chars from title

aidan-:
### Motivation for changes:
Some search sites have very simple search implementations that can't match with non-alphanumeric characters in the movie title as these characters are often not in the entry name on the site.  Removing these characters from the search string has a higher likelihood of resulting in a search match.

An example of this is `Spider-Man: No Way Home (2021)` -> `Spider Man No Way Home 2021`.

### Detailed changes:
- Add flag to `movie_list` that replaces non-alphanumeric characters with a space.

### Config:
```
test_list_get_strip_non_alphanum:
   movie_list:
     list_name: test_list
     strip_non_alphanum: yes
```

### Test output:
```
====================================================== test session starts =======================================================
platform linux -- Python 3.8.10, pytest-6.2.5, py-1.10.0, pluggy-1.0.0
rootdir: /home/aidan/Projects/Flexget, configfile: setup.cfg
plugins: cov-3.0.0, xdist-2.4.0, forked-1.3.0
collected 18 items                                                                                                               

flexget/tests/test_movie_list.py ..................                                                                        [100%]

======================================================== warnings summary ========================================================
<removed warnings>
================================================ 18 passed, 53 warnings in 5.38s =================================================
```

### DIFF
--- a/flexget/components/managed_lists/lists/movie_list/db.py
+++ b/flexget/components/managed_lists/lists/movie_list/db.py
@@ -8,6 +8,7 @@ from sqlalchemy.sql.elements import and_
 from flexget import plugin
 from flexget.db_schema import versioned_base, with_session
 from flexget.entry import Entry
+from flexget.utils.tools import strip_non_alphanum as strip_non_alphanum_fn
 
 try:
     # NOTE: Importing other plugins is discouraged!
@@ -65,7 +66,7 @@ class MovieListMovie(Base):
             self.list_id,
         )
 
-    def to_entry(self, strip_year=False):
+    def to_entry(self, strip_year=False, strip_non_alphanum=False):
         entry = Entry()
         entry['title'] = entry['movie_name'] = self.title
         entry['url'] = 'mock://localhost/movie_list/%d' % self.id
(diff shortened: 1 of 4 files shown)


gazpachoking:
Which sites are these? Ideally we'd resolve this on the search side, that way it can work for every input plugin, and not require extra configuration.

aidan-:
> Which sites are these?
I've hit it with two sites, one is a gazelle based tracker (haven't checked on other gazelle trackers) and the other is [Airdc++ Web](https://airdcpp-web.github.io/) which I am playing around with as a new search/output plugin.

> Ideally we'd resolve this on the search side
I agree, I think that would be a better approach especially if users aren't having issues with this on other sites.

Happy if you want to close this PR and I can open another one for the specific site I'm seeing the 'issue' with.

gazpachoking:
> Happy if you want to close this PR and I can open another one for the specific site I'm seeing the 'issue' with.

I think that would be the ideal solution, thanks! Especially because we can already strip the characters before adding to movie_list to achieve a similar result to this.

PR #3401, opened 2022-03-29
Upgrade all deps

gazpachoking:
### Detailed changes:
- 

### Addressed issues:
- Fixes # .

### Implemented feature requests:
- Feathub #[XX](https://feathub.com/Flexget/Flexget/+XX).

### Config usage if relevant (new plugin or updated schema):
```
paste_config_here
```
### Log and/or tests output (preferably both):
```
paste output here
```
#### To Do:

- [ ] Stuff..

PR #3412, opened 2022-04-03
Fix jaraco lib names in requirements.txt

aidan-:
### Motivation for changes:
I was unable to do a clean install from `master` due to some changes in #3401.  After running `bin/pip install -e .` and trying to run `bin/flexget -V` the error was:

```
pkg_resources.DistributionNotFound: The 'jaraco-text==3.7.0' distribution was not found and is required by FlexGet
```

Changing the `jaraco-*` dependency naming back to `jaraco.*` looks to fix this issue.

### Detailed changes:
- This is a similar change/revert like in 1604012738f1c011e50313b1605cfd029670d5d6

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -66,18 +66,18 @@ idna==3.3
     # via requests
 itsdangerous==2.0.1
     # via flask
-jaraco-classes==3.2.1
+jaraco.classes==3.2.1
     # via jaraco-collections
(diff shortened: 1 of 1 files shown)


gazpachoking:
Thanks for this. Not sure how these dots all got changed to dashes. I was using pip-compile to upgrade the deps when they got messed up.

PR #3415, opened 2022-04-04
[add] gotify: add support for setting content_type

aidan-:
### Motivation for changes:
Gotify notifications are plain text by default but [they support markdown](https://gotify.net/docs/msgextras#clientdisplay).  With markdown support, notifications can be enriched with images, bold text, etc which make them easier to read/highlight important information.

### Detailed changes:
- Add option `content_type` to gotify notifier.
- Default value is `text/plain` for backwards compatibility 

```
notify:
  task:
    always_send: yes
    title: "test!"
    template: gotify.template
    via:
      - gotify:
          url: "https://redacted-site.net/"
          token: "nope"
          priority: 1
          content_type: "text/markdown"
```

### DIFF
--- a/flexget/components/notify/notifiers/gotify.py
+++ b/flexget/components/notify/notifiers/gotify.py
@@ -33,7 +33,11 @@ class GotifyNotifier(object):
             'url': {'format': 'url'},
             'token': {'type': 'string'},
             'priority': {'type': 'integer', 'default': 4},
-            'content_type': {'type': 'string', 'enum': ['text/plain', 'text/markdown'], 'default': 'text/plain'},
+            'content_type': {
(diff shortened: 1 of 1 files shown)


aidan-:
> Are text/plain and text/markdown the only 2 valid options? Can we add that as an enum in the config schema? https://json-schema.org/understanding-json-schema/reference/generic.html#enumerated-values

@gazpachoking That's correct, they're the only two valid values for now.  I've just updated the schema to use an enum 👍

ksurl:
@aidan- I noticed the wiki doesn't have any docs for this plugin in the notifiers section. if you can provide a sample config, I can add it to the wiki

PR #3425, opened 2022-04-27
[change] Added pagination to `my_anime_list`

### LINKED ISSUE #3421: Plugin my_anime_list, URL JSON result maxed at 300 entries.
### Expected behaviour:
For the Plugin: my_anime_list to return the requested categories, preferably with less traffic overhead.

### Actual behaviour:
The URL JSON result returns the first 300 results of ALL categories in order of status id.
This is a limitation of the JSON url used and not a bug in the flexget plugin code.

### Steps to reproduce:
- Step 1: Have more than 300 listed anime on the site, in any given category.
- Step 2: Request a category that is sorted last in the JSON result, like "plan_to_watch".
- Step 3: If you have more than 300 anime listed as "Completed" or "Dropped", the JSON result would never load the "plan_to_watch" entries.

#### Config:
```yaml
my_anime_list:
  username: <removed>
  status:
    - watching
    - plan_to_watch
```
  
#### Log:
<details>

```
na
```
</details>

### Additional information:

The JSON source used in the plugin, only gives the first 300 results from MAL. (https://myanimelist.net/animelist/flexget/load.json)
This is not an issue if you use "watching" as they are sorted first in the JSON reply, but "plan_to_watch" that many use is sorted last, after categories like "completed" and "dropped".

Solutions:
It's possible to specify what result you want in the URL but only for a single status id.
If it merged multiple status specific JSON results, it would probably require less traffic than the current parameterless "load.JSON", as most users just want status "1" & "6".
Ex. Status 6 is "plan_to_watch" only: https://myanimelist.net/animelist/flexget/load.json?status=6

It's also possible to offset the result by a given amount, but would require merging large lists until an empty result is given "[]".
Ex. JSON results from 300->600: https://myanimelist.net/animelist/flexget/load.json?offset=300

Alternative solution:
From what i read, Load.JSON is an internal tool for loading lists and is not meant to be used as an API.
There is a new API, but it appears to require an authorization token to use and it's in beta:
API auth: https://myanimelist.net/apiconfig
API doc.: https://myanimelist.net/apiconfig/references/api/v2

BrutuZ:
Allows for more than 300 results (may the lord have mercy on your soul)

### Motivation for changes:
The referenced issue
### Detailed changes:
- Added pagination logic with offsets
- Reordered some lines to group logic
- Updated to f-strings
- Updated as an entry generator (makes it compatible with `limit` plugin)

### Addressed issues:
- Fixes #3421 .

### DIFF
--- a/flexget/plugins/input/my_anime_list.py
+++ b/flexget/plugins/input/my_anime_list.py
@@ -56,60 +56,65 @@ class MyAnimeList:
 
     @cached('my_anime_list', persist='2 hours')
     def on_task_input(self, task, config):
-        entries = []
-        selected_status = config['status']
-        selected_airing_status = config['airing_status']
-        selected_types = config['type']
-
+        selected_status = config.get('status', ['all'])
         if not isinstance(selected_status, list):
             selected_status = [selected_status]
 
+        selected_airing_status = config.get('airing_status', ['all'])
         if not isinstance(selected_airing_status, list):
             selected_airing_status = [selected_airing_status]
 
+        selected_types = config.get('type', ['all'])
         if not isinstance(selected_types, list):
             selected_types = [selected_types]
 
         selected_status = [STATUS[s] for s in selected_status]
-
         selected_airing_status = [AIRING_STATUS[s] for s in selected_airing_status]
+        list_json = []
 
-        try:
-            list_response = task.requests.get(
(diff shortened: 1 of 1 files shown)


BrutuZ:
Tested as working per https://github.com/Flexget/Flexget/issues/3421#issuecomment-1120318446

PR #3440, opened 2022-05-15
[add] Discord Notifier: `timestamp` support and rate-limit prevention

BrutuZ:
### Motivation for changes:
Add support for the `timestamp` field on the embeds.
Got rate-limited when the task tried to send over 30 notifications at once.

### Detailed changes:
- Put the API request in a Session with TimedLimiter to prevent losing notifications to rate-limit by the server, queuing them instead.
- Added `timestamp` to the schema and added code to try to convert its value and validate it since the API will return an error otherwise

Was already documented

### DIFF
--- a/flexget/components/notify/notifiers/discord.py
+++ b/flexget/components/notify/notifiers/discord.py
@@ -1,14 +1,18 @@
+from datetime import datetime
+
+from dateutil.parser import ParserError, isoparse
 from loguru import logger
-from requests.exceptions import RequestException
 
 from flexget import plugin
 from flexget.event import event
 from flexget.plugin import PluginWarning
-from flexget.utils import requests
+from flexget.utils.requests import RequestException, Session, TimedLimiter
 
 plugin_name = 'discord'
 
 logger = logger.bind(name=plugin_name)
+session = Session()
+session.add_domain_limiter(TimedLimiter('discord.com', '3 seconds'))
 
 
 class DiscordNotifier:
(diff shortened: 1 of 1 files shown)


flexget/components/notify/notifiers/discord.py
```diff
@@ -115,15 +120,39 @@ def notify(self, title, message, config):
...
+                            embed['timestamp'] = ts
+                        except (ParserError, ValueError) as e:
+                            logger.info(f"'timestamp' is in an invalid format: {e}")
+                if not isinstance(ts, datetime):
+                    embed.pop('timestamp', None)
+                    logger.warning("'timestamp' is invalid, dropping it")
+                else:
+                    embed['timestamp'] = datetime.strftime(ts, r'%Y-%m-%dT%H:%M:%S%z')
+
+            web_hook = {
```
gazpachoking, line 147: Is this bit indented too much? Shouldn't this just happen once after the loop?

BrutuZ: There can be up to 10 embeds in a single message, the loop parses them and evaluates each `timestamp` property

gazpachoking: Right, but the web hook is only used once after all the embeds are processed, and contains all of them.

BrutuZ: Ah, I see now. Indeed the web_hook definition was a little too indented. The code preview misled me since it was the last line

PR #3443, opened 2022-05-20
[fix] pyparsing: lock pyparsing to 2.4.7 fixing webui log page

kristenmills:
### Motivation for changes:
The Log page on webui is broken after we updated pyparsing. I'm going to look into this further so we can update pyparsing properly but for now want to lock it back to 2.4.7.

### Detailed changes:
- lock pyparsing version 2.4.7

### Addressed issues:
- https://github.com/Flexget/webui/issues/155


gazpachoking:
I had a hell of a time fighting with pip-compile and getting everything working last time I messed with deps. I think we should just manually do the edits right now to only pin pyparsing so we don't have to mess with tweaks to get all the python versions working.

kristenmills:
Yeah agreed that makes sense, i'll do that.

PR #3447, opened 2022-05-23
cache apt packages in main workflow

ksurl:
### Motivation for changes:

speed up workflow by caching apt packages

### Detailed changes:
- use github action from https://github.com/awalsh128/cache-apt-pkgs-action/

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -18,12 +18,12 @@ jobs:
       fail-fast: false
     steps:
     - uses: actions/checkout@v2
-    # Can we cache the apt-get stuff somehow?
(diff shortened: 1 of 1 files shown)


gazpachoking:
Hmm, this seems nice, just a bit worried if it will break considering there are only 7 stars on that action.

ksurl:
I took a look at the repo and they're using github caching under the hood. And some shell scripts to load the contents to the appropriate paths after mounting the cache

gazpachoking:
I tried this action in #3572 . Didn't seem to work so well. Caused the tests to fail on the second run. https://github.com/Flexget/Flexget/actions/runs/3247141127/jobs/5326731642

PR #3452, opened 2022-05-25
[fix] skip all test_decompress tests if rarfile module missing

aidan-:
### Motivation for changes:
Running tests against a clean virtualenv (Python 3.10.4) results in failed tests in `flexget/tests/test_decompress.py` due to missing `rarfile` dep.  Virtualenv was setup (following the docs) by running:

```
pip install -e .
pip install -r dev-requirements.txt
```

### Detailed changes:
- Some tests in `test_decompress.py` are skipped if `rarfile` is missing, but some haven't been configured to use the `pytest.mark.skipif` decorator.  
- This change adds the decorator to the whole test suite, as all uses of `decompress` require the `rarfile` module.

```
=== 1349 passed, 34 skipped, 4 xfailed, 5 xpassed, 543 warnings in 32.60s ===
```

### DIFF
--- a/flexget/tests/test_decompress.py
+++ b/flexget/tests/test_decompress.py
@@ -8,6 +8,7 @@ except ImportError:
     rarfile = None
 
 
+@pytest.mark.skipif(rarfile is None, reason='rarfile module required')
 @pytest.mark.usefixtures('tmpdir')
 class TestExtract:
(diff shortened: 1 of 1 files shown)


gazpachoking:
LGTM, thanks!

Issue #3444, opened 2022-05-21
Pytz warning on startup


ksurl:
### Expected behaviour:

no warning

### Actual behaviour:
warning on startup for pytz zone deprecation

### Steps to reproduce:
- Step 1: run flexget as daemon on 3.3.11 with python 3.9
  
#### Log:
<details>
 <summary>(click to expand)</summary>

```
/usr/local/lib/python3.9/site-packages/flexget/components/scheduler/scheduler.py:129: PytzUsageWarning: The zone attribute is specific to pytz's interface; please migrate to a new time zone provider. For more details on how to do so, see https://pytz-deprecation-shim.readthedocs.io/en/latest/migration.html
  if timezone.zone == 'local':
```
</details>

### Additional information:

- FlexGet version: 3.3.11
- Python version: 3.9.9
- Installation method: pip via docker
- Using daemon (yes/no): yes
- OS and version: alpine 3.15 container on debian 11 host

NoSubstitute:
Same issue, slightly different message, which seems to indicate that it's a piece of code that needs changing, and not a setting in my environment.

```
$ ~/flexget/bin/flexget daemon start
/home/USER/flexget/lib/python3.8/site-packages/flexget/components/scheduler/scheduler.py:129: PytzUsageWarning: The zone attribute is specific to pytz's interface; please migrate to a new time zone provider. For more details on how to do so, see https://pytz-deprecation-shim.readthedocs.io/en/latest/migration.html
  if timezone.zone == 'local':
```

aidan-:
It looks like this issue is coming from the `apscheduler` dependency.  More information [here](https://github.com/agronholm/apscheduler/discussions/570).

mylovesaber:
Problem still exist :
```bash
root@Ubuntu-2204-jammy-amd64-base ~ # systemctl status flexget --no-pager -l
● flexget.service - flexget Daemon
     Loaded: loaded (/etc/systemd/system/flexget.service; enabled; vendor preset: enabled)
     Active: active (running) since Thu 2022-06-02 19:45:01 CST; 8min ago
    Process: 15505 ExecStartPre=/bin/rm -f /root/.config/flexget/.config-lock (code=exited, status=0/SUCCESS)
   Main PID: 15506 (flexget)
      Tasks: 15 (limit: 38231)
     Memory: 103.0M
        CPU: 1.419s
     CGroup: /system.slice/flexget.service
[...]

PR #3453, opened 2022-05-26
[fix] use datetime to detect running systems tzinfo

aidan-:
### Motivation for changes:
Attempting to address the issue raised in #3444.  The tldr is that `pytz` is being deprecated and there is now a more 'supported' way to get local `tzinfo`.
```
PytzUsageWarning: The zone attribute is specific to pytz's interface; please migrate to a new time zone provider. For more details on how to do so, see https://pytz-deprecation-shim.readthedocs.io/en/latest/migration.html
```

### Detailed changes:
- Use `datetime` to determine the running systems `tzinfo`.
- According to the `datetime` [docs](https://docs.python.org/3.6/library/datetime.html#datetime.datetime.astimezone) this method might only be reliable in Python >= 3.6.  I've tested this using 3.10 locally and it seems to be working.  I'll let the CI builds kick off to test the other versions.

### Addressed issues:
- #3444

#### Comments
- Turns our time zones in general are a bit of a rabbit hole in python.  [This](https://blog.ganssle.io/articles/2018/03/pytz-fastest-footgun.html) was an interesting read that provided a bit of context on why certain methods are being deprecated.

### DIFF
--- a/flexget/components/scheduler/scheduler.py
+++ b/flexget/components/scheduler/scheduler.py
@@ -1,9 +1,9 @@
+import datetime
 import hashlib
 import logging
 import os
 import struct
 
-import datetime
 from apscheduler.jobstores.sqlalchemy import SQLAlchemyJobStore
 from apscheduler.schedulers.background import BackgroundScheduler
 from apscheduler.triggers.cron import CronTrigger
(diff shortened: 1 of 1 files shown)


gazpachoking:
I remember struggling to eliminate timezone issues with apscheduler for quite a while. This looks much cleaner, hopefully nothing new pops up. Thanks!

PR #3495, opened 2022-07-06
[fix] Use "string" type for Aria2 options

gdm257:
### Motivation for changes:
The value of option must be a `string`, not `integer`.

### Detailed changes:
- allow `string` `number` `boolean` type for `options`
- auto convert non-string type to string for options before sending to aria2

> In the option struct, the name element is the option name (without the preceding `--`) and the value element is the argument as a **string**.    ——[Options - aria2 documentation](https://aria2.github.io/manual/en/html/aria2c.html#id3)

As document says, option value should be a string. If other types are used such as integer, the option will not take effect (already tested in aria2 v1.36.0).

Now option can be a boolean or number directly. Writing option as a string is also allowed. See the config below.

```yaml
    aria2:
      path: "/downloads"
      options:
        max-upload-limit: 0
        bt-stop-timeout: "2400"
        seed-time: 66
        seed-ratio: 4.4
        bt-remove-unselected-file: true
        allow-overwrite: "false"
```

### DIFF
--- a/flexget/plugins/clients/aria2.py
+++ b/flexget/plugins/clients/aria2.py
@@ -42,7 +42,9 @@ class OutputAria2:
             },
             'options': {
                 'type': 'object',
-                'additionalProperties': {'oneOf': [{'type': 'string'}, {'type': 'number'}, {'type': 'boolean'}]},
+                'additionalProperties': {
+                    'oneOf': [{'type': 'string'}, {'type': 'number'}, {'type': 'boolean'}]
+                },
             },
         },
         'required': ['path'],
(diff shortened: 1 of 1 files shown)


gazpachoking:
I think it would be better to allow the user to input numbers, then convert to a string before sending to their api. Requiring the user to put quotes around numbers in their config is non-intuitive.

luoyemengchen:
> I think it would be better to allow the user to input numbers, then convert to a string before sending to their api. Requiring the user to put quotes around numbers in their config is non-intuitive.

Good idea, I will convert them

gazpachoking:
Looks good to me, thanks!

PR #3519, opened 2022-08-02
Translation without title tifx

DDanii:
### Motivation for changes:
There are translations without title witch causes an error.


paranoidi:
There are changes to two different plugins here. Not clear if intentional.

DDanii:
@paranoidi  fixed it

PR #3529, opened 2022-08-23
adding matrix notifier

antibios:
### Motivation for changes:
Matrix is a secure msging platform that allows for updates via a REST call

### Detailed changes:
Addition of matrix.py to the notifiers folder. 

### Addressed issues:
- Fixes # .
- Feature request.

- Matrix support

```
Sample config:
      entries:
        via:
          - matrix:
                  server: 'https://matrix.org'
                  token: 'SECRETKEY-dsarasfaFASZgo'
                  roomid: '!ufPOdHzBzMRGDSASgf:matrix.org'

```
```
2022-08-23 14:32:49 VERBOSE  notify        sort-stuff     Successfully sent a notification to `matrix`
```
#### To Do:
- [ ] Group multiple notifications to a single msg to reduce number of requests.

### DIFF
--- a/flexget/components/notify/notifiers/matrix.py
+++ b/flexget/components/notify/notifiers/matrix.py
@@ -1,0 +1,76 @@
+from loguru import logger
+from requests.exceptions import RequestException
+
+from flexget import plugin
+from flexget.event import event
+from flexget.plugin import PluginWarning
+from flexget.utils.requests import Session as RequestSession
+
+requests = RequestSession(max_retries=3)
+
+plugin_name = 'matrix'
+
+logger = logger.bind(name=plugin_name)
+
+def urljoin(*args):
+    """
+    Author: Rune Kaagaard
+    Joins given arguments into an url. Trailing but not leading slashes are
+    stripped for each argument.
+    """
+    return "/".join(map(lambda x: str(x).rstrip('/'), args))
+
+class MatrixNotifier:
+    """
+    Sends messages via Matrix. The MatrixHttpApi library is required to be installed.
+    Install it with: `pip install matrix_client`
+
+    All fields are required.
+
+    token: Is available in Element under Help/About.
+    roomId: Is available in Element under room settings.
+
+    Example::
+
+      notify:
+        entries:
+          via:
+            - matrix:
+                server: "https://matrix.org"
+                token: sender's token
+                roomid: room identifier
+    """
+
+    schema = {
+        'type': 'object',
+        'properties': {
+            'server': {'type': 'string'},
+            'token': {'type': 'string'},
+            'roomid': {'type': 'string'},
+        },
+        'required': ['server', 'token', 'roomid'],
+        'additionalProperties': False,
+    }
+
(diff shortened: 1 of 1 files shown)


flexget/components/notify/notifiers/matrix.py
```diff
@@ -0,0 +1,78 @@
...
+
+class MatrixNotifier:
+    """
+    Sends messages via Matrix. The MatrixHttpApi library is required to be installed.
+    Install it with: `pip install matrix_client`
+
+    All fields are required.
+
+    token: Is available in Element under Help/About.
+    roomId: Is available in Element under room settings.
```
gazpachoking, line 33: Capitalization is different here than in actual usage. Maybe we want to standardize it to `room_id` everywhere?

gazpachoking (commented):
Looks pretty simple and good. Maybe tweak the room id name in config?

flexget/components/notify/notifiers/matrix.py
```diff
@@ -0,0 +1,78 @@
...
+    def notify(self, title, message, config):
+        """
+        Send notification to Matrix Room
+        """
+        notification = {'body': message, 'msgtype': "m.text"}
+        room = urljoin(
+            config['server'],
+            "_matrix/client/r0/rooms",
+            config['roomid'],
+            "send/m.room.message?access_token=" + config['token'],
```
paranoidi, line 68: probably should use

urllib.parse.urlencode(....)

flexget/components/notify/notifiers/matrix.py
```diff
@@ -0,0 +1,78 @@
...
+from flexget.utils.requests import Session as RequestSession
+
+requests = RequestSession(max_retries=3)
+
+plugin_name = 'matrix'
+
+logger = logger.bind(name=plugin_name)
+
+
+def urljoin(*args):
```
paranoidi, line 16: why not just use standard urljoin ?

antibios: urllib.urljoin() takes from 2 to 3 positional arguments but 4 were required.

flexget/components/notify/notifiers/matrix.py
antibios: done

flexget/components/notify/notifiers/matrix.py
antibios: rllib.urljoin() takes from 2 to 3 positional arguments but 4 were required.

paranoidi:
Awesome, let's merge this in. Can you update the wiki too?

I created the page at

https://flexget.com/en/Plugins/Notifiers/matrix

In case I have completed it, just proof-read it :)

antibios:
I've updated the wiki with instructions

paranoidi:
@antibios nice! I made some further formatting improvements to stay more consistent with other documentation

thanks for contributing

gyandoo:
@paranoidi is it possible to send text/markdown as matrix supports it, and if yes, how, thanks

PR #3553, opened 2022-10-10
Change the configuration format of the notifier Bark

RyoLee:
### Motivation for changes:
Change to the consistent identifier naming convention with the project

### Detailed changes:
- replace options
  - automaticallyCopy: string -> automatically_copy: boolean
  - isArchive: string -> is_archive: boolean
- add enum
  - level: 'active', 'timeSensitive', 'passive'

### Addressed issues:
- Flexget/wiki#11

- N/A

```
notify:
  entries:
    what: rejected  #only for test
    title: |
          {{task}}
    message: |
          {% if series_name is defined %}{{series_name}}{% if series_id is defined %} - {{series_id}}{% endif %}{% if trakt_ep_name is defined %} - {{trakt_ep_name}}{% endif %}
          {% elif imdb_name is defined%}{{movie_name}}
          {% else %}{{title}}
          {% endif %}
    via:
      - bark:
          server:  https://******
          device_key: ******
          level: passive
          automatically_copy: true
          is_archive: true
```
```
paste output here
```
#### To Do:

- N/A

### DIFF
--- a/flexget/components/notify/notifiers/bark.py
+++ b/flexget/components/notify/notifiers/bark.py
@@ -30,17 +30,9 @@ class BarkNotifier:
             webpost:
               server: <string>
               device_key: <string>
-              [level: <string>]
-              [badge: <integer>]
-              [automaticallyCopy: <string>]
-              [copy: <string>]
-              [sound: <string>]
-              [icon: <string>]
-              [group: <string>]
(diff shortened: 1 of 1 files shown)


gazpachoking:
👍

PR #3555, opened 2022-10-10
[add] docker: run as non-root user

ksurl:
### Motivation for changes:

allow running image as non-root user by using non protected folder /config. WORKDIR set to /config so the current directory is detected first for config location

~~add tzdata to support timezones via env~~

### Detailed changes:
- change config folder location from /root/.flexget to /config
~~- install tzdata to allow timezone setting~~

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -30,7 +30,8 @@ ENV PYTHONUNBUFFERED 1
 
 RUN apk add --no-cache --upgrade \
         ca-certificates \
-        nodejs && \
+        nodejs \
+        tzdata && \
     rm -rf /var/cache/apk/*
 
 COPY --from=0 /wheels /wheels
--- a/Dockerfile
+++ b/Dockerfile
@@ -47,7 +47,7 @@ RUN pip install -U pip && \
 
 COPY --from=0 /flexget-ui-v2 /usr/local/lib/python3.9/site-packages/flexget/ui/v2/
 
-RUN mkdir /root/.flexget
-VOLUME /root/.flexget
+VOLUME /config
+WORKDIR /config
 
 ENTRYPOINT ["flexget"]
--- a/Dockerfile
+++ b/Dockerfile
@@ -30,8 +30,7 @@ ENV PYTHONUNBUFFERED 1
 
 RUN apk add --no-cache --upgrade \
         ca-certificates \
-        nodejs \
-        tzdata && \
+        nodejs && \
     rm -rf /var/cache/apk/*
 
 COPY --from=0 /wheels /wheels


ksurl:
no need to tzdata and pre-init script to copy to /etc/localtime if you mount /etc/localtime from the docker host.

BrutuZ:
TZ support via Env Variable seemed like a nice addition seeing as Docker on Windows is a thing

ksurl:
> TZ support via Env Variable seemed like a nice addition seeing as Docker on Windows is a thing

that's true, but for now I'm respecting the simple "flexget" entrypoint instead of going straight to a more complex init process that does other stuff before running flexget. you have the flexibility of setting your own entrypoint script that does that before starting the daemon.

ksurl:
@trim21 again let me know if this breaks any tests or current usage for you. if this is merged, we may need to add it to upgrade actions on the wiki for those actually using the official image to change their mount location.

gazpachoking:
> that's true, but for now I'm respecting the simple "flexget" entrypoint instead of going straight to a more complex init process that does other stuff before running flexget. you have the flexibility of setting your own entrypoint script that does that before starting the daemon.

I don't think we need to add any init process to allow tz support. I think if we just install the tzdata package, the user can set the TZ env variable when running the container (which should probably still default to UTC if the user doesn't specify.) Let's put that in a different PR if we go for it though.

This PR seems fine to me, I like this folder better. Users probably will need to edit their docker run command or compose file though. Putting the notes in the upgrade actions makes sense, and we should probably bump the minor version number.

ksurl:
I don't think any changes needed for entrypoint or command in user side setup, just the host folder/volume mount location. Since I set workdir to the new folder, flexget already looks there as current directory for the config by default and -c is unnecessary. Anyone calling it with -c shouldn't be affected. The only thing is now they'll have a dangling randomly named volume since docker will create one if not specified when volume is defined in the Dockerfile.

gazpachoking:
> I don't think any changes needed for entrypoint or command in user side setup, just the host folder/volume mount location.

Right, that's what I meant, they'll have to change their run command (or volume in compose) from `docker run --mount type=bind,source="$(pwd)"/config.yml,target=/root/.flexget/` to the new folder, right? (or something along those lines)

ksurl:
Yes. I can do a separate pr for timezone support.

trim21:
If you asks… this may break my current usage, I'm current mounting my config to "/root/.flexget"

gazpachoking:
> If you ask… this may break my current usage, I'm current mounting my config to "/root/.flexget"

Hmm, yeah I was thinking that's how everyone would do it, and have their config/logs in a folder on the host system, that gets bound into the container. What does the `VOLUME` command in the dockerfile actually do if you haven't done that? Does it mount a docker volume there automatically? How do you edit your config in that case?

trim21:
> What does the VOLUME command in the dockerfile actually do if you haven't done that? Does it mount a docker volume there automatically?

Yes docker will create a docker volume if you don't have a VOLUME binding on that.

ksurl:
So moving to /config is ok?

I am using `user` to set a non root gid:uid in my compose file. 

My workaround is to mount /config and also an empty folder to /root/.flexget so it doesn't create a new volume.

trim21:
I think thay depends on if flexget will continue to search config in "/.root/flexget" , docker volume is not a big issue here, but the old mounting config may stop working is an issue here

ksurl:
It looks in /root/.flexget since it is set as the current directory via workdir. The workdir also changes in this commit, so it shouldn't look for the old folder.  I'm ok with maintaining a custom image based off the main one if this change is too risky with users not being aware of the change.

gazpachoking:
So.. about this running as non-root user. Does this change actually do that? Or just allow it? @trim21 If yours is still running as root user, shouldn't it find your config still since it's in the home folder?

ksurl:
> So.. about this running as non-root user. Does this change actually do that? Or just allow it? @trim21 If yours is still running as root user, shouldn't it find your config still since it's in the home folder?

No it still defaults as root. You have to specify a non root gid:uid when running the container. This would break root users since the working directory is no longer /root/.flexget but if it is still mounted and running as root, it should find it I guess. The other issue would be the new volume created for /config if not specified. I'd be open to just dropping the volume declaration in the image and letting the user decide for themselves. Then we can drop workdir also (or leave it). This would require the user who is running as nonroot to use -c to specify the config file.

gazpachoking:
> No it still defaults as root. You have to specify a non root gid:uid when running the container. This would break root users since the working directory is no longer /root/.flexget but if it is still mounted and running as root, it should find it I guess.

Yeah, I'm thinking it should still find configs mounted in /root/.flexget since it won't find one in the current directory, and it'll fall back to the home directory. I think this change is fine without a version bump if this is the case. (I should actually test...)

ksurl:
leaving VOLUME in will create the dangling volume for users mapping to the old location. if we remove it, it will still find the config in /config since it's the current directory via WORKDIR and also /root/.flexget since it's the home folder, so the run command or compose file doesn't need to specify the location.

X-dark:
I am not sure it make much sense doing that change without creating and using a non-root user inside the container. Yes that would be a breaking change but the impact on user side is just mounting a volume elsewhere.
For me there is no reason for this container to run as root.

ksurl:
That's an option to create a user inside (add shadow package) but I am just using `user: gid:uid` in my compose file.

Adding a user would probably mean adding an init system to read the uid and gid from env and running flexget as that user. I like leaving the default entrypoint as is.

If they want something more complex, there is wiserain's image which is based off linuxserver/baseimage-alpine.

trim21:
may need to take care of config folder permission

ksurl:
If you define user in the image, you would need a script as entrypoint to set custom gid, uid and fix permissions of the config folder since not everyone will use the same id.

trim21:
I think a pr simply changing workdir and volume path  is not a good idea here, because it make a break change and didn't solve any problem that can't be solve by changing `docker run` arguments, like setting working dir.

ksurl:
> I think a pr only changes workdir and volume path is not a good idea here.
> 
> Currently our docker image is simply shipping a python environment with our flexget package, I'm not against adding an init system or entry point script or some other things (like adding tzdata) to make it more feature-rich, but this PR doesn't add a feature IMO, It add a break change and didn't solve any problem that can't be solved by changing `docker run` arguments or docker-compose file. If anyone need this, they only need to change their working dir configuration

It can be non-breaking if we don't change but only remove workdir/volume. anyone still mounting to /root/.flexget will still work fine. I'd be ok with that. It is not really adding a "feature" but fixing the workaround of mounting a fake folder to the root location and then using a different mount for your config. nonroot user can't read/write to /root.

gazpachoking:
I like `/config` better as a default config directory. It both feels cleaner, and allows the user to easily change to non-root if they desire. I think @ksurl is right that this shouldn't break existing configs, since if you mounted your config in the root home directory, and still run as root, that will be in the search path as long as you don't also put a config in `/config`

ksurl:
> I like `/config` better as a default config directory. It both feels cleaner, and allows the user to easily change to non-root if they desire. I think @ksurl is right that this shouldn't break existing configs, since if you mounted your config in the root home directory, and still run as root, that will be in the search path as long as you don't also put a config in `/config`

this will create dangling volumes for users on the old path since `VOLUME /config` is declared. docker will autocreate a named volume mounted to /config on every new container start.

gazpachoking:
> this will create dangling volumes for users on the old path since `VOLUME /config` is declared. docker will autocreate a named volume mounted to /config on every new container start.

@ksurl Do you think we should roll back? Is this really a problem? I bumped to 3.4.0 so we can document it on the upgrade actions page if people want to move their mount, or delete the dangling volume.

ksurl:
I don't see a reason to roll back as long as it's clearly stated in the changelog and upgrade actions.

gazpachoking:
Care to draft the upgrade action? 😄

ksurl:
image

how does it look

gazpachoking:
👍

ksurl:
wiki updated

paranoidi:
I think https://flexget.com/InstallWizard/Docker requires some updates too?

gazpachoking:
> I think https://flexget.com/InstallWizard/Docker requires some updates too?

Good call. Done.

PR #3557, opened 2022-10-10
[add] docker: tzdata for timezone support via env

ksurl:
### Motivation for changes:

allow setting timezone for docker via environment variable

### Detailed changes:
- add tzdata package to image
- defaults to UTC

#### compose
```
environment:
  - TZ={TIMEZONE}
```
#### docker cli
```
docker run -e TZ={TIMEZONE}
```

Use name from tz database name [here](https://en.wikipedia.org/wiki/List_of_tz_database_time_zones)

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -30,7 +30,8 @@ ENV PYTHONUNBUFFERED 1
 
 RUN apk add --no-cache --upgrade \
         ca-certificates \
-        nodejs && \
+        nodejs \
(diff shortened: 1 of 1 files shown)


ksurl:
I confirmed just adding the package is enough to work when setting TZ to the desired timezone. no need to mess with /usr/share/zoneinfo, /etc/localtime or /etc/timezone in the image

PR #3559, opened 2022-10-11
Use poetry for dependency management/dev environment creation

gazpachoking:
I now have a competing PR for this using pdm instead of poetry. Looking for any feedback one way or the other #3573 

### Motivation for changes:

There were several problems with pip-tools for dependency management:
- When running on a particular platform, platform specific requirements would get baked in to the requirements.txt
- Dependabot updates were wacky, and wouldn't be the same result as updating manually (no idea if this is better with poetry yet or not)
- Conflicts between version requirements for multiple dependencies were not handled well
- No way to make sure optional dependencies don't introduce version incompatibilities

Dependency management was further complicated because installing flexget in editable mode in a dev environment still pinned the dependencies. This made trying new versions of deps annoying, as you had to pin to new version, then install again.

### Detailed changes:
- Switches to using poetry for dependency managment. the requirements*.in files went away, in favor of defining the loose version requirements in pyproject.toml. Poetry does the dependency resolving, and finds versions compatible with all transitive dependencies to pin in poetry.lock. poetry export is used to create cross-platform requirements*.txt files to be used by end users, or setuptools to create the pinned build
- Also drops python 3.6, as that was another source of pain for updating deps after it lost support
- Added pre-commit hooks which check the poetry lock file when requirements in pyproject.toml change
- Allows using `poetry install` to set up a virtual environment, install deps into it, and install flexget into it all at once. This dev environment will allow the looser dep version requirements specified in pyproject.toml so that dep versions can be adjust and tested before commiting the pinned versions back to the repo (via poetry.lock)
- pre-commit hooks are added such that `requirements.txt`, `dev-requirements.txt`, and `dev-requirements-extra.txt` all get automatically updated on commit based on the versions in the lock file. These can be used by people not doing a poetry install, and by the release build, which is still done by setuptools at the moment.
- Switch setuptools to using PEP 621 metadata, and being configured in pyproject.toml. setup.py, setup.cfg, and MANIFEST.in all get to go away.

### Usage
I haven't changed to poetry for building or publishing the packages, as poetry doesn't allow publishing based on pinned dependencies from the lock file at the moment. Might try out some solutions later for that, but for now I think this can work on its own to just generate the requirements.txt files like pip-tools was.

Updating our hard requirements should be done in pyproject.toml now. To update a pinned requirement:
`poetry update <package>` (`--lock` can be used to just update the lock file without installing anything) 
`poetry update` will update all packages (respecting constraints from pyproject.toml still)
(`--lock` can be used with either of these to update the lock file without installing anything) 

#### To Do:

- [ ] See if we like it?
- [ ] Add optional dependencies in a dependency group to make sure they are compatible with required dep versions.
- [x] ~Maybe use poeblix poetry plugin to switch to poetry to build the release rather than setuptools?~
- [x] Update dev docs (sphinx)


.pre-commit-config.yaml
```diff
@@ -9,7 +9,20 @@ repos:
   rev: 22.10.0
   hooks:
   - id: black
+- repo: https://github.com/python-poetry/poetry
+  rev: '1.2.2'
+  hooks:
+    - id: poetry-check
+    - id: poetry-lock
```
trim21, line 16: Is it necessary to run this hook before every commit?

gazpachoking: Yes, renovate will probably do the main updating. This is so that if manual changes are made it's sure to get synced to the lock file and requirements files.

trim21: If you want to make sure lock file are synched with pyproject.toml, you only need `poetry-check`

gazpachoking: poetry-check only checks the structure of pyproject.toml https://python-poetry.org/docs/master/cli/#check

poetry-lock makes sure the lock file is up to date.

It didn't seem too slow while I was committing, a second or so, but I changed it now so that it only runs when pyproject.toml has been changed. Fairly certain that's the only time we'd need to update the lock file, and I'm actually a bit confused why that isn't the default for the hook.

trim21: I'm embarrassing... 

I'm actually suggesting `poetry-lock` with `--check` args, not currently `--no-update`, in this case it only check if lock file is synced with main file, won't do deps resolving.

gazpachoking: Hmm. That will fail in the same cases, but not fix the lock file, no? That would speed up the check, but we'd want to run `poery lock` afterward anyway to fix the lock file wouldn't we? I think it's okay that it's a little bit slow, now that I've changed it to only run when pyproject or the lock file changes. It might be a good idea to add `poetry check --lock` as well, so that it runs on the CI as well (where the poetry lock hook is disabled because there is no network), and fails PRs that don't match though...

gazpachoking: I added the bit for `poetry lock --check` on the CI. What do you think about the current state?

trim21: It looks fine.

It's just poetry dependencies resolving is really slow, so only check lock in pre-commit hooks and let developer to manually run `poetry lock` after they manually change `pyproject.toml#poetry` is not a bad idea.

This is how I take to run `poetry lock --no-update` on my machine for the first time, without any poetry cache:

```text
$ time poetry lock --no-update

Writing lock file

real    7m56.862s
user    0m0.046s
sys     0m0.015s
```

gazpachoking: 8 minutes! Is that on poetry 1.2.2? My cache must have already been good or something, because I haven't seen that sort of time on any recent poetry versions. I agree though, if it can take that long, we don't want it as part of the hook.

trim21: I'm using poetry 1.2.1

gazpachoking: I'll take it out just in case. The failure with --check should be good enough, and we don't want anybody wondering what the hell is taking their commit so long if anything like that can happen.

gazpachoking:
Okay, after those other iterations, this became my top choice again. One of my main issues I was trying to solve was that pip-tools produced different outputs depending on the OS/python version it was run on, so the pinned requirements were not cross-platform. Poetry is the only solution I tried that fully takes care of that in the current state while preserving the other bits of our workflow. It's also the only solution compatible with dependabot or renovate, which is also nice. This change will also cause the least disruption, because nobody is required to use poetry, and all previous install methods are still supported.

I also switched over setuptools to be fully configured in pyproject.toml now, with PEP 621 metadata which is pretty nice, and will make switching build backend easier in the future if something seems nicer.

PR #3572, opened 2022-10-16
Switch to building and releasing using poetry

gazpachoking:
Not super sold on this yet, was just testing out what it might look like, and starting discussion.
This is based on the poetry dependency management branch, here's the real diff from there:

https://github.com/Flexget/Flexget/compare/poetry...poetry-build

Good things:
- Don't have to maintain the requirements.txt files separately.
- Don't need our custom version bumping script anymore.
- No more MANIFEST.in or setuptools
- Hmm. Thought this pros list would be longer. Now I'm totally not sure.
- Good things I did in this branch that can be separated from build/release changes and merged separately:
  - Tweaked the importing of `__version__` from needing to be out of order in `__init__.py` (actually not really certain if this is good)
  - Moved most of the settings in setup.cfg into pyproject.toml

Issues:
- Poetry doesn't have an equivalent of 'gui_scripts' at the moment (for flexget-headless on Windows)
- Version info is static in pyproject.toml. This makes loading it for --version and `flexget.__version__` tougher.
  - Currently I'm using poetry-bumpversion plugin to update it in pyproject and `__init__.py` at the same time.
- Poetry plugin requirements can't be added to the pyproject. This means if a dev wants to use them, the ones we use need to be documented separately for them to install. They are really only being used for build/release though, so it should be fine in just the CI.
- Maybe some things I'm forgetting at the moment.


gazpachoking:
Going to stick with setuptools for releasing for now. losing flexget-headless was a blocker for me.

PR #3573, opened 2022-10-16
Use pdm for dependency management/dev environment creation/release building

gazpachoking:
### Motivation for changes:
After writing the poetry branch (#3559 #3572 ), and making some compromises and hacks, (especially after trying to build our package with poetry,) I decided to investigate and try some other solutions. pdm and hatch both seemed interesting, but hatch doesn't have any lock file support at the moment, which put it out of contention.

Pros:
- Uses new standards like PEP 621 to define metadata and dependencies. This makes it easier to use, or switch to other tools in the future. (poetry will support this as well in a future version)
- Has a github action that feels a bit simpler and more integrated than setting up poetry (though I haven't tried out any caching with it yet)
- ...

Cons:
- Isn't supported in dependabot or renovate.
- ...

### Detailed changes:
- Switch to pdm for dependency managment
- Use it to install the environment for CI runs

### TODO
- [x] Update docs for installing dev environment with this method
- [x] try out building/releasing


trim21:
I have a github action to install poetry deps with cache, don't know if you are interested 

https://github.com/trim21/install-poetry-project

gazpachoking:
> I have a github action to install poetry deps with cache, don't know if you are interested

Thanks! I think I'm going to go with pdm rather than poetry for now. I think setup-python actually has poetry venv caching built in now, although I've read there are a few issues where it keeps cache when it really should be purged.

liiight:
@gazpachoking why even change from pip-tools? It's actively maintained and relatively straight forward.

gazpachoking:
> why even change from pip-tools? It's actively maintained and relatively straight forward.

@liiight 
A few reasons:
- I was having a hell of a time updating deps recently. Largely because many libraries dropped 3.6 support while we were still trying to maintain it. pip-tools doesn't take this in to account when resolving deps
- pip-tools doesn't take any other platform dependent environment markers into account when resolving, so windows only dependencies get added to the requirements.txt if you run pip-compile on windows. python 3.7 only deps get skipped if you run it on python 3.8
- Optional deps were having conflicts with pinned versions in our lock file. Using poetry or pdm means we can put those optional deps in a dependency group, and they will be used to resolve dependencies that work with everything, even though we don't have to include those optional deps in our requirements.

gazpachoking:
The one problem I just realized though: Having removed requirements.txt means people who do a source install without pdm won't get our pinned dependencies, and  have no way to get them other than to use pdm. (Not totally sure if that's a problem, pdm is really easy to get. If you are doing your own virtualenv install you could just install it in that virtualenv.)(The same thing applies to poetry.) If we want to still support that, we could export a requirements.txt file. pdm's export has many of the same problems as pip-tools though, where it loses all environment markers, so requirements.txt isn't cross platform. poetry has the ability to keep those, even when exporting to requirements.txt.

liiight:
Cool, thanks for the explanation

gazpachoking:
BTW. I'm really excited for poetry 2.0, and even more so for a PEP-665 (standard lockfile) successor. Poetry 2.0 will support PEP-621 pyproject standard metadata, which means you could do your dev install with hatch, pdm, or poetry without any changes to the pyproject in this branch. Once a lockfile standard is approved, (and implemented in the managers,) even the pins will transfer between all the tools. The only thing tool specific would be the configuration of the build backend for releases, which will only matter in our release CI (where we are free to choose whatever we want.)

liiight:
Dunno about poetry, they had some controversy which drastically reduced my confidence in them: https://www.youtube.com/watch?v=Gr9o8MW_pb0

ksurl:
> The one problem I just realized though: Having removed requirements.txt means people who do a source install without pdm won't get our pinned dependencies, and have no way to get them other than to use pdm. (Not totally sure if that's a problem, pdm is really easy to get. If you are doing your own virtualenv install you could just install it in that virtualenv.)(The same thing applies to poetry.) If we want to still support that, we could export a requirements.txt file. pdm's export has many of the same problems as pip-tools though, where it loses all environment markers, so requirements.txt isn't cross platform. poetry has the ability to keep those, even when exporting to requirements.txt.

so once pdm is merged, this will be another breaking change? will need to update instructions on wiki as well as Dockerfile possibly

gazpachoking:
> so once pdm is merged, this will be another breaking change? will need to update instructions on wiki as well as Dockerfile possibly

Probably most things wouldn't break, but some behavior would change.
- Python 3.6 installs would break, already have a warning appearing for those users.
- For the Dockerfile we could switch to something like this https://pdm.fming.dev/latest/usage/advanced/#use-pdm-in-a-multi-stage-dockerfile
- I don't think other source installs (not using pdm) would break, but I think they would switch from having our pinned deps, to just the loose ones. (Provided people are using a recent pip version. PEP 517 and PEP660 are what allow this.)

I'm still a little bit torn on the whole thing overall. I'm frustrated with pip-compile, but I'm not sure if anything else really fits perfectly at the moment.

Poetry allows for exporting a cross-platform requirements.txt, which we could still ship to ease use for people not using poetry. (They'd still have to switch workflow to explicitly `pip install -r requirements.txt` though, and if workflow has to be switched anyway, I don't feel adding pdm is a huge burden.) Poetry doesn't support gui-scripts though, which is how we ship `flexget-headless` to run flexget on windows without a command prompt popup. The issue @liiight mentioned also sours me a bit on poetry.

gazpachoking:
Now I'm back to leaning towards poetry. There are a few advantages over pdm I think.
- It does a really good job of flattening the environment markers for all dependencies, including transitive ones. This means that the exported requirements.txt, or the pinned dep wheel work much nicer cross platform.
- It's older and more popular, so supported by dependabot and renovate.

The blocker would be that we can't generate the flexget-headless script if used for building the distribution. We could keep setuptools now as the builder though, and use it like we are using pip-compile right now to manage deps. This would minimize the impact of the change right now too, leaving all old workflows the same, give us a better dependency manager, and allow poetry as your environment manager if you so choose. We can revisit changing out how we build the project at a later date.

gazpachoking:
Yeah, I think I'm going to go with poetry, but leave our build process using setuptools for now. #3559  Some exiting things are happening with standardization and other tools though, I think we could revisit again later with more options. #3559

PR #3574, opened 2022-10-19
Dockerfile: switch to python 3.10

X-dark:
### Motivation for changes:

Python 3.10 is supported by Flexget, no reason to keep using 3.9

### Detailed changes:
- change image version to use python 3.10


gazpachoking:
👍 Thanks!

PR #3576, opened 2022-10-19
[fix] release: add changelog to release body with changelog-builder-action

ksurl:
### Motivation for changes:
github releases have empty body with no notes or changelog on new versions

### Detailed changes:
- replace run step with changelog builder action

#### To Do:

- verify github release notes
- test changelog workflow for wiki changelog

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -40,14 +40,12 @@ jobs:
         TWINE_PASSWORD: ${{ secrets.pypi_token }}
       run: |
         ./release.sh
-    - name: Prep Changelog
+    - name: Build Changelog
       id: changelog
       if: env.release_tag
-      run: |
-        version=$(echo ${{ env.release_tag }} | sed -e 's/v//' -e 's/\.//g')
-        date=$(date +%Y-%m-%d)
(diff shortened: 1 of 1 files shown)


gazpachoking:
Not sure how this will behave compared to our own changelog builder, but happy to find out. Thanks!

ksurl:
how does it currently generate the changelog for the wiki? just reading the commit messages since the last release? I believe this action does the same.

gazpachoking:
It only adds lines from commit messages that were prefixed by one of the compatible tags in brackets.

https://github.com/Flexget/Flexget/blob/f55093c976a26e0fd1e46d8cd43143717c3b536e/update-changelog.py#L13-L17

gazpachoking:
Plus, the wiki page can be edited manually to add changes that weren't in commit messages, or edit them. (Although I don't think we've ever done a manual edit 😝 )

ksurl:
looks like github releases can also be edited after the fact.

gazpachoking:
It would be nice if we could have one changelog that got copied in one
direction or another on release, but something is better than nothing, so
I'm not too concerned. If we find a clean way to embed log from wiki into
release in the future, is like to switch to that.

ksurl:
I will look at it tomorrow but I think it's possible to wrangle this to handle both and replace the wiki changelog workflow

PR #3578, opened 2022-10-20
[fix]: use existing script to generate github release changelog

ksurl:
### Motivation for changes:
Add changelog to github releases as well as the wiki

### Detailed changes:
- Dropped changelog action
- Use update-changelog.py to create additional ChangeLog.md file in main repo

#### To Do:

- verify changelog on next release

### DIFF
--- a/.github/workflows/changelog.yml
+++ b/.github/workflows/changelog.yml
@@ -40,10 +40,16 @@ jobs:
     - name: Update the Changelog
       run: |
         set -x
-        python update-changelog.py wiki/ChangeLog.md
+        python update-changelog.py wiki/ChangeLog.md ChangeLog.md
         cd wiki
         git add ChangeLog.md
         if ! git diff --cached --exit-code; then
             git commit -m "Automated ChangeLog update"
             git push origin main
+        fi
+        cd ..
+        git add ChangeLog.md
+        if ! git diff --cached --exit-code; then
+            git commit -m "Automated ChangeLog update"
+            git push origin develop
         fi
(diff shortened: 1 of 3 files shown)


gazpachoking:
Isn't this going to be the entire changelog, and not just the bit for the current  version?

ksurl:
I left the existing file so that commits to the wiki repo with the full page. The 2nd file write is a new file in this repo with `cur_ver` (which I hope is just the current version changelog). I assume it is since it's between the comment lines in the 1st file write.

gazpachoking:
I don't think I'm a fan of commiting this to the main repo. I'd rather have it pull from the changelog at release time.

gazpachoking:
I think let's make a command in dev_tools that pulls the relevant section of the changelog.

ksurl:
What do you think of writing a 2nd file in the wiki then? Something like changelog_diff.md. My main concern is edge cases not covered by the regex. Having the changelog output be saved secondarily eliminates this.

gazpachoking:
What do you think about this? I haven't tested anything, see if any problems pop out at you.
https://github.com/Flexget/Flexget/compare/658f88619e49def5060e8006bc2409fa68ca6c30...develop

ksurl:
looks good. python will read # without needing to escape anything?

gazpachoking:
Yep, it's just reading the change log as a text file, so no characters will have special meanings. I did test the cli tool to get the changelog items, and that was working. The bit I'm less sure of is the github action workflow changes.

gazpachoking:
Alternate method is now working to add changelog to releases.

PR #3593, opened 2022-11-03
Python 3.11 Support

gazpachoking:
### Motivation for changes:
Support FlexGet on python 3.11

#### To Do:

- [ ] Stuff..


gazpachoking:
Hmm. Was it that easy?

PR #3604, opened 2022-11-06
Add activityDate and isStalled to transmission input

sarendsen:
### Motivation for changes:
Some useful fields for cleaning up stalled torrents.


gazpachoking:
Thanks!

PR #3628, opened 2022-12-06
Fix icon path in `toast` notifier plugin

skrattaren:
### Motivation for changes:

FlexGet's favicon is installed to `<FLEXGET-UI-PATH>/v1/app/favicon.ico`, but `toast` plugin expects it at `<FLEXGET-UI-PATH>/src/favicon.ico`


paranoidi:
thanks for the second fix too :)

PR #3629, opened 2022-12-06
Fix handling of empty `url` in *toast* on macOS

skrattaren:
### Motivation for changes:
`pync` package cannot handle empty string as an argument for `open`, so is `url` isn't configured for *toast*, no notification was displayed.


paranoidi:
Thanks for the fix

PR #3637, opened 2022-12-16
tmdb_lookup release dates

lostb1t:
Initial WIP of adding release dates to the tmdb lookup plugin. 
Few notes:

- Added `timed_lru_cache` for lru_cache with cache expiration tho might be better to cache to disk or db to be more memory friendly
- No estimator in this PR. As it doesn’t make a lot of sense because the est_movie_release is already looking at the primary release date of tmdb. The primary release date includes theatric releases. It would actually be helpfull to be able to define a type of release(s) for estimators.


gazpachoking:
Wow, sorry about my lack of response here. I'm thinking if we add this, we should store it directly in our db, rather than relying on a memory cache.

Issue #3619, opened 2022-11-29
sftp_list not handling known_hosts correctly


kingamajick:
### Expected behaviour:

Successfully open a SFTP connection `No hostkey for host $host found.`

### Actual behaviour:

Failure opening SFTP connection

### Steps to reproduce:
- Step 1: Run docker container `docker build -t flexget . && docker run -v "some/path/to/flexget:/home/flexget/.flexget" -v "some/path/to/ssh:/home/flexget/.ssh" -it flexget`

The root of this problem appears to be that pysftp doesn't [support port numbers in known_hosts](https://stackoverflow.com/a/64439662/131693), and is also appears to be abandoned (last update made in 2016). I tried the work around of not defining the port number in known hosts, but ran into `unpack requires a buffer of 4 bytes`

The `known_hosts` file is valid, `ssh-keygen -l -f ~/some/path/to/ssh/known_hosts -F $host` lists the fingerprints correctly.

Apparently [Paramiko](https://www.paramiko.org/) on which pysftp is based doesn't have these issues, so potentilly converting [sftp_client](https://github.com/Flexget/Flexget/blob/develop/flexget/components/ftp/sftp_client.py) to use Paramiko directly would be a fix.

#### Config:
```yaml
templates:
  bar:
    sftp_list:
     host: '{? sftp.host ?}'
      port: '{? sftp.port ?}'
      username: '{? sftp.username ?}'
      private_key: '{? sftp.private_key ?}'
      recursive: True
      get_size: True
      files_only: False
      dirs:
        - "/srv/file"

tasks:

  foo task:
    template:
      - bar
```
  
#### Log:
<details>
 <summary>(click to expand)</summary>

Salient parts of the logs

With port defined in known_hosts
```
2022-11-29 10:55:23 DEBUG    sftp_client   foo task        Connecting to $host
2022-11-29 10:55:23 DEBUG    sftp_client   foo task        Caught exception: No hostkey for host $host found.
2022-11-29 10:55:23 WARNING  sftp_client   foo task        Failed to connect to $host; waiting 15 seconds before retrying.
```

Without port defined 'workaround' in known_hosts
```
2022-11-29 10:57:27 DEBUG    sftp_client   foo task        Connecting to $host
2022-11-29 10:57:27 DEBUG    sftp_client   foo task        Caught exception: unpack requires a buffer of 4 bytes
[...]

PR #3641, opened 2022-12-22
Add suite of test cases to the sftp_{list, download, upload} plugins.

kingamajick:
### Motivation for changes:
Test suite missing for sftp plugins

### Detailed changes:
This change uses a local test sftp server to assert the behaviour of the various sftp plugins. By using an actually test server it opens the door to moving to Paramiko if needed (see https://github.com/Flexget/Flexget/issues/3619) while ensuring that the current functionality doesn't break.

Note: In addition it also enables a host hey to be specified via the plugin config for the list and upload plugins. This makes it possible to use the test sftp server without adding an entry to known_hosts.

First step towards fixing https://github.com/Flexget/Flexget/issues/3619

Following config added for list and upload plugins.
```
host_key:
    key_type: 'ssh-rsa'
    public_key: 'base64_encoded_host_key'
               
```


gazpachoking:
Excellent, thanks!

PR #3645, opened 2022-12-27
If a sftp_download entry represents a symlink and delete_origin is true, only remove the symlink.

kingamajick:
### Motivation for changes:
Currently if a entry represents a symlink and delete_origin is true, the target of the symlink gets removed, but not the symlink itself. This changes the behaviour such that if an entry represents a symlink, only the symlink is removed.

### Detailed changes:
If a sftp_download entry represents a symlink and delete_origin is true, only remove the symlink.

Currently is a entry represents a symlink, the target of the symlink gets removed rather than the symlink itself. This change corrects that and ensure that in the case of a symlink entry, only the symlink itself gets removed.

This also contains a fix for the test_sftp_server to use absolute() rather than resolve() when canonizing paths. Without this change lstat (which shouldn't follow symlinks which resolve does) fails to work correctly.

### DIFF
--- a/flexget/components/ftp/sftp_client.py
+++ b/flexget/components/ftp/sftp_client.py
@@ -5,6 +5,7 @@ from base64 import b64decode
 from dataclasses import dataclass
 from functools import partial
 from pathlib import Path, PurePath, PurePosixPath
+from stat import S_ISLNK
 from typing import Callable, List, Optional
 from urllib.parse import quote, urljoin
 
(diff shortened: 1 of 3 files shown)


kingamajick:
Just need to hold off on this request for now, it appears that the sftp_list plugin uses normalise to follow symlinks in the first place (i missed this in my tests cases as it uses the symlink name as the title. I'm going to update the sftp_list to be configurable to not follow symlinks and roll it into this change to make it more complete.

PR #3647, opened 2022-12-31
[add] Flood plugin

Winter:
### Motivation for changes:
Wanted to interface with [Flood](https://github.com/jesec/flood). Originally created a plugin that I've been using for 18 monthes at this point but the PR went stale. 
See #3072.

### Detailed changes:
- Adds new dependency `pyjwt`, for parsing Flood's JWT token. 
- Adds two new plugins, `flood` and `from_flood` for interfacing with Flood.

### Config
```
flood:
  url: '{? flood.url ?}'
  username: '{? flood.username ?}'
  password: '{? flood.password ?}'
  action: 'add'
  path: '/download/'
  tags:
    - Flexget
    
from_flood:
  url: '{? flood.url ?}'
  username: '{? flood.username ?}'
  password: '{? flood.password ?}'
```

#### To Do:

- [x] Download action for OutputFlood

### DIFF
--- a/flexget/plugins/clients/flood.py
+++ b/flexget/plugins/clients/flood.py
@@ -218,9 +218,9 @@ class InputFlood:
 
     Example:
         from_flood:
-            url: http://localhost:3000
-            username: username
-            password: password
+            url: http://localhost:3000  # Required. Url for Flexget to connect to Flood.
+            username: flexget           # Required. Username for authentication with Flood.
+            password: password          # Required. Password for authentication with Flood.
     """
 
     schema = {
@@ -231,7 +231,7 @@ class InputFlood:
             'password': {'type': 'string'},
         },
         'additionalProperties': False,
-        'required': ['url'],
+        'required': ['url', 'username', 'password'],
     }
 
     def generate_entry(self, config: dict, torrent: dict) -> Entry:
@@ -268,12 +268,12 @@ class OutputFlood:
 
     Example:
         flood:
-            url: http://localhost:3000
-            username: username
-            password: password
-            path: /downloads/Flexget
-            tags: Flexget
-            action: add
+            url: http://localhost:3000  # Required. Url for Flexget to connect to Flood. 
+            username: flexget           # Required. Username for authentication with Flood.
+            password: password          # Required. Password for authentication with Flood.
+            action: add                 # Required. The action to perform. Can be 'add', 'remove', 'delete', 'start', or 'stop'.
+            path: /downloads            # If the action is set to 'add', the path is relative to the Flood download directory.
+            tags: [ 'Flexget' ]         # If the action is set to 'add', the tags to add to the torrent.
     """
 
     schema = {
@@ -287,7 +287,7 @@ class OutputFlood:
             'tags': {'type': 'array', 'items': {'type': 'string'}},
         },
         'additionalProperties': False,
-        'required': ['url', 'action'],
+        'required': ['url', 'username', 'password', 'action'],
     }
 
     def _add_entry_file(
--- a/flexget/plugins/clients/flood.py
+++ b/flexget/plugins/clients/flood.py
@@ -1,9 +1,10 @@
 import base64
 import os
 import time
+import urllib.parse
 
 from loguru import logger
-from requests import get, post
+from requests import Response
 from requests.exceptions import RequestException
 
 from flexget import plugin
@@ -23,56 +24,66 @@ logger = logger.bind(name='flood')
 
 
 class FloodClient:
-    def __init__(self) -> None:
-        self.jwt = ""
+    @staticmethod
+    def request(task: Task, config: dict, method: str, url: str, success_list: list, **kwargs) -> Response:
+        """
+        Send a request to Flood.
+        Returns None if the response status_code is not in the success_list.
+        """
 
-    def is_jwt_expired(self) -> bool:
+        try:
+            response = task.requests.request(method, urllib.parse.urljoin(config["url"], url), **kwargs)
+        except RequestException as e:
+            raise PluginError(f"An error occurred while attempting to send a request to Flood: {e}")
+
+        if not response.status_code in success_list:
+            raise PluginError(f"Failed to send request to Flood. Received status {response.status_code} expected one of {success_list}.")
+        
+        return response
+
+    @staticmethod
+    def is_jwt_expired(task: Task) -> bool:
         """
         Check if the JWT token is expired.
         Returns True if the JWT token is expired or if the JWT token is not set.
         """
 
-        if not self.jwt:
+        jwt = task.requests.cookies.get('jwt')
+
+        if not jwt:
             return True
 
         try:
-            decoded_jwt = decode(self.jwt, algorithms="HS256", options={"verify_signature": False})
+            decoded_jwt = decode(jwt, algorithms="HS256", options={"verify_signature": False})
         except:
             return True
 
         return decoded_jwt.get("exp", 0) <= time.time()
 
-    def authenticate(self, config: dict) -> None:
+    @staticmethod
+    def authenticate(task: Task, config: dict) -> None:
         """
         Authenticate with Flood and obtain a JWT token.
         """
 
-        try:
-            response = post(
-                f"{config['url']}/api/auth/authenticate",
-                json={"username": config["username"], "password": config["password"]},
-            )
+        # Attempt to authenticate with Flood.
+        FloodClient.request(task, config, "post", "api/auth/authenticate", success_list=[200], json={"username": config["username"], "password": config["password"]})
 
-            if response.status_code == 200 and response.cookies.get("jwt"):
-                self.jwt = response.cookies.get("jwt")
-                logger.debug("Successfully authenticated with Flood.")
-            else:
-                raise PluginError(
-                    f"Failed to authenticate with Flood. (status={response.status_code})"
-                )
-        except RequestException as e:
-            raise PluginError(
-                f"An error occurred while attempting to authenticate with Flood: {e}"
-            )
+        # Check if the JWT token is expired or unset.
+        if FloodClient.is_jwt_expired(task):
+            raise PluginError("Failed to authenticate with Flood. No valid JWT was found.")
+        
+        logger.debug("Successfully authenticated with Flood.")
 
+    @staticmethod
     def add_torrent_urls(
-        self,
+        task: Task,
         config: dict,
         urls: list,
         destination: str = None,
         tags: list = None,
         start: bool = True,
-    ) -> bool:
+    ) -> None:
         """
         Add a list of torrent URLs to Flood.
         Will only attempt to add torrents if the JWT token is not expired or unset.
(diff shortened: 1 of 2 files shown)


flexget/plugins/clients/flood.py
```diff
@@ -0,0 +1,442 @@
...
+            data["destination"] = destination
+        if tags:
+            data["tags"] = tags
+
+        try:
+            response = post(
+                f"{config['url']}/api/torrents/add-files", json=data, cookies={"jwt": self.jwt}
+            )
+
+            if response.status_code == 200:
```
ianstalk, line 142: I'd consider using `response.ok` instead unless `200` is really the only successful response. Also it kind of muddies the try/except block to have this here. I would move everything but getting the response out.

flexget/plugins/clients/flood.py
```diff
@@ -0,0 +1,442 @@
...
+        Will only attempt to list torrents if the JWT token is not expired or unset.
+        """
+
+        if self.is_jwt_expired():
+            raise PluginError("Not authenticated with Flood.")
+
+        try:
+            response = get(f"{config['url']}/api/torrents", cookies={"jwt": self.jwt})
+
+            if response.status_code == 200:
```
ianstalk, line 167: Same feedback as above

flexget/plugins/clients/flood.py
```diff
@@ -0,0 +1,442 @@
...
+
+class InputFlood:
+    """Creates entries from torrents in Flood based
+
+    Example:
+        from_flood:
+            url: http://localhost:3000
+            username: username
+            password: password
+    """
```
ianstalk, line 275: Maybe add descriptions for the config values? I know they're pretty self-explanatory but it lets people know what the options are.

flexget/plugins/clients/flood.py
```diff
@@ -0,0 +1,442 @@
...
+    """Add, remove, delete, start, or stop torrents in Flood.
+
+    Example:
+        flood:
+            url: http://localhost:3000
+            username: username
+            password: password
+            path: /downloads/Flexget
+            tags: Flexget
+            action: add
```
ianstalk, line 330: Again listing the config values and what they represent would be nice.

flexget/plugins/clients/flood.py
```diff
@@ -0,0 +1,442 @@
...
+
+        if task.manager.options.test:
+            logger.info('Would add {} to Flood with options:', entry['title'])
+
+            if entry.get('file'):
+                logger.info('  File: {}', entry['file'])
+            elif entry.get('url'):
+                logger.info('  Url: {}', entry['url'])
+            else:
+                logger.info('  No URL or File found.')
```
ianstalk, line 376: I'd consider extracting this to a helper method to reduce nesting.

flexget/plugins/clients/flood.py
```diff
@@ -0,0 +1,442 @@
...
+
+            logger.info('  Destination: {}', destination)
+            logger.info('  Tags: {}', ', '.join(tags) or 'None')
+            logger.info('  Start: {}', config.get('start', True))
+            return
+
+        if entry.get('file'):
```
ianstalk, line 383: Same comment as above.

flexget/plugins/clients/flood.py
```diff
@@ -0,0 +1,442 @@
...
+                self.add_entry(config, task, entry)
+            elif 'flood_hash' in entry:
+                if config['action'] == 'remove':
+                    self.flood.delete_torrents(config, [entry['flood_hash']], False)
+                elif config['action'] == 'delete':
+                    self.flood.delete_torrents(config, [entry['flood_hash']], True)
+                elif config['action'] == 'start':
+                    self.flood.start_torrents(config, [entry['flood_hash']])
+                elif config['action'] == 'stop':
+                    self.flood.stop_torrents(config, [entry['flood_hash']])
```
ianstalk, line 436: Consider extracting this to remove the nested if/else.

ianstalk (commented):
This looks pretty good overall. Some minor comments.

flexget/plugins/clients/flood.py
Winter: I actually did not know that you can create variables inside a  block and use them outside of the block (without first declaring it outside the block). It makes sense given the way Python works, and that you're not really "creating" variables. 

I was actually going to write here that 200 is the only acceptable answer as it can return 207 which indicates that some succeeded and some failed but it can also respond with 202 which is also acceptable, so I will need to allow only 200 and 202. Good thing I double checked!

The other thing that I did have here was additional logging that a 207 occurred, but did remove it.

ianstalk: If that's the case then I'd make a function or method to check the status code so you don't repeat yourself.

flexget/plugins/clients/flood.py
Winter: Extracted this into it's own method in winter/Flexget@584f9b7

flexget/plugins/clients/flood.py
Winter: I decided to keep the test option logging as it was given that it's so small and the logic being quite simple.

flexget/plugins/clients/flood.py
Winter: Introduced new method in Winter/Flexget@79fae7e for creating requests using the task requests session. It requires a list of successful status codes otherwise it will raise a `PluginError` halting the plugin.

flexget/plugins/clients/flood.py
Winter: Introduced new method in https://github.com/Winter/Flexget/commit/79fae7e for creating requests using the task requests session. It requires a list of successful status codes otherwise it will raise a PluginError halting the plugin.

flexget/plugins/clients/flood.py
Winter: Added descriptions to module docstring examples in Winter/Flexget@c067577

flexget/plugins/clients/flood.py
Winter: Added descriptions to module docstring examples in Winter/Flexget@c067577

Winter:
Whoops! I didn't add Winter/Flexget@c067577 to my push earlier, hence weird merge. Thanks VS Code.

flexget/plugins/clients/flood.py
```diff
@@ -0,0 +1,446 @@
...
+
+        if not response.status_code in success_list:
+            raise PluginError(
+                f"Failed to send request to Flood. Received status {response.status_code} expected one of {success_list}."
+            )
+
+        return response
+
+    @staticmethod
+    def is_jwt_expired(task: Task) -> bool:
```
ianstalk, line 53: `is_jwt_valid` seems like a better name here, since it checks both existence and expiration.

flexget/plugins/clients/flood.py
```diff
@@ -0,0 +1,446 @@
...
+            self.add_entry(task, config, entry)
+        elif 'flood_hash' in entry:
+            if config['action'] == 'remove':
+                FloodClient.delete_torrents(task, config, [entry['flood_hash']], False)
+            elif config['action'] == 'delete':
+                FloodClient.delete_torrents(task, config, [entry['flood_hash']], True)
+            elif config['action'] == 'start':
+                FloodClient.start_torrents(task, config, [entry['flood_hash']])
+            elif config['action'] == 'stop':
+                FloodClient.stop_torrents(task, config, [entry['flood_hash']])
```
ianstalk, line 373: Maybe log a warning or throw an exception if `'flood_hash'` isn't available?

flexget/plugins/clients/flood.py
```diff
@@ -0,0 +1,446 @@
...
+        data = {"urls": urls, "start": start}
+
+        if destination:
+            data["destination"] = destination
+        if tags:
+            data["tags"] = tags
+
+        # Attempt to add the torrent URLs to Flood.
+        response = FloodClient.request(
+            task, config, "post", "api/torrents/add-urls", success_list=[200, 202], json=data
```
ianstalk, line 121: You could probably make the list of acceptable response codes a named `Final` variable at the top of the file for readability (and maybe a tiny performance boost).

flexget/plugins/clients/flood.py
```diff
@@ -0,0 +1,446 @@
...
+        try:
+            FloodClient.add_torrent_urls(task, config, [entry['url']], destination, tags, start)
+        except PluginError as e:
+            return entry.fail(e)
+
+    def _process_accepted_entry(self, task: Task, config: dict, entry: Entry) -> None:
+        if config['action'] == 'add':
+            self.add_entry(task, config, entry)
+        elif 'flood_hash' in entry:
+            if config['action'] == 'remove':
```
ianstalk, line 366: A pity you can't use `match` here, but it's only available in Python 3.10 and newer

ianstalk (commented):
A few more minor points, but this looks pretty good to me.

flexget/plugins/clients/flood.py
Winter: I did originally have a check to see if the torrent exists before trying to perform any of the actions that require a hash, but Flood will instead throw a status 500, I figured that it would be picked up by L37's try-catch block

flexget/plugins/clients/flood.py
Winter: I like it, does make a lot more sense!

flexget/plugins/clients/flood.py
Winter: I guess the `[ 200, 202 ]` is acceptable under every scenario. I'll do that.

flexget/plugins/clients/flood.py
Winter: Unless you've got any other idea's for how I can handle different actions, if-else is probably here to stay.. haha

flexget/plugins/clients/flood.py
Winter: Renamed `is_jwt_expired` to `is_jwt_invalid` in Winter/Flexget@b3cb221

flexget/plugins/clients/flood.py
Winter: Added a entry.fail if no `flood_hash` was found in Winter/Flexget@6f32337

paranoidi:
Looking good! Maybe add docstrings to couple methods with altered priority why those were needed.

Winter:
> Looking good! Maybe add docstrings to couple methods with altered priority why those were needed.

Added in Winter/Flexget@950eee5

flexget/plugins/clients/flood.py
```diff
@@ -0,0 +1,531 @@
...
+
+        if not response.status_code in FloodClient.success_list:
+            raise PluginError(
+                f"Failed to send request to Flood. Received status {response.status_code} expected one of {FloodClient.success_list}."
+            )
+
+        return response
+
+    @staticmethod
+    def is_jwt_invalid(task: Task) -> bool:
```
ianstalk, line 53: I'm not in love with negative conditionals like this. I'd probably make it `is_jwt_valid()` and return true for the positive case but I suppose that's a matter of taste.

Winter: Yeah I don't like it either.. I'm not sure what was going through my head if I'm being completely honest. I think I just pressed F2 and went easy peasy without a second thought.

Winter:
Decided it's not worth caring about the PR. Opted to instead just add an overides folder to my Docker setup rather than building my own container and keeping it up to date.

kronibos:
> Decided it's not worth caring about the PR. Opted to instead just add an overides folder to my Docker setup rather than building my own container and keeping it up to date.

will u ever finish the plugin?

PR #3650, opened 2023-01-03
[change] estimate_release - make configurable

### LINKED ISSUE #3614: Feature request: a combination of loose and optimistic for release estimations
https://flexget.com/Plugins/discover

> loose will check for release dates, if release date can not be determined perform search anyway.
> optimistic sets the estimation mode to strict but starts searching interval units (eg. 7 days) before release date.

The recommended option is `optimistic` but this will prevent flexget from searching for movies that don't have a release date on blu-ray.com. This should be covered by `loose` - it will perform the search anyway but apparently honors the release date reported by blu-ray.com - and I've observed that sometimes releases are available much sooner than that.
Since neither of these options suit me I'm forced to use `ignore` which will make flexget search for movies which have obviously not yet released - making dozens of unnecessary requests.

Ideally I would like to have a combination of loose and optimistic - i.e. have `loose` accept interval units just like `optimistic` does.

aidan-:
### Motivation for changes:
Primarily driven by the feature request in #3614.

### Detailed changes:
Before this PR, the `estimate_release` plugin, used by other plugins like the `discover` plugin, was not user configurable.  Some config in the `discover` plugin allowed users to config modify the estimated release, but the functionality was limited to specific modes of operation.

Rather than complicating the `discover` plugin with logic to control what should be the responsibility of the `estimate_release` plugin, this PR extends the `estimate_release` plugin to allow some basic configuration - such as modifying the calculated estimated release date and restricting which estimator providers to use.

This PR also removes the `optimistic` mode from `discover` as that functionality is now supported through the configuration of `estimate_release` across all modes.

- Fixes #3614

```
        tasks:
          test_estimates:
            discover:
              interval: 0 seconds
              what:
              - mock:
                - title: Foo
              from:
              - test_search: yes
            estimate_release:
              offset: -7 days
              providers: ["movies_bluray"]
```
#### To Do:
- [ ] Update wiki
- [ ] Further integration testing if the approach is on track

### DIFF
--- a/flexget/components/estimate_release/estimate_release.py
+++ b/flexget/components/estimate_release/estimate_release.py
@@ -39,7 +39,7 @@ class EstimateRelease:
 
     @property
     def schema(self) -> Dict[str, Any]:
-        """Create schema for that allows configuring estimator providers and
+        """Create schema that allows configuring estimator providers and
         related settings.
         """
 
--- a/flexget/components/estimate_release/estimate_release.py
+++ b/flexget/components/estimate_release/estimate_release.py
@@ -22,9 +22,8 @@ task_estimate_config: Dict[str, Any] = {}
 def init_estimators(manager) -> None:
     """Prepare the list of available estimator plugins."""
 
-    estimators = {
-        p.name.replace('est_', ''): p for p in plugin.get_plugins(interface=ESTIMATOR_INTERFACE)
-    }
+    for provider in plugin.get_plugins(interface=ESTIMATOR_INTERFACE):
+        estimators[provider.name.replace('est_', '')] = provider
 
     logger.debug('setting default estimators to {}', list(estimators.keys()))
 
@@ -34,6 +33,9 @@ class EstimateRelease:
     Front-end for estimator plugins that estimate release times
     for various things (series, movies).
     """
+
+    def __init__(self):
+        self.task_estimate_config = {}
 
     @property
     def schema(self) -> Dict[str, Any]:
@@ -83,12 +85,16 @@ class EstimateRelease:
         """
         if "providers" in self.task_estimate_config:
             # Map task configured providers to plugin instance map
-            task_estimators = [
-                estimators[p].instance.estimate for p in self.task_estimate_config['providers']
-            ]
+            try:
+                task_estimators = [
+                    estimators[p].instance.estimate for p in self.task_estimate_config['providers']
+                ]
+            except KeyError as error:
+                logger.error(f"invalid provider plugin given: {error}")
+                raise
         else:
             # Use all loaded estimator plugins
-            task_estimators = [e.instance.estimate for e in estimators]
+            task_estimators = [e.instance.estimate for e in estimators.values()]
 
         return sorted(
             task_estimators,
--- a/flexget/components/estimate_release/estimate_release.py
+++ b/flexget/components/estimate_release/estimate_release.py
@@ -1,9 +1,32 @@
+from datetime import datetime
+from typing import Any, Dict, List, Optional, Union
+
 from loguru import logger
 
 from flexget import plugin
 from flexget.event import event
+from flexget.plugin import PluginInfo
+from flexget.task import Task
+from flexget.utils.tools import parse_timedelta
 
 logger = logger.bind(name='est_released')
+ESTIMATOR_INTERFACE = "estimate_release"
+
+# Mapping of available estimate providers to plugin instance
+estimators = {}
+# Task specific estimate configuration
+task_estimate_config: Dict[str, Any] = {}
+
+# We need to wait until manager startup to access other plugin instances, to make sure they have all been loaded
+@event('manager.startup')
+def init_estimators(manager) -> None:
+    """Prepare the list of available estimator plugins."""
+
+    estimators = {
+        p.name.replace('est_', ''): p for p in plugin.get_plugins(interface=ESTIMATOR_INTERFACE)
+    }
+
+    logger.debug('setting default estimators to {}', list(estimators.keys()))
 
 
 class EstimateRelease:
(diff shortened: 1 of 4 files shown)


paranoidi:
I'm a bit torn how this adds the release estimation configuration to root level of the plugin. Although it's definitely not a deal breaker here. 

I might prefer if it was value for each mode instead, like with `optimistic` option was but being optional.

Good work though so far :)

Any thoughts from other coders and users?

gazpachoking (commented):
Yeah, I think this is an improvement from what we have.

It made me realize though, that we have a modular system for estimating releases, but at the moment the user has no control over said system. I wonder if we should add some contol over that similar to how we did for the parsing system. Such that a task could look something like:
```yaml
discover:
  what: blah
  from: blah
release_estimation:
  providers: [bluray, tvmaze]
  adjustment: -7 days
```
That would probably be a task for a separate PR though.

paranoidi:
> Yeah, I think this is an improvement from what we have.
> 
> It made me realize though, that we have a modular system for estimating releases, but at the moment the user has no control over said system. I wonder if we should add some contol over that similar to how we did for the parsing system. Such that a task could look something like:
> 
> ```yaml
> discover:
>   what: blah
>   from: blah
> release_estimation:
>   providers: [bluray, tvmaze]
>   adjustment: -7 days
> ```
> 
> That would probably be a task for a separate PR though.

That makes much more sense and I like the configuration format much more as it moves all estimation related configuration under release_estimations key. Better for context, better name (adjustment). If we want to go with that route though, then this PR should probably not be merged at all.

@aidan- Would you be up for implementing what gazpachoking outlined?

flexget/plugins/input/discover.py
```diff
@@ -292,12 +289,22 @@ def interval_expired(self, config, task, entries):
...
-        config['release_estimations'].setdefault('mode', 'strict')
-        config['release_estimations'].setdefault('optimistic', '0 days')
+        if not isinstance(config['release_estimations'], dict):
+            # 'optimistic' mode is being deprecated to add support for modifying the
+            # resultant release date through a consistent modifier used by all modes.
+            if 'optimistic' in config['release_estimations']:
+                # Identify configurations that have both the deprecated optimistic
+                # mode enabled and the new configuration option.
+                if config['release_date_modifier'] != '0 days':
+                    logger.critical("optimistic is deprecated - remove conflictig configuration.")
```
BrutuZ, line 301: Small typo here: conflictig > conflicting
Maybe give a little more context on "optimistic" telling it's about release estimations?

flexget/plugins/input/discover.py
```diff
@@ -90,15 +90,17 @@ class Discover:
                     {
                         'type': 'string',
                         'default': 'strict',
-                        'enum': ['loose', 'strict', 'ignore', 'smart'],
+                        'enum': ['loose', 'strict', 'ignore', 'smart', 'optimistic'],
```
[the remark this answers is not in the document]
aidan-: Good spot, left over from the initial refactor before I realized that `optimistic` is no longer required if we can do the date modification on all the other modes.

flexget/plugins/input/discover.py
```diff
@@ -292,12 +289,22 @@ def interval_expired(self, config, task, entries):
...
+                # Identify configurations that have both the deprecated optimistic
+                # mode enabled and the new configuration option.
+                if config['release_date_modifier'] != '0 days':
+                    logger.critical("optimistic is deprecated - remove conflictig configuration.")
+                else:
+                    # Convert deprecated optimistic intervals to new property by
+                    # negating it.
+                    config[
+                        'release_date_modifier'
+                    ] = f"-{config['release_estimations']['optimistic']}"
```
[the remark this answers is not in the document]
aidan-: 🤦 I only did happy path integration testing - I'll add some more unit tests and do more integration tests.

paranoidi:
> Yeah for sure! I think that would look much cleaner and more logical.

Awesome, I think we all are in agreement. Waiting for PR updates when you have the time :)

paranoidi:
Another possible name for the option would be "offset", which I kinda like even more.

BrutuZ:
> Another possible name for the option would be "offset", which I kinda like even more.

Maybe `estimator_offset` to keep _what_ it is offsetting clear, otherwise it could be thought to be affecting the `interval` setting of `discover`

aidan-:
@paranoidi bit slow on this one, real life got in the way.  I still need to do some integration testing to make sure its all functional but figured I'd get some eyes on it now to make sure its in the right direction.

flexget/components/estimate_release/estimate_release.py
```diff
@@ -12,32 +34,105 @@ class EstimateRelease:
     for various things (series, movies).
     """
 
-    def estimate(self, entry):
+    def __init__(self):
+        self.task_estimate_config = {}
+
+    @property
+    def schema(self) -> Dict[str, Any]:
```
paranoidi, line 41: This is not the way to handle sub plugins & interfaces, please look at how discover plugin handles search interface.

flexget/components/estimate_release/estimate_release.py
```diff
@@ -12,32 +34,105 @@ class EstimateRelease:
...
+        return schema
+
+    def on_task_start(self, task: Task, config) -> None:
+        # Load task specific estimator configuration
+        if config:
+            self.task_estimate_config = config
+
+    def on_task_exit(self, task: Task, config) -> None:
+        # Restore default estimator configuration for next task run
+        self.task_estimate_config = {}
```
paranoidi, line 73: This modifying configuration and restoring later is not a good idea. Generally speaking plugins should be stateless and not rely on __init__.

paranoidi (changes requested):

flexget/components/estimate_release/estimate_release.py
aidan-: Interesting, this is how the `parsing` plugin does it which was mentioned in one of the above comments as a reference.

flexget/components/estimate_release/estimate_release.py
aidan-: Sure.  I'll take a look.

flexget/components/estimate_release/estimate_release.py
paranoidi: Hmm, apologies. I guess there's no better way for it in these kinds of situations then. There is still issue with having `self.task_estimate_config` and the one declared in module level (L18). I don't think the module one is used and can be removed.

gazpachoking: Yeah, we have to do some hacky stuff for bits that don't directly get a task parameter, there isn't a better way at the moment. I was thinking of making a plugin method that worked like a context manager, to make plugins that do setup and teardown around tasks a bit easier. Something like:
```python
def on_task_context(self, task, config):
    if config:
        self.task_estimate_config = config
    try:
        yield
    finally:
        self.task_estimate_config = {}
```
That way you don't have to hook task_start, task_exit, and task_abort trying to manage state. @paranoidi Thoughts?

I also had this crazier idea based on that, but my gut tells me it's probably bad. It would allow a plugin to retain state during a task without storing things to self.
```python
def on_task_phases(self, task, config):
    yield 'start'
    setupstuff = somefunc()
    yield 'filter'
    for entry in task.all_entries:
        do_something(entry, setupstuff)
    yield 'exit'
    teardown(setupstuff)
```

paranoidi: Heh, sure. I have some thoughts. Not overly sold on the idea. 

I've been thinking years that we should have a baseclass for plugins and in there have a `prepare_config` method that plugins can override to format config with default values etc. Then passing the already prepared config into the usual phase methods. Also this would make it possible to declare the phase methods in the baseclass making them much more explicit. Last time I checked there was only one dynamically added phase which would not be a hard to get rid of.

gazpachoking: > I've been thinking years that we should have a baseclass for plugins and in there have a `prepare_config` method that plugins can override to format config with default values etc. Then passing the already prepared config into the usual phase methods.

Yeah, I love this idea.
I think it's totally separate than a context manager type thing to do cleanup, and I'm not totally sold on that idea either. The problem of doing setup and teardown in on_task_start/exit is that you can't guarantee that they run in pairs. I just fixed an issue with the parsing plugin where the on_task_abort cleanup ran even though the task was aborted before its on_task_start even set up the thing it was trying to clean up. You could also run into the issue of an abort happening after on_task_exit handler was run, and it tries to clean up twice. Here is a [naive search](https://github.com/search?q=repo%3AFlexget%2FFlexget+%22on_task_abort+%3D+on_task_exit%22&type=code) that shows some plugins that might be able to use a thing like the context manager hook.

paranoidi: Maybe the baseclass could introduce setup/teardown methods (although I kinda dislike those names as they are heavily unit test related) that are guaranteed to run as expected. It would be much more explicit than relying on on_task_start etc.

PR #3668, opened 2023-01-28
Update morethantv categories

spgomez:
Categories were obsolete. Updated to reflect the current six categories.

Morethantv plugin was not working correctly

### Detailed changes:
removed previous categories : TV, Movies, Other
added new categories: HD Episode, HD Movies, HD Season, SD Episode, SD Movies, SD Season


gazpachoking:
My only concern is that people's configs will break (not pass validation.) I guess they were a bit broken anyway without this, but not sure how we should handle. Maybe we just bump the version and add an upgrade action.

spgomez:
I couldn't get it working with the current category settings. Seems like MoreThanTv changed things when they re-launch. I'd be interested to see if anyone currently has the plugin working. 

I'm currenting using a filtered RSS feed and it's working just fine, so I'd understand if this wasn't something you felt like messing with.

Thanks for the feedback.

gazpachoking:
Merged in 2711281c177fd665f9dfe1d60521ba716fcfb7dd

gazpachoking:
Updated the wiki page and UpgradeActions page to mention the new categories.

PR #3693, opened 2023-02-19
Add the support of 'labels' for Transmission.

### LINKED ISSUE #3520: Adding labels to torrents in transmission
Since version 3.00 transmission now supports labels for torrents. They can be provided at add-time, see [Sec. 3.4 of the RPC spec](https://github.com/transmission/transmission/blob/main/docs/rpc-spec.md#32-torrent-mutator-torrent-set).

It would be neat if the transmission plugin accepted the `labels` keyword so Flexget could automatically label torrents., E.g. one could do
```yml
set:
    path: '/var/Downloads/{{series_name}}'
    label:
        - '{{series_name}}'
        - flexget
```

ProteinPig:
### Motivation for changes:
I need this
### Detailed changes:
- Add the support of 'labels' for Transmission.

- Fixes #3520 

```
    transmission:
      # transmission settings
      labels: 
        - test
        - flexget
```
```
transmission_isStalled         : False
transmission_labels            : ['test', 'flexget']
transmission_leftUntilDone     : 0
```

### DIFF
--- a/flexget/plugins/clients/transmission.py
+++ b/flexget/plugins/clients/transmission.py
@@ -226,6 +226,7 @@ class PluginTransmissionInput(TransmissionBase):
                 'secondsDownloading',
                 'secondsSeeding',
                 'torrentFile',
+                'labels',
             ]:
                 try:
                     value = getattr(torrent, attr)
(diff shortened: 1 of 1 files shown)


gazpachoking:
Hmm. I was thinking we should make it consistent with how deluge plugin works (deluge plugin uses 'label' rather than 'labels',) but I guess it's inherently different if transmission can have more than one label.

paranoidi:
Yeah, and qbittorrent is different too. It has category and tags. I think it's fine that they are differences per client, they're prefixed anyway.

paranoidi:
Good motivation for change ;)

Thanks for contributing!

PR #3694, opened 2023-02-20
[fix] sftp_list - check host_key before accessing keys of dict

aidan-:
### Motivation for changes:
`sftp_list` configurations that do not set `host_key` via Flexget are currently broken.

### Detailed changes:
- #3641 introduced the `host_key` configuration that allows the setting of the SFTP server host key via Flexget rather than relying on the system lookup.
- The `host_key` config key is defaulted to `None` and no check is currently done to ensure its populated before accessing sub-keys.  This causes an exception: `TypeError: 'NoneType' object is not subscriptable`.

- Fixes the above mentioned issue.

n/a

This is challenging to test, which is probably the whole point of #3641 doing the refactor (and adding the new config options) to make this more testable.


paranoidi:
Thanks!

PR #3706, opened 2023-03-04
Update sftp list and download symlink handling

kingamajick:
### Motivation for changes:

Previously it was impossible to get the path to a symlink as the sftp_list plugin would resolve the symlink to it's target (in the same way that sftp_download would delete the target rather than the symlink).

### Detailed changes:

This pull request contains two main changes, firstly, the sftp_list will return the path of symlinks rather than there targets. Secondly if a symlink is targeted to be deleted, the symlink will be remove rather than it's target deleted.


kingamajick:
Hi, do you have any feedback on the PR?

flexget/components/ftp/sftp_client.py
```diff
@@ -169,21 +178,25 @@ def download(self, source: str, to: str, recursive: bool, delete_origin: bool) -
...
             raise SftpError(f'Remote path does not exist: {source}')
 
+        is_symlink: bool = self.is_link(source)
         if self.is_file(source):
             source_file: str = parsed_path.name
             source_dir: str = parsed_path.parent.as_posix()
             try:
                 self._sftp.cwd(source_dir)
-                self._download_file(to, delete_origin, source_file)
+                self._download_file(to, delete_origin and not is_symlink, source_file)
```
ianstalk, line 187: I'd consider binding `delete_origin and not is_symlink` to a named variable to make it clearer what the combined value means.

flexget/tests/test_sftp_download.py
```diff
@@ -14,63 +14,57 @@ class TestSftpDownload:
     _config = """
         templates:
           anchors:
-            _sftp_list: &base_sftp_download
-                to: {{ download_path }}
-                socket_timeout_sec: 2
-                connection_tries: 1
+            _sftp_download: &base_sftp_download
```
ianstalk, line 17: Smart use of anchors to reduce repetition. If you wanted to get really fancy, I think anchors expand recursively, but this is probably a good balance of DRYness and simplicity.

ianstalk (approved):
A few minor things but overall LGTM

PR #3716, opened 2023-03-11
upgrade to transmission-rpc v4

trim21:
### Detailed changes:
- add transmission-rpc to deps, expected version will be installed with flexget

- Fixes # .


BrutuZ:
I vote against adding yet another dependency to the base install. It's fine for the dockerfile as overhead is expected.

Keep it as a note for its requirement on the plugin's page as already is the case for a few others

gazpachoking:
Well, first of all, great work on the update. I think I agree with @BrutuZ though on including it directly right now, especially since 3.7 isn't eol quite yet, and this would be our only reason to drop support. We have discussed a few times having plugins be able to declare their dependencies more formally, and have them be automatically installed when used. Would be really great if we could move to that system eventually. I think for now though we still leave this as optional, and document the requirement on the wiki page and in an error message when it's used without the proper dep.

trim21:
> We have discussed a few times having plugins be able to declare their dependencies more formally, and have them be automatically installed when used

I think pip extra requirements is for this...

gazpachoking:
> I think pip extra requirements is for this...

Yeah, we could add a pip extra to let an user install flexget[transmission] or something, but that doesn't feel any better than just having them install both dependencies separately. The real goal that I'm talking about would be to let someone just install flexget, but then when they include transmission plugin in their config flexget would automatically install (a compatible version of) transmission-rpc into their environment without any added steps.

> I can help upgrading to transmission-rpc v4 after you drop python3.7(EOL 06 Dec 2022) 🤔

I'm pretty sure the code in this branch is compatible with both transmission-rpc 3.x and 4.x? Couldn't we do this before adding the hard requirement of transmission-rpc 4.0 and python 3.7?

trim21:
> then when they include transmission plugin in their config flexget would automatically install (a compatible version of) transmission-rpc into their environment without any added steps.

Don't do that... Even add all deps to requirements is a better idea than install pip deps in python program ...

> I'm pretty sure the code in this branch is compatible with both transmission-rpc 3.x and 4.x?

I'm pretty sure it doesn't work with v3

gazpachoking:
> Don't do that... Even add all deps to requirements is a better idea than install pip deps in python program ...

The other idea I had was some sort of more explicit command like `flexget deps install` which did the installation, based on what plugins you had configured. I still am not sure about adding everything as explicit deps though no matter what our future better plan might be.

> I'm pretty sure it doesn't work with v3

Oh. I had only looked at a couple of attributes that were compatible with both 3.x and 4.x. Just noticed now that you are the author of transmission-rpc 😅 you probably know better than me!

trim21:
I backport transmission-rpc v4 to python3.7 now.

and also remove transmission-rpc from deps

gazpachoking:
Oh, excellent! I think this should make things much easier for our users finding the right combination of versions. I think with these changes we are good to merge and can just update all our docs to use transmission-rpc 4+
I guess we should bump our minor version and document the required dep update in an upgrade action.
Anyone have any concerns?

trim21:
add runtime package version check

paranoidi:
@trim21 Is this in a state where it should be merged? I think you know better than anyone :)

trim21:
yes

gazpachoking:
I think we should bump our minor flexget version, otherwise I'm 👍

trim21:
> I think we should bump our minor flexget version, otherwise I'm 👍

Yes

trim21:
> I think we should bump our minor flexget version, otherwise I'm 👍

Sorry, do you mean to bump version in this PR? I thought you will do it in trunk

gazpachoking:
> Sorry, do you mean to bump version in this PR? I thought you will do it in trunk

I bumped it in here. I think we are probably good to go now. We'll just put an upgrade action and update the wiki with the new version requirements.

gazpachoking:
I updated the wiki page as well as created an [upgrade action](https://flexget.com/en/UpgradeActions#h-360-2023-04-09) for this. Anyone should feel free to update more places if it's needed for more clarity.

gazpachoking:
Thanks for all your work on this (and transmission-rpc itself) @trim21 ! 👏

trim21:
@gazpachoking it require 4.1.4 or greater， not 4.0.0

gazpachoking:
Gotcha. Updated the wiki pages 👍

PR #3737, opened 2023-04-09
add qb package to Docker

trim21:
### Motivation for changes:

I'm using both qb and transmission…

### Detailed changes:
- add qbittorrent-api python package to docker image

- Fixes # .

```
paste_config_here
```
```
paste output here
```
#### To Do:

- [ ] Stuff..


trim21:
looks like dockerfile is not actually tested

gazpachoking:
I'm good with this.
It made me start thinking though, should we add a dependency group to the pyproject for optional deps that are included in the docker image rather than putting them directly in the dockerfile? Would mean that poetry could resolve the specific versions that get installed and make sure they are compatible with our other deps. @paranoidi @X-dark @ksurl ?

ksurl:
does that mean it will install at runtime? if it can pin versions rather than latest, it might be ok. unpredictable packages is against the philosophy of static images, but I'm sure other projects break this rule all the time if it works and is limited in scope.

gazpachoking:
No, not proposing it installs at runtime. We could either have it generate a docker-requirements.txt similar to how we have other requirements files (which the dockerfile would then use) or we could switch the dockerfile to install flexget with poetry ratther than pip. In any case, the deps versions would still be resolved at image creation time.

gazpachoking:
> looks like dockerfile is not actually tested

Yeah. Do you (or anyone) have a good idea of what we should test/how we should test it?

trim21:
> > looks like dockerfile is not actually tested
> 
> Yeah. Do you (or anyone) have a good idea of what we should test/how we should test it?

build and run flexget --help

ksurl:
> No, not proposing it installs at runtime. We could either have it generate a docker-requirements.txt similar to how we have other requirements files (which the dockerfile would then use) or we could switch the dockerfile to install flexget with poetry ratther than pip. In any case, the deps versions would still be resolved at image creation time.

Haha I actually made a local branch to pr a separate requirement file for docker but never pushed it.

trim21:
I'll just let poetry to export a requiremenets-docker.txt to be requirements.txt with docker group

trim21:
and why we install `nodejs` in docker image? which plugin need it?

.pre-commit-config.yaml
```diff
@@ -22,8 +22,8 @@ repos:
     - id: poetry-export
       args: ["--only", "dev", "-f", "requirements.txt", "-o", "dev-requirements.txt", "--without-hashes"]
     - id: poetry-export
-      args: ["--only", "plugin-test", "-f", "requirements.txt", "-o", "dev-requirements-extras.txt", "--without-hashes"]
```
gazpachoking, line 25: Did you mean this one to come out?

trim21: Sorry, I want to add a new one, didn't want to change existing

.github/workflows/docker-image.yml
```diff
@@ -60,12 +60,11 @@ jobs:
       - name: Build Base Docker Image
         id: docker_build
         uses: docker/build-push-action@v4
-        if: github.event_name == 'push' && github.ref != 'refs/heads/master'
         with:
           context: ./
           file: ./Dockerfile
           builder: ${{ steps.buildx.output.name }}
-          push: true
+          push: ${{ github.event_name == 'push' && github.ref != 'refs/heads/master' }}
```
gazpachoking, line 67: What's the motivation for this change? Does it change anything functionally?

ksurl: Looks like it only pushes on tags rather than every commit?

gazpachoking: It got changed from being in the 'if' block to down here. Should have put my comment on the range so that the diff showed up better. Just curious the motivation.
image

trim21: it doesn't build docker image with previous `if` field， dockerfile is not tested in the pr

trim21: it doesn't actually change anything， previous ci only push and build on the condition， now it build in any condition， and push in the condition

trim21: https://github.com/Flexget/Flexget/actions/runs/4659863819/jobs/8247244212

gazpachoking: Ahh, I get it now. Sounds good!

PR #3743, opened 2023-04-18
Cross build docker images for more arch

maxcanna:
### Detailed changes:
- 

- Fixes # .

```
paste_config_here
```
```
paste output here
```
#### To Do:

- [ ] Stuff..

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -7,6 +7,7 @@ RUN apk add --no-cache --upgrade \
         build-base \
         libffi-dev \
         openssl-dev \
+        git \
         cargo \
         rust \
         unzip && \


trim21:
it failed in these lines:


https://github.com/Flexget/Flexget/blob/bf0fd6d311f66fd89f7490a3682f1ce8de43a48f/Dockerfile#L17-L18

You will need a cross arch way to bundle-webui

maxcanna:
@trim21 It's working now. Perhaps I'm too naive here but wouldn't something like this:

https://github.com/maxcanna/dockerfiles/blob/master/flexget/Dockerfile

suffice?

maxcanna:
@trim21 any updates on this? I saw #3748 but perhaps we can go with this in the meantime. Wdyt?

trim21:
I don't have write permission, you need to ask @gazpachoking

gazpachoking:
Seems good to me. We could probably split our `dev` requirements group into `dev` and `release` (to eliminate the ruff install, and the need for cargo here) but that could come in a separate PR. LGTM, thanks!

gazpachoking:
@maxcanna Which platforms were you actually needing? I pulled some out that were harder to support due to lack of prebuilt wheels on pypi. Left arm64 in there other than the default amd64.

maxcanna:
@gazpachoking I need `linux/arm/v7` in order to be able to run it on raspberry PI on 32 bit OS. Thank you for taking this into account

PR #3750, opened 2023-04-28
Add template filter to parse human-readable file size to bytes

mnixry:
### Motivation for changes:

Some of the plugin, for example `rss` does not produce the `torrent_size` entry field, but for some sites, it actually contains human-readable size like `114.5MB`, `14GB`, or `1.91TB` in the torrent title.
In some cases, we need to filter out the torrents that are larger or smaller than we expected size. In this case, a filter to parse human-readable string to actual bytes number is necessary.

### Detailed changes:
- Added a jinja filter `parse_size`  to support parse the string like `114.5GB` to `122943438848` bytes int

```yaml
tasks:
  download_filtered_size:
    rss: https://example.com/?token=xxxx
    accept_all: yes
    no_entries_ok: yes
    if: # rejects torrents size which are larger than 100GB
      - "title|re_search('\d+(\.\d+)?\s*\w+B')|parse_size > 100 * (1024**3)": reject
```

#### To Do:

Current PR is a proof of concept, if you have any ideas about it, welcome comment.

- [ ] tests
- [ ] document

### DIFF
--- a/flexget/utils/template.py
+++ b/flexget/utils/template.py
@@ -187,7 +187,9 @@ def filter_get_year(name: str) -> str:
     return split_title_year(name).year
 
 
-def filter_parse_size(val: str, match_re: Optional[str] = None, si: bool = False) -> int:
+def filter_parse_size(
+    val: str, match_re: Optional[str] = None, si: bool = False, case: bool = True
+) -> int:
     """Parse human-readable file size to bytes"""
     if not isinstance(val, str):
         return val
@@ -208,16 +210,17 @@ def filter_parse_size(val: str, match_re: Optional[str] = None, si: bool = False
         'PiB': 1024**5,
         'EiB': 1024**6,
     }
+    size_map = {k.casefold(): v for k, v in size_map.items()}
 
     match_re = match_re or r'(?P<digit>\d+(?:\.\d+)?)\s*(?P<unit>[KMGTPE]?i?B)'
-    matched_size = re.search(match_re, val)
+    matched_size = re.search(match_re, val, flags=0 if case else re.IGNORECASE)
 
     if matched_size:
-        unit = matched_size['unit']
-
-        if unit in size_map:
+        unit = matched_size['unit'].casefold()
+        unit_base = size_map.get(unit)
+        if unit_base is not None:
             size = float(matched_size['digit'])
-            return int(size * size_map[unit])
+            return int(size * unit_base)
     return 0
 
 
--- a/flexget/utils/template.py
+++ b/flexget/utils/template.py
@@ -187,29 +187,34 @@ def filter_get_year(name: str) -> str:
     return split_title_year(name).year
 
 
-def filter_parse_size(val: str, match_re: Optional[str] = None) -> int:
+def filter_parse_size(val: str, match_re: Optional[str] = None, si: bool = False) -> int:
     """Parse human-readable file size to bytes"""
     if not isinstance(val, str):
         return val
 
-    match_re = match_re or r'^(?P<digit>\d+(?:\.\d+)?)\s*(?P<unit>[a-zA-Z]+B)$'
-    matched_size = re.match(match_re, val)
+    carry = 1000 if si else 1024
+    size_map = {
+        'B': 1,
+        'KB': carry,
+        'MB': carry**2,
+        'GB': carry**3,
+        'TB': carry**4,
+        'PB': carry**5,
+        'EB': carry**6,
+        'KiB': 1024,
+        'MiB': 1024**2,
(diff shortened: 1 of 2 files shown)


flexget/utils/template.py
```diff
@@ -187,6 +187,35 @@ def filter_get_year(name: str) -> str:
...
+    match_re = match_re or r'^(?P<digit>\d+(?:\.\d+)?)\s*(?P<unit>[a-zA-Z]+B)$'
+    matched_size = re.match(match_re, val)
+
+    if matched_size:
+        unit = matched_size['unit']
+        size_map = {
+            'B': 1,
+            'KB': 1024,
+            'KiB': 1024,
+            'MB': 1024**2,
```
trim21, line 204: should this be `1000**2`?

mnixry: > should this be `1000**2`?

idk :( It may depends on platform that produce this string. 

In my opinion, it is providing an approximate value of torrent size because of the limited decimal digits, so 1024 or 1000 doesn't matter.

gazpachoking:
I like the idea. Maybe we should remove the anchors from the regex so that less regex in the template itself is needed? And maybe we make an optional parameter to control whether it assumes SI units or not, e.g. `1 kb|parse_size == 1024`, but `1 kb|parse_size(si=true) == 1000`. kib would of course always be 1024.

mnixry:
> Maybe we should remove the anchors from the regex so that less regex in the template itself is needed? And maybe we make an optional parameter to control whether it assumes SI units or not

@gazpachoking Updated. Please check again and feel free to comment :)

flexget/utils/template.py
```diff
@@ -187,6 +187,40 @@ def filter_get_year(name: str) -> str:
...
+        'EB': carry**6,
+        'KiB': 1024,
+        'MiB': 1024**2,
+        'GiB': 1024**3,
+        'TiB': 1024**4,
+        'PiB': 1024**5,
+        'EiB': 1024**6,
+    }
+
+    match_re = match_re or r'(?P<digit>\d+(?:\.\d+)?)\s*(?P<unit>[A-Z]?i?B)'
```
gazpachoking, line 212: I'd probably tighten up the `[A-Z]` to only `[KMGTPE]`

flexget/utils/template.py
```diff
@@ -187,6 +187,40 @@ def filter_get_year(name: str) -> str:
     return split_title_year(name).year
 
 
+def filter_parse_size(val: str, match_re: Optional[str] = None, si: bool = False) -> int:
+    """Parse human-readable file size to bytes"""
+    if not isinstance(val, str):
+        return val
+
+    carry = 1000 if si else 1024
```
gazpachoking, line 195: Maybe this is better called `base` than `carry`?

gazpachoking (commented):
Couple small comments, but looks good to me! And I think a few tests would be a great idea if you are up for it.

gazpachoking:
I thought of one more question... should it be case sensitive? I'm thinking people will only use this filter on fields that are very likely to have a size in it, so we probably don't need to be so strict about the casing.

mnixry:
> I thought of one more question... should it be case sensitive? I'm thinking people will only use this filter on fields that are very likely to have a size in it, so we probably don't need to be so strict about the casing.

Updated. I've added an argument that can let user match case-insensitive, and false by default. 

Case-insensitive matching will significantly increase the wrong match rate, considering a title like this, which are really common in most RSS feed:
```
Symphony No.9 - Ludwig van Beethoven [FLAC][96kbps][5.78G]
```
A case-insensitive matching will match `96kb` instead of `7.58G`.

gazpachoking:
That's a good point. We probably could have gone with it several iterations ago, but I can't help tweaking. 😄  What about a negative lookahead at the end of the regex to eliminate that kind of error? Something like `(?![A-Za-z])`. Or maybe just use a word boundary at the end `\b`

mnixry:
> What about a negative lookahead at the end of the regex to eliminate that kind of error? Something like `(?![A-Za-z])`. Or maybe just use a word boundary at the end `\b`

I adjusted the regex with lookahead syntax, it should work without wrong match:

```regex
(?P<digit>\d+(?:\.\d+)?)\s*(?P<unit>[KMGTPE]?i?B(?!\w))
```

Snipaste_2023-05-04_02-20-12

If you think there is no problem with this regexp, please reply and I will push the new commit later.

---

I believe this regex should work for 99% of entries. And for scenarios that this filter can't handle, users can chain the `re_search` and other jinja syntax to make it work for their scenarios.

gazpachoking:
Awesome, thanks!

PR #3751, opened 2023-04-29
build: only bundle v2 webui

trim21:
### Detailed changes:
- 

- Fixes # .

```
paste_config_here
```
```
paste output here
```
#### To Do:

- [ ] Stuff..


paranoidi:
This is okay change, but I think the correct step would be to deprecate and remove whole v1 and bump up the version number :)

trim21:
this is actually what we already did before
https://github.com/Flexget/Flexget/pull/3739/files

PR #3759, opened 2023-05-03
add `provenance: false` to docker buildx

trim21:
image

remove unknown/unknown


gazpachoking:
> > docker ci is being too slow ...
> 
> It seemed to be stuck. I cancelled it and restarted it a bit ago.

Oh geez. Actually it's the cross arch build from #3743. It  went from taking 2 minutes to over an hour...

trim21:
just remove pysftp from dev requirements and remove compiler from docker file should fix it

PR #3760, opened 2023-05-03
Streamline docker image build/distribution

### LINKED ISSUE #3748: docker: reuse base docker image
we have locked all requirements, so it's possible to reuse base docker image between versions.

we can build a base docker image based on `${{ hashFiles('requirements-docker.txt') }}` and re-used it, just build flexget wheel if previous base docker image exists. generate base docker image tag hash, try pull, and build new one if not exist, then push it.

# Pro

So if we we only change one source file and push a new version, users only need to pull very small layers with only source code change, most size of docker image will be reused from previous pulled docker image.

```text
base --- requiremenets --- src # version 0.0.1
                |-----  new src # version 0.0.2
```

# Con

developers can't build docker image from repo directly, them need to build base docker image like ci workflow.

gazpachoking:
### Motivation for changes:
Reduce build time and distribution size for updates to docker image. #3748 
Alternate implementation to #3749

### Detailed changes:
- Optimize layers of docker images such that dep and webui layers don't need to be rebuilt/distributed when deps haven't changed
- Add a cache to github actions so that reused docker layers don't need to be rebuilt. More testing is needed how much this helps. Hopefully a lot, because with the cross-arch building that step takes a full hour to do without caching.
- Remove installation of cryptography library on builder image, which took a long time on multi-platform builds because it needed rust installed and to be compiled from scratch
- Split release requirements from dev requirements (turns out this was sort of unrelated to docker)

- Fixes #3748

#### To Do:

- [ ] Figure out if this actually does what I think it does and makes things better


gazpachoking:
Okay, latest changes eliminate the need for cryptography (which required rust and to be compiled on many platforms) and has reduced the full build time down from an hour to about 17 minutes. With the layer cache now being stored in github actions, full build should be rare, and normal builds much shorter.

PR #3766, opened 2023-05-10
Update to sqlalchemy 2.0 (and fix some other deprecation warnings)

gazpachoking:
### Motivation for changes:
Getting ready for sqlachemy 2.0
Ended up finding other deprecated things unrelated to sqlalchemy as well
May merge this PR before we are fully sqlalchemy 2.0 ready, as I'm not sure how extensive those changes will end up being.

### Detailed changes:
- Remove old deprecated sqlalchemy behavior
- Update sqlalchemy to 2.0
- Remove other things that were causing deprecationwarnings while debugging


gazpachoking:
Would be great if people could back up their database then try this out. Seems like things are working, but I'm not sure we have 100% test coverage on everything necessary.

gazpachoking:
> apscheduler needs a version >= 3.10.0 to fix incompatibility with sqlachemy 2.0 (cf [agronholm/apscheduler#705](https://github.com/agronholm/apscheduler/issues/705))

Updated, thanks!

PR #3772, opened 2023-05-22
[fix] emby: solve reconnect error

soloam:
### Motivation for changes:

When the login failed with saved data (for example changed ip or login data), it would give an error, it should authenticate again

### Detailed changes:
- added a check to try to relogin

### DIFF
--- a/flexget/components/emby/api_emby.py
+++ b/flexget/components/emby/api_emby.py
@@ -155,6 +155,7 @@ class EmbyAuth(EmbyApiBase):
 
     _userid = ''
     _token = ''
+    _host_name = ''
     _connect_token = ''
     _connect_token_link = ''
     _connect_username = ''
@@ -242,7 +243,10 @@ class EmbyAuth(EmbyApiBase):
                             raise PluginError(
                                 f'Could not login to Emby Connect account `{self._connect_username}`, no server list'
                             )
-                        if server['Name'].lower() == self.host.lower():
+                        if (
+                            server['Name'].lower() == self.host.lower()
+                            or server['Name'].lower() == self.host_name.lower()
+                        ):
                             connect_server = server
                             break
                     else:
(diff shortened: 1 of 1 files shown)


flexget/components/emby/api_emby.py
```diff
@@ -242,7 +243,10 @@ def login(self, optional=False):
                             raise PluginError(
                                 f'Could not login to Emby Connect account `{self._connect_username}`, no server list'
                             )
-                        if server['Name'].lower() == self.host.lower():
+                        if (
+                            server['Name'].lower() == self.host.lower()
+                            or server['Name'].lower() == self.host_name.lower()
```
gazpachoking, line 248: What are the cases when `self.host_name` is different than `self.host`? It looks like they just get set the same down on line 358.

flexget/components/emby/api_emby.py
```diff
@@ -373,12 +378,12 @@ def check_token_data(self, token_data, login_type):
...
         self._userid = token_data.get('userid')
         self._token = token_data.get('token')
-        self._login_type = token_data.get('type')
+        self._login_type = token_data.get('login_type')
         self._connect_username = token_data.get('connect_username', '')
         self.host = token_data.get('host', '')
         self._logged = True
         endpoint = EMBY_ENDPOINT_USERINFO.format(userid=token_data['userid'])
-        response = EmbyApi.resquest_emby(endpoint, self, 'GET')
+        response = EmbyApi.resquest_emby(endpoint, self, 'GET', plugin_error=False)
```
gazpachoking, line 386: Would it be nicer to just catch `PluginError` here, rather than adding another keyword argument?

flexget/components/emby/api_emby.py
soloam: When the user sets the config, the host can a address, when we are connecting directly to the server, or a name, that is used to match the server name in emby connect (a proxy hosted by the emby team). To make the code readable, I replace the emby name with the actual address (supplied by emby connect on login), but with that I need to save the server name to use latter on, for example on the reconnect! That was the main issue this PR was solving!

flexget/components/emby/api_emby.py
soloam: I need the login requested to return false, so that I'm able to try to make a login again! For example if the ip on emby connect changed! Probably you are asking why not use a try catch on this... If that's the case, you probably are right! That would be a better approach! I'll change it.

Issue #3767, opened 2023-05-11
Error while sending notification to `telegram`: missing python-telegram-bot pkg


evilbulgarian:
### Expected behaviour:

Working telegram notifications from docker hub image. Please add python-telegram-bot to requirements.

### Actual behaviour:

Error while sending notification to `telegram`: missing python-telegram-bot pkg

### Steps to reproduce:

set up a telegram notification in config.yml

#### Config:
```yaml
        via:
          - telegram:
              bot_token: xxxxx
              parse_mode: markdown
              recipients:
                - group: xxxx

```
  
#### Log:
```
Error while sending notification to `telegram`: missing python-telegram-bot pkg
```
</details>

### Additional information:

- FlexGet version: 
- Python version: 3.11
- Installation method: dockerhub
- Using daemon (yes/no): yes
- OS and version: gentoo gnu/linux
- Link to crash log:

gazpachoking:
We are including a bunch of the optional deps in our docker image now, probably could just add this there.

evilbulgarian:
Yeah, I don't want to be building a custom docker image for a feature that should be supported since its included in the docs. Especially since there are already plugins for other things in the image that are also "optional" for some of the users.

PR #3776, opened 2023-05-24
Update requirements-docker.txt to add telegram-bot

evilbulgarian:
### Motivation for changes:
Support the ability to use Telegram notifications without building custom container and having automatic updates to flexget container releases.

### Detailed changes:
- added telegram-bot to docker build requirements

- Fixes https://github.com/Flexget/Flexget/issues/3767


gazpachoking:
The change should be added in pyproject.toml there is a docker dependency group in there which is responsible for generating the txt file

evilbulgarian:
> The change should be added in pyproject.toml there is a docker dependency group in there which is responsible for generating the txt file

roger thanks will do!

PR #3780, opened 2023-05-26
[fix] emby: multiple providers lookup

soloam:
### Motivation for changes:

The emby api allows to search by providers imdb, tmdb and the movie db. The problem is that it does not handle well searching with multiple providers at the same time. This fix handles this issue, making requests with the providers separated and in the end returns the best selection.

Also, some shows don't provide the "Year" field, this is used when we match more that one show (normally when searched by name). But the field "PremiereDate" is available, so we extract the year from the "PremiereDate"

Also I made a update to the pattern used to decode dates, depending on emby server I found different formats

### Detailed changes:
- Fixed process of  searches with providers if search fails
- Fixed year determination to help filter multiple results
- Fixed date pattern

### DIFF
--- a/flexget/components/emby/api_emby.py
+++ b/flexget/components/emby/api_emby.py
@@ -1831,10 +1831,11 @@ class EmbyApiSerie(EmbyApiMedia):
                 filter(lambda s: s.get('Name') == args.get('SearchTerm'), series['Items'])
             )
 
-            if args.get('Years'):
-                serie_filter = list(
-                    filter(lambda s: s.get('Year') == args.get('Years'), serie_filter)
-                )
+            if len(serie_filter) > 1:
+                if args.get('Years'):
+                    serie_filter = list(
+                        filter(lambda s: s.get('Year') == args.get('Years'), serie_filter)
+                    )
 
             if len(serie_filter) == 1:
                 serie = serie_filter[0]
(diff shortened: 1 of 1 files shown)


flexget/components/emby/api_emby.py
```diff
@@ -2705,11 +2709,22 @@ def strtotime(date) -> 'datetime':
...
+        # Check if the offset is present
+        if '+' in date:
+            date = date.split('+')[0]
+
+        # Check if ms is present
+        if '.' in date:
+            date = date.split('.')[0]
+
+        # Define the format
+        format_str = '%Y-%m-%dT%H:%M:%S'
```
gazpachoking, line 2724: Do we need all this custom date wrangling? Seems like it's probably just an iso date and we can use https://dateutil.readthedocs.io/en/stable/parser.html#dateutil.parser.isoparse

soloam: I have found that the api is not very accurate returning the date format, I have found more that one format, and had some errors because of that, so i tried to decode the formats and act accordingly! 

https://dateutil.readthedocs.io/en/stable/parser.html#dateutil.parser.isoparse seams to map all the versions! I can try to replace it!

I have other improvements to the lookup! Can I push them in this PR? Since it's related. I've been making some tuning to the lookup, to help getting the matching more accurate.

soloam:
Delayed but not forgotten!

soloam:
Still not forgotten

PR #3791, opened 2023-06-09
[fix] pyload: Fix pyload-ng api support

patrickegli:
### Motivation for changes:

Adding packages to pyload-ng doesn't work. There are several issues that prevent adding packages:
- Typo in variable `set_package_data_command`
- Parameter name for package_id changed in pyload-ng
- Key for folder name changed in pyload-ng

### Detailed changes:
- Fix typo in set_package_data_command: set_package_dat**a** (instead of set_package_dat**e**)
- Different name for package_id parameter (pyload stable: `pid`, pyload-ng: `package_id`)
- Different name for folder key (pyload stable: `folder`, pyload-ng: `_folder`)

Without these changes, adding a package to pyload-ng fails with error:
```
500 Server Error: INTERNAL SERVER ERROR for url: http://x.x.x.x:8000/api/set_package_date
```

### DIFF
--- a/flexget/plugins/clients/pyload.py
+++ b/flexget/plugins/clients/pyload.py
@@ -141,13 +141,17 @@ class PluginPyLoad:
         parse_urls_command = 'parseURLs'
         add_package_command = 'addPackage'
         set_package_data_command = 'setPackageData'
+        package_id_parameter = 'pid'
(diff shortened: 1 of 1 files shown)


gazpachoking:
Thanks!

PR #3792, opened 2023-06-09
add mutagen to docker requirements

ksurl:
### Motivation for changes:

support metadata editing via python in exec plugin

### Detailed changes:
- add mutagen package in docker image


ksurl:
https://github.com/python-poetry/poetry/pull/7637 this change in poetry 1.5.x  removed the category from the lock file. does it cause any issues?

gazpachoking:
I think the poetry changes to the lock file are fine. Not sure how I feel about more deps that aren't even part of included plugins. Starting to think we need some sort of easier solution for users to add deps more easily to their container.

ksurl:
good point, shouldn't include stuff that isn't used by flexget. I can get by with either a custom entrypoint script to install it on run or a custom image using this as a base image.

gazpachoking:
I was thinking about entry point scripts, but I think custom Dockerfiles end up being a bit simpler. Perhaps we should add a wiki page with instructions for this, as well as instructions for how to update. (Like reminding people to do a `build --pull` to grab the latest base image when upgrading.)

ksurl:
I was thinking of updating the wiki page for it.

ksurl:
https://flexget.com/InstallWizard/Docker#custom-dockerfile

gazpachoking:
Awesome! There was more info already in there than I realized. I added a bit to your custom image section about --pull to do an upgrade to the latest flexget version.

PR #3795, opened 2023-06-16
nzbnab_parameters

ianstalk:
### Motivation for changes:
Fix issue where newznab plugin loses the API key and category on successive searches.

### Detailed changes:
- Stopped mutating config in favor of returning URL and parameters

### DIFF
--- a/flexget/components/sites/sites/newznab.py
+++ b/flexget/components/sites/sites/newznab.py
@@ -14,9 +14,10 @@ logger = logger.bind(name='newznab')
 class Newznab:
     """
     Newznab search plugin
-    Provide a url or your website + apikey and a category
+    Provide a url or your website + apikey and a category.
+    When a URL is provided it will override website, apikey, and category.
 
-    Config example::
+    Config example:
 
         newznab:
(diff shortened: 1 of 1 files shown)


flexget/components/sites/sites/newznab.py
```diff
@@ -34,32 +36,38 @@ class Newznab:
                 'type': 'string',
                 'enum': ['movie', 'tvsearch', 'tv', 'music', 'book', 'all'],
             },
-            'url': {'type': 'string', 'format': 'url'},
-            'website': {'type': 'string', 'format': 'url'},
```
gazpachoking, line 38: I think better to keep the old properties, then add the oneOf up a level like:
```python
oneOf: [{'required': 'url'}, {'required': 'website'}]
```

ianstalk: Good idea. Didn't know you could do that.

gazpachoking:
Looks good to me! Thanks! That error on python 3.11 isn't to do with your change, I think it's a python 3.11.4 difference that just started showing up. Should go away on merge.

PR #3796, opened 2023-06-20
Fix small typos in documentation

pelmers:
This fixes two small typos in a couple of method docstrings.

Background: I'm doing a research study on function comments that may be incorrect. You can find more info and a short and optional anonymous survey [here](https://forms.office.com/e/yHv4PRXM2i).
 
Thanks for reviewing!


gazpachoking:
Thanks!

PR #3820, opened 2023-08-18
[Discord] Changes to Rate-Limit and Embed Color

BrutuZ:
### Motivation for changes:
I needed to use Jinja expressions to contextually set the embed color.
Rate-limit was overly cautious
### Detailed changes:
- Added STR as an allowed value for embed color
  - Parse the STR and convert it to Hexadecimal Int. Gracefully drop it with a warning if fails (invalid value) rather than raising exception
- Switched requests from fixed time rate to token based. Numbers from own responses since the one in the [API docs](https://discord.com/developers/docs/topics/rate-limits#rate-limits) (50 hits/second) would always trigger the rate-limit way before reaching them.
  - Added a handler for the rate-limit response header anyways.

- My own

- Fully backwards compatible

### DIFF
--- a/flexget/components/notify/notifiers/discord.py
+++ b/flexget/components/notify/notifiers/discord.py
@@ -43,7 +43,12 @@ class DiscordNotifier:
                         'title': {'type': 'string'},
                         'description': {'type': 'string'},
                         'url': {'type': 'string', 'format': 'uri'},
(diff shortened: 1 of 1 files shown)


gazpachoking:
Looks good to me, thanks!

PR #3831, opened 2023-09-01
parallel docker platform builds

ksurl:
### Motivation for changes:

speed up docker build workflow

### Detailed changes:
- use matrix for multi runner builds

- Fixes # .

```
paste_config_here
```
```
paste output here
```
#### To Do:

- [ ] Stuff..


gazpachoking:
Hmm. Seems like a good idea. Is it just coincidence that this build took longer than all the previous ones though?

ksurl:
> Hmm. Seems like a good idea. Is it just coincidence that this build took longer than all the previous ones though?

Maybe since they are all separate now, there was no previous cache?

ksurl:
Oh it's not merged yet so it wouldn't be any faster i dont think

ksurl:
My test repo has issues pushing the image (only pushes arm, not amd64), so this needs some debugging before it can be merged.

ksurl:
I can't figure it out, but the sha doesn't match with independent platform builds so, I think we'll have to abandon this for now. 7-8min for the build is not bad anyways. I was working on another project where builds were 3 hours.

PR #3835, opened 2023-09-03
Fix letterboxd.py movie data retrieving

eSoares:
Currently letterboxd in their watchlist the data-film-slug is just the end part of the URL to the film.

As a quick example for the movie "/film/asteroid-city/" the `data-film-slug` will be populated with "asteroid-city".  Without this fix the generated path would be `http://letterboxd.comasteroid-city` causing failure due to DNS resolve failing.

### Motivation for changes:

Currently broken letterboxd plugin.


dev-requirements.txt
```diff
@@ -53,7 +53,7 @@ sphinxcontrib-jsmath==1.0.1 ; python_version >= "3.7" and python_version < "4.0"
 sphinxcontrib-qthelp==1.0.3 ; python_version >= "3.7" and python_version < "4.0"
 sphinxcontrib-serializinghtml==1.1.5 ; python_version >= "3.7" and python_version < "4.0"
 sqlalchemy-stubs==0.4 ; python_version >= "3.7" and python_version < "4.0"
-tomli==2.0.1 ; python_version >= "3.7" and python_version < "3.11"
+tomli==2.0.1 ; python_version >= "3.7" and python_full_version <= "3.11.0a6"
```
paranoidi, line 56: Is this intentional change? What's 0a6 ?

gazpachoking: This is generated automatically. I suspect one of our deps changed their requirements to have that.

PR #3861, opened 2023-10-16
Fix deprecated use of jsonschema library

rsekman:
### Motivation for changes:

As of jsonschema version 4.18.0 `jsonschema.exceptions.RefResolutionError` is deprecated and we should catch/raise `referencing.exceptions.Unresolvable` instead

RefResolver usage was also switched to use the new referencing.Registry

### Detailed changes:

Simple 1-for-1 replacements

### DIFF
--- a/flexget/api/app.py
+++ b/flexget/api/app.py
@@ -12,9 +12,9 @@ from flask_cors import CORS
 from flask_restx import Api as RestxAPI
 from flask_restx import Model, Resource
 from flask_restx.reqparse import RequestParser
-from jsonschema import RefResolutionError
 from jsonschema.exceptions import ValidationError as SchemaValidationError
 from loguru import logger
+from referencing.exceptions import Unresolvable
 from sqlalchemy.orm import Session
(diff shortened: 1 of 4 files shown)


gazpachoking:
I don't think we are on that jsonschema version yet.

rsekman:
> I don't think we are on that jsonschema version yet.

We [are now](https://github.com/Flexget/Flexget/commit/a55678f315006f82c60feec7171558e36863f28c#diff-4d7c51b1efe9043e44439a949dfd92e5827321b34082903477fd04876edb7552R39).

gazpachoking:
Oh, good call. I guess we should update our usage of RefResolver as well now. I haven't looked at the new referencing library yet.

gazpachoking:
I removed the use of RefResolver also. I'm wondering if we should switch the the schema registering to using the new `Registry` class directly instead of still storing it separately and instantiating the Registry with a retriever function.

gazpachoking:
Thanks!

PR #3877, opened 2023-11-08
ncore site fixes

DDanii:
### Motivation for changes:
Fixing minor bugs

### Detailed changes:
- Seed and leech text converted to numbers
- Torrent title now uses tool-tip text instead of displayed text as it is sometimes shortened
- Using proper size conversion

### DIFF
--- a/flexget/components/sites/sites/ncore.py
+++ b/flexget/components/sites/sites/ncore.py
@@ -8,6 +8,7 @@ from flexget.entry import Entry
 from flexget.event import event
 from flexget.utils.requests import TokenBucketLimiter
 from flexget.utils.soup import get_soup
+from flexget.utils.tools import parse_filesize
 
 logger = logger.bind(name='nCore')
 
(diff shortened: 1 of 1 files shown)


flexget/components/sites/sites/ncore.py
```diff
@@ -111,15 +112,15 @@ def search(self, task, entry, config):
...
                     parent = a.parent.parent.parent.parent
 
-                    e['torrent_seeds'] = parent.find('div', class_='box_s2').string
-                    e['torrent_leeches'] = parent.find('div', class_='box_l2').string
-                    e['content_size'] = (
-                        float(parent.find('div', class_='box_meret2').string) * 1024**2
+                    e['torrent_seeds'] = int(parent.find('div', class_='box_s2').string)
+                    e['torrent_leeches'] = int(parent.find('div', class_='box_l2').string)
+                    e['content_size'] = parse_filesize(
+                        parent.find('div', class_='box_meret2').string
```
gazpachoking, line 123: The `parse_filesize` function expects there to be a unit in the string. The previous code looks like it was just expecting a bare number. Did you test this works?

DDanii: Yep, I tested it, and there is unit in the string. It was not tested when it was created.

gazpachoking:
Looks good to me, thank you!

PR #3878, opened 2023-11-11
copy files fails when copying between certain file systems.

barp:
### Motivation for changes:
On certain filesystem shutil.copy can fail due to difference in filesystems between the source and destination (like in truenas when the target fs has ACLs).

this change makes the mode change warn instead of failing the command allowing it to function in such cases.

### Detailed changes:
Create an implementation of the copy command from shutil which does not fail on copymode and just warn instead

### DIFF
--- a/flexget/plugins/output/file_operations.py
+++ b/flexget/plugins/output/file_operations.py
@@ -219,6 +219,20 @@ class TransformingOps(BaseFileOps):
     move = None
     destination_field = None
 
+    def copy_permissive(self, src, dst, follow_symlinks=True):
+        # this function is like shutil.copy but in case copy mode fails we ignore the error
+        if os.path.isdir(dst):
+            dst = os.path.join(dst, os.path.basename(src))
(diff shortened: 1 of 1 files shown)


flexget/plugins/output/file_operations.py
```diff
@@ -219,6 +219,17 @@ class TransformingOps(BaseFileOps):
...
 
+    def copy_permissive(self, src, dst, follow_symlinks=True):
+        # this function is like shutil.copy but in case copy mode fails we ignore the error
+        if os.path.isdir(dst):
+            dst = os.path.join(dst, os.path.basename(src))
+        shutil.copyfile(src, dst, follow_symlinks=follow_symlinks)
+
+        try:
+            shutil.copymode(src, dst, follow_symlinks=follow_symlinks)
+        except Exception as err:
```
gazpachoking, line 230: Can we be more explicit here? Is it just one exception type we are expecting?

gazpachoking:
Looks good to me, thanks!

PR #3882, opened 2023-11-16
FIX PTP plugin to use apiuser and apikey authentication and some improvements to search and match

bagobones:
### Motivation for changes:

Current plugin impersonates a user with their primary credentials and uses complicated cookie refresh auth. This is in violation of current site rules and can cause an auto ban. Current site rules require all scripts to use apiuser / apikey auth generated by the user as separate credentials.

### Detailed changes:
- removed username / password / cookie auth
- added apiuser / apikey header auth
- added some mappings for odd quality values from ptp to make the quality plugin match more often
- added search parameter for year if input has a movie year and NO IMDB ID
- plugin should raise an error if search fails.
- changed the schema default for the grouping parameter as it seems to provide better matches
- schema replaced username and password with apiuser and apikey

- Fixes authentication / user getting banned by old cookie auth

```
    schema = {
        'type': 'object',
        'properties': {
            'apiuser': {'type': 'string'},
            'apikey': {'type': 'string'},
            'passkey': {'type': 'string'},
            'tags': one_or_more({'type': 'string', 'enum': TAGS}, unique_items=True),
            'order_by': {'type': 'string', 'enum': list(ORDERING.keys()), 'default': 'Time added'},
            'order_desc': {'type': 'boolean', 'default': True},
            'freeleech': {'type': 'boolean'},
            'release_type': {'type': 'string', 'enum': list(RELEASE_TYPES.keys())},
            'grouping': {'type': 'boolean', 'default': False},
        },
        'required': ['apiuser', 'apikey', 'passkey'],
    }
```

### DIFF
--- a/flexget/components/sites/sites/passthepopcorn.py
+++ b/flexget/components/sites/sites/passthepopcorn.py
@@ -79,7 +79,9 @@ RELEASE_TYPES = {'non-scene': 0, 'scene': 1, 'golden popcorn': 2}
 
 @db_schema.upgrade('passthepopcorn')
 def upgrade(ver, session):
-    if ver is None or ver < 2:
+    if ver is None:
+        ver = 0
+    if ver < 2:
         if table_exists('passthepopcorn_cookie', session):
             logger.info('Removing old Cookie Tracking Table')
             drop_tables(['passthepopcorn_cookie'], session)
--- a/flexget/components/sites/sites/passthepopcorn.py
+++ b/flexget/components/sites/sites/passthepopcorn.py
@@ -79,14 +79,12 @@ RELEASE_TYPES = {'non-scene': 0, 'scene': 1, 'golden popcorn': 2}
 @db_schema.upgrade('passthepopcorn')
 def upgrade(ver, session):
     if ver is None:
-        ver = 0
-    if ver == 0:
+        raise db_schema.UpgradeImpossible
+    else:
         if table_exists('passthepopcorn_cookie', session):
             logger.info('Removing old Cookie Tracking Table')
             # Drop the deprecated data
             drop_tables(['passthepopcorn_cookie'], session)
-        logger.info('DB Schema is now v1')
-        ver = 1
     return ver
 
 
--- a/flexget/components/sites/sites/passthepopcorn.py
+++ b/flexget/components/sites/sites/passthepopcorn.py
@@ -201,7 +201,7 @@ class SearchPassThePopcorn:
                     release_res = torrent['Resolution']
                     release_res = release_res.replace("PAL", "576p") #Common PAL DVD vertial resolution
                     release_res = release_res.replace("NTSC", "480p") #Common PAL DVD vertial resolution
-                    # many older realses have a resolution defined as 624x480 for example this will split the value at take the hight
+                    # many older releases have a resolution defined as 624x480 for example this will split the value at take the hight
                     tsplit = release_res.split("x", 1)                    
                     if len(tsplit)>1:
                         release_res = tsplit[1] + 'p' 
--- a/flexget/components/sites/sites/passthepopcorn.py
+++ b/flexget/components/sites/sites/passthepopcorn.py
@@ -8,6 +8,10 @@ from flexget.event import event
 from flexget.utils.requests import RequestException, TimedLimiter
 from flexget.utils.requests import Session as RequestSession
 from flexget.utils.tools import parse_filesize
+from flexget.utils.sqlalchemy_utils import (
+    drop_tables,
+    table_exists
+)
 
 logger = logger.bind(name='passthepopcorn')
 Base = db_schema.versioned_base('passthepopcorn', 1)
--- a/flexget/components/sites/sites/passthepopcorn.py
+++ b/flexget/components/sites/sites/passthepopcorn.py
@@ -205,7 +205,9 @@ class SearchPassThePopcorn:
                     release_name = torrent['ReleaseName'].replace(
                         "4K", ""
                     )  # Remove 4K from release name as that is often the source not the torrent resolution it confuses the quality plugin
-                    e['title'] = f"{release_name} [{release_res} {torrent['Source']} {torrent['Codec']} {torrent['Container']}]"
+                    e[
+                        'title'
+                    ] = f"{release_name} [{release_res} {torrent['Source']} {torrent['Codec']} {torrent['Container']}]"
 
                     e['imdb_id'] = entry.get('imdb_id')
 
--- a/flexget/components/sites/sites/passthepopcorn.py
+++ b/flexget/components/sites/sites/passthepopcorn.py
@@ -132,7 +132,7 @@ class SearchPassThePopcorn:
 
         if config.get('freeleech'):
             params['freetorrent'] = int(config['freeleech'])
-
+        
         ordering = 'desc' if config['order_desc'] else 'asc'
 
         grouping = int(config['grouping'])
@@ -156,6 +156,10 @@ class SearchPassThePopcorn:
         # searching with imdb id is much more precise
         if entry.get('imdb_id'):
             search_strings = [entry['imdb_id']]
+        else:
+            # use the movie year if available to improve search results.
+            if 'movie_year' in entry:
+                params['year'] = int(entry['movie_year'])
 
         for search_string in search_strings:
             params['searchstr'] = search_string
@@ -239,7 +243,7 @@ class SearchPassThePopcorn:
                             e['torrent_id'], authkey, passkey
                         )
                     )
-                    logger.debug('Add Entry: {} S:{}', e['title'],e['torrent_seeds'])
+                    logger.debug('Add Entry: {} Seeds:{}', e['title'],e['torrent_seeds'])
                     entries.add(e)
 
         return entries
--- a/flexget/components/sites/sites/passthepopcorn.py
+++ b/flexget/components/sites/sites/passthepopcorn.py
@@ -90,16 +90,6 @@ def upgrade(ver, session):
         raise db_schema.UpgradeImpossible
     return ver
 
-
-class PassThePopcornCookie(Base):
-    __tablename__ = 'passthepopcorn_cookie'
-
-    username = Column(Unicode, primary_key=True)
-    _cookie = Column('cookie', Unicode)
-    cookie = json_synonym('_cookie')
-    expires = Column(DateTime)
-
-
 class SearchPassThePopcorn:
     """
     PassThePopcorn search plugin.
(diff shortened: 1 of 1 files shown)


flexget/components/sites/sites/passthepopcorn.py
```diff
@@ -311,7 +198,34 @@ def search(self, task, entry, config):
...
+                        + ' ['
+                        + release_res
+                        + ' '
+                        + torrent['Source']
+                        + ' '
+                        + torrent['Codec']
+                        + ' '
+                        + torrent['Container']
+                        + ']'
+                    )
```
BrutuZ, line 228: inb4: Use an f-String instead

bagobones: It now uses an f-string. however there are now about a hundred ways you could interpret f-string formatting in python.

bagobones: further googled what an f-string is should be fixed.. not really a regular python dev so any bad habits are being copy and pasted from other existing code that has formatted strings everywhere. didn't get the distinction or benefit of the f string without further reading.

bagobones:
@gazpachoking Mind taking a look at this. It should fix what is broken with the current plugin

flexget/components/sites/sites/passthepopcorn.py
```diff
@@ -82,24 +76,6 @@
 RELEASE_TYPES = {'non-scene': 0, 'scene': 1, 'golden popcorn': 2}
 
 
-@db_schema.upgrade('passthepopcorn')
-def upgrade(ver, session):
-    if ver is None:
-        ver = 0
-    if ver == 0:
-        raise db_schema.UpgradeImpossible
-    return ver
```
gazpachoking, line 91: I guess technically a new version should be added that drops the cookie table. We do have a utility function to do that dropping. https://github.com/Flexget/Flexget/blob/ba878634b022cb5e94d32e4d2e593e38a19714a4/flexget/utils/sqlalchemy_utils.py#L98

flexget/components/sites/sites/passthepopcorn.py
```diff
@@ -262,27 +137,30 @@ def search(self, task, entry, config):
...
             logger.debug('Using search params: {}', params)
             try:
-                result = self.get(
-                    self.base_url + 'torrents.php',
-                    params,
-                    config['username'],
-                    config['password'],
-                    config['passkey'],
-                ).json()
+                siteresponse = requests.get(
```
gazpachoking, line 154: Should we be using `task.requests` here, to make sure plugins modifying the session work?

gazpachoking (commented):
Seems pretty good. We should probably drop the old db table, then we'll bump the minor version to indicate that configs will need to be changed.

flexget/components/sites/sites/passthepopcorn.py
bagobones: Seems like a better idea than abandoning it.. I will try and re-insert something.

flexget/components/sites/sites/passthepopcorn.py
bagobones: You tell me? Mostly just replicating the existing style, calls of the original plugin and similar site plugins. If you think this is of high value then it... "looks" simple enough to test.

One thing I do have a consern about / may need more testing is that I would like the plugin to try and follow enforce rate limiting rules and not let the user over ride them. At the moment that is just the same domain request per second the old plugin (but with fewer calls due to no cookie login loop) had but it may need a further PR with more restrictions if anything is discovered after heavier use.

gazpachoking: I think we should be using it. Other plugins may not be, because I think originally the search api didn't have access to the task. The good thing about using `task.requests` is that the user could customize the limiting behavior by using the domain_delay plugin.

gazpachoking: Do they publish what the actual limit is?

bagobones: > Do they publish what the actual limit is?

No, from discussions on the forum for other scripts it appears they don't want to disclose the exact values / methods they are using to detect abuse. In general it was stated that no faster than 1 request per second and HALT your script if you get an HTTP error. (no specifics on any errors are given).

Since removing the old login logic I have not actually hit the rate limit with the new code as it really only makes one request per title and waits 5s. The old code would hammer the auth URL if the cookies didn't work, and with the new abuse filters the old code will auto lock you out of the site quite quickly.

flexget/components/sites/sites/passthepopcorn.py
bagobones: I THINK? I fixed it the old code went up to version 0 so I am only doing a table check once and removing it if it exists.. Let me know if I got this right in the new commit.

flexget/components/sites/sites/passthepopcorn.py
bagobones: change commited.

flexget/components/sites/sites/passthepopcorn.py
bagobones: Going to stop playing with the import upgrade I am guessing it is fine? Is actually the expected outcome? not sure why it shows as FAILED however then?

`FAILED flexget/tests/test_migrate.py::TestMigrate::test_upgrade - ValueError: The plugin passthepopcorn has no stored schema to reset. `

gazpachoking: I think we need to do a full 'upgrade' i.e. change the current schema version to 2 in the `Base` declaration, then we can remove the old upgrades and change the if in the upgrade function to do `if ver is None or ver < 2` in order to capture all previous states. Inside that block you'll do the table drop.

gazpachoking: We need the `ver is None or ver < 2` to prevent that current error of the tests, and you were right to have the `ver=2` in there.

bagobones: Did a little more reading, don't think we can call UPGRADE unless we are declaring a NEW base schema but since what we WANT to do is clean up and remove the table  I think we want to call reset_schema instead of upgrade.

bagobones: I have removed all the upgrade code and DB import references as well as the plugin isn't using the db at all now.. If you have a way to clean up all tables I am all ears (probably will need to import the DB libraries again) but as it stands it works without the upgrade code.

flexget/components/sites/sites/passthepopcorn.py
bagobones: This seems to have removed the rate limit completely, I was just doing a more extensive stress test on the plugin and noticed it was not rate limiting at all now.  I am guessing as the rate limit is not applied to the correct object now.. looking into it.

bagobones: Verified a fix to correctly attach the limit to the task.requests, it is working now.

flexget/components/sites/sites/passthepopcorn.py
gazpachoking: Oh, I realized I was complicating things a bit. To clean up the db we can just put the function in an `@event('manager.upgrade')` rather than the `@db_schema.upgrade`. That way we can leave all the other DB stuff you already cleaned up totally out still.

flexget/components/sites/sites/passthepopcorn.py
gazpachoking: Hmm. Something I wondered while looking at this change. If there are 2 limiters on the same site, should they both apply? Right now only the first one does. I think at a minimum it should be changed so that the most recent applies rather than the oldest. (i.e. If the plugin includes a limiter, as in this case, one the user adds via domain_delay should override.) Potentially we should go further and just have _both_ of them apply. If they were both just timers between requests it would mean the longer wait always applied, but if a regular timed limiter was combined with a token bucket limiter that might actually be useful.

BrutuZ: I don't think so, pretty sure it's basically just a dict that gets overwritten since the hostname is the index. And the TimedLimiter essentially calls the TokenLimiter with a token count of 1, so they're actually the same

bagobones: Testing the change as is the coded delay overrides the users domain_delay: if they have provided one.

If you want to provide your own a fully working code snippet example for me to use instead that is fine, however this is a problem with ALL existing plugins that call add_domain_limiter inside them. 

There were two main goals in my rewrite.. replace the auth and ensure a safe default rate limit was in place..

I think those are met.. 

There are going to be a select few that use this plugin anyway.

flexget/components/sites/sites/passthepopcorn.py
bagobones: > To clean up the db we can just put the function in an @event('manager.upgrade') rather than the @db_schema.upgrade. That way we can leave all the other DB stuff you already cleaned up totally out still.

I don't see any existing plugins that do this, also in order to have the table search and drop calls I would have to import the schema and DB functions still..

Looking into my SQL lite DB this table should have next to nothing in it and only exist for a small number of users.

I would prefer to abandon it and move on.

flexget/components/sites/sites/passthepopcorn.py
gazpachoking: @bagobones Yes, sorry, I didn't mean you should do something here. I was just talking the opportunity to discuss the desired behavior

bagobones: > @bagobones Yes, sorry, I didn't mean you should do something here. I was just talking the opportunity to discuss the desired behavior

Thanks @gazpachoking , I am focused on trying to get this fix merged and I am happy to correct or implement improvements, but keep in mind this is 100% hobby time for me and I am not a full time coder.. I abandon my ombi pull request a few years ago as it seemed to be stuck in code review (might try and help with the new ombi_list one that is ALSO stuck in review).

Hoping this fix can get merged if it isn't going to create issues and helps others that want to use it.

gazpachoking:
Thanks for sticking with it @bagobones! I added an upgrade action and updated the wiki page, but feel free to add some more detail there. Maybe add on the ptp wiki page where to get the api user and key from?
https://flexget.com/en/UpgradeActions#h-3100-2023-11-19 
https://flexget.com/Searches/passthepopcorn

flexget/components/sites/sites/passthepopcorn.py
gazpachoking: For posterity, I ended up changing this after merge to explicitly specify not overwriting existing limiters. I changed domain_delay to explicitly overwrite existing ones, so it should be that this (and other plugins adding their own limiter) can be properly overridden by the user domain_delay now.

PR #3884, opened 2023-11-20
Replace status index with something smaller

ianstalk:
### Motivation for changes:
The index `ix_status_execution_task_id_start_end_succeeded` is almost 93% the size of the actual table. An index on just `task_id` should be similarly performant but much smaller.

### Detailed changes:
- Add DB util functions to manage indexes
- Update DB migration to drop `ix_status_execution_task_id_start_end_succeeded` if it exists and then create the new index

- Fixes # .

```
paste_config_here
```
```
paste output here
```
#### To Do:

- [ ] Stuff..

### DIFF
--- a/flexget/components/status/db.py
+++ b/flexget/components/status/db.py
@@ -107,10 +107,7 @@ class TaskExecution(Base):
         }
 
 
-Index(
-    'ix_status_execution_task_id',
-    TaskExecution.task_id
-)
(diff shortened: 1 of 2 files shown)


gazpachoking:
I think we need to remove the old index declaration from the schema, as well as add the new one still.

gazpachoking:
LGTM, thanks!

PR #3886, opened 2023-11-21
Add Jinja templating to Transmission labels

g-regor:
### Motivation for changes:
With this, some series properties can be automatically added as a label

### Detailed changes:
- `labels` can be either a string or a list of strings
- If a label contains a comma, it's split into multiple labels
- Commas from rendered Jinja templates are removed (Transmission doesn't allow them)
- Whitespace is trimmed and empty labels are removed
- Labels defined in the series entry are concatenated with the labels defined under transmission, instead of overridden

```
tasks:
  Task-name:
    transmission:
      ...
      labels: "Source: xyz, Series: {{series_name}}, Quality: {{quality}}"
    series:
      - "The Show":
          set:
            labels:
              - Drama
              - Funny
```

This results in `["Source: xyz", "Series: The Show", "Quality: 1080p", "Drama", "Funny"]`


ianstalk:
Are you certain that Transmission doesn't support commas in labels? Also I would update the docstring.

g-regor:
> Are you certain that Transmission doesn't support commas in labels?

[Yes](https://github.com/transmission/transmission/blob/4.0.4/libtransmission/rpcimpl.cc#L985-L988)

gazpachoking:
I'm good with this, the only thing I'm dubious about is the string formatted comma delimited multi-label configuration. I think we should just require multiple labels to be in a list.
i.e. Instead of this:
```
      labels: "Source: xyz, Series: {{series_name}}, Quality: {{quality}}"
```
require this:
```yaml
      labels:
        - "Source: xyz"
        - "Series: {{series_name}}"
        - "Quality: {{quality}}"
```

g-regor:
So, is concatenating labels from config and entry ok, since it's goes against how that usually works?

gazpachoking:
> So, is concatenating labels from config and entry ok, since it's goes against how that usually works?

Yeah, I guess that's good. Seems like it's a bit more flexible than just overriding config labels with entry labels.

flexget/plugins/clients/transmission.py
```diff
@@ -756,7 +756,18 @@ def _make_torrent_options_dict(self, config, entry):
...
         if 'labels' in opt_dic:
-            change['labels'] = opt_dic['labels']
+            labels = set()
+            for obj in [config, entry]:
+                obj_labels = obj.get('labels', [])
+                if not isinstance(obj_labels, list):
+                    obj_labels = [obj_labels]
+                labels.update(obj_labels)
+            labels = [entry.render(label) for label in labels]
+            labels = [re.sub(r"(\s),|,(\s)", r"\1\2", label).replace(",", " ") for label in labels]
```
gazpachoking, line 766: What is the regex meant to do? It strips out commas that have whitespace next to them? Why do that separate than the regular comma replacement?

g-regor: If a comma is next to a whitespace, it's just removed, otherwise it's replaced with space.

It's to reduce unnecessary whitespace. So `1,2, 3, 4 , 5` becomes `1 2 3 4  5` instead of `1 2  3  4   5`.

gazpachoking: I know it's sorta nitpicky and doesn't matter much, but my inclination would be to make the code simple and leave the formatting to the user. I think replacing the ',' is fine since things could break without that, but really it should be up to the user to format the label how they want. If you really want to avoid duplicate whitespace while removing commas it could just be `re.sub("\s*,\s*", " ", label)` which would replace commas and all surrounding whitespace with just a space, and then the second `replace` call isn't needed.

I think also a comment is warranted here like `# Transmission doesn't allow commas in labels`, just so we remember why that's happening in the future.

g-regor: My intent was to not change user formatting too much.
`re.sub("\s*,\s*", " ", label)` would remove user added whitespace. Also `\s` matches more than just space.
`label.replace(",", " ")` would create double spaces.
`label.replace(",", "")` would merge words.

And user doesn't control formatting in text added with templating.
Titles with commas usually have just comma followed by space, where simple replacement would work.

g-regor:
Are certain fields always available in templates?  
Can "quality" not be there?

[The documentation](https://flexget.com/Jinja) says that it will produce an empty string, but I noticed that invalid fields cause an exception.
And [in code](https://github.com/Flexget/Flexget/blob/v3.10.2/flexget/plugins/clients/transmission.py#L774-L777), it actually just captures the exception and logs the error.

It should probably be done here the same.

flexget/plugins/clients/transmission.py
gazpachoking: Maybe `re.sub(", ?", " ", label)` then? I'm good with whatever you think is best here though.

gazpachoking:
> [The documentation](https://flexget.com/Jinja) says that it will produce an empty string, but I noticed that invalid fields cause an exception.
> And [in code](https://github.com/Flexget/Flexget/blob/v3.10.2/flexget/plugins/clients/transmission.py#L774-L777), it actually just captures the exception and logs the error.

Yeah, that documentation is probably oversimplified/innacurate depending on where templates are being used. I'd say we should catch errors and log a warning then ignore that label for ones that the rendering failed here.

flexget/plugins/clients/transmission.py
g-regor: Ok, I used this. It should be good enough for most use cases.

gazpachoking:
Looks good. Thanks for this!

PR #3892, opened 2023-12-05
change: update deliverybot action

ksurl:
### Motivation for changes:

original action is using deprecated node 12 base

### Detailed changes:
- switch to forked action with node 16 base 
- https://github.com/justacodename/deployment-status


gazpachoking:
Sweet, thanks!

PR #3900, opened 2023-12-22
Removed special characters from alpharatio.py search

kigon001:
the movie-list plugin that uses imdb adds certain special characters to the search, which alpharatio can't handle, like parentheses and colon. This commit makes sure that only alphanumerical characters is used.

Top Gun: Maverick (2023)
South Park: Not Suitable for Children (2023)

Will now work since this commit will strip the colon and the parentheses.

### Detailed changes:
- 

- Fixes # .

```
paste_config_here
```
```
paste output here
```
#### To Do:

- [ ] Stuff..


gazpachoking:
Awesome, thanks!

PR #3943, opened 2024-02-28
[Change] New interfaces for EZTV plugin

BrutuZ:
### Motivation for changes:
- EZTV removed search functionality from their RSS URL
- RARBG closed
### Detailed changes:
- Added Input and Search interfaces to EZTV plugin
  - Generators, compatible with the `limit` plugin, hard-capped at 1500 entries (up to 15 pages with 100 entries) otherwise.
  - Search requires the entry to have the `imdb_id` field set due to an API limitation, it only allows searching by IMDB IDs without the `tt` prefix
- Removed RARBG plugin

Input and Search need to be called with a simple boolean
```yaml
task:
  eztv: yes # Input
  discover:
    what:
      - entry-list: my-series
    from:
      - eztv: yes  # Search - Will search for each entry's `imdb_id`, skips them if it's not defined
```


gazpachoking:
Looks good to me, thanks!

PR #3959, opened 2024-03-23
[change] Update best_quality.py

diogosena:
### Motivation for changes:

The best_quality plugin was just considering the first entry as the best quality, and that's not always true, many entries can have the same quality as the best.

### Detailed changes:

Added comparison of every entry with the best quality.

### DIFF
--- a/flexget/plugins/filter/best_quality.py
+++ b/flexget/plugins/filter/best_quality.py
@@ -25,6 +25,10 @@ class FilterBestQuality:
                 'enum': ['accept', 'reject', 'do_nothing'],
                 'default': 'reject',
             },
+            'single_best': {
+                'type': 'boolean',
+                'default': True,
+            }
         },
         'additionalProperties': False,
     }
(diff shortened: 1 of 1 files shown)


gazpachoking:
Hmm, does this change the plugin from only ever acting on one entry to now acting on multiple ones? I can see how this might be desirable, but I'm also worried it's changing the behavior a bit much. Maybe I'm wrong though. Anyone else using this plugin that wants to weigh in?

diogosena:
> Hmm, does this change the plugin from only ever acting on one entry to now acting on multiple ones? I can see how this might be desirable, but I'm also worried it's changing the behavior a bit much. Maybe I'm wrong though. Anyone else using this plugin that wants to weigh in?

Yes, "on best" can now act on every entry. I understand your concern, but the old behavior seems wrong and is very confusing, "on_best" could act just on one entry, and "on lower" can act on all the others, even if they all are the same quality as  the best .
Besides, if someone needs only first entry, there's the unique plugin to use.

diogosena:
@gazpachoking maybe you prefer an option "single_best: [yes|no]", defaulting to yes, to have the same behavior as before?

diogosena:
> Oh, yes, sorry for the delay. I meant to suggest something like that. I think one of the most common uses for this plugin would be to grab a single copy of something in the best quality available, so it makes sense to keep that as the default. Adding an option sounds fine though to have it operate in this new mode though.

Just updated the pull request, and updated the wiki (thought that will need some kind of approval), feel free to revert if needed.

gazpachoking:
Sounds good, thanks!

PR #3994, opened 2024-05-21
[nfo_lookup] Fallback to movie.nfo if same name .nfo not found

diogosena:
### Motivation for changes:
Latest version of jellyfin stopped generating .nfo with same name of movies, and is using movie.nfo instead.

### Detailed changes:
Added a fallback to search for movie.nfo if same name .nfo not found.


gazpachoking:
Looks good, thanks!

PR #4003, opened 2024-06-10
[fix] Import error preventing telegram notifier to initialize correctly

georgeyk:
### Motivation for changes:

Telegram notifier failing to initialize due to an import error, leading to an incorrect error report (https://github.com/georgeyk/Flexget/blob/develop/flexget/components/notify/notifiers/telegram.py#L276). 

### Detailed changes:
- Import `NetworkError` from the right location

This change is compatible with the defined versions of python-telegram-bot:
- v12.8: https://github.com/python-telegram-bot/python-telegram-bot/blob/v12.8/telegram/error.py#L64
- v21.3: https://github.com/python-telegram-bot/python-telegram-bot/blob/v21.3/telegram/error.py#L147
- using this reference: https://github.com/Flexget/Flexget/blob/develop/pyproject.toml#L122

---

I'm fairly new to the codebase, so I'm a bit hesitant about this test. We only need to check `_enforce_telegram_plugin_ver` (https://github.com/georgeyk/Flexget/blob/develop/flexget/components/notify/notifiers/telegram.py#L274), but I'm assuming its a private api, and didn't find another way to properly check this error. The test should fail with `python-telegram-bot==21.0.1`, but it's kind generic.


gazpachoking:
Fix looks good. I'm not sure about that test either. What exactly is it testing?

georgeyk:
hi @gazpachoking - in short, if it fails with an KeyError, it means it didn't fail to load telegram module. Again, not great, but couldn't think of anything else. We could just try to import telegram from that module, since it would fail at import time, then checking if `telegram is not None`. I'm ok to remove the test too, if that's preferable.

 Also, just realized that the only supported version is v12.8 (documented here too: https://flexget.com/en/Plugins/Notifiers/telegram) - so, I updated the dependencies in the last commit.

gazpachoking:
It's supposed to fail if the telegram-bot library isn't installed? Not sure it's actually doing that, since that isn't installed with our github actions tests right now, but all the tests still seem to have passed. If we have a more meaningful test (that we are sure works as expected,) we can add it to the list of deps that get installed before the test suite is run.
https://github.com/Flexget/Flexget/blob/4c5c5346ba17ecaa5f943165d842d33d1a87748f/pyproject.toml#L110-L115

georgeyk:
it's installed, or am I looking in the wrong place? https://github.com/Flexget/Flexget/actions/runs/9470787685/job/26092491908?pr=4003

georgeyk:
if you remove the dependency, then i can change the test to check for `DependencyError` instead.

gazpachoking:
> it's installed, or am I looking in the wrong place? https://github.com/Flexget/Flexget/actions/runs/9470787685/job/26092491908?pr=4003

Ahh, you are correct. I didn't realize we are installing the deps from all the dependency groups for the test job.

georgeyk:
so, i just removed the test. hope this gets in.

gazpachoking:
Oh, I'm sorry! Just got busy, didn't mean to ignore this. Thank you!

gazpachoking:
Oh shoot. I didn't notice this took the imports out of the try/except block. Can you add that back in? Right now this causes flexget to crash for anyone that doesn't have the dep installed.

georgeyk:
@gazpachoking sure, i'll open a new PR in a moment.

PR #4007, opened 2024-06-19
[fix]: Wrap telegram imports to avoid runtime crashes

georgeyk:
### Motivation for changes:
Flexget is crashing when `python-telegram-bot` package is not installed, per https://github.com/Flexget/Flexget/pull/4003#issuecomment-2179346431

### Detailed changes:
- Revert part of the changes introduced in https://github.com/Flexget/Flexget/pull/4003

### DIFF
--- a/flexget/components/notify/notifiers/telegram.py
+++ b/flexget/components/notify/notifiers/telegram.py
@@ -1,13 +1,18 @@
-import telegram
 from loguru import logger
 from packaging import version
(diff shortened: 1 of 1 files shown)


georgeyk:
@gazpachoking here^. I apologize for such mistake.

gazpachoking:
No problem, I didn't notice either when I looked over. Thanks for the quick fix!

PR #4015, opened 2024-07-13
Anilist: Handle rate-limit and add more metadata

BrutuZ:
### Motivation for changes:
I needed it

### Detailed changes:
- Token based rate limiter and wait based on response header if we still hit it since sometimes they restrict it further from the documented limit.
- Include almost all metadata fields available.
- Call the `v2` endpoint of the relations API, providing `TheTVDB` and `TMDB` IDs (no season though)

- N/A

```
N/A
```


gazpachoking:
Looks good, thanks!

PR #4017, opened 2024-07-14
[add] Discord Notifier: silent notification property

Winter:
### Motivation for changes:
Wanted to have a notification go out when something is grabbed and another go out when it's available, but didn't want to be notified that it was grabbed, only that it was available; without creating two separate channels.

### Detailed changes:
- Added new property called `silent` onto discord notifier plugin.
- Check if `silent` is true, set property `flags` onto web_hook object with the suppress notification bitfield set.

```
notify:
  task:
    message: <message>
    via:
      - discord:
          web_hook_url: <web_hook_url>
          silent: true
```

https://discord.com/developers/docs/resources/webhook#execute-webhook-jsonform-params

### DIFF
--- a/flexget/components/notify/notifiers/discord.py
+++ b/flexget/components/notify/notifiers/discord.py
@@ -26,6 +26,7 @@ class DiscordNotifier:
           via:
             - discord:
                 web_hook_url: <string>
+                [silent: <boolean>] (suppress notification)
                 [username: <string>] (override the default username of the webhook)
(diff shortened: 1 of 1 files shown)


BrutuZ:
Huh, I was just considering changing my notifications to conditionally include a ping to my user/role but this might be more elegant, and easier too since I won't have to make the already _very_ messy message even worse with more Jinja formulas

gazpachoking:
LGTM!

PR #4038, opened 2024-08-31
qBittorrent: add options for ratio limit and seeding time limit

lingfish:
### Motivation for changes:
This adds config options for specifying ratio and seeding time limits.

### Detailed changes:
- Adds two config options: `ratio_limit`, and `seeding_time_limit`

None.

```
    qbittorrent:
      label: some-label
      path: /home/user/download
      ratio_limit: 1.5
      seeding_time_limit: 10
```

`ratio_limit` is a float, and `seeding_time_limit` is in minutes.

### DIFF
--- a/flexget/plugins/clients/qbittorrent.py
+++ b/flexget/plugins/clients/qbittorrent.py
@@ -274,7 +274,7 @@ class OutputQBitTorrent:
 
             seeding_time_limit = entry.get('seeding_time_limit', config.get('seeding_time_limit'))
             if seeding_time_limit:
-                form_data['seedingTimeLimit'] = parse_timedelta(seeding_time_limit) * 60
+                form_data['seedingTimeLimit'] = int(parse_timedelta(seeding_time_limit).total_seconds() / 60)
 
             is_magnet = entry['url'].startswith('magnet:')
 
--- a/flexget/plugins/clients/qbittorrent.py
+++ b/flexget/plugins/clients/qbittorrent.py
@@ -225,8 +225,6 @@ class OutputQBitTorrent:
         config.setdefault('maxupspeed', 0)
         config.setdefault('maxdownspeed', 0)
         config.setdefault('fail_html', True)
-        config.setdefault('ratio_limit', -2)
-        config.setdefault('seeding_time_limit', -1)
         return config
 
     def add_entries(self, task, config):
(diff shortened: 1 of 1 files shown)


lingfish:
Hrm, need advice here please?  How should I handle `float`s?  Use a `string` in the schema or what?

BrutuZ:
Keep it simple, percentage integer

lingfish:
> Keep it simple, percentage integer

Hrm, looking at [the schema](https://github.com/Flexget/Flexget/blob/develop/docs/develop/schemas.rst), it would seem `number` maps fine, yes?

gazpachoking:
Sorry for the delay on this. I think @BrutuZ idea to allow intervals for the seed time is nice. So the user can write e.g "3 days" or "1 week"

You can see how we validate it in the config, and how we can turn it into a timedelta easily here.

https://github.com/Flexget/Flexget/blob/58d75fbf9e46cea38668e2f45c172223c1d6e495/flexget/plugins/filter/age.py#L30

https://github.com/Flexget/Flexget/blob/58d75fbf9e46cea38668e2f45c172223c1d6e495/flexget/plugins/filter/age.py#L63

gazpachoking:
Also curious, does -2 mean something special to qbittorrent for ratio limit?

BrutuZ:
From the docs:

The maximum seeding ratio for the torrent. -2 means the global limit should be used, -1 means no limit.

lingfish:
Ugh, I mangled that commit, pushing another now.

flexget/plugins/clients/qbittorrent.py
```diff
@@ -261,6 +268,16 @@ def add_entries(self, task, config):
...
                 form_data['dlLimit'] = maxdownspeed * 1024
 
+            ratio_limit = entry.get('ratio_limit', config.get('ratio_limit'))
+            if ratio_limit:
+                form_data['ratioLimit'] = ratio_limit
+
+            seeding_time_limit = entry.get('seeding_time_limit', config.get('seeding_time_limit'))
+            if seeding_time_limit:
+                form_data['seedingTimeLimit'] = int(
+                    parse_timedelta(seeding_time_limit).total_seconds() / 60
```
gazpachoking, line 278: Hmm, when this ends up as the default of -1, is this going to fail? Maybe take away the default, and check if the option `is None` here, and in that case set it to -1 instead of trying to parse.

lingfish: That's a good point actually... both options are optional.

gazpachoking:
Looks good to me, thanks for sticking with it through the delays! Want to add these options to the wiki as well?

lingfish:
Thanks for helping out too... apologise for a few bad pushes there!  Yep, I was wondering how the doco worked, was mislead by the sphinx stuff.  Will do the wiki now.

PR #4045, opened 2024-09-08
Update Deps

gazpachoking:
### Motivation for changes:
#3940 was having trouble as many tests were breaking. @carno figured out it was pytest 8+ causing the issue. This PR caps pytest version and updates all the other pinned deps. We will have to investigate the pytest issue later.


gazpachoking:
I think the issue with the new pytest must be something to do with teardown not happening properly. Certain tests fail when run with the whole suite, but succeed when run individually. There are also pysftp tests that are failing due to the socket already being used, I'm assuming because the last test didn't tear down properly.

gazpachoking:
Yeah, I had a look at that pytest changelog and nothing popped out to me as requiring us to update, but I suspect you are right that something in our conftest needs to change.

PR #4054, opened 2024-10-09
add stopped to add_paused for qbit5

whiteout12:
### Motivation for changes:
In qbittorrent 5 the add_paused option has changed to add_stopped. To handle this I have added stopped to the form data that is entered when adding torrent to qbittorrent. form_data['paused'] is still there to handle older versions of the webapi.
This change follows the same as had been done with label/category for different versions earlier
- Fixes # .

```

```
```

```

### DIFF
--- a/flexget/plugins/clients/qbittorrent.py
+++ b/flexget/plugins/clients/qbittorrent.py
@@ -247,7 +247,8 @@ class OutputQBitTorrent:
 
             add_paused = entry.get('add_paused', config.get('add_paused'))
             if add_paused:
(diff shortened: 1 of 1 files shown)


gazpachoking:
Did you test this against both versions? Neither version has a problem with the extra key? Looks great if so.

whiteout12:
@gazpachoking Yes I have tested locally against qbit 4.6.7 and 5.0.0 and both works fine using "add_paused" with this change.

gazpachoking:
Excellent, thanks!

PR #4078, opened 2024-12-02
Drop support for Python 3.8

### LINKED ISSUE #4077: Drop support for Python 3.8
### Summary:
As the Python community has moved towards supporting newer Python versions, it is becoming increasingly difficult to maintain compatibility with older versions like Python 3.8. 

Given that Python 3.8 has reached its end of life (EOL) and the FlexGet project relies on dependencies and features that are better supported in later Python versions, it would be beneficial for the project to drop support for Python 3.8.

### Rationale:
**Security**: Python 3.8 is no longer receiving security updates, which could expose the project to potential vulnerabilities.

**Maintenance**: Maintaining support for deprecated versions can be resource-intensive, and focusing on newer versions will allow us to take advantage of improvements in performance, security, and new features.

**Future-proofing**: By supporting only actively maintained versions of Python, we ensure that FlexGet remains compatible with the latest advancements in the Python ecosystem.

vivodi:
### Motivation for changes:

Fixes #4077

As the Python community has moved towards supporting newer Python versions, it is becoming increasingly difficult to maintain compatibility with older versions like Python 3.8.

Given that Python 3.8 has reached its end of life (EOL) on October 14, 2024 and the FlexGet project relies on dependencies and features that are better supported in later Python versions, it would be beneficial for the project to drop support for Python 3.8.


vivodi:

gazpachoking:
Thanks! Looks good to me, and I'm good with this. Only thing I'm thinking is that we should bump the minor version to indicate that users might need to update something.

vivodi:
Done @gazpachoking

gazpachoking:
Awesome, thanks!

PR #4079, opened 2024-12-03
Bump ruff to 0.8.1

vivodi:
### Motivation for changes:

Supersedes #4073

Upgrading ruff to 0.8.1 requires manual work and therefore cannot be done by bots.


vivodi:

PR #4081, opened 2024-12-07
Bump pytest to 8.x

vivodi:
### Motivation for changes:

Supersedes #4046

Upgrading pytest to 8.x requires manual work and therefore cannot be done by bots.


vivodi:

vivodi:
> Was it xdist stuff that was causing the failures with the automatic upgrade?

Yes.

> I see changes to use pathlib as well, but it doesn't feel like that would have broken anything.

It did not break anything, but according to [the document](https://docs.pytest.org/en/stable/how-to/tmp_path.html#the-tmpdir-and-tmpdir-factory-fixtures), these days, it is preferred to use `tmp_path` and `tmp_path_factor` instead of `tmpdir` and `tmpdir_factory`.

>[!TIP]
>The `tmpdir` and `tmpdir`_factory fixtures are similar to `tmp_path` and `tmp_path_factory`, but use/return legacy [py.path.local](https://py.readthedocs.io/en/latest/path.html) objects rather than standard [pathlib.Path](https://docs.python.org/3/library/pathlib.html#pathlib.Path) objects.

gazpachoking:
Yep, I like the pathlib change as well, was just theorizing about what actually fixed the tests. Thanks again!

PR #4084, opened 2024-12-10
Move pytest configuration so developers can run tests locally with the same configuration as GitHub Actions.

vivodi:
### Motivation for changes:
Move pytest configuration so developers can run tests locally with the same configuration as GitHub Actions.


gazpachoking:
Makes sense to me, thanks!

PR #4086, opened 2024-12-12
Easily install extra packages

vivodi:
### Usage
With this PR, users can easily install extra packages:
```
pip install flexget[deluge]
pip install flexget[qbittorrent]
pip install flexget[telegram]
pip install flexget[transmission]
```
Or shorthanded:
```
pip install flexget[de]
pip install flexget[qb]
pip install flexget[tg]
pip install flexget[tr]
```
Install multiple packages at once:
```
pip install flexget[de,qb,tg]
```
To install all:
```
pip install flexget[all]
```
### Benefits

With this PR, end users don't need to care which version is compatible with FlexGet if FlexGet uses a specific version of a package. 
```diff
- pip install flexget python-telegram-bot==12.8
+ pip install flexget[tg]
```

In additon, if FlexGet uses a different underlying package to implement telegram plugin functionality, end users won't have to know that.


vivodi:
Because currently `cloudscraper` actually cannot bypass cloudflare verification, and `cloudscraper` is no longer maintained, and there is no good replacement, I suggest removing all related code.

gazpachoking:
I like this idea. I'm not sure about the shortened aliases, I'm sorta inclined to just include the long form of the extras unless you or anyone feels strongly that the short forms are useful.

Do you think it's a good idea to add some sort of test where we don't install any extras, and make sure we don't crash to prevent issues like we had with the telegram plugin? #4088

If cloudscraper is broken, I'm good with another PR to rip it out. That one probably needs a version bump because even though it wasn't working, we'll need to let users know to take it out of their config to pass config validation.

vivodi:
> Do you think it's a good idea to add some sort of test where we don't install any extras, and make sure we don't crash to prevent issues like we had with the telegram plugin?

I tried to write a test using only `pytest` but failed. I don't know if there is a better way to test without changing the workflow.

gazpachoking:
Yeah, I'll think about that. I think you are right, we'd need an extra workflow

.github/workflows/main.yml
```diff
@@ -53,11 +53,12 @@ jobs:
...
         key: venv-${{ runner.os }}-${{ steps.setup-python.outputs.python-version }}-${{ hashFiles('**/poetry.lock') }}
-    - name: Install dependencies
-      if: steps.cached-poetry-dependencies.outputs.cache-hit != 'true'
-      run: poetry install --no-interaction --no-root
     - name: Install project
       run: poetry install --no-interaction
+    - name: Verify functionality without extras
+      run: poetry run python flexget/tests/verify_functionality_without_extras.py
+    - name: Install project extras
+      run: poetry install --all-extras --no-interaction
```
vivodi, line 61: I have added this check to the existing workflow.

vivodi:
Currently using setuptools as the build backend and poetry to manage dependencies, distributing FlexGet with extras is not as easy as I thought.

Extras have to be specified in setuptools:
```toml
[project.optional-dependencies]
telegram = ["python-telegram-bot>=21.9"]
```
but this defeats the point of poetry managing dependencies.

Or dynamically:
```toml
[tool.setuptools.dynamic.optional-dependencies]
telegram = { file = ["requirements/telegram.txt"] }
```
but this way you can't use poetry's `extras`, but can only use multiple `groups`, which make things complicated.

I think a better solution is to use poetry as the build backend.
```toml
[build-system]
requires = ["poetry-core"]
build-backend = "poetry.core.masonry.api"
```

vivodi:
What do you think is best? @gazpachoking

gazpachoking:
Yeah, we have a sort of complicated setup for dependency management right now. Much of it is because we want to release to pypi with pinned deps, and most of the existing tools don't really allow for that. I was looking at moving to uv to do dependency management, but I ran into the issue if we are using a pep 621 tool for dep management, we can't also use a pep 621 tool for the build backend, because we want our build backend to use pinned deps, but our dependency manager to use looser ones. I'm really not sure what the best route is. I was tempted to stop having fully pinned deps in our pip release, but we don't have rigorous testing on different dep versions, so that might introduce more user issues.

vivodi:
### Loosening Dependency Constraints in pyproject.toml Is Redundant
If the final package pins dependencies, specifying looser versions in `pyproject.toml` is unnecessary. The following two configurations will have no effect on the PyPI package:
```toml
[tool.poetry.dependencies]
apscheduler = ">=3.10.0"
rpyc = ">=5.0,!=5.2.*,!=5.3.0"
```
```toml
[tool.poetry.dependencies]
apscheduler = "==3.11.0"
rpyc = "==6.0.1"
```
### Conclusion
> I'm really not sure what the best route is. I was tempted to stop having fully pinned deps in our pip release, but we don't have rigorous testing on different dep versions, so that might introduce more user issues.

A reasonable compromise using a pep 621 tool for both dependency management and build backend, could be to loosen the dependency constraints in pyproject.toml to specify only the major version numbers (e.g., `^1.0`), rather than fully pinning every version. This would provide some leeway while still minimizing the risk of introducing breaking changes or incompatible versions.
```toml
[tool.poetry.dependencies]
apscheduler = "^3.10.0"
rpyc = "^5.4"
```

vivodi:
### Loosening Dependency Constraints in pyproject.toml Is Redundant
If the final package pins dependencies, specifying looser versions in `pyproject.toml` is unnecessary. The following two configurations will have no effect on the PyPI package:
```toml
[tool.poetry.dependencies]
apscheduler = ">=3.10.0"
rpyc = ">=5.0,!=5.2.*,!=5.3.0"
```
```toml
[tool.poetry.dependencies]
apscheduler = "==3.11.0"
rpyc = "==6.0.1"
```
### Conclusion
> I'm really not sure what the best route is. I was tempted to stop having fully pinned deps in our pip release, but we don't have rigorous testing on different dep versions, so that might introduce more user issues.

A reasonable compromise using a pep 621 tool for both dependency management and the build backend, could be to loosen the dependency constraints in pyproject.toml to specify only the major version numbers (e.g., `^1.0`), rather than fully pinning every version. This would provide some leeway while still minimizing the risk of introducing breaking changes or incompatible versions.
```toml
[tool.poetry.dependencies]
apscheduler = "^3.10.0"
rpyc = "^5.4"
```
I believe it's possible to switch to UV using the above solution. However, without using the same tool for both dependency management and the build backend, implementing this PR will be challenging.

gazpachoking:
Yeah, the reason behind having looser constraints for development and pinned for release is because we can put in the known requirements loosely, and have dependabot (or renovate in our case) update the pins within our defined loose ranges.  These new pinned versions get tested via our CI before we merge them for release. If we do a pypi release with the looser constraints that means users end up getting dep versions that were not tested in CI first, and if we do pinned constraints in pyproject that makes testing and automated updating by a bot harder. (the bot that updates the pins is a bit broken right now though, which is a whole different problem)

vivodi:
> Yeah, the reason behind having looser constraints for development and pinned for release is because we can put in the known requirements loosely, and have dependabot (or renovate in our case) update the pins within our defined loose ranges. These new pinned versions get tested via our CI before we merge them for release. If we do a pypi release with the looser constraints that means users end up getting dep versions that were not tested in CI first, and if we do pinned constraints in pyproject that makes testing and automated updating by a bot harder. (the bot that updates the pins is a bit broken right now though, which is a whole different problem)

The proposed compromise, #4098, should effectively address your concerns.

vivodi:
The following extras are provided:
```
pip install flexget[cloudscraper]
pip install flexget[deluge]
pip install flexget[locked]
pip install flexget[qbittorrent]
pip install flexget[telegram]
pip install flexget[transmission]
pip install flexget[all]
```

gazpachoking:
Sorta sucks we have to do the requirements export dance for these other extras when using dynamic extras. Maybe I'll look in to that plugin to handle it automatically in the future without the explicit export and including of requirements file as extra.

I like this, except I think we should keep requirements.txt and dev-requirements.txt named the same and in the root folder. Both for backward compatibility (people may be using these in their workflow,) and because it's a sort of standard (although.. I think requirements-dev.txt is probably more common than dev-requirements.txt maybe that one could change.)

vivodi:
It's done, but I still feel that putting too many files in the root directory makes it look messy.

gazpachoking:
> It's done, but I still feel that putting too many files in the root directory makes it look messy.

The requirements.txt is a standard that I feel good about keeping there. The requirements-dev.txt I could have gone either way on. I was thinking for backwards compatibility keeping it, but we renamed it anyway so 🤷 That one could probably be hidden away. It's only really used for rtd workflow I think. Developers can use uv directly now to do the venv sync.

PR #4087, opened 2024-12-12
Legacy key/value format with whitespace separator should not be used

vivodi:
### Motivation for changes:
"ENV key=value" should be used instead of legacy "ENV key value" format
More info: https://docs.docker.com/go/dockerfile/rule/legacy-key-value-format/


gazpachoking:
Thanks!

Issue #4085, opened 2024-12-11
Flexget 3.12.4 Docker image doesn't start


dangmai:
### Expected behaviour:

Flexget Docker image should be able to start.

### Actual behaviour:

For some reason, `ghcr.io/flexget/flexget:3.12.4` cannot be started, even though earlier versions work fine.

### Steps to reproduce:
- Step 1:

```
docker run -it ghcr.io/flexget/flexget:3.12.4
echo $?
```

Notice that the container has exited with code 139, with no output.

Meanwhile, if we run the same command but with `3.12.3`:
```
docker run -it ghcr.io/flexget/flexget:3.12.3
```

The container does run (albeit with errors because there is no config file), and it prints out logs to the output.

(My setup is more complicated than this of course, but I want to show a minimally-reproducible example).

#### Config:

No config needed, because Flexget can't be brought up at all.
  
#### Log:

No log, because the container can't be brought up at all

</details>

### Additional information:

- FlexGet version: 3.12.4
- Python version:
- Installation method: Docker
- Using daemon (yes/no): yes
- OS and version: Ubuntu 22.04 x64
- Link to crash log: No crash log

dangmai:
I don't see anything in the commits of this repo that could lead to this issue. Some folks are reporting issue with the base image (python-alpine) https://github.com/docker-library/python/issues/993, that may be worth looking into. I don't even Segmentation fault in the log though.

vivodi:
> docker run -it ghct.io/flexget/flexget:3.12.4
> docker run -it ghct.io/flexget/flexget:3.12.3

There are two typos, it is `ghcr` instead of `ghct`

1. Run:
`docker run -it --entrypoint sh ghcr.io/flexget/flexget`
2. Run:
`flexget`
3. You'll see
`Segmentation fault (core dumped)`

This should indeed be a problem with the base image.

dangmai:
> > docker run -it ghct.io/flexget/flexget:3.12.4
> > docker run -it ghct.io/flexget/flexget:3.12.3
> 
> There are two typos, it is `ghcr` instead of `ghct`
> 
>     1. Run:
>        `docker run -it --entrypoint sh ghcr.io/flexget/flexget`
> 
>     2. Run:
>        `flexget`
> 
>     3. You'll see
>        `Segmentation fault (core dumped)`
> 
> 
> This should indeed be a problem with the base image.

[...]

PR #4090, opened 2024-12-15
[Fix] Telegram preventing Flexget launch on dependency errors

BrutuZ:
### Motivation for changes:
Flexget would not start in a VENV with older `python-telegram-bot` and a config with the `telegram` plugin in an _unused_  template, giving a stack trace instead

### Detailed changes:
- Moved exception raising in case of no / obsolete dependency inside the class, so the plugin properly registers and the existing handlers take care of it

- Fixes https://discord.com/channels/536690097056120833/536690097496784906/1316804237656588368

```
paste_config_here
```
With the previously recommended dependency version installed:
```yaml
tasks:
  test:
    mock: ['Test Entry']
    accept_all: yes
    exec: echo Ran!
    notify:
      abort:
        message: Abort!
        via:
          - telegram:
              bot_token: "foo"
              recipients:
                - username: "bar"
      entries:
        message: Entry!
        via:
          - telegram:
              bot_token: "foo"
              recipients:
                - username: "bar"
      task:
        message: Task!
        via:
          - telegram:
              bot_token: "foo"
              recipients:
                - username: "bar"

```
```
2024-12-14 23:50:20 INFO     manager                       Test mode, creating a copy from database ...
2024-12-14 23:50:23 VERBOSE  manager                       Creating new database C:\Users\BrutuZ\flexget\test-test.sqlite - DO NOT INTERRUPT ...
2024-12-14 23:50:26 VERBOSE  task_queue                    There are 1 tasks to execute. Shutdown will commence when they have completed.
2024-12-14 23:50:26 VERBOSE  details       test            Produced 1 entries.
2024-12-14 23:50:26 VERBOSE  task          test            ACCEPTED: `Test Entry` by accept_all plugin
2024-12-14 23:50:26 VERBOSE  details       test            Summary - Accepted: 1 (Rejected: 0 Undecided: 0 Failed: 0)
2024-12-14 23:50:26 INFO     exec          test            Would execute: echo Ran!
2024-12-14 23:50:26 ERROR    notify_entry  test            Plugin `telegram` requires dependency `python-telegram-bot >= 21.9`
2024-12-14 23:50:26 ERROR    notify_entry  test            Plugin `telegram` requires dependency `python-telegram-bot >= 21.9`
2024-12-14 23:50:27 INFO     manager                       Removed test database
```

### DIFF
--- a/flexget/components/notify/notifiers/telegram.py
+++ b/flexget/components/notify/notifiers/telegram.py
@@ -15,18 +15,19 @@ from sqlalchemy import Column, Integer, String
 from flexget import db_schema, plugin
 from flexget.event import event
 from flexget.manager import Session
-from flexget.plugin import PluginError
+from flexget.plugin import DependencyError, PluginError
 
 if TYPE_CHECKING:
     from collections.abc import AsyncGenerator
 
 try:
-    if Version(version("python-telegram-bot")) < Version('21.9'):
-        raise plugin.PluginWarning('obsolete python-telegram-bot pkg')
-except PackageNotFoundError:
-    pass
-else:
     import telegram
+
+    if Version(version("python-telegram-bot")) < Version('21.9'):
+        raise ImportError
+except (ImportError, PackageNotFoundError):
+    telegram = None
+else:
     from telegram.error import ChatMigrated, NetworkError, TelegramError
     from telegram.ext import ApplicationBuilder
 
@@ -201,6 +202,11 @@ class TelegramNotifier:
     }
 
     def notify(self, title: str, message: str, config: dict) -> None:
+        if telegram == None:
+            raise PluginError(
+                DependencyError(_PLUGIN_NAME, 'python-telegram-bot >= 21.9')._get_message()
+            )
+
         self._load_config(config)
         asyncio.run(self.main(message))
 
@@ -352,7 +358,7 @@ class TelegramNotifier:
         usernames: list[str],
         fullnames: list[tuple[str, str]],
         groups: list[str],
-    ) -> (list[ChatIdEntry], bool):
+    ) -> tuple[list[ChatIdEntry], bool]:
         """get chat ids for `usernames`, `fullnames` & `groups`.
         entries with a matching chat ids will be removed from the input lists.
         """
@@ -462,7 +468,9 @@ class TelegramNotifier:
 
     async def _get_bot_updates(
         self,
-    ) -> (dict[str, telegram.Chat], dict[(str, str), telegram.Chat], dict[str, telegram.Chat]):
+    ) -> tuple[
+        dict[str, telegram.Chat], dict[(str, str), telegram.Chat], dict[str, telegram.Chat]  # type: ignore
+    ]:
         """get updated chats info from telegram"""
         # highly unlikely, but if there are more than `telegram.constants.PollingLimit.MAX_LIMIT`
         # msgs waiting for the bot, we should not miss one


vivodi:
If a user has already installed `python-telegram-bot` and telegram is in the config, then there is a high probability that the user will need to use the telegram plugin, so I am inclined to leave this PR unmerged.

BrutuZ:
It might also be just a secondary notifier, or not used on all tasks. Doesn't justify having Flexget crash on launch

________________________________

vivodi:
But it’s equally bad for users to find out that after completing all tasks they can’t see the results.

BrutuZ:
At least the files will still be there once the user investigates why there have been no notifications in a while, as opposed to no task running at all, potentially causing them to lose entries on fast-moving RSS feeds or IRC channels, not to mention the seeding "prime-time" many value

vivodi:
I installed outdated `python-telegram-bot` and measured the elasped time it took for FlexGet to fail.
```console
> Measure-Command { flexget execute }

Days              : 0
Hours             : 0
Minutes           : 0
Seconds           : 4
Milliseconds      : 456
Ticks             : 44565728
TotalDays         : 5.15807037037037E-05
TotalHours        : 0.00123793688888889
TotalMinutes      : 0.0742762133333333
TotalSeconds      : 4.4565728
TotalMilliseconds : 4456.5728
```
I think it fails fast enough so the user will immediately notice the failure.

BrutuZ:
How long it takes from launch to crash is not the issue, but how long user might take to notice it crashed with notifications as their main feedback from an otherwise silent background application that requires no regular interaction.

For example, this user didn't notice something was wrong and Flexget wasn't runningfor 4 days https://github.com/Flexget/Flexget/issues/4085#issuecomment-2544117824

gazpachoking:
What's the current behavior before this change? My feeling is:
- If you don't have telegram plugin in your config, you should not get any warnings or errors.
- If you do have telegram plugin in your config, you should get an error on startup. (Ideally also when running the check command)
- Any changes we make in required versions of the package should result in a bumped minor version and an upgrade action, as with other changes that can break config.

gazpachoking:
> * If you do have telegram plugin in your config, you should get an error on startup. (Ideally also when running the check command)

Now that I said that, I'm not sure what the best way to implement something like that would be. Maybe a custom key in the schema that declares the requirements, that way our validation could assert the requirement was there?

BrutuZ:
> What's the current behavior before this change? My feeling is:
> - If you don't have telegram plugin in your config, you should not get any warnings or errors.
> - If you do have telegram plugin in your config, you should get an error on startup. (Ideally also when running the check command)

It crashes if you have the wrong version of the dependency installed at all. So if you used the Telegram Notifier at some point and removed it from your config, it crashes unless you either uninstall or update the unused dependency.
 - This PR does that. Only shows an error if it tries to send a notification and you don't have a supported dependency installed
 - Failure / Error on startup is what triggered this PR

gazpachoking:
I think the best answer is a configuration error that prevents startup. `Tried to configure plugin telegram, but you do not have the required dependency: telegram > 3.2` or whatever. This would show up immediately and prevent startup _only_ when an user had the plugin in their config, just like other configuration errors. It would be nice to switch all of our optional deps to a system like that. I get that you don't like failing at startup for a plugin that might be a 'bonus' plugin in a task and not required for it's main function, but I don't want to start classing the plugins that way, people use them in different ways.
This PR might be better than the current behavior though. Preventing startup if you don't even have the plugin in your config is not great. I'll try to think of how to implement my idea and go with this or something similar if I can't find a nice way.

PR #4093, opened 2024-12-16
Remove `cloudscraper`

vivodi:
### Motivation for changes:
Currently, `cloudscraper` actually cannot bypass cloudflare verification, and cloudscraper is no longer maintained.


BrutuZ:
Suggestion: add a deprecation warning now and leave removal to the next version bump so xonfigs don't break out of the blue again

vivodi:
> Suggestion: add a deprecation warning now and leave removal to the next version bump so xonfigs don't break out of the blue again

It's not necessary, because removing `cloudscraper` won't break anything.

BrutuZ:
You're removing a plugin, so if it's already in the config file it won't pass validation anymore and Flexget will fail to launch

vivodi:
> You're removing a plugin, so if it's already in the config file it won't pass validation anymore and Flexget will fail to launch

You are right, I misunderstood it before. 

However, I have  bumped the minor version.

BrutuZ:
Hence the suggestion, make the plugin give users a deprecation warning before removing it and breaking configs.

vivodi:
Because @BrutuZ  strongly recommends it, we first prompt users that the plugin has been deprecated, rather than removing it directly.

vivodi:
Because @BrutuZ strongly recommends it, we first prompt users that the plugin has been deprecated, rather than removing it directly.

PR #4094, opened 2024-12-19
Check functionality without extras

vivodi:
### Motivation for changes:
It's a good idea to add some sort of test where we don't install any extras, and make sure we don't crash to prevent issues like we had with the telegram plugin. https://github.com/Flexget/Flexget/pull/4088

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -47,6 +47,8 @@ jobs:
           version: ${{ env.UV_VERSION }}
       - name: Install dependencies
         run: uv sync --group plugin-test --frozen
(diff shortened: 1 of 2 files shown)


vivodi:
It looks like this PR is redundant because `pytest` is currently running without extras. If there are any issues, `pytest` tests will fail.

> [!TIP]
Note that the `--all-extras` argument is not used below.

https://github.com/Flexget/Flexget/blob/b6d0776b2188308456c9f0135b02f21d4cc6bc3d/.github/workflows/test.yml#L48-L52

If you still think it's necessary, I can submit the PR again. @gazpachoking

PR #4098, opened 2024-12-25
Pin major version numbers for dependencies

vivodi:
_Originally posted by @gazpachoking in https://github.com/Flexget/Flexget/issues/4086#issuecomment-2561894130_

Pinning major version numbers for dependencies is a middle-ground approach between pinning full version numbers (which locks everything tightly) and not pinning version numbers at all (which allows for automatic updates).

Pinning major version numbers provides a good balance between stability and keeping dependencies up-to-date. It helps avoid breaking changes while still benefiting from incremental improvements.

After this PR is merged, #4062 can be merged.


vivodi:

gazpachoking:
I think that this would be a good compromise to simplify our dependency workflow but still have things be mostly tested. There are a few things I think we need to figure out, some of them before we pull the trigger on something like this:
- Renovate is sorta busted right now. I think we would want renovate to be sending PRs with updated requirements when new versions come out. This would take the place of the lockfile maintenance PRs (which haven't come out in a while due to afformentioned renovate problems)
- Are we going to widen our ranges when a new version comes out of our deps, or require only the new version? I guess probably the latter to make it more similar to pinning.
- How do we deal with testing the latest minor dep releases? Still have the tests run against the lockfile and keep the lockfile maintenance PRs from renovate to update? I guess that's still best, at least then we can see when test failures are due to dep updates.
- Something seems up with this branch, it's modifying way more files than it should, I think a rebase went wrong or something.

vivodi:
> Renovate is sorta busted right now. I think we would want renovate to be sending PRs with updated requirements when new versions come out. This would take the place of the lockfile maintenance PRs (which haven't come out in a while due to afformentioned renovate problems)

Renovate is now working as expected and will send PRs with updated requirements when new versions are released.

> Are we going to widen our ranges when a new major version comes out of our deps, or require only the new version? I guess probably the latter to make it more similar to pinning.

For major version updates, we’ll stick to requiring only the new version, similar to pinning.

> How do we deal with testing the latest minor dep releases? Still have the tests run against the lockfile and keep the lockfile maintenance PRs from renovate to update? I guess that's still best, at least then we can see when test failures are due to dep updates.

We’ll continue testing against the lockfile and allow Renovate to handle the necessary updates. This way, we can identify any test failures linked to dependency updates.

> Something seems up with this branch, it's modifying way more files than it should, I think a rebase went wrong or something.

It has now been corrected.

Everything should be in order now!

gazpachoking:
Let's go for it and see how things shake out! Thanks!

gazpachoking:
I feel like renovate should have closed or updated the lock file maintenance branch/PR. It has been stuck for a long time. I don't see any errors in the renovate log anymore, but it still didn't handle that branch. I manually closed the ticket and deleted the branch, we'll see if it opens one again after a dep updates.

gazpachoking:
Oh wait, I guess we disabled lock file maintenance. We should have that enabled, no?

vivodi:
> Oh wait, I guess we disabled lock file maintenance. We should have that enabled, no?

The pre-commit hook `uv-lock` requires an internet connection, while pre-commit CI only has network support during installation time, not during runtime. Therefore, it can only be skipped.

vivodi:
I am working on resolving the lockfile issue, but it may take some time.

gazpachoking:
> The pre-commit hook `uv-lock` requires an internet connection, while pre-commit CI only has network support during installation time, not during runtime. Therefore, it can only be skipped.

Oh right, I ran into that when checking out uv. I was considering having github actions do the lock checking for PRs. I still haven't quite figured how we want everything to work. This is my current feeling:
- The lock check that should happen on pre-commit need only check that the lock file is consistent with pyproject.toml, not update anything. Perhaps we need to have a github action to do this check on PRs for developers that don't have pre-commit enabled on the dev machine, since pre-commit CI can't do it online.
- Renovate's lockFileMaintenance will update our lockfile with minor version updates for everything once a week. (I'm going to try to re-enable this and see what it does)
- Renovate normal manager will make a PR to update our pinned major versions when a new major version comes out.

Is that the way you are thinking?

vivodi:
Sorry, I misunderstood your question before.

lockFileMaintenance is no longer needed, 
https://github.com/Flexget/Flexget/blob/71495dee73287a63eca4e759f9ad32c986a320f1/renovate.json#L12
This configuration already ensures that `uv.lock` stays up to date.

> [!TIP]
> update-lockfile = Update the lock file when in-range updates are available, otherwise replace for updates out of range. Works for bundler, cargo, composer, gleam, npm, yarn, pnpm, terraform and poetry so far

gazpachoking:
I think that range strategy doesn't work for the uv manager. This is what we are getting in the log:
```
DEBUG: Unsupported rangeStrategy: update-lockfile. Using "replace" instead.
```

gazpachoking:
I'm also seeing that our lock check action is failing [here](https://github.com/Flexget/Flexget/pull/4108), I think because the lock file containes flexget's own version number now. I guess we need to bump that in our release script?

vivodi:
You can see this PR in my fork repo: https://github.com/vivodi/Flexget/pull/23, which indicates that the setting is working for uv.

That log might be for other managers, as we have `github-actions`, `docker` and `regex` managers, several of which don't support the setting `"rangeStrategy": "update-lockfile"`.

gazpachoking:
Ahh, gotcha. I'll back out my last change then.

gazpachoking:
I'm thinking when we bump the version number in the release script, we probably also need to run `uv lock --upgrade-package flexget`

vivodi:
I'm not sure if bumping the version in the release script can guarantee that the versions in `uv.lock` are always consistent with those in `pyproject.yaml` in all cases. 

Since `pre-commit ci` can't run `uv-lock`, would it be better to run it within GitHub Actions and then write the modified `uv.lock` file to the repository afterward?

gazpachoking:
> I'm not sure if bumping the version in the release script can guarantee that the versions in `uv.lock` are always consistent with those in `pyproject.yaml` in all cases.

Yeah, I think letting github actions do it instead of pre-commit CI would be fine, as long as the only thing that's getting updated is the flexget version itself.

vivodi:
> Ahh, gotcha. I'll back out my last change then.

I take back what I said earlier.

Using `lockFileMaintenance` is better because `"rangeStrategy": "update-lockfile"` can only update FlexGet's own dependencies, not the dependencies of those dependencies. You can change the `rangeStrategy` to `replace` while keeping `lockFileMaintenance` enabled.

vivodi:
> Yeah, I think letting github actions do it instead of pre-commit CI would be fine, as long as the only thing that's getting updated is the flexget version itself.

I am working on this, but it may take some time.

gazpachoking:
Okay, made a few changes, hopefully everything is working as intended.

The other reason I think lock file maintenance is good is that there is only one PR per week (or however often we schedule it) having constant prs for each dep gets a little overwhelming and is one reason I switched to this method in the first place.

gazpachoking:
Ohhh... I just remembered another difference we had when pinning... the transitive deps were also pinned. Now they aren't even constrained to major versions like our direct dependencies...

I guess we just roll with it for a bit and see how things go? 🤷 Maybe we should also reinstate the requirements.txt file with the tested deps for people who don't have uv and want to use the pinned deps.

vivodi:
> maybe also bring back the `locked` extra idea I had in the uv PR? How did you feel about that?
I believe there is no need to provide a `locked` extra for fixed dependencies. 

I believe there is no need to provide a `locked` extra for fixed dependencies.

First, we are not responsible for managing the versions of transitive dependencies. These versions should be determined by the direct dependencies that require them. Package managers will resolve dependencies that meet the specified conditions, so we don’t need to concern ourselves with the exact versions of transitive dependencies.

Second, if Flexget has previously worked with older versions of its dependencies and continues to function with the latest versions—without utilizing new features introduced during updates—it is reasonable to believe that Flexget can operate with any dependency version, as long as the major version is consistent.

Finally, many prominent Python projects, such as [Matplotlib](https://github.com/matplotlib/matplotlib/blob/5a4345217bc2293ed5996ac5ff4f097dacc2368f/pyproject.toml#L34-L42), [Pandas](https://github.com/pandas-dev/pandas/blob/1be26374dd7ef43bc709c4bc6db2daf7bfd606c8/pyproject.toml#L61-L122) and [Pytest](https://github.com/pytest-dev/pytest/blob/5d58b1fa7822fb1b8ff28cacc8a3c37e521e5cc3/pyproject.toml#L49-L64) (click the links to review their dependency versioning practices), do not specify exact dependency versions but instead use version ranges, and these projects also do not (and cannot) test every dependency version during testing. Additionally, these projects do not provide a `locked` extra option. The fact that very few projects enforce fully locked versions is precisely why `uv` does not offer a convenient `locked` extra feature either.

For these reasons, I believe there is no need to provide an option for fixed dependencies. Or could you provide examples of well-known Python projects that fully pin dependency versions or offer a `locked` extra?

gazpachoking:
> I believe there is no need to provide a `locked` extra for fixed dependencies.
> 
> First, we are not responsible for managing the versions of transitive dependencies. These versions should be determined by the direct dependencies that require them. Package managers will resolve dependencies that meet the specified conditions, so we don’t need to concern ourselves with the exact versions of transitive dependencies.
> 
> Second, if Flexget has previously worked with older versions of its dependencies and continues to function with the latest versions—without utilizing new features introduced during updates—it is reasonable to believe that Flexget can operate with any dependency version, as long as the major version is consistent.

Yeah, I think that's mostly true. The problem is that there are times when this isn't true, and we end up getting support tickets. The idea with the pinned dependencies is that a given flexget version implies a given set of dependencies, which we know was run through CI. It reduces a lot of the variables when trying to reproduce or help someone with an issue.

> Finally, many prominent Python projects, such as [Matplotlib](https://github.com/matplotlib/matplotlib/blob/5a4345217bc2293ed5996ac5ff4f097dacc2368f/pyproject.toml#L34-L42), [Pandas](https://github.com/pandas-dev/pandas/blob/1be26374dd7ef43bc709c4bc6db2daf7bfd606c8/pyproject.toml#L61-L122) and [Pytest](https://github.com/pytest-dev/pytest/blob/5d58b1fa7822fb1b8ff28cacc8a3c37e521e5cc3/pyproject.toml#L49-L64) (click the links to review their dependency versioning practices), do not specify exact dependency versions but instead use version ranges, and these projects also do not (and cannot) test every dependency version during testing. Additionally, these projects do not provide a `locked` extra option. The fact that very few projects enforce fully locked versions is precisely why `uv` does not offer a convenient `locked` extra feature either.

This are libraries, meant for developers to use. Flexget is an application meant for end users. Applications are very often distributed with locked down deps. The crux of the issue is that pypi is not really meant to distribute applications, it's designed for libraries, which is why or release and dependency management strategy has always been a bit weird. I don't mind loosening these pinned deps, but for support requests I would like there to be an easy way for end users to install a consistent set of dependencies, both direct and transitive. We could then include those instructions on the issue template to reduce support requests and increase ease of helping people.

I think the simplest to implement solution would be to include a requirements.txt and/or a constraints.txt that could be used by the end user to ensure they were on the tested pinned deps (and could be used by any person attempting to reproduce someone else's problem.) This has the problem of not being easy for someone installing via pip/pipx to use, since they'd have to match the constraints file to the version of flexget they were installing. (something like `pip install flexget==3.13.2 -c https://raw.githubusercontent.com/Flexget/Flexget/refs/tags/v3.13.2/constraints.txt`) Having 'locked' extras was my idea to make it easier for end users, but it complicates our dep management/release workflow.

I'm actually starting to doubt whether it was a good idea to unpin things.. but I'm willing to wait to see how many issues it causes (or doesn't cause) before making a full judgement.

gazpachoking:
Okay, I'm thinking make the 'locked' extra like I had it in the uv PR and include our requirements.txt again (some people may have been using that in their workflows anyway.) I think if we do extras for some of the optional depends, we can directly lock the version in that extra.

gazpachoking:
@vivodi BTW, feel free to come into the chat room for more real time discussion. https://flexget.com/Chat

vivodi:
Your point makes sense to some extent. Thank you for your patient reply.

PR #4123, opened 2025-01-07
Separate doc dependencies

vivodi:
> That one could probably be hidden away. It's only really used for rtd workflow I think. Developers can use uv directly now to do the venv sync.
_Originally posted by @gazpachoking in https://github.com/Flexget/Flexget/issues/4086#issuecomment-2574272985_
            
The PR was completed based on the guidance from @gazpachoking.

Separating documentation dependencies has several benefits:

1. **Reduced Build Size**: By keeping documentation dependencies separate, the main project build is not burdened with unnecessary packages, resulting in a smaller and more efficient build.

2. **Faster Development Process**: Developers can avoid unnecessary installations of documentation-related packages when working on the core functionality of the project, leading to faster setup and iteration times.

3. **Cleaner Dependency Management**: By isolating doc dependencies, it becomes easier to manage and update them without affecting the main dependencies, reducing the risk of conflicts.

4. **Improved Security**: Documentation dependencies often include tools that are not required for production. Keeping them separate reduces the attack surface and limits potential security vulnerabilities.

5. **Easier CI/CD Maintenance**: In continuous integration or continuous deployment workflows, separating doc dependencies helps to avoid running unnecessary documentation tools in production or testing environments, streamlining the process.

6. **Better Versioning Control**: When doc dependencies are kept separate, they can have their own versioning scheme, independent of the main project, ensuring compatibility and making updates more straightforward. 

Overall, it improves project maintainability, security, and efficiency.


vivodi:
`requirements/dev.txt` or `requirements/requirements-dev.txt`? I changed it to the former because the latter might be too verbose.

vivodi:
> Since we renamed it already, I'm okay with putting it in the requirements folder.

`requirements/dev.txt` or `requirements/requirements-dev.txt`? I changed it to the former because the latter might be too verbose.

gazpachoking:
Yep, I'm fine with that rename. @BrutuZ I think we should be done renaming that file on you now. 😄

PR #4134, opened 2025-01-08
Enable the E711/E712 rules and use the `noqa` marker for lines related to SQLAlchemy.

vivodi:
Enable the E711/E712 rules and use the `noqa` marker for lines related to SQLAlchemy, because comparisons to `True`/`False`/`None` are valid for sqlalchemy.

E711: https://docs.astral.sh/ruff/rules/none-comparison/
E712: https://docs.astral.sh/ruff/rules/true-false-comparison/

### DIFF
--- a/flexget/components/estimate_release/estimators/est_series_tvmaze.py
+++ b/flexget/components/estimate_release/estimators/est_series_tvmaze.py
@@ -61,7 +61,7 @@ class EstimatesSeriesTVMaze:
             logger.debug('received air-date: {}', entity.airdate)
             entity_data['entity_date'] = entity.airdate
 
-        if entity_data['data_exists'] == False:
+        if entity_data['data_exists'] is False:
             # Make Lookup to series to see if failed because of no episode or no data
             lookup = api_tvmaze.series_lookup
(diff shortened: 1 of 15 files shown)


vivodi:

gazpachoking:
These aren't set to autofix, right? I don't mind manually marking the #noqa when we do it, just want to make sure the pre-commit won't try to automatically fix these.

vivodi:
Pre-commit won't autofix because they are unsafe.

vivodi:
https://docs.astral.sh/ruff/rules/none-comparison/#fix-safety: This rule's fix is marked as unsafe, as it may alter runtime behavior when used with libraries that override the ==/__eq__ or !=/__ne__ operators. In these cases, is/is not may not be equivalent to ==/!=. 

https://docs.astral.sh/ruff/rules/true-false-comparison/#fix-safety: This rule's fix is marked as unsafe, as it may alter runtime behavior when used with libraries that override the ==/__eq__ or !=/__ne__ operators. In these cases, is/is not may not be equivalent to ==/!=.

PR #4138, opened 2025-01-08
Clean up comparisons to None/True/False

gazpachoking:
### Motivation for changes:

Use more idiomatic comparisons rather than add noqas in #4134


vivodi:
You did a great job. My PR pales in comparison to yours.

PR #4141, opened 2025-01-09
Enforce the PGH rules to prevent the indiscriminate use of `noqa`

vivodi:
### Motivation for changes:
Suppressing all diagnostics can hide issues in the code.

Blanket `noqa` annotations are also more difficult to interpret and maintain, as the annotation does not clarify which diagnostics are intended to be suppressed.

PGH rules: https://docs.astral.sh/ruff/rules/#pygrep-hooks-pgh


vivodi:

PR #4143, opened 2025-01-09
Pin the major version of urllib3

vivodi:
### Motivation for changes:
Fix the major version of urllib3 to avoid PRs like #4115 (the final versions resolved by uv for `urllib3<2` and `urllib3<3` are the same), which make no actual changes.


gazpachoking:
I'm okay with this, but will this actually avoid the renovate PR? Won't it just make a different one to update to the latest version?

vivodi:
Yes, it will make a different one.

However, this PR avoid false positives like #4115, which has no practical effect.

Renovate will make a new PR that truly bumps the `urllib3` version, and we can see if all the tests pass with `urllib3` V2+.

gazpachoking:
Makes sense, and sounds good. Thanks!

PR #4144, opened 2025-01-09
More fork-friendly

vivodi:
### Motivation for changes:
When switching to uv, I noticed during testing of the Docker and release pipelines that some adjustments were needed:
- The name and email of FlexGet-Bot should not be considered secret, as they can be directly seen in the Git history. This PR makes them more straightforward and thus less configurations are required for forked versions of FlexGet to PyPI or Docker Hub.
-  The Docker image name should not be hardcoded.

This PR simplifies the process of releasing forked versions of FlexGet to PyPI or Docker Hub.

If you feel that displaying your email here is inappropriate, you can also consider showing it as `github-actions`.

### DIFF
--- a/.github/workflows/docker-image.yml
+++ b/.github/workflows/docker-image.yml
@@ -51,8 +51,8 @@ jobs:
         uses: docker/metadata-action@369eb591f429131d6889c46b94e711f089e6ca96 # v5
         with:
           images: |
-            flexget/flexget
-            ghcr.io/flexget/flexget
+            ${{ secrets.DOCKERHUB_USERNAME }}/flexget
+            ghcr.io/${{ github.repository_owner }}/flexget
           tags: |
             type=semver,pattern={{version}}
             type=semver,pattern={{major}}.{{minor}}
--- a/.github/workflows/changelog.yml
+++ b/.github/workflows/changelog.yml
@@ -29,8 +29,8 @@ jobs:
           ref: main
       - name: Setup Git User
         run: |
-          git config --global user.email ${{ secrets.git_email }}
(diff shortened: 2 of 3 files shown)


vivodi:

gazpachoking:
Ehh, my email isn't a secret, but I'd prefer not to check it into the repo anyway.

vivodi:
How about displaying it as a more neutral GitHub Actions?
```
git config user.email github-actions[bot]@users.noreply.github.com
git config user.name GitHub Actions
```
image

gazpachoking:
> How about displaying it as a more neutral GitHub Actions?

That's totally fine. I think I set up my own account because that's the only way I could figure out how to do it before.

vivodi:
> That's totally fine. I think I set up my own account because that's the only way I could figure out how to do it before.

Done.

.github/workflows/docker-image.yml
```diff
@@ -51,8 +51,8 @@ jobs:
         uses: docker/metadata-action@369eb591f429131d6889c46b94e711f089e6ca96 # v5
         with:
           images: |
-            flexget/flexget
-            ghcr.io/flexget/flexget
+            ${{ secrets.DOCKERHUB_USERNAME }}/flexget
```
gazpachoking, line 54: I just realized this isn't quite right, and is why we haven't had a release on docker hub in a few days. This should be 'flexget' but my docker username is 'gazpachoking'

vivodi: I'm sorry for causing such an error. I've submitted a PR #4160 for this.

PR #4149, opened 2025-01-10
Possible fix for `uv lock` loop

vivodi:
### Motivation for changes:
A possible fix for `uv lock` loop problem. The rationale is to add extras only in `release.sh`.

https://github.com/Flexget/Flexget/pull/4115#issuecomment-2581807168


gazpachoking:
Hmm, I think maybe the error we were getting is misleading and I misdiagnosed the problem. I think the real problem is that we can't update to urllib3 v2 and still satisfy our other requirements. I disabled all the extras and still got a (much more informative) error about urllib3. I also tried to change one of our other deps that would conflict with the requirements files, and it seemed to go fine. I think let's hold off on more changes until we can get a more clear read on whether there is a problem, and what exactly it is.

vivodi:
> Hmm, I think maybe the error we were getting is misleading and I misdiagnosed the problem. I think the real problem is that we can't update to urllib3 v2 and still satisfy our other requirements. I disabled all the extras and still got a (much more informative) error about urllib3. I also tried to change one of our other deps that would conflict with the requirements files, and it seemed to go fine. I think let's hold off on more changes until we can get a more clear read on whether there is a problem, and what exactly it is.

Your initial judgment is correct. It indeed causes a dependency cycle at the moment. Recent "Lock file maintenance" (#4164, #4133, #4109) updates only include dev dependencies that do not have dependency cycle issues. Moreover, this issue can be easily reproduced:
1. Run:
    ```
    uv lock --upgrade
    ```
2. Comment out the following code:
    ```
    #[tool.hatch.metadata.hooks.requirements_txt.optional-dependencies]
    #deluge = ['requirements/deluge.txt']
    #locked = ['requirements.txt']
    #qbittorrent = ['requirements/qbittorrent.txt']
    #telegram = ['requirements/telegram.txt']
    #transmission = ['requirements/transmission.txt']
    #all = ['requirements/deluge.txt', 'requirements/qbittorrent.txt', 'requirements/telegram.txt', 'requirements/transmission.txt']
    ```
3. Run:
    ```
    uv lock --upgrade
    ```
4. Now you can see many new dependencies.

vivodi:
> Hmm, I think maybe the error we were getting is misleading and I misdiagnosed the problem. I think the real problem is that we can't update to urllib3 v2 and still satisfy our other requirements. I disabled all the extras and still got a (much more informative) error about urllib3. I also tried to change one of our other deps that would conflict with the requirements files, and it seemed to go fine. I think let's hold off on more changes until we can get a more clear read on whether there is a problem, and what exactly it is.

Your initial judgment is correct. It indeed causes a dependency cycle at the moment. Recent "Lock file maintenance" (#4164, #4133, #4109) updates only include dev dependencies that do not have dependency cycle issues. Moreover, this issue can be easily reproduced:
1. Run:
    ```
    uv lock --upgrade
    ```
2. Comment out the following code:
    ```
    #[tool.hatch.metadata.hooks.requirements_txt.optional-dependencies]
    #deluge = ['requirements/deluge.txt']
    #locked = ['requirements.txt']
    #qbittorrent = ['requirements/qbittorrent.txt']
    #telegram = ['requirements/telegram.txt']
    #transmission = ['requirements/transmission.txt']
    #all = ['requirements/deluge.txt', 'requirements/qbittorrent.txt', 'requirements/telegram.txt', 'requirements/transmission.txt']
    ```
3. Run:
    ```
    uv lock --upgrade
    ```
4. Now you can see many new dependencies.

If you believe the issue does exist and the approach in this PR is feasible, I will resolve the merge conflicts and make improvements.

vivodi:
> Hmm, I think maybe the error we were getting is misleading and I misdiagnosed the problem. I think the real problem is that we can't update to urllib3 v2 and still satisfy our other requirements. I disabled all the extras and still got a (much more informative) error about urllib3. I also tried to change one of our other deps that would conflict with the requirements files, and it seemed to go fine. I think let's hold off on more changes until we can get a more clear read on whether there is a problem, and what exactly it is.

Your initial judgment is correct. It indeed causes a dependency cycle at the moment. Recent "Lock file maintenance" (#4164, #4133, #4109) updates only include dev dependencies that do not have dependency cycle issues. Moreover, this issue can be easily reproduced:
1. Run:
    ```
    uv lock --upgrade
    ```
2. Comment out the following code:
    ```
    #[tool.hatch.metadata.hooks.requirements_txt.optional-dependencies]
    #deluge = ['requirements/deluge.txt']
    #locked = ['requirements.txt']
    #qbittorrent = ['requirements/qbittorrent.txt']
    #telegram = ['requirements/telegram.txt']
    #transmission = ['requirements/transmission.txt']
    #all = ['requirements/deluge.txt', 'requirements/qbittorrent.txt', 'requirements/telegram.txt', 'requirements/transmission.txt']
    ```
3. Run:
    ```
    uv lock --upgrade
    ```
4. Now you can see many new dependencies.

gazpachoking:
I think I'm going to have a crack at a hatchling plugin which dynamically generates these first. If that doesn't work out I'll come back to this.

gazpachoking:
Okay, #4190 is what I had in mind, and I think it's a bit more elegant. What do you think?

vivodi:
> Okay, #4190 is what I had in mind, and I think it's a bit more elegant. What do you think?

Your solution is much better than mine. Mine feels more like a temporary fix.

PR #4165, opened 2025-01-13
Run uv-lock in github actions

gazpachoking:
### Motivation for changes:
pre-commit.ci won't run our uv-lock because it requires network access. Make sure things are up to date by running it on github actions

### DIFF
--- a/.github/workflows/uv-lock.yml
+++ b/.github/workflows/uv-lock.yml
@@ -21,3 +21,4 @@ jobs:
     - uses: stefanzweifel/git-auto-commit-action@v5
       with:
         commit_message: Update uv lock file
+        token: ${{ secrets.flexgetbot_pat }}
--- a/.github/workflows/uv-lock.yml
+++ b/.github/workflows/uv-lock.yml
@@ -3,6 +3,7 @@ name: Update uv.lock
 
 on:
   pull_request:
+  pull_request_target:
   push:
     branches: [develop]
 
@@ -13,9 +14,10 @@ jobs:
     - uses: actions/checkout@v4
       with:
         ref: ${{ github.head_ref }}
-    - uses: tox-dev/action-pre-commit-uv@v1
-      with:
-        extra_args: uv-lock
+    - name: Install uv
+      uses: astral-sh/setup-uv@v5
+    - name: Update lock
+      run: uv lock
     - uses: stefanzweifel/git-auto-commit-action@v5
       with:
         commit_message: Update uv lock file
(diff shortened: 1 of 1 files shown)


.github/workflows/uv-lock.yml
```diff
@@ -0,0 +1,21 @@
+# pre-commit.ci does not allow network access, so we run this on github actions
+name: Update uv.lock
+
+on:
+  pull_request:
```
vivodi, line 5: The `pull_request` event can only handle the lockfile for PRs from branches within the same repository, not for PRs from branches in a forked repository. To handle the lockfile for PRs from a forked repository, the `pull_request_target` event is needed, and that is the most complicated part. I can get it done, just need some time.

gazpachoking:
I think everything here would be compatible with the pull_request_target event, but I'm trying to figure out the security implications. I think the issue is that it would run the trusted workflow against the untrusted repo, and somebody could edit the pre-commit-config in order to have our secrets run against their untrusted code. I guess the way to prevent that is to not use the pre-commit-action, such that the action doesn't use any other code from the fork repo when running. Guess that's what you are working on in the other PR. I get it now.

.github/workflows/uv-lock.yml
```diff
@@ -0,0 +1,24 @@
...
+  pre-commit:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v4
+      with:
+        ref: ${{ github.head_ref }}
+    - name: Install uv
+      uses: astral-sh/setup-uv@v5
+    - name: Update lock
+      run: uv lock
```
vivodi, line 20: `uv lock` should be run on the merge commit between the PR and the base branch (that's how pre-commit.ci does it), rather than running `uv lock` on the PR's head commit. Running it on the head commit might result in a situation where all tests pass for the PR, but fail after the merge.

gazpachoking:
Hmm, this feels close now, but clearly something is wrong. 
image

.github/workflows/uv-lock.yml
```diff
@@ -0,0 +1,24 @@
+# pre-commit.ci does not allow network access, so we run this on github actions
+name: Update uv.lock
+
+on:
+  pull_request_target:
```
vivodi, line 5: The `pull_request_target` event runs before GitHub detecting merge conflicts (it runs regardless of whether there are merge conflicts or not), so it’s necessary to check for merge conflicts. This PR should only run if there are no conflicts.

gazpachoking:
I'll hand this back off to you, there are a lot of intricacies I hadn't thought about. 😅

vivodi:
Okay, I'll do it when I have time.

PR #4174, opened 2025-01-17
Resolve the issue of not being able to handle `uv.lock `outdated problem caused by a `uv` version bump within a PR

vivodi:
At present, `autofix.yml` can't handle `uv.lock` outdated problem caused by a `uv` version bump within a PR because the version of `uv` used during execution comes from the `autofix.yml` in the current `develop` branch, not the uv version in a PR as the `pull_request_target` event is based on the base branch.

See discussions in https://github.com/Flexget/Flexget/pull/4169#issuecomment-2596135030

### DIFF
--- a/.github/workflows/autofix.yml
+++ b/.github/workflows/autofix.yml
@@ -1,3 +1,4 @@
+# If you rename or move this file, you MUST update the filename used for YAML parsing below.
 name: Autofix
 on:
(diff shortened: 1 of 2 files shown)


vivodi:

gazpachoking:
Good thinking!

Issue #4167, opened 2025-01-14
Cached_input crashes


paranoidi:
Tried to remedy by deleting clearing tables input_cache and input_cache_entry, did not help.

#### Log:
<details>
 <summary>(click to expand)</summary>

```
BUG: Unhandled error in plugin crossmatch: 'NoneType' object is not iterable
Traceback (most recent call last):

  File "/home/paranoidi/.local/share/uv/python/cpython-3.12.8-linux-x86_64-gnu/lib/python3.12/threading.py", line 1032, in _bootstrap
    self._bootstrap_inner()
    │    └ <function Thread._bootstrap_inner at 0x7f8f7c646ca0>
    └ <Thread(task_queue, started daemon 140254156617472)>
  File "/home/paranoidi/.local/share/uv/python/cpython-3.12.8-linux-x86_64-gnu/lib/python3.12/threading.py", line 1075, in _bootstrap_inner
    self.run()
    │    └ <function Thread.run at 0x7f8f7c646980>
    └ <Thread(task_queue, started daemon 140254156617472)>
  File "/home/paranoidi/.local/share/uv/python/cpython-3.12.8-linux-x86_64-gnu/lib/python3.12/threading.py", line 1012, in run
    self._target(*self._args, **self._kwargs)
    │    │        │    │        │    └ {}
    │    │        │    │        └ <Thread(task_queue, started daemon 140254156617472)>
    │    │        │    └ ()
    │    │        └ <Thread(task_queue, started daemon 140254156617472)>
    │    └ <bound method TaskQueue.run of <flexget.task_queue.TaskQueue object at 0x7f8f77e3e600>>
    └ <Thread(task_queue, started daemon 140254156617472)>

  File "/home/paranoidi/.flexget/flexget/task_queue.py", line 46, in run
    self.current_task.execute()
    │    │            └ <function Task.execute at 0x7f8f79ff82c0>
    │    └ <flexget.task.Task object at 0x7f8f775b6150>
    └ <flexget.task_queue.TaskQueue object at 0x7f8f77e3e600>

  File "/home/paranoidi/.flexget/flexget/task.py", line 89, in wrapper
    return func(self, *args, **kw)
           │    │      │       └ {}
           │    │      └ ()
           │    └ <flexget.task.Task object at 0x7f8f775b6150>
           └ <function Task.execute at 0x7f8f79ff8220>

  File "/home/paranoidi/.flexget/flexget/task.py", line 717, in execute
    self._execute()
[...]

PR #4180, opened 2025-01-20
Adopt the RET rule (all automatic fixes)

vivodi:
### Motivation for changes:
Adopt the RET rule (all automatic fixes)

https://docs.astral.sh/ruff/rules/#flake8-return-ret


gazpachoking:
Looks good! I might have even spotted the issue causing #4167 because of this. 😄

PR #4190, opened 2025-01-23
Add a hatchling hook to create the locked extras.

gazpachoking:
### Motivation for changes:
- Creating the requirements files for all of our extras seemed inelegant.
- The way things were working, our locked extras got added to our requirements when uv was doing resolution, effectively preventing all upgrades.

### Detailed changes:
- Hatch build hook added which adds the pinned extras based on a `uv export`
- Only generate the extras when BUILD_LOCKED_EXTRAS env variable is defined. This lets `uv lock` commands work without knowing about the pinned extras.
- Release job turns on the env variable.
- Users running on a git checkout will no longer have the extras, but this shouldn't be a problem. Doing a `uv sync` will already use the lock file to get the pinned deps. And `uv sync --group groupname` can be used to get the deps for the other extras we provide.


gazpachoking:
Ahh, tomllib isn't on older python, that's why the tests are failing. Perhaps we can just rip out all the toml reading stuff if this uv behavior changes. https://github.com/astral-sh/uv/issues/10882

PR #4195, opened 2025-01-24
Adopt the `PT` rules.

vivodi:
### Motivation for changes:
https://docs.astral.sh/ruff/rules/#flake8-pytest-style-pt

All modifications are automatic fixes.


vivodi:

PR #4208, opened 2025-02-02
Pydantic plugin config models

gazpachoking:
### Motivation for changes
- There have been some issues with defaults not getting set in plugin configs that are set via the schema.
- When there are multiple formats for the config it can be hard to reason about what is guaranteed to be there.

One possible improvement would be to define the plugin configs as pydantic models. This lets editor inspections work as well as opening the possibility of a type checker on the usage. It also does default setting separate than the schema (which I think we should fix even if we don't go with something like this.)

~~This implementation doesn't change anything in core, but it could be moved up to core such that the config gets validated and parsed by a declared model and passed straight to the plugin already in the appropriate format. Didn't want to try that for this quick test.~~ I did a small amount of integration with the core. API v3 requires a `config_model` attribute of the plugin class which is a pydantic model. The core will use this model to generate the schema, as well as validate using the model and hand the plugin back the validated instance rather than the raw config. The pydantic validation happens at task run time though, not config validation time. Still not optimal.

### Current Thoughts
I not sure I'm a fan of this after doing this small test:
- ~~We still needed a prepare_config to get the multiple formats possible for the config into shape. In order for this to be moved up into core we'd probably want to take requirement away. Perhaps it could be done with custom validators on the model, but that might make things even more complicated.~~ I'm happier with the custom validators on the model.
- Switching to pydantic doing the validation might be quite a bit harder. This current implementation just uses it to generate a json schema which does the validation. There is a possibility that the json schema is valid but the instantiation still fails when done later.
- Declaring the json_schema_extra is not so nice for things like dependentRequired as done here. Plus, I think pydantic is on json schema 2020.12, and we are still on draft 4 atm. It is still nice to have the json schema even if we switch to doing pydantic validation, so that we can still publish our schema for use in vscode or pycharm.

#### Good stuff
- I noticed this plugin was referencing multiple keys that weren't actually declared in the schema. Wasn't crashing, but that's pretty confusing. I imagine we have more of those throughout the code base, and this would make it easier to spot.
- Defaults are guaranteed to be set.
- You can reason about all the types, because pydantic guarantees it before handing it off.

Wanted to put this out for discussion though before abandoning the idea.

### Possible ways forward
- Abort completely
- Pydantic things:
  - Do something like this on plugins where we want some more type checking without changing core
  - Change core to do validation using pydantic
- Possible improvements without doing pydantic things:
  - Skip pydantic but change so that 'default' keyword doesn't even try to set defaults. Fix all the plugins that are relying on it.
  - Update to a newer version of json schema in our configs


flexget/plugins/operate/free_space.py
```diff
@@ -1,19 +1,59 @@
...
 
-
-def get_free_space(config, task):
+class AbortMode(str, Enum):
+    BELOW = 'below'
+    ABOVE = 'above'
+
+
+class Config(BaseModel):
+    space: ByteSize | float
```
gazpachoking, line 21: Don't really love this ByteSize | float. The | float is just to generate the schema to allow floats from the user, but when we make an instance of this class it will always be a ByteSize

flexget/plugins/operate/free_space.py
```diff
@@ -1,19 +1,59 @@
...
+        """Converts to the full form of the config from simplified form."""
+        if isinstance(data, (float, int)):
+            data = {'space': data}
+        return data
+
+    model_config = ConfigDict(
+        extra='forbid',
+        json_schema_extra={
+            "type": ["object", "string", "number"],
+            "dependentRequired": {"host": ["user", "ssh_key_filepath", "path"]},
```
gazpachoking, line 51: I don't really love customizing the json schema here, but it is nicer than above, as we are properly describing our schema input differently than our validated output.

flexget/plugins/operate/free_space.py
```diff
@@ -22,100 +62,66 @@ def get_free_space(config, task):
         ssh.set_missing_host_key_policy(paramiko.AutoAddPolicy())
         try:
             ssh.connect(
-                config.get('host'),
-                config.get('port', 22),
-                config.get('user'),
-                config.get('password', None),
-                config.get('pkey', None),
```
gazpachoking, line 29: Password and pkey were never in the schema. Unstructured config didn't really let us see that easily.

flexget/plugins/operate/free_space.py
```diff
@@ -1,19 +1,65 @@
...
+    if isinstance(val, (int, float)):
+        return val * 1024 * 1024
+    return val
+
+
+BackCompatByteSize: TypeAlias = Annotated[
+    ByteSize,
+    Field(json_schema_extra=lambda s: s['anyOf'].append({'type': 'number'})),
+    BeforeValidator(convert_to_bytes),
+]
```
gazpachoking, line 32: I think probably the route to go for this specific change would be to drop backwards compatibility and require all sizes declare their unit. I sorta like that it's possible to describe exactly what we want here though. Schema allows numbers or string with unit, but the parsed config will only be one type. No need to handle the case separate after validation.

gazpachoking:
Don't have an appetite for this now, will probably leave the branch for reference if there is desire in the future.

PR #4224, opened 2025-02-07
[remove] Remove the `cfscraper` plugin

vivodi:
### Motivation for changes:
The `cfscraper` plugin should have been removed in v3.14.0.


gazpachoking:
macos tests are not stable. I'm inclined to just disable them unless you have any idea why

vivodi:
I've disabled them.

gazpachoking:
I guess we need to bump the version again since this will be a breaking change for anyone who didn't already remove this plugin.

PR #4234, opened 2025-02-10
Adopt the `N` rules

vivodi:
### Motivation for Changes:

The `N` rules ensure our code adheres to the [PEP 8](https://www.python.org/dev/peps/pep-0008/) naming conventions, which we should follow.

All fixes have been applied manually.

For more information on the `N` rules, refer to the documentation: [Ruff PEP 8 Naming Rules](https://docs.astral.sh/ruff/rules/#pep8-naming-n).

### DIFF
--- a/flexget/components/ftp/sftp_client.py
+++ b/flexget/components/ftp/sftp_client.py
@@ -346,7 +346,7 @@ class SftpClient:
     def _get_cnopts(self) -> Optional['pysftp.CnOpts']:
         if not self.host_key:
             return None
-        KeyClass = getattr(
+        KeyClass = getattr(  # noqa: N806 It's a class
             importlib.import_module("paramiko"), HOST_KEY_TYPES[self.host_key.key_type]
         )
         key = KeyClass(data=b64decode(self.host_key.public_key))
--- a/flexget/components/managed_lists/lists/imdb_list.py
+++ b/flexget/components/managed_lists/lists/imdb_list.py
@@ -357,7 +357,8 @@ class ImdbEntrySet(MutableSet):
             return '{} list is not modifiable'.format(self.config['list'])
         return None
 
-    def _from_iterable(cls, it):
+    @staticmethod
+    def _from_iterable(it):
         # TODO: is this the right answer? the returned object won't have our custom __contains__ logic
         return set(it)
 
--- a/flexget/components/managed_lists/lists/ombi_list.py
+++ b/flexget/components/managed_lists/lists/ombi_list.py
@@ -152,7 +152,7 @@ class OmbiEntry:
             self.ombi_title = self.ombi_title + ' E' + str(data['tmdb_episode']).zfill(2)
 
     @property
-    def requestId(self) -> str:
+    def request_id(self) -> str:
         if "requestId" in self.data:
             return self.data['requestId']
 
@@ -160,8 +160,8 @@ class OmbiEntry:
         self.data = new_data
         return self.data['requestId']
 
-    @requestId.setter
-    def requestId(self, value: str):
+    @request_id.setter
+    def request_id(self, value: str):
         self.data['requestId'] = value
 
     def already_requested(self) -> tuple[bool, str]:
@@ -197,7 +197,7 @@ class OmbiEntry:
                 endpoint=endpoint, data=data, headers=headers
             )
 
-            self.requestId = response['requestId']
+            self.request_id = response['requestId']
 
             log.info(f"{self.ombi_title} was requested in Ombi.")
             return True
@@ -229,7 +229,7 @@ class OmbiEntry:
 
         api_endpoint = f"api/v1/Request/{self.entry_type}/available"
 
-        data = {"id": self.requestId}
+        data = {"id": self.request_id}
 
         headers = self._request.create_json_headers()
 
@@ -252,7 +252,7 @@ class OmbiEntry:
 
         log.info(f"Marking {self.ombi_title} as deleted in Ombi.")
 
-        api_endpoint = f"api/v1/Request/{self.entry_type}/{self.requestId}"
+        api_endpoint = f"api/v1/Request/{self.entry_type}/{self.request_id}"
 
         headers = self._request.create_json_headers()
 
@@ -281,7 +281,7 @@ class OmbiEntry:
 
         api_endpoint = f"api/v1/Request/{self.entry_type}/unavailable"
 
-        data = {"id": self.requestId}
+        data = {"id": self.request_id}
 
         headers = self._request.create_json_headers()
 
@@ -307,7 +307,7 @@ class OmbiEntry:
         api_endpoint = f"api/v1/Request/{self.entry_type}/deny"
 
         # In the future, we might want to allow the user to specify a reason.
-        data = {"id": self.requestId, "reason": "Denied by Flexget automation."}
+        data = {"id": self.request_id, "reason": "Denied by Flexget automation."}
 
         headers = self._request.create_json_headers()
 
@@ -333,7 +333,7 @@ class OmbiEntry:
         api_endpoint = f"api/v1/Request/{self.entry_type}/approve"
 
         # In the future, we might want to allow the user to specify a reason.
-        data = {"id": self.requestId}
+        data = {"id": self.request_id}
 
         headers = self._request.create_json_headers()
 
--- a/flexget/components/seen/cli.py
+++ b/flexget/components/seen/cli.py
@@ -55,7 +55,7 @@ def seen_forget(manager: Manager, options):
 
 
 def seen_add(manager: Manager, options):
-    DEFAULT_TASK = 'cli_add'
+    default_task = 'cli_add'
 
     seen_name = options.add_value
     if is_imdb_url(seen_name):
@@ -78,13 +78,13 @@ def seen_add(manager: Manager, options):
             console('No specified tasks in config.')
             return
     else:
-        tasks = [DEFAULT_TASK]
+        tasks = [default_task]
         local = None
 
     for task in tasks:
         db.add(seen_name, task, {'cli_add': seen_name}, local=local)
 
-    if tasks == [DEFAULT_TASK]:
+    if tasks == [default_task]:
         console(f'Added `{seen_name}` as seen. This will affect all tasks.')
     else:
         console(
--- a/flexget/components/series/db.py
+++ b/flexget/components/series/db.py
@@ -172,10 +172,10 @@ class Season(Base):
         return min(release.first_seen for release in self.releases)
 
     @first_seen.expression
-    def first_seen(cls):
+    def first_seen(self):
         return (
             select(func.min(SeasonRelease.first_seen))
-            .where(SeasonRelease.season_id == cls.id)
+            .where(SeasonRelease.season_id == self.id)
             .correlate(Season.__table__)
             .label('first_seen')
         )
@@ -281,10 +281,10 @@ class Episode(Base):
         return min(release.first_seen for release in self.releases)
 
     @first_seen.expression
-    def first_seen(cls):
+    def first_seen(self):
         return (
             select(func.min(EpisodeRelease.first_seen))
-            .where(EpisodeRelease.episode_id == cls.id)
+            .where(EpisodeRelease.episode_id == self.id)
             .correlate(Episode.__table__)
             .label('first_seen')
         )
--- a/flexget/components/series/db.py
+++ b/flexget/components/series/db.py
@@ -172,10 +172,11 @@ class Season(Base):
         return min(release.first_seen for release in self.releases)
 
     @first_seen.expression
-    def first_seen(self):
+    @classmethod
+    def first_seen(cls):
         return (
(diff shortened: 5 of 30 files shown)


vivodi:

flexget/components/ftp/sftp_client.py
```diff
@@ -346,10 +346,10 @@ def _connect(self, connection_tries: int) -> 'pysftp.Connection':
     def _get_cnopts(self) -> Optional['pysftp.CnOpts']:
         if not self.host_key:
             return None
-        KeyClass = getattr(
+        key_class = getattr(
             importlib.import_module("paramiko"), HOST_KEY_TYPES[self.host_key.key_type]
         )
-        key = KeyClass(data=b64decode(self.host_key.public_key))
+        key = key_class(data=b64decode(self.host_key.public_key))
```
gazpachoking, line 352: Not familiar with this code, but seems like this might be a class.

flexget/components/imdb/db.py
```diff
@@ -225,7 +225,7 @@ def upgrade(ver, session):
     # v9  Added Metacritic score exftraction/filtering
     # v10 Added plot keywords to the DB schema
     if ver is None or ver <= 9:
-        raise UpgradeImpossible(
+        raise UpgradeImpossibleError(
```
gazpachoking, line 228: Don't think I like this better.

flexget/components/managed_lists/lists/imdb_list.py
```diff
@@ -357,7 +357,7 @@ def immutable(self):
             return '{} list is not modifiable'.format(self.config['list'])
         return None
 
-    def _from_iterable(cls, it):
+    def _from_iterable(self, it):
```
gazpachoking, line 360: Adding the rule for this seems okay.

flexget/components/managed_lists/lists/ombi_list.py
```diff
@@ -152,16 +152,16 @@ def __init__(
             self.ombi_title = self.ombi_title + ' E' + str(data['tmdb_episode']).zfill(2)
 
     @property
-    def requestId(self) -> str:
+    def request_id(self) -> str:
```
gazpachoking, line 155: I'm a bit torn on this. Sometimes I think it's okay to use camelcase when we are mirroring another api.

flexget/components/seen/cli.py
```diff
@@ -55,7 +55,7 @@ def seen_forget(manager: Manager, options):
 
 
 def seen_add(manager: Manager, options):
-    DEFAULT_TASK = 'cli_add'
+    default_task = 'cli_add'
```
gazpachoking, line 58: This feels like a constant.

flexget/plugin.py
```diff
@@ -108,7 +108,7 @@ def __str__(self):
 
 
 # TODO: move to utils or somewhere more appropriate
-class internet:
+class Internet:
```
gazpachoking, line 111: Don't think I like this. Whether it's implemented as a class or function seems to be an implementation detail to me. It acts like a function in usage.

flexget/utils/cached_input.py
```diff
@@ -104,7 +104,7 @@ def db_cleanup(manager, session: DBSession) -> None:
         logger.verbose('Removed {} old input caches.', result)
 
 
-class cached:
+class Cached:
```
gazpachoking, line 107: Yeah, this seems like another one where usage is as if it's a function even though it's declared as a class.

flexget/task.py
```diff
@@ -159,13 +159,13 @@ def __repr__(self) -> str:
         return f'<EntryContainer({list.__repr__(self)})>'
 
 
-class TaskAbort(Exception):
+class TaskAbortError(Exception):
```
gazpachoking, line 162: Don't think this is better. It's not necessarily an error, it can be similar to GeneratorExit, which also doesn't have the Error suffix.

flexget/tests/api_tests/test_database_api.py
```diff
@@ -37,5 +37,5 @@ def test_database_get_plugins(self, api_client, schema_match):
         assert rsp.status_code == 200
         data = json.loads(rsp.get_data(as_text=True))
 
-        errors = schema_match(OC.plugin_list, data)
+        errors = schema_match(ObjectsContainer.plugin_list, data)
```
gazpachoking, line 40: What rule is this even? About not being camelcase? I'd argue OC is camelcase.

gazpachoking (changes requested):
Don't think I'm a fan of these ones. I didn't look over everything exhaustively, but I'm neutral to negative on most of these changes. I don't want to have to litter #noqa s about in order to name things as we want either. There might be a subset of these rules that are okay. The only one I noticed so far was about naming the first argument to a method cls or self, that one seems good.

flexget/api/app.py
```diff
@@ -257,27 +257,27 @@ class NotFoundError(APIError):
...
 
 
-class PreconditionFailed(APIError):
+class PreconditionFailedError(APIError):
     status_code = 412
     description = 'Precondition failed'
 
 
-class NotModified(APIError):
+class NotModifiedError(APIError):
```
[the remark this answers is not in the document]
vivodi: Okay, I'll revert this change.

flexget/components/ftp/sftp_client.py
vivodi: KeyClass is indeed a class.  
```py
KeyClass = getattr(importlib.import_module("paramiko"), HOST_KEY_TYPES[self.host_key.key_type])
```  
The code above violates the rule '[N806](https://docs.astral.sh/ruff/rules/non-lowercase-variable-in-function/) Variable `KeyClass` in function should be lowercase'. In this case, I tend to use `# noqa` because this rule applies in most cases.

flexget/plugin.py
vivodi: Most classes should have capitalized first letters. Should I use `noqa` here or disable the rule?

flexget/tests/api_tests/test_database_api.py
vivodi: It is [N817](https://docs.astral.sh/ruff/rules/camelcase-imported-as-acronym/) `camelcase-imported-as-acronym`. Ruff considers fully uppercase `OC` to be potentially mistaken for a constant by readers. If you don't like this rule, I will disable it.

flexget/components/managed_lists/lists/ombi_list.py
vivodi: Since this rule applies in most cases, `noqa` can be used for the rare exceptions.

flexget/api/app.py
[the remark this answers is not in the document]
vivodi: Okay, I'll disable the corresponding rule.

flexget/components/seen/cli.py
vivodi: According to PEP 8’s [Constants](https://peps.python.org/pep-0008/#constants), constants are defined at the module level, not within functions. This is a constant defined inside a function. Should I use `noqa`, disable the rule, or keep this constant in lowercase?

vivodi:
Thank you for carefully reviewing these changes!

There are still a few areas where I’m unsure how to handle them, so I’d appreciate your final decision @gazpachoking .

Since many rules apply in most cases, using `noqa` for rare exceptions seems reasonable.

flexget/components/seen/cli.py
gazpachoking: I've come around, I think this instance is fine in lower case.

flexget/config_schema.py
```diff
@@ -436,13 +436,13 @@ def validate_properties_w_defaults(validator, properties, instance, schema):
     yield from BaseValidator.VALIDATORS["properties"](validator, properties, instance, schema)
 
 
-def validate_anyOf(validator, anyOf, instance, schema):
-    errors = BaseValidator.VALIDATORS["anyOf"](validator, anyOf, instance, schema)
+def validate_any_of(validator, any_of, instance, schema):
+    errors = BaseValidator.VALIDATORS["anyOf"](validator, any_of, instance, schema)
```
gazpachoking, line 440: I think leaving the camelcase in this file is fine, since it's emulating the jsonschema keywords.

flexget/plugin.py
gazpachoking: I think noqa is okay for this one and `cache`

flexget/plugins/clients/aria2.py
```diff
@@ -61,12 +61,12 @@ def _get_req_params(self, method, params=None):
             del req_params['params']
         return req_params
 
-    def _default_error_handle(code, message):
-        logger.critical('Fault code {} message {}', code, message)
-        raise plugin.PluginError(f'Fault code {code} message {message}', logger)
+    def _default_error_handle(self, message):
+        logger.critical('Fault code {} message {}', self, message)
+        raise plugin.PluginError(f'Fault code {self} message {message}', logger)
```
gazpachoking, line 66: I think in this one it's secretly being used as a `@staticmethod` by virtue of referencing these functions from the function signatures before the class actually gets to make them into methods. I believe it won't break anything to just put `@staticmethod` on the two methods in this file and leave the variable names the same.

flexget/tests/api_tests/test_database_api.py
gazpachoking: I think we can disable this rule.

flexget/tests/test_sort_by.py
```diff
@@ -96,9 +96,9 @@ class TestSortBy:
               dict_field: {a: 2, b: 0}
     """
 
-    def generate_test_ids(param):
-        if param[0:5] == 'test_' or not isinstance(param, list):
-            return param
+    def generate_test_ids(self):
```
gazpachoking, line 99: I think this one is also a `@staticmethod`

flexget/utils/cached_input.py
gazpachoking: Yeah, noqa is good here.

flexget/utils/database.py
```diff
@@ -34,10 +34,10 @@ def wrapper(*args, **kwargs):
     if len(args) == 1 and not kwargs and callable(args[0]):
         # Used without arguments, e.g. @with_session
         # We default to expire_on_commit being false, in case the decorated function returns db instances
-        _Session = functools.partial(Session, expire_on_commit=False)
+        _session = functools.partial(Session, expire_on_commit=False)
```
gazpachoking, line 37: noqa on this one, it's acting as a class

flexget/components/managed_lists/lists/ombi_list.py
gazpachoking: I think actually this particular instance is fine to just be lower case.

gazpachoking (commented):
Okay, I've come around on most of this. I think we should disable the rule for exception naming. Maybe a few of those instances would be better with the Error suffix, but I don't like it on a lot of them. Easier just to leave that rule disabled than go through all of them and noqa or rename. `internet` and `cache` should be noqaed, along with a few other things I left comments on. There are a couple instances where methods need to be marked as static rather than rename the first argument that I've noted as well. The renames in config_schema are the ones I'm most torn about, as it's emulating the camelcase from json schema, but looking at it again I think snake_case is nicer than noqas.

flexget/config_schema.py
gazpachoking: Oops, I conflicted with myself on this comment. I've changed my mind and think snake_case is nicer than noqaing them.

flexget/tests/test_sort_by.py
vivodi: The modification failed the tests on Python 3.9, so I added `#noqa`. After dropping support for Python 3.9 later this year, `#noqa` can be removed and `@staticmethod` can be added.

vivodi:
I have made the changes everywhere as you suggested.

flexget/components/series/db.py
```diff
@@ -281,10 +281,10 @@ def first_seen(self):
         return min(release.first_seen for release in self.releases)
 
     @first_seen.expression
-    def first_seen(cls):
+    def first_seen(self):
```
gazpachoking, line 284: This one needs @classmethod too

gazpachoking (commented):
Oops, noticed one more thing, otherwise looks good! Thanks

flexget/components/series/db.py
```diff
@@ -172,10 +172,10 @@ def first_seen(self):
...
```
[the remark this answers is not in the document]
vivodi: Done.

flexget/components/status/db.py
```diff
@@ -47,10 +47,10 @@ def last_execution_time(self):
         return max(execution.start for execution in self.executions)
 
     @last_execution_time.expression
-    def last_execution_time(cls):
+    def last_execution_time(self):
```
gazpachoking, line 50: Doh. One more classmethod.

vivodi: Done.

gazpachoking:
Looks good, thanks!

PR #4238, opened 2025-02-10
[Change] Make `exists_movie` recurse subfolders

BrutuZ:
### Motivation for changes:
The following `exists_movie` config 
```yaml
    exists_movie:
      allow_different_qualities: better
      type: files
      path:
        - /media/Movies/
```
didn't behave as expected because it doesn't support the following folder structure.
```
/media/Movies/
├─── Cool Movie [2025]
│   ├─── Cool.Movie.2025.BluRay.2160p.HDR.x265.AAC.mkv
```
Because it wouldn't look into the folders themselves that contain the files, so the `allow_different_qualities: better` and `type: files` would never match, causing no entries to be rejected and duplicates being downloaded on every run.

### Detailed changes:
- Changed the `iterdir` call for a `rglob` that does look recursively into the path if the new `recursive` boolean parameter is set to `true`.

- Personal gripe

```diff
exists_movie:
  path: /path/to/movies
  [type: {dirs|files}]
  [allow_different_qualities: {better|yes|no}]
  [lookup: {imdb|no}]
+ [recursive: {yes|no}]
```

### Ideas
 - [x] Make it a setting to prevent introducing slowdowns on existing configs checking large trees?

### DIFF
--- a/flexget/plugins/filter/exists_movie.py
+++ b/flexget/plugins/filter/exists_movie.py
@@ -28,6 +28,7 @@ class FilterExistsMovie:
         [type: {dirs|files}]
         [allow_different_qualities: {better|yes|no}]
         [lookup: {imdb|no}]
+        [recursive: {yes|no}]
     """
 
     schema = {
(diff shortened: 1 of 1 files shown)


gazpachoking:
Seems good to me! Care to add it to the wiki as well?

PR #4241, opened 2025-02-11
Adopt the `G` rules

vivodi:
### Motivation for changes:
All changes were made manually.

G rules: https://docs.astral.sh/ruff/rules/#flake8-logging-format-g

### DIFF
--- a/flexget/components/archives/decompress.py
+++ b/flexget/components/archives/decompress.py
@@ -188,7 +188,7 @@ class Decompress:
             if not test:
                 archive.delete()
             else:
-                logger.info(f'Would delete archive {archive.path}')
+                logger.info('Would delete archive {}', archive.path)
                 archive.close()
         else:
             archive.close()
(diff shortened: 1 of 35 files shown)


vivodi:

gazpachoking:
I sorta feel like the f strings are more readable, I'd be inclined to disable the rule that disallows those. How much of a performance impact does this really make? I suspect it's low enough to not worry about it. The other rules seem fine.

vivodi:
@paranoidi said [here](https://github.com/Flexget/Flexget/pull/1549#discussion_r92422969):
> Don't use string formatting with logger, use params

 and @gazpachoking agreed with this point, saying:
> The params way also avoids some extra work in certain cases. When a message won't be displayed at the current logging level, the system never even does the string replacement. If you don't use parameters, the string replacement work has to be done no matter if the log message is going to be displayed or not. This is normally not a big issue, but could be bad inside some heavy loops.

I don't deny that f-strings may have a slight advantage in readability, but for the sake of consistency and performance, adopting the rule is better. The existence of the rule itself justifies its validity, and I hope you'll reconsider.

gazpachoking:
Okay, I'm good. It is technically more efficient, even if that might be minor, and we can use the named format route if it's important to have the variable names inline. Thanks!

PR #4242, opened 2025-02-11
Fix default value for list kind in imdb_watchlist

pogman-code:
### Motivation for changes:

I realised that, in my previous PR, I wrongfully set the default value of `kind` to "predefinedList" instead of "list".  
"predefinedList" is used for IMDB's watchlist only, other list use the "list" key so it should be the default `kind`.

My bad!


gazpachoking:
Thanks!

PR #4247, opened 2025-02-13
Adopt the `PIE` rules

vivodi:
### Motivation for changes:

All fixes were applied automatically.

PIE: https://docs.astral.sh/ruff/rules/#flake8-pie-pie


vivodi:

PR #4248, opened 2025-02-13
Adopt the `PYI` rules

vivodi:
PYI: https://docs.astral.sh/ruff/rules/#flake8-pyi-pyi


vivodi:

PR #4249, opened 2025-02-13
Adopt the `TRY` rule (This PR only includes automatic fixes)

vivodi:
### Motivation for changes:
This PR only includes automatic fixes for the [TRY400](https://docs.astral.sh/ruff/rules/error-instead-of-exception/) rule. The benefits of adopting this rule are explained in the [linked document](https://docs.astral.sh/ruff/rules/error-instead-of-exception/#why-is-this-bad). Manual fixes for the `TRY` rule will be included in a separate PR.


vivodi:

PR #4250, opened 2025-02-13
Update `readthedocs` Python to v3.13 (manual intervention required)

vivodi:
### Motivation for changes:
Fix #4221


vivodi:

gazpachoking:
It looks like pendulum 3.1 with python 3.13 support might be imminent. Maybe we can just wait for that given that it doesn't really matter what version of python rtd runs.

vivodi:
Alright, I'm looking forward to the upcoming Pendulum version.

PR #4254, opened 2025-02-13
Add strip_dates property and more Entry fields to imdb_watchlist

pogman-code:
### Motivation for changes:

Since most *_list plugins have it, adds strip_dates property to imdb_watchlist
Also adds more Entry fields when parsing items.

### Detailed changes:
- add `strip_dates` property
- append the date to the title depending on the config
- add `imdb_url`, `imdb_original_name`, `movie_name`, `series_name`, `movie_year`, `series_year`, `imdb_score` and `imdb_votes`

### DIFF
--- a/flexget/components/imdb/imdb_watchlist.py
+++ b/flexget/components/imdb/imdb_watchlist.py
@@ -136,7 +136,6 @@ class ImdbWatchlist:
             logger.verbose('Nothing found in imdb list: {}', config['list'])
             return []
 
-        entries = []
         page_no = 1
 
         while len(items) < total_item_count:
@@ -156,10 +155,7 @@ class ImdbWatchlist:
             except Exception:
                 raise plugin.PluginError('Received invalid list data')
 
-        for item in items:
-            entries.append(self.parse_entry(item['listItem'], config))
-
-        return entries
+        return [self.parse_entry(item['listItem'], config) for item in items]
 
     def parse_entry(self, item, config) -> Entry:
         entry = Entry()
--- a/flexget/components/imdb/imdb_watchlist.py
+++ b/flexget/components/imdb/imdb_watchlist.py
@@ -1,5 +1,3 @@
-import contextlib
-
 from loguru import logger
 
 from flexget import plugin
--- a/flexget/components/imdb/imdb_watchlist.py
+++ b/flexget/components/imdb/imdb_watchlist.py
@@ -29,7 +29,7 @@ TITLE_TYPE_MAP = {
 
 
 class ImdbWatchlist:
-    """ "Creates an entry for each movie in your imdb list."""
+    """Creates an entry for each movie in your imdb list."""
 
     schema = {
         'type': 'object',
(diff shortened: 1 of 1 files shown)


pogman-code:
While I am at it, can someone tell me if there is some Entry fields that I should add (like `series_name`, `movie_name`, etc.) that would be relevant?

Also, regarding `strip_dates`, beside the fact that it removes the date from the `title` field, what does it imply?  
I added it because I used it with `trakt_list` but now I am wondering what is the point of it really :D

flexget/components/imdb/imdb_watchlist.py
```diff
@@ -155,25 +155,43 @@ def parse_html_list(self, task, config, url, params, headers, kind='list') -> li
...
+        entry['imdb_id'] = item['id']
+
+        entry['imdb_name'] = title
+        entry['imdb_original_name'] = item['originalTitleText']['text']
+        if title_type in ['movie', 'tvMovie']:
+            entry['movie_name'] = title
+        elif title_type in ['tvSeries', 'tvMiniSeries']:
+            entry['series_name'] = title
+
+        with contextlib.suppress(ValueError, TypeError):
```
gazpachoking, line 176: What errors are we expecting this might be catching?

gazpachoking (commented):
Looks good to me, except I'm not fully sure what errors we are suppressing and why for the year setting part. As to why strip dates is useful... it's probably only useful in very limited situations. In configure_series, leaving the date off makes the matching of the series name a bit looser, though usually not desirable. There might be a couple more cases.

flexget/components/imdb/imdb_watchlist.py
pogman-code: For movies/series that are not released yet, `item['releaseYear']['year']` is not availble, that's why I added this suppression.

There is the same thing with ratings few lines down :)

pogman-code:
> As to why strip dates is useful... it's probably only useful in very limited situations. In configure_series, leaving the date off makes the matching of the series name a bit looser, though usually not desirable. There might be a couple more cases. (nobody has been clamoring for it in this plugin, but it's probably nice to include for parity)

I tried to play with it a bit and noticed that some search plugins does not handle well when the year is present, so I understand the need of it

flexget/components/imdb/imdb_watchlist.py
gazpachoking: Wouldn't it be a `KeyError` if this key isn't present? Maybe it's better to just check? `if 'releaseYear' in item`

gazpachoking: This suppress just feels ambiguous to me, as I don't know which call within here would cause either of those errors. The one below makes slightly more sense, as `float()` could cause those errors. (but down there I don't know why the votecount line is inside that supress)

pogman-code: You are right, I went a bit head first/brain off when refactoring this part.

For released movies/series those properties look like this in the JSON:
```
"releaseYear":{
  "year":2025,
  "endYear":"None",
  "__typename":"YearRange"
},
"ratingsSummary":{
  "aggregateRating":7.1,
  "voteCount":3345,
  "__typename":"RatingsSummary"
},
```

And like this whe unreleased:
```
  "releaseYear":"None",
  "ratingsSummary":{
    "aggregateRating":"None",
    "voteCount":0,
    "__typename":"RatingsSummary"
  },
```

So i replaced both `suppress` by a more fitting `if` statement in 43851f7f5603633da32c75c30616a2cd97876b2f.

Thanks for your feedback :)

gazpachoking:
Thanks!

pogman-code:
> hi @pogman-code, thanks for adding this, but with the latest v3.15.14, I see an exception at the IMDb task:
> 
> Logs
> Can you please take a look and fix it?
Hi, sorry about that, it has been fixed #4275
It should be available in next build (3.15.15), my bad 😞

PR #4255, opened 2025-02-14
Adopt the `D` rules (automatic fixes only)

vivodi:
Manual fixes will be submitted in another PR.

`D` rules: https://docs.astral.sh/ruff/rules/#pydocstyle-d


vivodi:

PR #4261, opened 2025-02-17
[remove] Remove unused or redundant dependencies

vivodi:
### Motivation for changes:
Remove unused or redundant dependencies.


gazpachoking:
Don't we need sphinx for rtd? How is that working now?

Moving requirements-dev again? (@BrutuZ 😅)

BrutuZ:
I've learned my lesson since and moved to UV xD
Thanks for the heads up though o7

vivodi:
> Don't we need sphinx for rtd? How is that working now?
> 
> Moving requirements-dev again? (@BrutuZ 😅)

According to the [Read the Docs documentation](https://docs.readthedocs.com/platform/stable/config-file/index.html#adjust-sphinx-and-mkdocs-version), Read the Docs automatically installs Sphinx, so there's no need to explicitly specify Sphinx. If there are other dependencies required for building the documentation, they should be specified explicitly.

vivodi:
> I've learned my lesson since and moved to UV xD Thanks for the heads up though o7

Since @BrutuZ has decided to use uv, is it still necessary to keep the `requirements-dev.txt`?

gazpachoking:
> Since @BrutuZ has decided to use uv, is it still necessary to keep the `requirements-dev.txt`?

Probably not strictly necessary. Might be fine to remove it. We probably would have heard about it if more people were using it since we broke that workflow a couple times already. 😜

vivodi:
> Probably not strictly necessary. Might be fine to remove it. We probably would have heard about it if more people were using it since we broke that workflow a couple times already. 😜

I've removed it.

gazpachoking:
Thanks!

PR #4262, opened 2025-02-18
preserve transmission labels order

fedej:
### Motivation for changes:

I have a task that adds labels to a torrent and another one that reads labels from torrents, expecting them to be in the same order they were added. Currently, because a regular set is used, the order in which they are defined in the config is not repected.

### Detailed changes:
- Used a dictionary instead of a set, since it preserves insertion order. Alternatively, it could be done with a list and an inclusion check.

- Fixes # . Didn't create an issue for this, but can do so if needed

Added a unit test that was failing and now passing. Although the random nature of the set could eventually make it pass
```
...
flexget/tests/test_transmissiontorrent.py::TestTransmissionTorrentPlugin::test_output_labels_order
<redacted>/__init__.py:3142: DeprecationWarning: Deprecated call to `pkg_resources.declare_namespace('zc')`.
  Implementing implicit namespace packages (as specified in PEP 420) is preferred to `pkg_resources.declare_namespace`. See https://setuptools.pypa.io/en/latest/references/keywords.html#keyword-namespace-packages
    declare_namespace(pkg)

-- Docs: https://docs.pytest.org/en/stable/how-to/capture-warnings.html
========================================================== 1 passed, 4 warnings in 1.39s ===========================================================
```
#### To Do:

- [ ] Stuff..

### DIFF
--- a/flexget/plugins/clients/transmission.py
+++ b/flexget/plugins/clients/transmission.py
@@ -744,12 +744,12 @@ class PluginTransmission(TransmissionBase):
             change['queuePosition'] = opt_dic['queue_position']
 
         if 'labels' in opt_dic:
-            labels = set()
+            labels = {}
             for obj in [config, entry]:
                 obj_labels = obj.get('labels', [])
                 if not isinstance(obj_labels, list):
                     obj_labels = [obj_labels]
-                labels.update(obj_labels)
+                labels.update({obj_label: None for obj_label in obj_labels})
             rendered_labels = []
             for label in labels:
                 try:
--- a/flexget/tests/test_transmission.py
+++ b/flexget/tests/test_transmission.py
@@ -1,0 +1,35 @@
+from unittest import mock
+
+
+@mock.patch('flexget.plugins.clients.transmission.transmission_rpc')
+class TestTransmissionTorrentPlugin:
+    config = """
+        tasks:
+            test_input:
+                mock:
+                    - {title: 'title1', url: 'url1', transmission_id: '321', file: 'file1'}
+                accept_all: yes
+                transmission:
+                    host: localhost
+                    port: 9091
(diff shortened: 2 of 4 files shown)


fedej:
Transmissionrpc lib is not installed in the runner environment, apologies for the noise I'll try to look at this again tomorrow

vivodi:
> Transmissionrpc lib is not installed in the runner environment, apologies for the noise I'll try to look at this again tomorrow

We intentionally avoided installing optional dependencies in the CI to ensure that other functionalities would work properly even without them, in order to avoid issues like https://github.com/Flexget/Flexget/pull/4088.

I am trying to add support for testing extras without breaking the above.

fedej:
> > Transmissionrpc lib is not installed in the runner environment, apologies for the noise I'll try to look at this again tomorrow
> 
> We intentionally avoided installing optional dependencies in the CI to ensure that other functionalities would work properly even without them, in order to avoid issues like #4088.
> 
> I am working on adding support for testing extras while preserving the integrity of the current setup.

Sounds good, I'll wait for that and rebase when ready. Thanks!

gazpachoking:
> Sounds good, I'll wait for that and rebase when ready. Thanks!

Okay, we merged the branch for testing plugins with optional deps. Should just need to add the mark `@pytest.mark.requires_optional_deps` and you'll be good to go.

fedej:
> > Sounds good, I'll wait for that and rebase when ready. Thanks!
> 
> Okay, we merged the branch for testing plugins with optional deps. Should just need to add the mark `@pytest.mark.requires_optional_deps` and you'll be good to go.

@gazpachoking @vivodi I updated the test and dependencies configuration. Thanks!

gazpachoking:
Thanks!

PR #4264, opened 2025-02-18
[add] Add support for testing extras

vivodi:
### Motivation for changes:
Previously, we deliberately avoided installing optional dependencies in the CI to ensure that other functionalities would work properly even without them, in order to avoid issues like #4088. However, now with #4262, there is a need to add testing for extras, which will require changes to the workflow.

To ensure that other functionalities work properly without optional dependencies, we require the addition of `@pytest.mark.extra` decorator for testing extras. This way, we can first test the non-extras functionality and then test the extras.

```py
@pytest.mark.extra
def test_telegram():
    pass
```


vivodi:

This PR is not intended to fully test the Telegram notifier. The purpose of this PR is to provide support for testing extras.

.github/workflows/test.yml
```diff
@@ -30,11 +30,14 @@ jobs:
           enable-cache: true
           python-version: ${{ matrix.python-version }}
           version: ${{ env.UV_VERSION }}
-      - name: Install dependencies
-        run: uv sync --group plugin-test --frozen
-      - name: Test with pytest
-        run: |
-          uv run pytest -n logical --dist loadgroup --junit-xml=test-results.xml
+      - name: Install mandatory dependencies
+        run: uv sync --frozen --group plugin-test
```
gazpachoking, line 34: I like the idea of this PR. Maybe we should mark the tests that need the plugins in this group as well, so that we can just do a bare sync for the main tests? Maybe even just have this as the only group that gets installed later instead of also 'all'

gazpachoking: Is there an advantage to having 2 separate marks and groups here? I was thinking we just have one mark, something like `@pytest.mark.needs_optional_deps`
Then we'd add all the optional deps that get tested to the plugin-test group (or perhaps there is a more descriptive name that would be good... optional-deps-for-tests?)
From my perspective the groups that turn into 'extras' aren't really special here, and we don't have tests for most of them anyway.

vivodi: I misunderstood your meaning earlier, and now I’ve marked them with `@pytest.mark.require_optional_deps`.

.github/workflows/test.yml
```diff
@@ -30,11 +30,14 @@ jobs:
...
-        run: uv sync --group plugin-test --frozen
-      - name: Test with pytest
-        run: |
-          uv run pytest -n logical --dist loadgroup --junit-xml=test-results.xml
+      - name: Install mandatory dependencies
+        run: uv sync --frozen
+      - name: Run tests not requiring optional dependencies
+        run: uv run pytest -m 'not require_optional_deps' -n logical --dist loadgroup --junit-xml test-results.xml
+      - name: Install optional dependencies
+        run: uv sync --frozen --group all --group plugin-test
```
gazpachoking, line 38: Nice. This is getting pretty close to what I had in mind, the only other thing I was thinking is that we take out the `--group all` here. Then we'd just add telegram to the plugin-test group. (Unless there is some advantage I'm not seeing to installing all the extras.) We could add it as `{ include-group = "telegram" }` (like is done for the 'all' group) to avoid listing the version requirements separately again.

vivodi: I've made the changes according to your suggestion.

gazpachoking:
Looks good! Great idea, and thanks for sticking with the changes!

PR #4269, opened 2025-02-21
Adopt the `D` rules (manual fixes)

vivodi:
### Motivation for changes:
We should follow [PEP 257 – Docstring Conventions](https://peps.python.org/pep-0257/), so we should adopt the `D` rules.

D rules: https://docs.astral.sh/ruff/rules/#pydocstyle-d


vivodi:

PR #4301, opened 2025-03-12
Remove the broken dev container

vivodi:
### Motivation for changes:
The extant dev container configuration has become defunct. It relies on `pyenv`, `black` formatter, `flake8`, and `dev-requirements.txt`—a toolchain now incongruous with our actual development stack. Furthermore, since its inception, the dev container has languished without updates, casting serious doubt on its viability.  

Eschewing the dev container in favor of a streamlined setup appears unequivocally advantageous: simply installing `uv` and executing `uv sync` suffices to provision the development environment.  

In summation, retaining an obsolete and ostensibly unusable dev container serves no discernible purpose.


vivodi:
If the primary developers do not utilize the dev container and it remains neglected, I believe it should be removed. It would be even worse if new developers attempted to use it, only to be met with frustration upon discovering its dysfunction. Moreover, setting up the development environment without the dev container is sufficiently straightforward.

PR #4303, opened 2025-03-12
Exclude Test Files from Docker Image

vivodi:
### Motivation for changes:

This pull request addresses an inefficiency in the current project structure: test files are inadvertently included in the Docker image, leading to unnecessary image bloat and prolonged pull times.

```console
$ docker run --entrypoint sh flexget/flexget -c 'du -sh /flexget/flexget/tests'

24.4M   /flexget/flexget/tests
```

To resolve this issue, the `flexget/tests` directory has been relocated to a top-level `tests` directory. This change is underpinned by three key considerations:  

1. **Obviation of Accidental Inclusion**  
   By structurally decoupling test files from the package directory, this change **eliminates the possibility** of their accidental inclusion in both the Docker image and PyPI distributions—regardless of future modifications to the release pipeline. Unlike ad-hoc file exclusions, this method is inherently resilient and does not rely on continuous oversight.  

2. **Aligning with Industry Best Practices**  
   Test files are **not** an intrinsic part of the `flexget` package, and it is standard practice to place the `tests` directory at the root level rather than inside `src`. Established projects such as [Pytest](https://github.com/pytest-dev/pytest), [Pillow](https://github.com/python-pillow/Pillow), and [Flask](https://github.com/pallets/flask) adhere to this convention, reinforcing its efficacy in maintaining a clear separation between core functionality and test code.  

3. **Streamlined Directory Architecture**  
   Reducing directory nesting simplifies repository navigation and expedites modifications to test-related files. A flatter directory structure enhances clarity and reduces cognitive overhead, leading to a **more streamlined and developer-friendly** workflow.  

In summary, this change not only eliminates unintended Docker image expansion but also fortifies the project against accidental test file distribution, aligns with widely accepted structural paradigms, and enhances overall maintainability.


vivodi:

gazpachoking:
I think I'm good with this. @paranoidi Do you remember why tests got moved into the package in the first place? ee52cb6

In _theory_ the source distribution should have the tests included, but I don't think that's actually too important, and may be better to exclude them so we don't fill up our quota on pypi as fast. I struggled with that decision last time I fixed our releases.

paranoidi:
@gazpachoking  yes I do remember, I could not get the tests to run with extras when they were in the correct place for some reason. I don't think that is relevant that much as extras is not being used?

@vivodi Yeah, it's fine to move tests into correct place from my point of view. Preferable even.

gazpachoking:
Oh, yeah, I forgot about those extras. Given that the last commit there was 9 years ago, I think it's probably safe to assume nobody is using them.

That was my only concern, I'm happy with this change too.

PR #4307, opened 2025-03-14
Adopt the `T20` rules

vivodi:
### Motivation for changes:

To ensure consistent and professional logging practices, the use of `print` should be strictly avoided across the codebase. Instead, `loguru` must be utilized for all logging purposes, except in explicitly defined contexts such as standalone scripts and test cases, where direct output may be appropriate. This approach enhances maintainability, facilitates debugging, and ensures better control over log management.

`T20` rules: https://docs.astral.sh/ruff/rules/#flake8-print-t20

### DIFF
--- a/flexget/__init__.py
+++ b/flexget/__init__.py
@@ -2,6 +2,8 @@ import os
 import sys
 from typing import TYPE_CHECKING, Optional
 
+from loguru import logger
+
 # __version__ import need to be first in order to avoid circular import within logger
 from ._version import __version__  # noqa: F401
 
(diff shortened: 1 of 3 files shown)


vivodi:

gazpachoking:
I think the manager ones might be because an error can happen that prevents logging setup. Perhaps that doesn't apply anymore due to refactorings that have happened with logging since then, so this might be totally okay now. I'll try to take a closer peek and remember if there are any issues with using logging in these places.

PR #4309, opened 2025-03-16
Set `quote-style` to 'single'

vivodi:
This PR enforces a consistent `quote-style` to enhance code maintainability and readability.  

#### **Key Benefits:**  

1. **Avoids unnecessary escape characters**  
   - Example transformation:  
https://github.com/Flexget/Flexget/blob/2167aa8784afc4c759baf6d20252cf73a17e98df/flexget/components/notify/notifiers/rapidpush.py#L57
     becomes:  
     ```python
     'error_not': "Cannot use 'channel' with 'group', 'category', or 'priority'"
     ```  
   - This improves readability and reduces visual clutter caused by escape characters.  

2. **Simplifies batch find-and-replace operations**  
   - When performing large-scale refactoring (e.g., `pathlib` migration), maintaining a uniform quote style reduces the need to search for multiple variations (e.g., `entry['location']` vs. `entry["location"]`).  
   - This makes code search, linting, and automated refactoring more efficient.  

3. **Aligns with best practices in many codebases**  
   - Many style guides (e.g., PEP 8, Black) encourage consistency in string quoting.  
   - Enforcing a single quote style ensures that all contributors follow the same convention.  

4. **Reduces potential errors in copy-pasting or code reviews**  
   - Inconsistent quoting may lead to confusion when modifying strings, especially when dealing with nested quotes.  
   - A single, predictable quote style minimizes the likelihood of errors during manual edits or code reviews.  

5. **Improves diff quality in version control**  
   - Unnecessary quote changes in PRs can create noise in diffs, making real logic changes harder to spot.  
   - A consistent style prevents unnecessary churn in version control.  

By setting `quote-style`, we improve code readability, maintainability, and collaboration efficiency. Let me know if any adjustments are needed! 🚀


gazpachoking:
If we are going to standardize, I'm torn between whether single or double should be the standard. I think maybe I like double better? But that might be the more invasive change, since our old standard was single.

BrutuZ:
I'm partial to single quotes, and given how the quote conversion was disabled since Black and still on Ruff, I'd have assumed you were too 😅

vivodi:
I have created a variant of this PR #4312, setting the quote style to double. Only one needs to be merged.

vivodi:
> > I'm partial to single quotes, and given how the quote conversion was disabled since Black and still on Ruff, I'd have assumed you were too 😅
> 
> lol, I've kept using single quotes because that was our original guideline here. I sorta think double quotes look better, and it's nice that they can contain strings with apostrophes. Single quotes are easier to type though, but if we are having our pre-commit hook make things consistent for us that's a moot point.
> 
> I guess I'm inclined to stick with single just because that was our original guideline and it would be the less invasive change, though I'm not sure why I care about that when it's all automatic. 🤷

If you prefer double quotes, then we should switch to using them—even if it's an invasive change, it's not an issue since everything is handled automatically. Switching to double quotes won't introduce any bugs.

That's a great point about being able to contain apostrophes, and I agree we should make the switch. I also hadn’t considered that we don’t need to worry about single quotes being easier to type since the process is fully automated.  

Plus, double quotes are the default for both `Black` and `Ruff`, so it might indeed be the better choice overall.

paranoidi:
Yeah, I liked single quotes more because of visibility on ease of typing. No other reason really. I don't lose my sleep if switch is made.

paranoidi:
Additionally, for this reason FlexGet often uses backticks to encapsulate fields. Then you do not need to escape things in such extreme ways as in the RapidPush example. Backticks are also generally never used in titles etc making them distinguishable what is quoted by FlexGet.

gazpachoking:
After spending way too much time debating what I liked better I just went with you guys. 😄

PR #4318, opened 2025-03-21
from_piratebay: add custom query property 

corngood:
### Motivation for changes:

I want to query an individual user, which is possible with the pirate bay API (`https://apibay.org/q.php?q=user:[name]`), but there's no way to do it in the current plugin.

### Detailed changes:
- added a `query` string property which takes precedence over `list`/`category` when set

I decided to go with the general purpose query parameter since it's possible to use it for keyword searches and other functionality of the API, which is undocumented and probably unstable.

- I also added 'member' back to the list of ranks because my queries were returning results that used it, and the plugin was crashing

```
--- a/flexget/plugins/input/from_piratebay.py
+++ b/flexget/plugins/input/from_piratebay.py
@@ -118,6 +118,7 @@ class FromPirateBay:
                     {'type': 'integer'},
                 ]
             },
+            'query': {'type': ['string']},
             'list': {'type': 'string', 'enum': list(LISTS), 'default': 'top'},
             'rank': {'type': 'string', 'enum': list(RANKS), 'default': 'all'},
         },
```


flexget/plugins/input/from_piratebay.py
```diff
@@ -145,13 +150,14 @@ def on_task_input(self, task, config):
             else:
                 list_url = f'{url}/precompiled/data_top100_48h.json'
         elif category:
-            list_url = f'{url}/q.php?q=category:{category}'
+            list_url = f'{url}/q.php'
```
corngood, line 153: This could be done in terms of `query`, but I didn't want to change the precedence of anything here, so this seemed like the cleanest option for now.

corngood:
I'm still using this and it's useful to me.  I just rebased it.

vivodi:
I noticed that the coverage check for this PR failed, which caused a slight drop in the overall project coverage. Could you please take a look and see if you can improve the tests to maintain our coverage standard?

For guidance on how to write tests for FlexGet, detailed instructions are available at: https://flexget.readthedocs.io/en/latest/contributor/Topic%20guides/Running%20and%20writing%20tests.html.

corngood:
@vivodi sorry for the delay on this.  I must have missed your last message.

I've added tests, including for the new functionality.

tests/test_input_sites.py
```diff
@@ -19,7 +19,12 @@ class TestInputSites:
               genres: ['Action and Adventure']
           test_apple_trailers_simple:
             apple_trailers: 720p
-
+          test_from_piratebay:
+            from_piratebay:
+              category: HD - Movies
+          test_from_piratebay_query:
+            from_piratebay:
+              query: user:metheguy
```
corngood, line 40: I picked a user that has a bunch of popular linux ISOs

tests/cassettes/test_input_sites.TestInputSites.test_from_piratebay.yml
```diff
@@ -0,0 +1,142 @@
+interactions:
```
corngood: This is real recent data from TPB.  Is there a problem with putting this in the repo?  I don't see other cassette files that obviously hit torrent trackers.

We could replace it with synthetic data.

tests/test_input_sites.py
```diff
@@ -35,3 +53,12 @@ def test_apple_trailers(self, execute_task):
     def test_apple_trailers_simple(self, execute_task):
         task = execute_task('test_apple_trailers_simple')
         assert task.entries, 'no entries created / site may be down'
+
+    @pytest.mark.online
+    def test_from_piratebay(self, execute_task):
+        for task in ['all', 'cat', 'all_48h', 'cat_48h']:
```
corngood: These could be separate tests if you prefer

corngood: I decided this was being too clever and split it up

paranoidi:
Sorry for the slowness, merged. Thank you for your efforts!

PR #4326, opened 2025-03-29
Revamp the developer documentation website

vivodi:
### Motivation for changes:
The current website has display issues and outdated content—for example, it still mentions Poetry for package management, while we have switched to uv. This PR completely revamps the developer documentation website in both style and content.

Display issue:
image


vivodi:

paranoidi:
I don't think having plugin documentation in there is really necessary anymore. It could be removed and thus simplifying maintenance.

vivodi:
> I don't think having plugin documentation in there is really necessary anymore. It could be removed and thus simplifying maintenance.

Did you look at the website built from this PR? If not, you can take a look first. The website built by this PR doesn’t have dedicated plugin documentation, and the API Reference is fully auto-generated with no maintenance required.

https://flexget--4326.org.readthedocs.build/en/4326/

paranoidi:
@vivodi Apologies, I just skimmed the diff. Looking pretty sweet.

vivodi:
@gazpachoking The content seems fairly complete now. Would you be able to review it when you have some time?

gazpachoking:
Well, this for sure looks like an improvement. It's a lot to look at to even know if I fully agree with everything, but I'd say let's put it in and fix/simplify later if needed. There is certainly overlap with what's in here and on the wiki, and I don't know where is the right place for everything.

PR #4333, opened 2025-04-03
Change `entry['location']` to use the `pathlib.Path` type

vivodi:
### Motivation for changes:

```
tasks:
  task:
    filesystem:
      path:
        - "/home/k/test"
      mask: '*.torrent'
      retrieve: files
    accept_all: yes
    exec:
      auto_escape: yes
      on_output:
        for_accepted:
          - rm -f -- "{{ location }}"
          - echo "location {{location}}"
          - echo "pathbase {{location|pathbase}}"
          - echo 'pathdir {{location | pathdir}}'
          - echo "pathbase {{location|pathdir|pathbase}}"
    set:
      label: "tv-{{location | pathdir | pathbase }}"
```
In the above configuration, the bug occurred because `entry['location']` was of type `str`, while `filter_pathdir()` expected a `Path` type parameter. To resolve this, `entry['location']` is now stored as a `Path` type.

### DIFF
--- a/flexget/entry.py
+++ b/flexget/entry.py
@@ -3,6 +3,7 @@ import types
 import warnings
 from datetime import date, datetime
 from enum import Enum
+from pathlib import Path
 from typing import TYPE_CHECKING, Callable, Optional, Union
 
 import pendulum
@@ -252,6 +253,10 @@ class Entry(LazyDict, Serializer):
             if not isinstance(value, (str, LazyLookup)):
                 raise plugin.PluginError(f'Tried to set title to {value!r}')
             self.setdefault('original_title', value)
+
+        # location handling
+        if key == 'location' and isinstance(value, str) and value != '':
+            value = Path(value)
 
         try:
             logger.trace('ENTRY SET: {} = {!r}', key, value)
--- a/tests/test_jinja_filters.py
+++ b/tests/test_jinja_filters.py
@@ -1,3 +1,5 @@
+from pathlib import Path
+
 import jinja2
 import pytest
 
@@ -32,6 +34,16 @@ class TestJinjaFilters:
             set:
               size: "{{title|parse_size}}"
               size_si: "{{title|parse_size(si=True)}}"
+          path:
+            mock:
+              - {"title":"path", "location": "/a/b/c.d"}
+            accept_all: yes
+            set:
+              pathdir: "{{location|pathdir}}"
+              pathbase: "{{location|pathbase}}"
+              pathbase2: "{{location|pathdir|pathbase}}"
+              pathname: "{{location|pathname}}"
+              pathext: "{{location|pathext}}"
     """
 
     custom_filters = [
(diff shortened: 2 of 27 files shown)


tests/test_jinja_filters.py
```diff
@@ -99,6 +111,14 @@ def test_parse_size(self, execute_task):
         assert task.accepted[4]['size'] == int(task.accepted[4]['actual'] * 1024**3)
         assert task.accepted[4]['size_si'] == int(task.accepted[4]['actual'] * 1000**3)
 
+    def test_path(self, execute_task):
+        task = execute_task('path')
+        assert Path(task.accepted[0]['pathdir']) == Path('/a/b')
+        assert task.accepted[0]['pathbase'] == 'c.d'
+        assert task.accepted[0]['pathbase2'] == 'b'
+        assert task.accepted[0]['pathname'] == 'c'
+        assert task.accepted[0]['pathext'] == '.d'
```
vivodi, line 120: I added a test for this bug to prevent future regressions.

gazpachoking:
Hmm. I don't think we should assume the types of any entry fields, since they could come from places we don't control (e.g. set plugin, mock plugin, user created plugin.) But... since the location field has become sort of a standard, we _could_ enforce it to always be a Path object in the Entry class itself and then be able to count on its type. I'm not sure if I like that either though. We have avoided doing anything like that except for with title and url so far. https://github.com/Flexget/Flexget/blob/0d01ca5e2e6933537c346cc85fae14d80fcfcbf9/flexget/entry.py#L251

flexget/entry.py
```diff
@@ -253,6 +254,10 @@ def __setitem__(self, key, value):
                 raise plugin.PluginError(f'Tried to set title to {value!r}')
             self.setdefault('original_title', value)
 
+        # location handling
+        if key == 'location' and isinstance(value, str) and value != '':
+            value = Path(value)
+
```
vivodi, line 260: Now in `Entry.__setitem__`, I convert the `location` field into a `pathlib.Path` object. This ensures that even if the value comes from places we don't control (e.g. third-party plugins, mock plugins, or user-created plugins), it won't cause issues. Storing a `pathlib.Path` object in `entry['location']` also simplifies the code logic and is an important step toward going "all in `pathlib`".

Allowing `entry['location']` to store either a `str` or a `pathlib.Path` introduces ambiguity and complexity. Every time the field is accessed, we have to check its type and potentially convert it. This increases the chance of bugs, makes the code harder to reason about, and leads to inconsistent behavior depending on the input. By standardizing on `pathlib.Path`, we reduce cognitive overhead and make the interface more predictable.

gazpachoking:
Oh! I didn't notice when we merged the other pr that the jinja filters only accept Paths now. Those should for sure be changed to accept strings or paths. Even if we enforce the location field as a Path, there are other strings that users may want to operate on with the jinja path filters.

PR #4342, opened 2025-04-11
Fix release pipeline

vivodi:
### Motivation for changes:
Errors occurred in the deployment workflow after updating `uv`. Updated related code to be compatible with `uv` changes.

### DIFF
--- a/scripts/build_locked_extras.py
+++ b/scripts/build_locked_extras.py
@@ -44,7 +44,7 @@ def update_metadata_with_locked(
         except (subprocess.TimeoutExpired, subprocess.CalledProcessError) as exc:
             print(f'Failed to export locked dependencies for `{group}`: {exc.stderr}')
             sys.exit(1)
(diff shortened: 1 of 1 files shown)


scripts/build_locked_extras.py
```diff
@@ -44,7 +45,7 @@ def update_metadata_with_locked(
         except (subprocess.TimeoutExpired, subprocess.CalledProcessError) as exc:
             print(f'Failed to export locked dependencies for `{group}`: {exc.stderr}')
             sys.exit(1)
-        requirements = [line for line in export.splitlines() if not line.startswith('#')]
+        requirements = [line for line in export.splitlines() if re.search('^(?! *#)', line)]
```
gazpachoking, line 48: Is this just `not line.lstrip().startswith('#')`? Is that easier to read?

gazpachoking: Thanks!

Issue #4329, opened 2025-03-31
All filters regarding PATH stop working pathdir pathbase after v3.15.27


Anian-igor:
### Expected behaviour:

Previously worked template {{location | pathdir }} or {{location | pathdir | pathbase }} stopped worked with error 
`Error during rendering: (AttributeError) 'str' object has no attribute 'parent'`

### Actual behaviour:

cannor parse location

### Steps to reproduce:
```
exec:
  - echo "location {{location}}"
  - echo "pathbase{{location|pathbase}}"
```
worked on version 3.15.26, not worked after 3.15.27

#### Config:
```dropbox_series_settings:
    filesystem:
      path:
        - "/tv/share/flexget/nextcloud/tvseries/"
      mask: '*.torrent'
      retrieve: files
    accept_all: yes
    exec:
      auto_escape: yes
      on_output:
        for_accepted: 
          - rm -f -- "{{ location }}"
          - echo "series_name {{series_name}}"
          - echo "location {{location}}"
          - echo "pathbase{{location|pathbase}}"
          - echo 'pathdir {{location | pathdir}}'
          - echo "pathbase{{location|pathdir|pathbase}}"
    set:
      label: "tv-{{location | pathdir | pathbase }}"
```
  
#### Log:
<details>
 <summary>(click to expand)</summary>

```
tv-DROPBOX-season-one phase_name: on_output operation: for_accepted cmd: rm -f -- "/tv/share/flexget/nextcloud/tvseries/Roadies s01.torrent"
2025-03-31 14:17:16.877902+00:002025-03-31 17:17:16 VERBOSE  exec          tv-DROPBOX-season-one Executing: rm -f -- "/tv/share/flexget/nextcloud/tvseries/Roadies s01.torrent"
2025-03-31 14:17:16.881561+00:002025-03-31 17:17:16 DEBUG    exec          tv-DROPBOX-season-one phase_name: on_output operation: for_accepted cmd: echo "series_name Roadies"
2025-03-31 14:17:16.881603+00:002025-03-31 17:17:16 VERBOSE  exec          tv-DROPBOX-season-one Executing: echo "series_name Roadies"
2025-03-31 14:17:16.882515+00:002025-03-31 17:17:16 INFO     exec          tv-DROPBOX-season-one Stdout: series_name Roadies
[...]

PR #4356, opened 2025-04-17
Change the jinja filters to accept both `PathLike` and `str`

vivodi:
This PR only includes the fix for #4329. The change of `entry['location']` to use the `Path` type will continue in #4333.

### DIFF
--- a/flexget/utils/template.py
+++ b/flexget/utils/template.py
@@ -3,6 +3,7 @@ import re
 from contextlib import suppress
 from copy import copy
 from datetime import date, datetime, time
+from pathlib import Path, PurePosixPath, PureWindowsPath
 from typing import TYPE_CHECKING, Any, AnyStr, Optional, Union, cast
 from unicodedata import normalize
 
@@ -30,7 +31,6 @@ from flexget.utils.tools import format_filesize, parse_filesize, split_title_yea
 
 if TYPE_CHECKING:
     from collections.abc import Mapping
-    from pathlib import Path
 
     from flexget.entry import Entry
     from flexget.manager import Manager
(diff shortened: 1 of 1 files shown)


gazpachoking:
Why `PathLike`? Just curious, not implying that it's the wrong thing.

Is there some good way we can make those first 4 filters work correctly on both windows and posix paths no matter which OS flexget is running on? Maybe by using `PureWindowsPath` rather than `Path` (when the input is a string). That would technically do the wrong thing with a posix path that contained backslashes, but is that common enough to worry about?

vivodi:
> Why `PathLike`? Just curious, not implying that it's the wrong thing.
> 
> Is there some good way we can make those first 4 filters work correctly on both windows and posix paths no matter which OS flexget is running on? Maybe by using `PureWindowsPath` rather than `Path` (when the input is a string). That would technically do the wrong thing with a posix path that contained backslashes, but is that common enough to worry about?

Good question! The use of `PathLike` is mainly for completeness—it ensures the function can accept any object implementing the `__fspath__` protocol, not just strings. While such cases might be rare (like `os.DirEntry`), it's a low-cost way to make the function more flexible.

About the Windows vs POSIX path handling—I'm not sure I fully understand the use case you're referring to. Why would FlexGet, when running on Linux, need to interpret a Windows-style path (or vice versa)? Since FlexGet typically operates on the local filesystem, wouldn't it make the most sense to handle paths according to the current OS? 

Introducing logic to parse paths from a different OS could actually lead to more subtle issues—for example, treating a POSIX path with a backslash as a Windows path might misinterpret the input rather than help. Unless there's a clear use case where FlexGet regularly processes paths from other systems, it might be safer to stick with the default `Path` behavior, which aligns with the host OS. I'd love to hear more if you've encountered such cross-platform scenarios in practice.

gazpachoking:
> About the Windows vs POSIX path handling—I'm not sure I fully understand the use case you're referring to. Why would FlexGet, when running on Linux, need to interpret a Windows-style path (or vice versa)? Since FlexGet typically operates on the local filesystem, wouldn't it make the most sense to handle paths according to the current OS?

You very well could be running against entries created from another machine, say a remote torrent client, or ftp server. The 'location' field will always be a reference to a local file (because we've defined it that way) but there are other fields that could contain paths on a remote filesystem.

vivodi:
The current code should be able to handle cross-platform paths now.

For simplicity, `PathLike` was changed to `Path`.

gazpachoking:
I think I may have steered us wrong on 'pathdir' I haven't tested yet, but the problem I foresee would be that when it's a PureWindowsPath it'll use windows separators (backslashes) when outputting the path. The rest of them don't deal with multiple paths of directories, so I don't think it's a problem there.

flexget/utils/template.py
```diff
@@ -135,23 +135,31 @@ def diff(self, dt: Optional[datetime] = None, abs: bool = True) -> Interval:
...
+        val = PureWindowsPath(val)
     return val.suffix
 
 
-def filter_pathdir(val: 'Path') -> 'Path':
+def filter_pathdir(val: Union[Path, str] = '') -> Union[PureWindowsPath, PurePosixPath, Path]:
     """Directory containing the given path."""
+    if isinstance(val, str):
+        val = PureWindowsPath(val) if '\\' in val and val[1] == ':' else PurePosixPath(val)
     return val.parent
```
vivodi, line 163: Would it work if we changed `filter_pathdir` to this?

PR #4360, opened 2025-04-20
Full Optional Dependency Support via Extras and Docker Integration

vivodi:
### Motivation for changes:

Currently, we only provide a limited set of extras: `deluge`, `qbittorrent`, `telegram`, and `transmission`. This PR extends support by exposing **all** optional dependencies as extras and bundling them directly into the Docker image. This approach offers several advantages:

1. **Improved dependency resolution** — All optional dependencies are included in UV's resolution process, ensuring full compatibility across the entire dependency tree.

2. **Simplified installation** — Users can easily install the required dependencies using `flexget[extra_1, extra_2]`, without needing to manage them individually.

3. **Streamlined Docker usage** — Bundling all dependencies into the Docker image simplifies usage. As first-party dependencies, there's no strong reason to exclude them, especially considering the minimal impact on image size. The current uncompressed image is 305MB, and with all dependencies included, it increases to just 313MB.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -37,7 +37,7 @@ jobs:
         run: uv run pytest -m 'not require_optional_deps' -n logical --dist loadgroup --junit-xml test-results.xml
       - name: Run tests requiring optional dependencies
         if: ${{!cancelled() && steps.core-tests.outcome != 'skipped'}}
-        run: uv run --group plugin-test pytest -m 'require_optional_deps' -n logical --dist loadgroup --junit-xml test-results.xml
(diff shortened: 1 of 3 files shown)


paranoidi:
Seems reasonable, please update the documentation at https://flexget.com/InstallWizard/Docker to match this new situation after the merge.

paranoidi:
Conflicts need resolving before merge.

vivodi:
I’ve already rebased it. @paranoidi

paranoidi:
Great, I'll merge it now.

PR #4368, opened 2025-04-21
Bump the version number to avoid a failure during the next release

vivodi:
Previously, there was an error during deployment, but a new release has already been published on PyPI. Therefore, we need to manually bump the version number to avoid a failure during the next release.


vivodi:
If this PR is not merged before the next release, the next release will fail. @gazpachoking

PR #4370, opened 2025-04-22
The "created" field is renamed to "ctime".

### LINKED ISSUE #4150: Filesystem plugin set created time from unrelated timestamp
Filesystem set "created" timestamp based on ctime which is not the creation time.

https://github.com/Flexget/Flexget/blob/4130bda694fe71c39dd9c7855e02ca6230ddf5a5/flexget/plugins/input/filesystem.py#L131

"created" timestamp should probably be discarded as there is no such timestamp on FS.

vivodi:
### Motivation for changes:
According to https://docs.python.org/3/library/stat.html#stat.ST_CTIME, `stat.ST_CTIME` is the “ctime” as reported by the operating system. On some systems (like Unix) is **the time of the last metadata change**, and, on others (like Windows), is the creation time (see platform documentation for details).

Since `ctime` is not always the creation time, renaming the "created" field to "ctime" can help avoid misunderstanding.

Closes #4150

### DIFF
--- a/flexget/_version.py
+++ b/flexget/_version.py
@@ -10,4 +10,4 @@ and update the version again for continued development.
 NOTE: Should always have all three parts of the version, even on major and minor bumps. i.e. 4.0.0.dev, not 4.0.dev
 """
 
-__version__ = '3.15.43.dev'
+__version__ = '3.16.0.dev'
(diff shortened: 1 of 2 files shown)


gazpachoking:
Should we detect the OS in use and try to do a 'friendly' (but accurate) name on *nixs? This would let us leave it as 'created' for windows, but something else on alternate systems. (not sure what a good name would be though, 'metadata_modified' seems a bit verbose) Or we could just rename all 3 to ctime, mtime and atime. Either way, we should do a version bump and add an upgrade action, since this could break configs.

vivodi:
I renamed all 3 to ctime, mtime and atime.

PR #4398, opened 2025-05-18
Return true even if a path doesn't exist

ianstalk:
This makes it clear that even a non-existent path will pass validation.

### DIFF
--- a/flexget/config_schema.py
+++ b/flexget/config_schema.py
@@ -274,11 +274,10 @@ def is_path(instance) -> bool:
     result = pat.search(instance)
     if result:
         instance = os.path.dirname(instance[0 : result.start()])
-    if os.path.isdir(os.path.expanduser(instance)):
-        return True
-    error: str = f'`{instance}` does not exist'
-    logger.warning(error)
-    return False
+    if not os.path.isdir(os.path.expanduser(instance)):
+        error: str = f'`{instance}` does not exist'
+        logger.warning(error)
+    return True
 
 
 # TODO: jsonschema has a format checker for uri if rfc3987 is installed, perhaps we should use that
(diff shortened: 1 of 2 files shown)


flexget/config_schema.py
```diff
@@ -274,11 +274,10 @@ def is_path(instance) -> bool:
...
     if result:
         instance = os.path.dirname(instance[0 : result.start()])
-    if os.path.isdir(os.path.expanduser(instance)):
-        return True
-    error: str = f'`{instance}` does not exist'
-    logger.warning(error)
-    return False
+    if not os.path.isdir(os.path.expanduser(instance)):
+        error: str = f'`{instance}` does not exist'
+        logger.warning(error)
```
vivodi, line 279: Hi there! Thanks for the contribution.

I have a couple of suggestions for this logging statement to improve its clarity, performance, and adherence to best practices.

### 1. Semantic Mismatch between Variable Name and Log Level

The variable is named `error`, but it's being logged as a `warning`. This can be a bit confusing for future readers, as the name `error` typically implies a more severe issue that might halt execution (corresponding to `logger.error` or an exception).

### 2. Use Deferred Formatting for Better Performance

The current code uses an f-string to create the log message. For logging, it's a best practice to use the logger's own substitution mechanism instead.

As the **[ruff documentation on `G004`](https://docs.astral.sh/ruff/rules/logging-f-string/)** explains, the logging module uses "lazy" or "deferred" formatting. This means it only constructs the full message string *if* the message's level (e.g., `WARNING`) is high enough to actually be processed by a handler.

When you use an f-string, the string is formatted *immediately*, even if the log level is configured to ignore warnings. This can waste CPU cycles, especially if the log statement is in a frequently called function.

### Suggested Improvement

Here’s how we could combine these points into a more robust and performant log statement:

```diff
- error: str = f'`{instance}` does not exist'
- logger.warning(error)
+ # By passing the instance as an argument, we let the logger handle string formatting.
+ # This avoids unnecessary work if the 'WARNING' level is disabled and also
+ # resolves the semantically confusing 'error' variable for a warning.
+ logger.warning("`{}` does not exist", instance)

```

This approach is more performant and directly clarifies the code's intent. Let me know what you think

vivodi: Fixed in #4458.

PR #4413, opened 2025-05-29
Pin `subliminal` to `<2.3` due to https://github.com/Diaoul/subliminal/issues/1302

vivodi:
### Motivation for changes:

The current lock file maintenance PR cannot be merged due to test failures caused by the subliminal update. After investigating, I found that the latest version of subliminal appears to have a bug ([https://github.com/Diaoul/subliminal/issues/1302](https://github.com/Diaoul/subliminal/issues/1302)).

For now, the subliminal version is pinned to avoid completely blocking lock file maintenance.


vivodi:
I submitted this PR as an example of the issue at https://github.com/codecov/codecov-action/issues/1831, so please hold off on merging it for now.

PR #4430, opened 2025-06-09
feat: Add support for Python 3.14

vivodi:
### Summary

This Pull Request introduces support for the upcoming Python 3.14 release. The goal is to ensure our project is compatible and can be tested against the new version proactively. This helps us identify and fix any potential compatibility issues before Python 3.14 becomes stable.

### Changes Made

- **CI Configuration (`.github/workflows/test.yml`)**:
  - Added `3.14` to the `python-version` matrix in our CI pipeline. This will run our full test suite against Python 3.14 on every PR and push.

- **Project Metadata (`pyproject.toml`)**:
  - Updated the `classifiers` to include `"Programming Language :: Python :: 3.14"`. This informs package indexes like PyPI that our project officially supports this version.


flexget/api/app.py
```diff
@@ -83,11 +83,11 @@ def __getattr__(self, item):
 
     __getitem__ = __getattr__
 
-    def __call__(self, data=None, method: Optional[str] = None):
+    def __call__(self, data=None, method: str | None = None):
         return self.caller(self.endpoint, data=data, method=method)
 
 
-def api_version(f: Callable[..., Response]):
+def api_version(f: 'Callable[..., Response]'):
```
paranoidi: Why are these switched to strings?

vivodi: Now that support for Python 3.9 has been dropped, we can transition to using `collections.abc.Callable` instead of its deprecated alias, `typing.Callable`.

In making this change, Ruff requires us to adopt a more optimized import strategy to minimize runtime overhead.

One common method is to place the import under a `TYPE_CHECKING` block. **However, this approach necessitates that forward references be enclosed in quotes (e.g., `'MyClass'`).**

An alternative, more modern approach is to add `from __future__ import annotations` to the top of each file. This allows for unquoted forward references and aligns with the language's future direction, as this behavior is slated to become the default in Python 3.14 and later.

If you find the quoted type hints visually distracting or cumbersome, I'm happy to adopt the second approach instead.

gazpachoking: Yeah, I prefer `from __future__ import annotations` over quoting the annotations. Ruff should be able to handle all that for us.

vivodi: Done.

pyproject.toml
```diff
@@ -216,31 +213,23 @@ ignore = [
...
-    "TRY400", # error-instead-of-exception, rejected by @gazpachoking in https://github.com/Flexget/Flexget/pull/4249#issuecomment-2656387500
+    "TRY400", # Rejected by @gazpachoking in https://github.com/Flexget/Flexget/pull/4249#issuecomment-2656387500
 ]
 
 [tool.ruff.lint.flake8-type-checking]
 quote-annotations = true
 
 [tool.ruff.lint.isort]
 known-first-party = ['flexget']
+required-imports = ["from __future__ import annotations"]
```
gazpachoking: Do we need to add this to every file right now? Maybe we can just mark the "FA" ruff fixes as safe? That should add it to any files where we are using annotations where it makes sense to add. (Not sure if there are other pertinent ones as well)

vivodi: The FA rules have already been enabled, but they are not helpful in resolving this particular issue.

At present, the only solution is to use the `required_imports` option to add `from __future__ import annotations` to the beginning of every file. The ability to add this import only where it is actually needed is an open issue in Ruff ([Add autofix that adds `from __future__ import annotations` only when it would be useful](https://github.com/astral-sh/ruff/issues/18502)) and is not something Ruff can currently handle.

If you don't want to add `from __future__ import annotations` to every file, the only alternative is to use string-based forward references (e.g., using quotes like `'ClassName'`).

vivodi:
I think the best approach is to wait until https://github.com/astral-sh/ruff/issues/18502 is resolved before adding `from __future__ import annotations`.

I've now removed the related changes from this PR.

gazpachoking:
That ruff ticket is exactly the behavior I was hoping they already had. Sounds good to me.

gazpachoking:
Looks like that new ruff behavior is in preview, want to test it and see how it turns out?

vivodi:
> Looks like that new ruff behavior is in preview, want to test it and see how it turns out?

@gazpachoking I have made the changes as you requested.

Would it be possible to merge my pull requests, especially this particular one? It would save a lot of effort from having to rebase it later.

gazpachoking:
Seems good to me! We are a bit before eol of python 3.9, but I don't think it matters. Thanks!

PR #4432, opened 2025-06-10
chore: Add `typos` to pre-commit hooks

vivodi:
Integrates the `typos` spell checker into our pre-commit configuration (`.pre-commit-config.yaml`).

This will automatically scan staged files for spelling errors before a commit is created, providing immediate feedback to developers. This helps maintain high-quality documentation and prevents typos from entering the repository in the first place.

Developers should run `pre-commit install` to enable the hook locally.

### DIFF
--- a/flexget/plugins/input/npo_watchlist.py
+++ b/flexget/plugins/input/npo_watchlist.py
@@ -45,8 +45,8 @@ class NPOWatchlist:
           auto_escape: yes
           on_output:
             for_accepted:
-              - download-npo -o "path/to/directory/{{series_name_plain}}" -f "{serie_titel} - {datum} \
-                  {aflevering_titel} ({episode_id})" -t {{url}}
+              - download-npo -o "path/to/directory/{{series_name_plain}}" -f "{series_title} - {datum} \
+                  {aflevering_title} ({episode_id})" -t {{url}}
     """
 
     schema = {
@@ -121,7 +121,7 @@ class NPOWatchlist:
 
             if 'isAuthenticatedUser' not in requests.cookies:
                 raise plugin.PluginError('Failed to login. Check username and password.')
-            logger.debug('Succesfully logged in: {}', email)
+            logger.debug('Successfully logged in: {}', email)
         except RequestException as e:
             raise plugin.PluginError(f'Request error: {e!s}')
 
--- a/flexget/plugins/output/subtitles_subliminal.py
+++ b/flexget/plugins/output/subtitles_subliminal.py
@@ -17,13 +17,14 @@ try:
     PROVIDERS = provider_manager.names()
 except ImportError:
     PROVIDERS = [
-        'argenteam',
-        'legendastv',
+        'addic7ed',
+        'gestdown',
+        'napiprojekt',
         'opensubtitles',
+        'opensubtitlescom',
+        'opensubtitlescomvip',
         'opensubtitlesvip',
         'podnapisi',
-        'shooter',
-        'thesubdb',
         'tvsubtitles',
     ]
 
@@ -48,14 +49,14 @@ class PluginSubliminal:
           alternatives:
             - eng
           exact_match: no
-          providers: legendastv, opensubtitles
+          providers: gestdown, opensubtitles
           single: no
           directory: /disk/subtitles
           hearing_impaired: yes
           authentication:
-            legendastv:
+            opensubtitles:
               username: myuser
-              passsword: mypassword
+              password: mypassword
     """
 
     schema = {
--- a/flexget/ui/v2/README.rst
+++ b/flexget/ui/v2/README.rst
@@ -2,6 +2,6 @@ FlexGet Web UI (v2)
 ===================
 
 The Flexget WebUI is currently being rewritten in ReactJS.
-The WebUI has been seperated into it's own repository.
+The WebUI has been separated into its own repository.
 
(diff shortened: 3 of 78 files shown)


flexget/components/sites/sites/descargas2020.py
```diff
@@ -66,7 +66,7 @@ def parse_download_page(self, url, task):
         url_format = DESCARGAS2020_TORRENT_FORMAT
 
         torrent_id_prog = re.compile(
-            r"(?:parametros\s*=\s*\n?)\s*{\s*\n(?:\s*'\w+'\s*:.*\n)+\s*'(?:torrentID|id)'\s*:\s*'(\d+)'"
+            r"(?:parameters\s*=\s*\n?)\s*{\s*\n(?:\s*'\w+'\s*:.*\n)+\s*'(?:torrentID|id)'\s*:\s*'(\d+)'"
```
BrutuZ, line 69: ⚠ Non-english site, makes sense the regular expression is in Spanish

flexget/plugins/input/npo_watchlist.py
```diff
@@ -45,8 +45,8 @@ class NPOWatchlist:
           auto_escape: yes
           on_output:
             for_accepted:
-              - download-npo -o "path/to/directory/{{series_name_plain}}" -f "{serie_titel} - {datum} \
-                  {aflevering_titel} ({episode_id})" -t {{url}}
+              - download-npo -o "path/to/directory/{{series_name_plain}}" -f "{series_title} - {datum} \
+                  {aflevering_title} ({episode_id})" -t {{url}}
```
BrutuZ, line 49: ⚠ Another non-english site

flexget/plugins/output/subtitles_subliminal.py
```diff
@@ -55,7 +55,7 @@ class PluginSubliminal:
           authentication:
             legendastv:
               username: myuser
-              passsword: mypassword
+              password: mypassword
```
BrutuZ, line 58: This can actually be removed altogether from the docs since legendas.tv no longer exists

flexget/ui/v2/README.rst
```diff
@@ -2,6 +2,6 @@ FlexGet Web UI (v2)
 ===================
 
 The Flexget WebUI is currently being rewritten in ReactJS.
-The WebUI has been seperated into it's own repository.
+The WebUI has been separated into it's own repository.
```
BrutuZ, line 5: While we're at it: `its`

BrutuZ (commented):
A couple of those might break plugins

flexget/components/sites/sites/descargas2020.py
vivodi: Thanks for pointing it out — I accidentally applied this change.

flexget/plugins/input/npo_watchlist.py
vivodi: If you take a look, you'll see that this is in the commented section. It is indeed a spelling mistake and should be corrected.

flexget/plugins/output/subtitles_subliminal.py
vivodi: Removed `legendastv` and updated the example to use `opensubtitles`.

gazpachoking:
Hmm. Not sure about this one. A lot of those typo fixes are good, but it also messes with identifiers which may change functionality. How does it notify when there are typos, does it fail the commit?

vivodi:
Under the current settings, it automatically corrects typos, but you can configure it to **report its findings only instead**. I noticed that in both the `uv` and `ruff` repositories, `typos` is configured to auto-correct by default.

https://github.com/astral-sh/uv/blob/f5305653238378376cd33f1a00d6b9eb511bef36/.pre-commit-config.yaml#L14-L17
https://github.com/astral-sh/ruff/blob/a863000cbc6a40d39553239ec33142245797bd89/.pre-commit-config.yaml#L69-L72

However, if you'd prefer to have `typos` only report issues without auto-correcting them, I'm happy to make that change.

vivodi:
The `typos` pre-commit hook is currently configured to automatically correct typographical errors, a practice consistent with the setups in both the `uv` and `ruff` repositories. This auto-correction is conditional: it only applies a fix when there is a single, unambiguous suggestion. In cases of ambiguity with multiple potential fixes, it will simply report the findings and fail the commit.

https://github.com/astral-sh/uv/blob/f5305653238378376cd33f1a00d6b9eb511bef36/.pre-commit-config.yaml#L14-L17
https://github.com/astral-sh/ruff/blob/a863000cbc6a40d39553239ec33142245797bd89/.pre-commit-config.yaml#L69-L72

It's important to note that even when auto-fixing is enabled, the commit will still fail, and you will need to commit the changes again. This is the intended behavior of pre-commit and is consistent with other hooks that have auto-fixing capabilities.

However, if you would prefer to disable automatic modifications, I can reconfigure the hook to a "report-only" mode. Please let me know your preference.

vivodi:
> but it also messes with identifiers which may change functionality.

The modifications for identifiers containing typos pertain exclusively to Emby. Therefore, it is only necessary to update the corresponding information on the https://flexget.com/en/Plugins/emby_lookup page. As these identifiers are arbitrary, they can be modified at will. @gazpachoking

paranoidi:
@vivodi Sorry it was left hanging around for a while. Great improvements all around :+1:

PR #4439, opened 2025-06-18
fix: Correctly append file extension instead of replacing it

vivodi:
The previous code used `pathlib.Path.with_suffix()` to add an extension. This method replaces the existing suffix, leading to incorrect behavior when the original filename already contained an extension.

For example, 'file.log' with extension '.bak' would become 'file.bak' instead of the intended 'file.log.bak'.

This commit changes the implementation to use an f-string for concatenation, ensuring the new extension is always appended to the full original filename.


gazpachoking:
Awesome, thanks!

PR #4442, opened 2025-06-20
chore(ci): Make patch coverage check non-blocking

vivodi:
The strict patch coverage check requires that all new and modified lines in a PR are accompanied by tests. Given the project's currently low overall test coverage, this often forces developers to write a substantial number of new tests even for minor changes, which creates a high barrier for contributions.

To unblock the development workflow, this change sets the patch coverage check to "informational". This means the check will now always pass, but it will continue to report on the coverage of each commit and pull request, providing valuable feedback to developers without halting the release process.

image

Crucially, the project-level coverage check remains active and blocking. Any PR that causes a decrease in the total project coverage will still fail the build.

image

### DIFF
--- a/tests/sftp/test_sftp_server.py
+++ b/tests/sftp/test_sftp_server.py
@@ -75,7 +75,7 @@ else:
 
         def __run_server(self):
             logger.debug('Starting sftp server 127.0.0.1:40022 with root fs at %s', self.__root)
-            try:
+            with contextlib.suppress(ConnectionAbortedError):
                 conn, addr = self.__server_socket.accept()
 
                 host_key = RSAKey.from_private_key_file('test_sftp_server.key')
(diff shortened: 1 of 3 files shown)


tests/sftp/test_sftp_server.py
```diff
@@ -93,10 +93,6 @@ def __run_server(self):
 
                 while transport.is_active():
                     time.sleep(1)
-            except ConnectionAbortedError:
-                pass
-            except Exception as e:
-                logger.critical(e, exc_info=True)
```
vivodi, line 99: This change is because I discovered that the `except Exception as e:` block was being executed occasionally. This caused the project coverage to fluctuate, which in turn led to failures in our CI coverage checks.

Considering this is only test code and does not affect users, we shouldn't blindly catch such a broad exception. By not catching this exception, future CI failures will reveal the actual traceback, allowing us to understand why this line of code is being executed intermittently.

PR #4443, opened 2025-06-21
ci: Skip redundant push builds after `merge_group` checks

vivodi:
When a pull request is merged via the merge queue, it triggers two separate CI workflow runs for the same commit:
1. One for the `merge_group` event.
2. A second, redundant one for the `push` event to the base branch.

This duplication consumes unnecessary CI minutes and slows down the feedback loop.

This commit adds a conditional `if` statement to our CI jobs. The jobs will now be skipped if the event is a `push` and the actor is `github-merge-queue[bot]`, which is the user responsible for merge queue merges.

This ensures that only the `merge_group` check is executed for commits entering the develop branch via the merge queue, thereby improving CI efficiency and saving resources.

### DIFF
--- a/.github/workflows/docker-image.yml
+++ b/.github/workflows/docker-image.yml
@@ -15,6 +15,7 @@ concurrency:
 jobs:
   build:
     name: Docker build
+    if: github.actor != 'github-merge-queue[bot]'
     permissions:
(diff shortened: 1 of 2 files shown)


vivodi:
Redundant checks:

image

PR #4454, opened 2025-06-23
refactor(deps): Remove ruff from pyproject.toml dev dependencies

vivodi:
The pre-commit configuration already includes and manages ruff for linting and formatting. Including it as a direct development dependency in `pyproject.toml` is unnecessary and can lead to version conflicts.

This commit removes ruff to rely solely on the version managed by pre-commit.

Benefits of this change:
- **Single Source of Truth for Consistency:** Establishes `.pre-commit-config.yaml` as the definitive source for the ruff version. This guarantees that every developer and the CI system uses the exact same linter, ensuring consistent and predictable checks across all environments.
- **Simplified Dependency Management:** Reduces the number of packages to manage in the main development environment.
- **Faster Environment Setup:** Slightly speeds up the creation of the development virtual environment (`uv sync`) as one less package needs to be resolved and installed.

### DIFF
--- a/.github/renovate.json5
+++ b/.github/renovate.json5
@@ -55,12 +55,6 @@
       extends: ['schedule:monthly'],
     },
     {
-      description: 'Group all ruff updates together. Wait before merge so both sources are out.',
-      groupName: 'ruff',
-      matchPackageNames: ['ruff', 'astral-sh/ruff-pre-commit'],
-      minimumReleaseAge: '6 hours',
-    },
-    {
       description: 'Group all uv updates together. Wait before merge so both sources are out.',
       groupName: 'uv',
       matchPackageNames: [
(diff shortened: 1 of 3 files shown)


BrutuZ:
This makes it so VS Code have to rely on its often outdated built-in version of Ruff to format on save/edits, currently incompatible with `future-annotations`

vivodi:
> This makes it so VS Code have to rely on its often outdated built-in version of Ruff to format on save/edits, currently incompatible with `future-annotations`

You can resolve the issue by installing Ruff globally: run `pip install -U ruff` outside of a venv or `uv tool install ruff@latest`.

For other global installation methods, see https://docs.astral.sh/ruff/installation

Another reason for not including Ruff in `dev-dependencies` is that if someone works on multiple projects that require Ruff, they may need to install Ruff repeatedly in multiple virtual environments, which wastes storage space. For such development tools, installing a single global copy is sufficient.

In another project that gazpachoking is actively maintaining, Ruff is also not included in the dev dependencies: https://github.com/starfederation/datastar-python/blob/e2e2e88e0f9fdbc905a6dd54ca04d3af24f6da71/pyproject.toml#L49-L50

PR #4478, opened 2025-07-06
[Change] AniList: Map more relations and make their lookup lazy

BrutuZ:
### Motivation for changes:
Utilize all provided IDs from the lookup.

### Detailed changes:

- Made the relations lookup lazy
- Mapped all possible IDs
- If the `limit` plugin is configured, use its `amount` to make smaller requests

Relation fields:
```diff
  anidb_id
+ anime_planet_id
+ anisearch_id
+ imdb_id
  kitsu_id
+ livechart_id
  mal_id
+ notify_moe_id
  tmdb_id
  tvdb_id
```
Caveat: `tmdb_id` could be incompatible with `tmdb_lookup` since it might be a TV ID while the latter only supports Movies.

- N/A.

### DIFF
--- a/flexget/plugins/input/anilist.py
+++ b/flexget/plugins/input/anilist.py
@@ -5,8 +5,9 @@ from loguru import logger
 
 from flexget import plugin
 from flexget.config_schema import one_or_more
-from flexget.entry import Entry
+from flexget.entry import Entry, register_lazy_lookup
 from flexget.event import event
+from flexget.utils import requests
 from flexget.utils.cached_input import cached
 from flexget.utils.requests import RequestException, TokenBucketLimiter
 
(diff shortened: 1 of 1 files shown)


flexget/plugins/input/anilist.py
```diff
@@ -272,11 +272,28 @@ def on_task_input(self, task, config):
...
+    ids = {}
+    try:
+        ids: dict[str, str | int] = requests.post(
+            'https://relations.yuna.moe/api/v2/ids',
+            json={'anilist': entry.get('al_id', eval_lazy=False)},
+        ).json()
+        logger.debug('Additional IDs: {}', ids)
+    except RequestException as e:
+        logger.verbose(f"Couldn't fetch additional IDs: {e}")
+    if not isinstance(ids, dict):
```
paranoidi: This might mask problems for unexpected returns, perhaps adding a debug log would be good idea.

BrutuZ: Added another exception and raised their log level to Warning

PR #4483, opened 2025-07-11
ci(deploy): improve release-readiness detection logic

vivodi:
The previous method of detecting new commits by comparing the `develop` and `master` branches was prone to failure. If the `master` branch was accidentally modified, causing it to be ahead of `develop`, our release script would fail.

This issue occurred when adopting GitHub Merge Queue, which committed directly to `master` but not to `develop`. This situation required a manual merge of `master` back into `develop` to proceed with the release (see commit https://github.com/flexget/flexget/commit/7414c3c8b7c46dad7ef07836d5d6a1289e7ceb01).

To prevent this issue from recurring and to increase the stability of the release process, this commit modifies the detection logic. The script now determines if a new release is needed by inspecting the tags of the most recent commits, rather than relying on a branch comparison.


gazpachoking:
Does this just eliminate the master branch then? I guess that's fine, the theory was that master branch would always have been the latest release, but that's a bit silly with our release schedule, and the tags will do that job.

vivodi:
> Does this just eliminate the master branch then?

Yes.

PR #4530, opened 2025-08-04
chore: remove `renovate-config-validator` pre-commit hook

vivodi:
@gazpachoking I just realized that the previous PR only added the check in GitHub Actions, but I forgot to remove `renovate-config-validator` from pre-commit.


gazpachoking:
Whoops, I totally didn't notice either

PR #4534, opened 2025-08-05
ci: reduce CI overhead by running autofix and zizmor only when specific files have changed

vivodi:
Detailed changes:
- Run `autofix.yml` only when `pyproject.toml` or `uv.lock` changes.
- Run `zizmor.yml` only when `.github/workflows` changes.

Limitation:

- Due to https://github.com/orgs/community/discussions/44490, these checks cannot be set as a required check for now unless it is run for all PRs, which would increase CI overhead.

- Due to https://github.com/orgs/community/discussions/45899, these checks cannot be set to run in the GitHub Merge Queue for now unless it is run for all PRs, which would increase CI overhead.

### DIFF
--- a/.github/workflows/autofix.yml
+++ b/.github/workflows/autofix.yml
@@ -4,7 +4,15 @@ on:
   push:
     branches:
       - develop
+    paths:
+      - pyproject.toml
+      - uv.lock
+      - .github/workflows/autofix.yml # This is needed because there might be `uv.lock` changes caused by a uv update.
   pull_request_target: # zizmor: ignore[dangerous-triggers]
+    paths:
+      - pyproject.toml
+      - uv.lock
+      - .github/workflows/autofix.yml # This is needed because there might be `uv.lock` changes caused by a uv update.
 permissions: {}
 concurrency:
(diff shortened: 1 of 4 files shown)


vivodi:
@gazpachoking I am planning to simplify the `autofix.yml` workflow by using the `autofix-ci/action`. This will require you to install the [autofix.ci](https://github.com/marketplace/autofix-ci) GitHub App for the Flexget repository.

The current `autofix.yml` I wrote to resolve inconsistencies between `pyproject.toml` and `uv.lock` is functional, but it is overly complex and difficult to maintain. This complexity primarily stems from the use of the `pull_request_target` trigger, which runs on the base branch.

By using `autofix-ci/action`, we can switch to a `pull_request` trigger. This change will allow us to handle pull requests from forks more simply and avoid the current complexities.

For your reference, the well-known JavaScript framework, Vue, [uses this GitHub App](https://github.com/vuejs/core/blob/c875019d49b4c36a88d929ccadc31ad414747c7b/.github/workflows/autofix.yml#L34), which should alleviate any security concerns.

gazpachoking:
I added autofix-ci to the org.

PR #4535, opened 2025-08-07
refactor: simplify autofix.yml using autofix-ci/action

vivodi:

### DIFF
--- a/.github/workflows/changelog.yml
+++ b/.github/workflows/changelog.yml
@@ -35,7 +35,6 @@ jobs:
       - name: Install uv and Python
         uses: astral-sh/setup-uv@e92bafb6253dcd438e0484186d7669ea7a8ca1cc # v6
         with:
-          enable-cache: true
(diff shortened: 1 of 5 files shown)


.github/workflows/changelog.yml
```diff
@@ -35,7 +35,6 @@ jobs:
       - name: Install uv and Python
         uses: astral-sh/setup-uv@e92bafb6253dcd438e0484186d7669ea7a8ca1cc # v6
         with:
-          enable-cache: true
```
vivodi, line 38: No need to explicitly specify; caching is enabled by default for GitHub-hosted runners.

See https://github.com/astral-sh/setup-uv/blob/142240426d0e82f3e584603c3aa194c582f5aed3/README.md?plain=1#L202 and https://github.com/astral-sh/setup-uv/blob/142240426d0e82f3e584603c3aa194c582f5aed3/src/utils/inputs.ts#L36-L38

gazpachoking:
Awesome, much simpler. Thanks!

PR #4558, opened 2025-08-16
fix: Flexget not send notifications with new versions of Matrix

### LINKED ISSUE #4513: BUG: Flexget not send notifications with new versions of Matrix.
### Expected behaviour:

When I try to send a notification to my Matrix server, I get an error saying the message cannot be sent.
From what I've been able to find out, newer versions of the Matrix servers no longer use the method currently used by FlexGet:
```python
config['server'],
"_matrix/client/r0/rooms",
config['roomid'],
"send/m.room.message?access_token=" + config['token'],
```

This method of sending is old and no server (most recent ones) accepts it. Now, with the newer versions (Conduit, Element, etc.), they use **v3** instead of **r0**.

Therefore, I understand that this option would need to be updated to take advantage of the improvements brought by the new versions of the Matrix server.

In my case, as a Matrix server, I use **Conduit**, and as a client, on the local computer, I use **Element** and on the mobile device, **FluffyChat**.

Likewise, to send notifications right now and run tests, I'm using the following bash script:
```bash
response=$(curl -s -XPUT \
  -H "Authorization: Bearer ${MATRIX_TOKEN}" \
  -H "Content-Type: application/json" \
  -d "{
    \"msgtype\": \"m.text\",
    \"body\": \"${msg}\",
    \"format\": \"org.matrix.custom.html\",
    \"formatted_body\": \"${formatted_msg}\"
  }" \
  "${MATRIX_SERVER}/_matrix/client/v3/rooms/${PODCAST_ROOM}/send/m.room.message/${txn_id}")
```
And as you can see, it uses the new version of the Matrix API (**v3**). I tried testing to send with the old version (**r0**) and no Matrix server would accept it. It was all errors.

### Steps to reproduce:

There aren't any specific steps; you just need to configure FlexGet notifications to use Matrix and send them to a current Matrix server. By reviewing the FlexGet logs, you can see the error I reported below.
And I know for sure it's not an error with my Matrix server, because with the exception of FlexGet, it works perfectly with other applications and with my current server configuration, and I receive all messages.

Even using the [Pushbits](https://www.pushbits.io) gateway, which mimics **Gotify**, you can send messages to FlexGet for applications that don't yet have this capability implemented.

#### Config:

```yaml
notify-all:
  notify:
    entries:
      message: 'moveVideos'
      via:
        - matrix:
            server: 'URL_MATRIX'
            token: 'TOKEN'
            room_id: 'ROOM_ID'
```

#### Log:

<details>

```
Error while sending notification to Matrix: 404 Client Error: Not Found for url: https://MATRIX_URL/_matrix/client/r0/rooms/ROOM_ID/send/m.room.message?access_token=TOKEN
```

</details>

### Additional information:

- FlexGet version: 3.15.26
- Python version: 3.12
- Installation method: Docker
- Using daemon (yes/no): Yes
- OS and version: unRaid ( 7.1.4 ) but not using the default template, but the correct code to use flexget in docker
  -   I try with another VPS ( Debian + Docker + Flexget ) with the same result. Error to send a notification with mathix.
- Link to crash log: `Error while sending notification to Matrix: 404 Client Error: Not Found for url: https://MATRIX_URL/_matrix/client/r0/rooms/ROOM_ID/send/m.room.message?access_token=TOKEN`

vivodi:
close #4513

### Changes in this PR

* **Improved API compatibility**
  The previous API was incompatible with the current Matrix server. Updated to ensure full compatibility.

* **Enhanced authentication**

  * Previously, only manual input of a shared client token was supported.
  * If the client logged out or the token expired, the token had to be updated manually.
  * Now supports direct login with **user + password**, with the option to specify a custom `device_name`.

* **Simplified room handling**

  * Previously, users had to manually enter a `room_id`.
  * When a room was upgraded, a new `room_id` had to be entered again.
  * Now supports using a `room_address`, avoiding this issue and making the process more convenient.

* **Added image upload support**

The usage instructions for the configuration have been documented in the docstring. Rendered results can be found here: https://flexget--4558.org.readthedocs.build/en/4558/api/flexget.components.notify.notifiers.matrix.html#flexget.components.notify.notifiers.matrix.MatrixNotifier

### DIFF
--- a/flexget/components/notify/notifiers/matrix.py
+++ b/flexget/components/notify/notifiers/matrix.py
@@ -1,47 +1,125 @@
+from __future__ import annotations
+
+import asyncio
+import contextlib
+from pathlib import Path
+
 from loguru import logger
-from requests.exceptions import RequestException
 
 from flexget import plugin
 from flexget.event import event
-from flexget.plugin import PluginWarning
-from flexget.utils.requests import Session as RequestSession
+from flexget.plugin import PluginError
+from flexget.utils.simple_persistence import SimplePersistence
 
-requests = RequestSession(max_retries=3)
+with contextlib.suppress(ImportError):
+    from nio import AsyncClient, JoinError, LoginResponse, RoomSendError, UploadError
+
+with contextlib.suppress(ImportError):
+    import aiofiles
+    import magic
+    from aiofiles import os
+    from PIL import Image
 
 plugin_name = 'matrix'
 
 logger = logger.bind(name=plugin_name)
+persist = SimplePersistence('matrix')
 
 
-def urljoin(*args):
-    """Join given arguments into a url.
-
-    Trailing but not leading slashes are stripped for each argument.
-    """
-    return '/'.join(str(x).rstrip('/') for x in args)
+class PotentialCacheError(PluginError):
+    """Exception to be thrown when the failure is suspected to result from cache inconsistency or invalidation."""
 
 
 class MatrixNotifier:
     """Send messages via Matrix.
 
-    The MatrixHttpApi library is required to be installed.
-    Install it with: `pip install matrix_client`
+    The ``matrix`` extra is required to be installed.
+    Install it with: ``pip install flexget[matrix]``
 
-    All fields are required.
+    Configuration:
 
-    server: Matrix hostname to integrate to (required)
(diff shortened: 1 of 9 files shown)


paranoidi:
Looking good. Your work is much appreciated!

PR #4560, opened 2025-08-17
Fix a formatting error in the issue template forms

vivodi:
In Markdown, bold syntax (`**text**`) must not have spaces between the asterisks and the text; otherwise, the bold formatting will not be applied.


vivodi:
The Docker build workflow occasionally fails, but it’s unrelated to this PR.

Could you please try again? @paranoidi

PR #4572, opened 2025-08-20
Disable network calls during tests to enhance stability and prevent unintended network access

vivodi:
**Problem:**

The current `no_requests` fixture is limited to intercepting network requests made by the `requests` and `http` libraries. This leaves a significant gap, as it doesn't prevent tests from making network calls using other popular libraries such as `httpx`, `aiohttp`, or others. This oversight can lead to the inadvertent addition of tests that rely on network access.

These network-dependent tests introduce several critical issues:
*   **Increased test execution time:** Waiting for network responses slows down the entire test suite.
*   **Test instability:** Network failures or timeouts can cause tests to fail intermittently, leading to flaky test runs.
*   **Inaccurate test coverage:** As seen in #4426, a test requiring network access for a `subliminal`-related feature caused significant fluctuations in the project's test coverage metrics.

**Solution:**

This pull request resolves these issues by completely disabling socket calls during the test suite. By blocking network access at the socket level, we ensure that no test can inadvertently make external network requests, regardless of the HTTP library used. This guarantees a more stable, reliable, and faster testing environment.

To achieve this, we have integrated the `pytest-socket` library, which effectively disables all network calls originating from Python's socket interface during test execution.

**CI Enhancement:**

Furthermore, the `VCR_RECORD_MODE` environment variable has been set to `none` in the CI environment. This will cause any test that attempts to make a network request without a pre-existing cassette to fail, thereby ensuring that all cassettes are intentionally and explicitly added to the repository using `git add`.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -22,6 +22,8 @@ jobs:
         python-version: ['3.10', '3.11', '3.12', '3.13', '3.14']
         operating-system: [windows-latest, macos-latest, ubuntu-latest]
       fail-fast: false
+    env:
+      VCR_RECORD_MODE: none
     steps:
       - name: Checkout
         uses: actions/checkout@08eba0b27e820071cde6df949e0beb9ba4906955 # v4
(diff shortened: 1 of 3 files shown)


vivodi:
Based on the [error log](https://github.com/Flexget/Flexget/actions/runs/17142646779/job/48632944297), the failure occurred while uploading the test coverage to Codecov, after the test results had already been uploaded successfully.

This issue is unrelated to this PR and appears to be an occasional error on Codecov’s side. If it happens again, we can report it at https://github.com/codecov/codecov-action.

PR #4576, opened 2025-08-23
Status plugin cli summary is now fast

paranoidi:
Command `flexget status` has always been slow. Claude was able to optimize it from 1m15s to 3s.

### DIFF
--- a/flexget/components/status/cli.py
+++ b/flexget/components/status/cli.py
@@ -70,8 +70,8 @@ def do_cli_summary(manager, options):
 
     with Session() as session:
         # Create aliases for different execution queries
-        LastExecution = aliased(db.TaskExecution)
-        LastSuccess = aliased(db.TaskExecution)
+        LastExecution = aliased(db.TaskExecution)  # noqa: N806
+        LastSuccess = aliased(db.TaskExecution)  # noqa: N806
 
         # Subquery to find the last execution time for each task
         last_execution_subq = (
@@ -85,7 +85,7 @@ def do_cli_summary(manager, options):
             session.query(
                 LastSuccess.task_id, func.max(LastSuccess.start).label('last_success_start')
             )
-            .filter(and_(LastSuccess.succeeded == True, LastSuccess.produced > 0))
+            .filter(and_(LastSuccess.succeeded, LastSuccess.produced > 0))
             .group_by(LastSuccess.task_id)
             .subquery()
         )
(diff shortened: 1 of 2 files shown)


vivodi:
I noticed that the ruff check in pre-commit has failed. For how to handle `==` and `!=` in SQLAlchemy queries, see https://github.com/Flexget/Flexget/pull/4138 for reference.

paranoidi:
@vivodi Yeah, I didn't have pre-commit initialized. What I still can't quite figure out is how to run ruff in this new situation. The only kind of mention in the docs is `uv run pre-commit run -a` but that does not do same checks as what is done when trying to actually commit. Any tips & updates to dev docs are appreciated :)

vivodi:
I am not entirely sure what issue you are encountering. I just reviewed the documentation and found that all the necessary information is already recorded there.

Running `uv sync` will install `pre-commit`, and running `uv run pre-commit install` will configure it so that every time you execute `git commit`, all pre-commit checks (including Ruff checks) will be run **automatically**. However, please note that the checks automatically triggered during `git commit` only apply to staged files. If you wish to run all pre-commit hooks against the entire codebase (for example, after modifying Ruff’s rule configuration), you will need to manually execute `uv run pre-commit run -a`, which is also the command executed in CI.

Since all of the above is already documented in the contributor guide, I am unsure what problem you are facing. Is there something specific you would like me to add to the contributor documentation? (I am the author of the development environment setup documentation.)

paranoidi:
@vivodi  Apologies for the noise, I started the feature branch from master and it is quite a bit behind develop. My mistake. Everything is working as expected.

vivodi:
I don’t think this has anything to do with your branch being behind `develop`, since the additional commits on `develop` are unrelated to Ruff.

> What I still can't quite figure out is how to run ruff in this new situation.

When you refer to a "new situation," do you mean that a commit containing code which would fail the pre-commit checks has already been submitted, and you now want to run pre-commit against that code?

First, I should point out that if pre-commit had been properly configured from the beginning with `uv run pre-commit install`, the situation you described would never occur.

If you are already in this situation, the only option is to run `uv run pre-commit run -a` to check the entire codebase; it is not possible to run the checks only against the last commit. _Alternatively_, you could revert the commit and run `git commit` again, but in my view, simply running `pre-commit run -a` is more convenient.

paranoidi:
I am not sure why the ruff did not install correctly on the first try. It installed fine with develop. And by "new situation" I mean whole project being under `uv`, it has been a while since I've pushed any code in here :)

vivodi:
> I started the feature branch from master and it is quite a bit behind develop.

The `master` branch is no longer in use. To prevent others from making similar mistakes, you can delete the `master` branch.

paranoidi:
Removed master to avoid such mistakes in the future.

PR #4580, opened 2025-08-23
docs: slightly widen the right sidebar for better readability

vivodi:
The right sidebar is too narrow at the moment, so headings often wrap onto two lines.


paranoidi:
Great work on the docs, they were much improved when I used them yesterday :)

PR #4590, opened 2025-08-24
fix(deps): use subliminal 2.3

matlads:
This commit clears a TODO item that was tied to the sublminal issue 1302 being completed.

https://github.com/Diaoul/subliminal/issues/1302 was resolved


vivodi:
@matlads Upgrading `subliminal` to 2.3 isn’t just about clearing out TODOs — the jump from 2.2 to 2.3 also comes with API changes. You’ll need to look into why the tests are failing and update the code accordingly to get them all passing.

gazpachoking:
Going to close this for now. Our bots are fighting over the formatting in here. I think I fixed the underlying issue for future PRs though. In any case, I believe we need some code updates to support the new subliminal version before we can bump it.

PR #4597, opened 2025-08-29
feat: Manipulate allows easier regex removals with erase

paranoidi:
More vibing. Reduce redundant configuration when stripping away unwanted texts.

```
manipulate:
  - title:
      replace:
        regexp: '^advertisement.'
        format: ''
  - title:
      replace:
        regexp: '^more.ads.'
        format: ''
```

to

```
manipulate:
  - title:
      erase:
        - '^advertisement.'
        - '^more.ads.'
```

### DIFF
--- a/flexget/plugins/modify/manipulate.py
+++ b/flexget/plugins/modify/manipulate.py
@@ -24,11 +24,16 @@ class Manipulate:
               regexp: <regexp>
               format: <regexp>
             [remove]: <boolean>
+            [erase]: <list of regexps>
 
     Example::
       manipulate:
         - title:
             extract: \[\d\d\d\d\](.*)
+        - title:
+            erase:
+              - "^unwanted.noise."
+              - "^more.advertisement."
     """
 
     schema = {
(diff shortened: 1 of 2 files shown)


flexget/plugins/modify/manipulate.py
```diff
@@ -116,10 +126,29 @@ def process(self, entry, jobs):
                     field_value,
                 )
                 if config.get('remove'):
+                    # Eemove entire field
```
gazpachoking: Where's that spell checker task when you need it. 😄

flexget/plugins/modify/manipulate.py
```diff
@@ -43,6 +48,11 @@ class Manipulate:
                     'extract': {'type': 'string', 'format': 'regex'},
                     'separator': {'type': 'string'},
                     'remove': {'type': 'boolean'},
+                    'erase': {
+                        'type': 'array',
+                        'items': {'type': 'string', 'format': 'regex'},
+                        'minItems': 1,
```
gazpachoking: Do we need the minItems? I know it doesn't make sense to include it if it isn't doing anything, just not sure it should cause the config to error out.

gazpachoking (commented):
Seems good to me. 👍 Couple small nitpicks, don't really care if they get changed, you can merge as is if you want.

PR #4604, opened 2025-09-03
ci: move `pyproject-fmt` in CI from `pre-commit.ci` to `GitHub Actions`

vivodi:
Because `pyproject-fmt` conflicts with Renovate, move `pyproject-fmt` in CI from `pre-commit.ci` to GitHub Actions. In GitHub Actions, we can run `pyproject-fmt` only when `github.event.pull_request.user.login != 'renovate[bot]'`.

### DIFF
--- a/.github/workflows/autofix-pyproject-fmt.yml
+++ b/.github/workflows/autofix-pyproject-fmt.yml
@@ -1,0 +1,37 @@
+name: autofix.ci
+on:
+  push:
+    branches: [develop]
+    paths:
+      - .github/workflows/autofix-pyproject-fmt.yml
+      - pyproject.toml
+  pull_request:
+    paths:
+      - .github/workflows/autofix-pyproject-fmt.yml
+      - pyproject.toml
+permissions: {}
+concurrency:
+  group: ${{ github.workflow_ref }}-${{ github.event.number || github.ref || github.run_id }}
+  cancel-in-progress: true
+env:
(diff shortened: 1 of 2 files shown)


.github/workflows/autofix-pyproject-fmt.yml
```diff
@@ -0,0 +1,37 @@
...
+concurrency:
+  group: ${{ github.workflow_ref }}-${{ github.event.number || github.ref || github.run_id }}
+  cancel-in-progress: true
+env:
+  PYTHON_VERSION: '3.13' # renovate: datasource=python-version depName=python
+  UV_VERSION: 0.8.13 # renovate: datasource=pypi depName=uv
+jobs:
+  uv-lock:
+    name: run pyproject-fmt
+    if: github.event.pull_request.user.login != 'renovate[bot]'
```
vivodi, line 22: I think this should solve the problem.

gazpachoking:
I think this should fix the conflict, does it seem right to you? https://github.com/Flexget/Flexget/commit/3a96254acaa7641869d30988d48118f7cebd4e97

vivodi:
> I think this should fix the conflict, does it seem right to you? [3a96254](https://github.com/Flexget/Flexget/commit/3a96254acaa7641869d30988d48118f7cebd4e97)

Thanks for pointing this out! I appreciate you finding a cleaner solution. Let’s go with that approach.

PR #4688, opened 2025-10-16
chore: make Renovate read the correct Read the Docs Python version

vivodi:
The Python used by Read the Docs for building is not updated immediately after a new CPython release.

This PR uses the correct datasource to prevent Renovate from creating incorrect PRs.

### DIFF
--- a/.github/renovate.json5
+++ b/.github/renovate.json5
@@ -30,11 +30,20 @@
     {
       customType: 'regex',
       managerFilePatterns: ['/^\\.readthedocs\\.yaml$/'],
-      matchStrings: [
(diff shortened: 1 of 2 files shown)


vivodi:
The Docker image build that’s completely unrelated to this PR always fails.

It used to fail only occasionally, but now it fails every time I try.

gazpachoking:
Do we need to update the sha for python:3.11-alpine to a newer version or something? Seems like something is wrong with alpine doing the apk add.

vivodi:
I don’t think it’s related to the SHA update of `python:3.11-alpine`, because a force push with **no changes** made the Docker image build succeed.

I still don’t understand why `apk update` sometimes fails.

**EDIT:** I submitted #4689 to try to fix this issue, though I'm not completely sure it’ll work or if it’s the best solution.

PR #4714, opened 2025-11-07
build: prevent deployment failures caused by out-of-sync commits during release workflow

vivodi:
#### Summary

The last two deployments failed due to a flaw in our release workflow. During the execution of `release.yml`, if new commits are pushed to the `develop` branch **after** `git checkout` but **before** `git push`, the workflow’s `git push` step fails because the remote branch has diverged — yet the new version still gets published to PyPI. This results in subsequent deployment workflows failing, since PyPI already contains the newly published version.

#### Changes

* Added a `git pull` step before `git push` to synchronize the local branch with the remote and prevent push failures.
* Moved the PyPI publish step to occur **after** the `git push` step.

  * This ensures that if a merge conflict occurs (for example, when `_version.py` has been updated in the new commits), the conflict will only cause a single failed run rather than breaking all subsequent automatic deployments, which eliminates the need for manual intervention.

#### Impact

This fix improves the robustness of our release process by ensuring:

* PyPI publishing only happens after successful synchronization with the remote branch.
* Future automated deployments continue to function even if a temporary failure occurs once due to merge conflicts.


.github/workflows/release.yml
```diff
@@ -18,49 +21,86 @@ jobs:
...
-      - name: Setup Git User
+      - name: Error if running on a branch other than develop
         run: |
-          git config user.email github-actions[bot]@users.noreply.github.com
-          git config user.name github-actions[bot]
+          if [ "$(git rev-parse HEAD)" != "$(git rev-parse origin/develop)" ]; then
+            echo "Release script should be run from develop branch."
+            exit 1
+          fi
+      - name: Cancel workflow run if there are no new commits
```
vivodi, line 30: Move the logic that skips deployment earlier to enable fail-fast behavior and reduce the complexity of subsequent conditional logic.

.github/workflows/release.yml
```diff
@@ -18,49 +21,86 @@ jobs:
...
+            core.setOutput('deployment_id', res.data.id);
+      - name: Set Deployment Status in_progress
+        uses: actions/github-script@ed597411d8f924073f98dfc5c65a23a2325f34cd # v8
+        with:
+          script: |
+            await github.rest.repos.createDeploymentStatus({
+                ...context.repo,
+                deployment_id: ${{steps.create-deployment.outputs.deployment_id}},
+                state: 'in_progress',
+            })
```
vivodi, line 65: The `trigger-deploy.yml` workflow has been merged into `release.yml` to simplify the deployment logic. However, the `justacodename/deployment-status` action requires the workflow to be triggered by a `deployment` event, so it cannot be used.

PR #4726, opened 2025-11-12
test: simplify pytest configuration using improvements from pytest v9

vivodi:

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -42,8 +42,8 @@ repos:
     rev: v0.11.0.1
     hooks:
       - id: shellcheck
-  - repo: https://github.com/henryiii/validate-pyproject-schema-store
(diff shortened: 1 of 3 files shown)


.pre-commit-config.yaml
```diff
@@ -42,8 +42,8 @@ repos:
     rev: v0.11.0.1
     hooks:
       - id: shellcheck
-  - repo: https://github.com/henryiii/validate-pyproject-schema-store
-    rev: 2025.08.07
+  - repo: https://github.com/abravalheri/validate-pyproject
```
vivodi, line 45: `validate-pyproject-schema-store` doesn’t use the latest pytest schema. Also, I don’t think this tool is particularly useful — if a tool causes more problems than it solves, I believe we should remove it.

PR #4728, opened 2025-11-12
add(plugin): Drop "input did not return anything" warning to verbose …

lingfish:
The change was seemingly made years ago to [plugins/input/inputs.py](https://github.com/Flexget/Flexget/blob/f7fba3f29412118e34ffd14a72d320616bdb0313/flexget/plugins/input/inputs.py#L47), but I'm getting warnings from utils.

For example, I'm getting it when a movie list from Trakt is currently empty, which is quite normal.

### DIFF
--- a/flexget/components/series/configure_series.py
+++ b/flexget/components/series/configure_series.py
@@ -54,7 +54,11 @@ class ConfigureSeries(plugin_series.FilterSeriesBase):
                 logger.warning('Error during input plugin {}: {}', input_name, e)
                 continue
             if not result:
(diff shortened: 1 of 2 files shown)


gazpachoking:
Hmm. I'm not sure this fix is quite right, but I don't think we have a good way to properly address it at the moment. The no_entries_ok plugin is meant to indicate that it's okay if the _task_ doesn't produce entries, but in these situations we are producing entries to configure other plugins, not directly into the task. I guess it's fine though, as I don't have any better ideas. Anyone have other feelings?

PR #4778, opened 2025-12-10
Fix flaky tests

vivodi:
In #4776, I used the new VCR.py v8 option `drop_unused_requests=True`, but it causes tests to fail when parameterization leads to different possible requests being made.


gazpachoking:
Ahh, I was confused what was happening there. Thanks!

PR #4781, opened 2025-12-13
feat: add 'normalize' config option to torznab plugin

pogman-code:
Hello,
As I discussed earlier in #4780, some torznab indexer are not very friendly with special chars.
So the proposed change is to add an optional config to the plugin to normalize strings for queries

```yaml
torznab:
  normalize: none/unicode/scene
```

### DIFF
--- a/flexget/plugins/input/torznab.py
+++ b/flexget/plugins/input/torznab.py
@@ -4,7 +4,7 @@ from xml.etree import ElementTree as ET
 from loguru import logger
 
 from flexget import options, plugin
-from flexget.components.sites.utils import torrent_availability
+from flexget.components.sites.utils import normalize_scene, normalize_unicode, torrent_availability
 from flexget.entry import Entry
 from flexget.event import event
 from flexget.plugin import PluginError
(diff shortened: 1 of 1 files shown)


paranoidi:
Wouldn't this be better in discover plugin so that it benefits any possible search plugin?

pogman-code:
> Wouldn't this be better in discover plugin so that it benefits any possible search plugin?

It's not that trivial because search plugins can use various entry properties (title, series_name, imdb_name, imdb_id, etc.)

pogman-code:
Hi, sorry to bump this but should I change something and/or is it a bad idea/approach?

paranoidi:
@pogman-code Hi, apologies for the delay. No worries about bumping the PR. It's not necessarily a bad idea, but this does come up occasionally in various different contexts and thus a more generic solution would be preferable.

Best idea that I've come up so far is normalizer plugin type which can be plugged into any (supporting) plugin. Aside from generic remove special characters these could be made for very specific use-cases like movie titles and have options such as strip or include the year.

@gazpachoking Thoughts?

paranoidi:
Just a note, I did end up trying couple different generic normalizers implemented. But I didn't yet come up with a solution that would be satisfactory on all fronts.

pogman-code:
> Just a note, I did end up trying couple different generic normalizers implemented. But I didn't yet come up with a solution that would be satisfactory on all fronts.

I poked around to find something too to no avail, but I blamed my lack of knowledge in the codebase :sweat_smile:

PR #4863, opened 2026-02-14
feat: add yamtrack_list input plugin

pogman-code:
### Motivation for changes:

Yamtrack is a self-hosted alternative to Trakt (see: https://github.com/FuzzyGrim/Yamtrack).  
The motivation here is to use a Yamtrack list as input, similar to other list input plugins.

### Detailed changes:

- The plugin scrapes a Yamtrack list page, it uses `session_id` cookie as of now because Yamtrack does not provides an API yet (see: https://github.com/FuzzyGrim/Yamtrack/pull/924 - no ETA)

```
yamtrack_list:
  host: http://localhost:8000 
  session_id: abcdefghijklmnopqrstuvwxyz
  list: 123 
  type: movie
```

### Future

- When the API will be released, it could be changed to be a "Managed List" plugin I guess.

### DIFF
--- a/flexget/plugins/input/yamtrack_list.py
+++ b/flexget/plugins/input/yamtrack_list.py
@@ -1,0 +1,122 @@
+import re
+
+from loguru import logger
+
+from flexget import plugin
+from flexget.entry import Entry
+from flexget.event import event
+from flexget.utils.requests import RequestException
+from flexget.utils.soup import get_soup
+
+logger = logger.bind(name='yamtrack_list')
+
+
+class YamtrackList:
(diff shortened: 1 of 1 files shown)


BrutuZ:
The API appears to have been merged, so maybe this could be converted to a managed list instead?

pogman-code:
> The API appears to have been merged, so maybe this could be converted to a managed list instead?

No quite, it has been merged in the main repository but not in the main branch and is not available in any release yet

But don't worry, I have not forgot this ;)

PR #4899, opened 2026-03-02
feat: make automatic configuration reloading the default behavior

vivodi:
### Motivation for changes:
Paranoidi pointed out in [#4860](https://github.com/Flexget/Flexget/discussions/4860#discussioncomment-15962259) that we should make `--autoreload-config` the default behavior and add `--no-autoreload-config` to disable it.
### Detailed changes:
Make `--autoreload-config` the default behavior and add `--no-autoreload-config` to disable it.

- Fixes #4860

### DIFF
--- a/flexget/manager.py
+++ b/flexget/manager.py
@@ -129,7 +129,7 @@ class Manager:
             logger.info('last manager was not torn down correctly')
 
         self.args = args
-        self.autoreload_config = False
+        self.autoreload_config = True
         self.config_file_hash: str | None = None
         self._config_path: Path | None = None
         self.log_filename: str = ''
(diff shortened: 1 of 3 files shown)


gazpachoking:
Should we leave the --autoreload-config so that we don't break people's scripts? (It wouldn't really do anything anymore, but at least you wouldn't get an error if your e.g. cron entry had it.)

flexget/manager.py
```diff
@@ -470,7 +470,11 @@ def daemon_command(self, options: argparse.Namespace) -> None:
             if options.daemonize:
                 self.daemonize()
             if options.autoreload_config:
-                self.autoreload_config = True
+                logger.warning(
+                    'The "--autoreload-config" flag is now default and deprecated. It is slated for removal in a future version, after which its use will cause an error. Please remove it from your command.'
+                )
```
vivodi, line 475: Now, when using the `--autoreload-config` option, a deprecation warning is displayed instead of causing an error.

gazpachoking:
Looks good, thanks!

PR #4929, opened 2026-03-22
ci: remove redundant release check

vivodi:
### Motivation for changes:
I’ve been reconsidering the necessity of the commit status check in our release workflow. Since all changes are integrated via pull requests and **managed through a merge queue**, this additional verification feels redundant.

The merge queue already ensures that every PR is tested against the latest base branch before merging, effectively guaranteeing a green build on the main branch. Most established projects don't verify commit statuses again at the release stage, and our current implementation relies on _private_ GitHub APIs, which poses a long-term maintenance risk.

As long as contributors with write access continue to use the queue **rather than pushing directly to the repository**, we can safely remove this check without compromising release quality.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -51,13 +51,6 @@ jobs:
           token: ${{ secrets.flexgetbot_pat }}
           persist-credentials: true
           fetch-depth: 0
-      - name: Check Commit Status # Currently, the GitHub frontend private API is being used, which may change at any time. Using the public API, however, is quite complex, as it requires querying both "Get the combined status for a specific reference" and "List check runs for a Git reference", and ensuring that both are 'success'.
-        run: |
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yml
```diff
@@ -51,13 +51,6 @@ jobs:
           token: ${{ secrets.flexgetbot_pat }}
           persist-credentials: true
           fetch-depth: 0
-      - name: Check Commit Status # Currently, the GitHub frontend private API is being used, which may change at any time. Using the public API, however, is quite complex, as it requires querying both "Get the combined status for a specific reference" and "List check runs for a Git reference", and ensuring that both are 'success'.
```
vivodi, line 54: Currently, we rely on GitHub’s internal frontend APIs to check commit statuses during the release process. Unlike public APIs, these are not guaranteed to be stable and could change at any moment without notice.

vivodi:
@gazpachoking By the way, we could technically make the coverage check informational, but I think the red '❌' is a good reality check for authors whose PRs tank the coverage. It keeps everyone accountable.

Plus, don't worry about 'breaking' the `develop` branch—it’s not a permanent fail. Once someone merges a PR that doesn't drop the coverage (even a simple Renovate one), the check will go back to green.

gazpachoking:
Makes sense, I'm good with it, thanks!

PR #4965, opened 2026-04-09
Migrate `sftp_list`, `sftp_download` and `sftp_upload` plugins from `pysftp` to `asyncssh`

### LINKED ISSUE #4532: Migrate SFTP plugins from unmaintained `pysftp` to `paramiko`
### Summary

Our `sftp_download`, `sftp_list`, and `sftp_upload` plugins currently rely on the `pysftp` library for SFTP operations. The last update to `pysftp` was in 2016, and the project is no longer maintained.

This presents a growing risk, including potential security vulnerabilities and the lack of support for modern features. For example, `pysftp` has known unfixed issues, particularly on Windows, and does not support newer SSH key types like Ed25519 and ECDSA, which are supported by its underlying dependency, `paramiko`.

### The Problem

The core dependency of `pysftp`, `paramiko`, is actively maintained and has recently released version 4.0. This new version is incompatible with the outdated `pysftp`, which will break our SFTP functionality if `paramiko` is upgraded. The issue is documented in the `pysftp` GitHub repository: [https://github.com/ryhsiao/pysftp/issues/10](https://github.com/ryhsiao/pysftp/issues/10).

Since `pysftp` is essentially a high-level wrapper for `paramiko`, the most direct and future-proof solution is to migrate our plugins to use `paramiko` directly. This will ensure continued compatibility, access to the latest security updates, and a wider range of features.

### Task

The goal of this task is to remove the `pysftp` dependency and refactor the `sftp_download`, `sftp_list`, and `sftp_upload` plugins to use the `paramiko` library directly for all SFTP operations.

**Acceptance Criteria:**
*   All calls to `pysftp` in the specified plugins are replaced with equivalent functionality from `paramiko`.
*   The `pysftp` library is removed from the project's dependencies.
*   The functionality of the `sftp_download`, `sftp_list`, and `sftp_upload` plugins remains intact and is fully tested after the migration.
*   The plugins now support the latest version of `paramiko`.

vivodi:
### Motivation for changes:

The current SFTP-related plugins (`sftp_list`, `sftp_download`, and `sftp_upload`) rely on `pysftp`, which is no longer maintained.

Resolves #4532.

### Detailed changes:

- **Library Migration**: Replaced `pysftp` with `asyncssh` across `sftp_list`, `sftp_download`, and `sftp_upload` plugins.
- **Dependency Update**: Updated requirements to include `asyncssh` and removed `pysftp`.

This PR does not modify the plugin configuration schema, so users do not need to change their configurations.


vivodi:
All previously planned functionality has been implemented, but there are still a few points I’d like feedback on:

1. The previous `pysftp`-based implementation could not correctly handle `known_hosts` files (https://github.com/Flexget/Flexget/issues/3619), which led to the introduction of the `host_key` configuration option (https://github.com/Flexget/Flexget/pull/3641). However, using `known_hosts` is generally more convenient for users, and `asyncssh` handles it correctly. Because of this, I’m considering removing the `host_key` configuration option and reverting back to standard `known_hosts` handling.

2. The previous implementation only retried on the initial connection failure, but did not retry subsequent SFTP operations when the connection was lost mid-session. This PR currently adds retry support for failed SFTP requests caused by lost connections. However, the PR also introduces concurrent requests, and combining retry logic with concurrency significantly increases the complexity of the codebase. I’m now considering removing the connection-loss retry feature entirely, since SFTP already runs over the reliable TCP protocol, making the practical benefit of such retries fairly limited. In addition, most of FlexGet’s codebase does not implement retry behavior for connection failures. Given the substantial reduction in complexity, would it make sense to remove this retry logic, even though the functionality has already been implemented in this PR?

3. This PR implements comprehensive concurrent request handling for the `sftp_list` plugin. However, for `sftp_upload` and `sftp_download`, only per-file concurrency (provided internally by `asyncssh`) and entry-level concurrency are implemented. Files within the same entry are still uploaded/downloaded sequentially. As a result, workloads involving a large number of very small files (e.g. a few KB each) may not fully utilize available bandwidth and can perform relatively slowly.

   Implementing true file-level concurrency together with the retry logic mentioned above would be considerably more difficult due to the increased complexity. Additionally, `asyncssh`’s upload/download APIs introduce extra overhead from file attribute queries. Achieving efficient file-level concurrency would likely require manually implementing recursive upload/download logic, and avoiding the API overhead entirely may require submitting upstream changes to `asyncssh`.
   
   Because of this, I also haven't exposed the concurrent request count configuration to users for now.

vivodi:
@gazpachoking Requesting a review of this PR.

Although, as mentioned in my previous comment, there are still some potential areas for improvement, this implementation is already the best solution we can achieve without introducing configuration changes.

The most notable improvement is the `sftp_list` plugin, which is now significantly faster than the previous implementation due to the use of coroutines. During my testing, I also found that the existing `sftp` plugin fails to connect to some SFTP servers, while the new implementation introduced by this PR works with those servers successfully.

Furthermore, thanks to the existing test coverage contributed by previous maintainers, we can verify that the new implementation preserves the behavior of the previous one.

Therefore, if we do not consider the potential improvements mentioned in my previous comment as blockers, I believe this PR is ready to be merged.

PR #5053, opened 2026-06-19
fix: Include playwright to all groups

paranoidi:
### Motivation for changes:

Seems like uv sync --all-groups uninstalls playwright

Is this the correct way to address the issue?

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -100,6 +100,7 @@ all = [
   { include-group = "boto3" },
   { include-group = "deluge" },
   { include-group = "ftp" },
+  { include-group = "imdb" },
   { include-group = "matrix" },
   { include-group = "plexapi" },
   { include-group = "qbittorrent" },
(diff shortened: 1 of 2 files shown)


vivodi:
This PR allows users to install Playwright with `pip install flexget[imdb]`. If you don't want to expose Playwright this way, you can instead use `uv sync --all-groups --inexact` to prevent `uv sync --all-groups` from uninstalling Playwright. See the [`--inexact` option for `uv sync`](https://docs.astral.sh/uv/reference/cli/#uv-sync--inexact) for more details.

However, I think this PR takes a pretty good approach to solving the problem, since most of FlexGet's optional dependencies can be installed via extras. This makes installation simpler for users: even when some optional dependencies have version constraints, users don't need to figure out which versions to install.

By the way, the pre-commit check failed because this PR unnecessarily changes the executable permission of `pyproject.toml`. The Docker build failure appears to have been an intermittent issue.

paranoidi:
My other PR removes playwright and solves WAF via python

PR #5071, opened 2026-07-02
Add a new (optional) parameter the '/tasks/execute/' API endpoint

logicalChimp:
### Motivation for changes:

A number of the series I watch (from a group I like) don't match nicely with the default guess-series logic... so when I call the Execute Task API, with 'entry_dump=true', the returned Entry list has no series data.

By default, the metadata 'guess' functionality is applied once, *before* the title manipulations are performed.

The added flag tells the Execute Task endpoint to re-do the Series / Movie 'guessing' process after the Tasks have run, in case the Tasks contain Title manipulations (or similar) that make the title more procesable.

Note: if the 'series_name' (and/or 'movie_name') is populated, the Entry is skipped and no 'second guess' attempt is made, for the respective plugin.

(This solution was chosen instead of reordering plugin execution order, to minimise the risk of changing existing behaviour/functionality).

### Detailed changes:

- Add a new 'second_guess_metadata' boolean flag (default: False) to the /api/core/tasks endpoint
- New parameter appears in the API documentation
- If the flag is set, then `finish_task` will call the 'apply_second_guess_metadata' function
  - loads the 'metainfo_series plugin, and the 'metainfo_movie' plugin
  - for each Entry
    - if series_name is not populated, re-run the series 'guess_entry' function against the Entry
    - if move_name is not populated, re-run the movie 'guess_entry' function against the Entry
- Added covering unit tests

- NA

NA

NA

### DIFF
--- a/flexget/api/core/tasks.py
+++ b/flexget/api/core/tasks.py
@@ -25,6 +25,7 @@ from flexget.entry import Entry
 from flexget.event import event
 from flexget.log import capture_logs
 from flexget.options import get_parser
+from flexget.plugin import get as get_plugin
 from flexget.task import task_phases
 from flexget.terminal import capture_console
 from flexget.utils import json, requests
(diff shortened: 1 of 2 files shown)


paranoidi:
I think you have a valid problem with the API here. I am wondering though if we could do the second pass intelligently automatically without having to resort into additional API arguments?

logicalChimp:
> I think you have a valid problem with the API here. I am wondering though if we could do the second pass intelligently automatically without having to resort into additional API arguments?

I did consider trying an 'intelligent' approach, but went with the extra flag to absolutely eliminate any risk of changing / breaking existing workflows etc.

In theory, you could just eliminate the flag, and leave the existing guards in place, so that it always tries to repopulate if the series name (and/or movie name) are blank, etc.

PR #5084, opened 2026-07-06
Fix failed checks in `lock file maintenance` PR

vivodi:
### Motivation for changes:

- Newer versions of pytest now require fixtures with a class scope to use `@classmethod`.
- The unmaintained `html5lib` library uses a deprecated module from `chardet`.

### Detailed changes:
- Applied `@classmethod` to class-scoped fixtures.
- Ignored `DeprecationWarning` from `chardet`.


gazpachoking:
LGTM, thanks!

PR #5130, opened 2026-08-04
Use PEP 702 @deprecated for deprecated APIs

vivodi:
## Motivation

Using [PEP 702 `@deprecated`](https://peps.python.org/pep-0702) provides better support for static analysis tools while making deprecation metadata part of the API definition. Explicitly declaring `typing_extensions` also makes the project's dependencies accurate and more maintainable.

## Changes

* Replace `warnings.warn` calls for deprecated APIs with the PEP 702 `@deprecated` decorator.

  * This allows type checkers and IDEs to detect deprecated APIs before runtime.
  * Keep deprecation information available through the standard deprecation mechanism.

* `typing_extensions` was already used in the project but was not listed as a project requirement — it was only installed as a transitive dependency of other packages. It is now explicitly declared as a direct dependency.

  * The project already uses `typing_extensions`, but it was previously only installed as a transitive dependency of other packages.
  * Declaring it explicitly ensures the project does not rely on indirect dependencies.

### DIFF
--- a/docs/contributor/Topic guides/Maintainer guide/Python version bump checklist.rst
+++ b/docs/contributor/Topic guides/Maintainer guide/Python version bump checklist.rst
@@ -27,7 +27,7 @@ Codebase verification
 
 - Search for obsolete version mentions:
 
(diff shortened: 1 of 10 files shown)


gazpachoking:
I think our comments on 'use api v3' are sorta wrong and that api v3 never materialized. But we didn't make anything worse here and I like the decorator warnings. Thanks!

PR #5135, opened 2026-08-06
Pyright corrections

paranoidi:
### Motivation for changes:

Asked llm nicely to fix easy pyright reported issues.

### DIFF
--- a/flexget/manager.py
+++ b/flexget/manager.py
@@ -32,7 +32,7 @@ from flexget.utils.sqlalchemy_utils import ContextSession
 from flexget.utils.tools import get_current_flexget_version, io_encoding, pid_exists
 
 Base = declarative_base()
-Session: type[ContextSession] = sessionmaker(class_=ContextSession)
+Session: sessionmaker[ContextSession] = sessionmaker(class_=ContextSession)
 
 import flexget.log  # noqa: E402
 from flexget import config_schema, db_schema, plugin  # noqa: E402
@@ -174,6 +174,7 @@ class Manager:
 
     @property
     def config_path(self) -> Path:
+        assert self._config_path is not None, 'find_config must be called before config_path is used'
         return self._config_path
 
     @property
@@ -250,7 +251,7 @@ class Manager:
         if db_schema.upgrade_required():
             logger.info('Database upgrade is required. Attempting now.')
             fire_event('manager.upgrade', self)
-            if manager.db_upgraded:
+            if self.db_upgraded:
                 fire_event('manager.db_upgraded', self)
         fire_event('manager.startup', self)
         self.initialized = True
@@ -282,11 +283,12 @@ class Manager:
             set when each respective task has finished running
         """
         if options is None:
-            options = copy.copy(self.options.execute)
+            task_options = copy.copy(self.options.execute)
         elif isinstance(options, dict):
-            options_namespace = copy.copy(self.options.execute)
-            options_namespace.__dict__.update(options)
-            options = options_namespace
+            task_options = copy.copy(self.options.execute)
+            task_options.__dict__.update(options)
+        else:
+            task_options = options
         task_names = self.tasks
         # Only reload config if daemon
         config_hash = self.hash_config()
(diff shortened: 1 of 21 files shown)


flexget/utils/database.py
```diff
@@ -124,7 +124,7 @@ def __init__(self, word: str | CaseInsensitiveWord):
         else:
             self.word = word
 
-    def lower(self) -> str:
+    def lower(self) -> Any:
```
vivodi, line 127: `Any` is too broad here. The correct type should be `str | sqlalchemy.Function`.

flexget/utils/database.py
```diff
@@ -134,7 +134,7 @@ def operate(self, op, other):
             other = CaseInsensitiveWord(other)
         return op(self.lower(), other.lower())
 
-    def __clause_element__(self) -> str:
+    def __clause_element__(self) -> Any:
```
vivodi, line 137: `Any` is too broad here. The correct type should be `str | sqlalchemy.Function`.

flexget/plugin.py
```diff
@@ -631,7 +638,7 @@ def get_plugin_by_name(name: str, issued_by: str = '???') -> PluginInfo:
     return plugins[name]
 
 
-def get(name: str, requested_by: str | object) -> object:
+def get(name: str, requested_by: Any) -> object:
```
vivodi, line 641: One way to avoid the overly broad `Any` is to define a `Protocol` for objects that provide a `plugin_info` attribute:

```python
from typing import Protocol

class HasPluginInfo(Protocol):
    plugin_info: PluginInfo

def get(name: str, requested_by: str | HasPluginInfo) -> object:
    if name not in plugins:
        who = requested_by if isinstance(requested_by, str) else requested_by.plugin_info.name
        raise DependencyError(issued_by=who, missing=name)
    ...
```

flexget/options.py
```diff
@@ -408,13 +417,13 @@ def add_subparser(self, name: str, **kwargs):
...
 
-    def get_subparser(self, name: str, default=_UNSET):
+    def get_subparser(self, name: str, default: object = _UNSET) -> ArgumentParser:
         if not self.subparsers:
             raise TypeError('This parser does not have subparsers')
         p = self.subparsers.choices.get(name, default)
         if p is _UNSET:
             raise ValueError(f'{name} is not an existing subparser name')
-        return p
+        return cast('ArgumentParser', p)
```
vivodi, line 426:
```suggestion
    def get_subparser(self, name: str) -> ArgumentParser:
        if not self.subparsers:
            raise TypeError('This parser does not have subparsers')
        p = self.subparsers.choices.get(name)
        if p is None:
            raise ValueError(f'{name} is not an existing subparser name')
        return p
```
I searched for every use of `get_subparser(self, name, default)` and found that the `default` parameter is never actually used. Therefore, it can be simplified to the implementation above.

This avoids the need for `cast`, which should generally be avoided because it only suppresses the type checker without providing any runtime safety, potentially masking real type errors.

flexget/manager.py
```diff
@@ -786,7 +792,9 @@ def init_sqlalchemy(self) -> None:
             self.database_uri = f'sqlite:///{filename}'
 
         if self.db_filename and not os.path.exists(self.db_filename):
-            logger.verbose('Creating new database {} - DO NOT INTERRUPT ...', self.db_filename)
+            logger.verbose(  # pyright: ignore[reportAttributeAccessIssue]
+                'Creating new database {} - DO NOT INTERRUPT ...', self.db_filename
+            )
```
vivodi, line 797: All uses of `logger.verbose` will trigger the following error:

```text
Cannot access attribute "verbose" for class "Logger"
  Attribute "verbose" is unknown
```

I think that instead of suppressing the error one by one, a better approach would be to migrate the log levels used by FlexGet to Loguru’s predefined log levels. **Using Loguru’s predefined levels avoids Pyright errors, since those levels are already part of Loguru’s type definitions.**

FlexGet currently uses a custom `verbose` level instead of Loguru’s predefined `succeed` level. In my opinion, FlexGet’s current levels — `critical`, `error`, `warning`, `info`, `verbose`, `debug`, and `trace` — are less clear than Loguru’s predefined levels — `critical`, `error`, `warning`, `succeed`, `info`, `debug`, and `trace`.

The reason is that it is difficult to clearly distinguish the semantic differences between `info`, `verbose`, `debug`, and `trace`. In contrast, the meaning of `succeed` is much more explicit and easier to understand.

vivodi (commented):
I haven't reviewed every detail of this PR, but I think we should make an effort to avoid excessive use of `Any`, `cast`, and Pyright warning suppressions. They can hide actual type problems, reduce the effectiveness of static type checking, make the code harder to maintain, and introduce technical debt over time. In most cases, it is better to improve the type design or use more precise types instead of bypassing the type checker.

gazpachoking:
I just peeked at the first file, and it adds a bunch of assert statements to narrow the type, not sure that's the right route

